PR #1, opened 2017-06-16
Add initial set of modules

ehelms:


modules/foreman_organization.py
```diff
@@ -0,0 +1,142 @@
...
+---
+module: foreman_organization
+short_description: Manage Foreman Organization
+description:
+    - Manage Foreman Organization
+author: "Eric D Helms (@ehelms)"
+requirements:
+    - "nailgun >= 0.28.0"
+    - "python >= 2.6"
+    - datetime
```
evgeni: `datetime` is core

modules/foreman_organization.py
```diff
@@ -0,0 +1,142 @@
...
+        module.fail_json(msg="Failed to connect to Foreman server: %s " % e)
+
+    try:
+        changed = ng.organization(name)
+        module.exit_json(changed=changed)
+    except Exception as e:
+        module.fail_json(msg=e)
+
+# import module snippets
+from ansible.module_utils.basic import *
```
evgeni: `import *` should not be used according to upstream (and I tend to agree)

modules/katello_upload.py
```diff
@@ -0,0 +1,188 @@
...
+    organization:
+        description:
+            - Organization that the Product is in
+        required: true
+'''
+
+EXAMPLES = '''
+- name: "Upload my.rpm"
+  local_action:
+      module: katello
```
zjhuntin: I think this should be `katello_upload`

modules/katello_repository.py
```diff
@@ -0,0 +1,189 @@
...
+    organization:
+        description:
+            - Organization that the Product is in
+        required: true
+'''
+
+EXAMPLES = '''
+- name: "Create repository"
+  local_action:
+      module: katello
```
zjhuntin: and this `katello_repository`

ehelms:
@zjhuntin @evgeni updated, thanks for the comments!

modules/foreman_organization.py
```diff
@@ -0,0 +1,139 @@
...
+            - Verify SSL of the Foreman server
+        required: false
+    name:
+        description:
+            - Name of the Foreman organization
+        required: true
+'''
+
+EXAMPLES = '''
+- name: "Create CI Organization"
```
evgeni: btw, I prefer to write `delefate_to: localhost` instead of `local_action`, both do exactly the same.
```yaml
- name: "Create CI Organization"
  foreman_organization:
      username: "admin"
      password: "admin"
      server_url: "https://fakeserver.com"
      name: "My Cool New Organization"
  delegate_to: localhost
```

ehelms: Any particular reason you prefer that?

evgeni: * one fewer level of indentation
* `action:` is deprecated, so reading `local_action:` feels weird

ehelms: Updated every module with this change

modules/katello_product.py
```diff
@@ -0,0 +1,156 @@
...
+        required: true
+    organization:
+        description:
+            - Organization that the Product is in
+        required: true
+'''
+
+EXAMPLES = '''
+- name: "Create Fedora product"
+  katello_product
```
evgeni: missing `:` after `katello_product`

modules/katello_publish.py
```diff
@@ -0,0 +1,154 @@
...
+        required: true
+    organization:
+        description:
+            - Organization that the Product is in
+        required: true
+'''
+
+EXAMPLES = '''
+- name: "Publish a content view"
+  katello_publish
```
evgeni: missing colon

modules/foreman_organization.py
```diff
@@ -0,0 +1,139 @@
...
+  foreman_organization
```
evgeni: missing colon

evgeni:
should the server be `foreman.example.com` or somesuch?

evgeni:
general ACK after the syntax errors are fixed

PR #3, opened 2017-06-21
Fix product update and add url to repository

ehelms:


modules/katello_repository.py
```diff
@@ -52,6 +52,10 @@
         description:
             - Organization that the Product is in
         required: true
+    url:
+        description:
+            - Repostiroy URL to sync from
```
evgeni: "Repository"

ehelms:
I fixed the spelling mistake and updated from `type` to `content_type` since that is how we tend to reference it more so. That also aligns with the Nailgun parameter.

PR #5, opened 2017-06-27
Fix repository to update URL if it changes

ehelms:


ehelms:
@akofink care to review?

ohadlevy:
trigging pr status to test travis.

akofink:
`type` isn't supported, even though it's used in the example. That kind of limits further testing. Mind fixing it while you're in here?

akofink:
I see now that `content_type` *does* work. Still, the example needs to be updated. Indeed, the URL change works now.

PR #6, opened 2017-06-28
Add katello_activation_key module

akofink:
Requires https://github.com/SatelliteQE/nailgun/pull/419

### DIFF
--- a/modules/katello_activation_key.py
+++ b/modules/katello_activation_key.py
@@ -1,0 +1,245 @@
+#!/usr/bin/python
+# -*- coding: utf-8 -*-
+# (c) 2017, Andrew Kofink <ajkofink@gmail.com>
+#
+# This file is part of Ansible
+#
+# Ansible is free software: you can redistribute it and/or modify
+# it under the terms of the GNU General Public License as published by
+# the Free Software Foundation, either version 3 of the License, or
+# (at your option) any later version.
+#
+# Ansible is distributed in the hope that it will be useful,
+# but WITHOUT ANY WARRANTY; without even the implied warranty of
+# MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+# GNU General Public License for more details.
+#
+# You should have received a copy of the GNU General Public License
+# along with Ansible.  If not, see <http://www.gnu.org/licenses/>.
+
+DOCUMENTATION = '''
+---
+module: katello_activation_key
+short_description: Create and Manage Katello activation keys
+description:
+    - Create and Manage Katello activation keys
+author: "Andrew Kofink (@akofink)"
+requirements:
+    - "nailgun >= 0.28.0"
+    - "python >= 2.6"
+options:
+    server_url:
+        description:
+            - URL of Foreman server
+        required: true
+    username:
+        description:
+            - Username on Foreman server
+        required: true
+    password:
+        description:
+            - Password for user accessing Foreman server
+        required: true
+    verify_ssl:
+        description:
+            - Verify SSL of the Foreman server
+        default: false
+    name:
+        description:
+            - Name of the activation key
+        required: true
+    organization:
+        description:
+            - Organization name that the activation key is in
+        required: true
+    lifecycle_environment:
+        description:
+            - Name of the lifecycle environment
+    content_view:
+        description:
+            - Name of the content view
+    subscriptions:
+        description:
+            - List of subscriptions that include name
+'''
+
+EXAMPLES = '''
+- name: "Create katello client activation key"
+  katello_activation_key:
+    username: "admin"
+    password: "admin"
+    server_url: "https://foreman.example.com"
+    name: "Katello Clients"
+    organization: "Default Organization"
+    lifecycle_environment: "Library"
+    content_view: 'client content view'
+    subscriptions:
+        - name: "Red Hat Enterprise Linux"
+  delegate_to: localhost
+'''
+
+RETURN = '''# '''
+
+try:
+    from nailgun import entities
+    from nailgun.config import ServerConfig
+    HAS_NAILGUN_PACKAGE = True
+except:
+    HAS_NAILGUN_PACKAGE = False
+
+
+class NailGun(object):
+
+    def __init__(self, server, entities, module):
+        self._server = server
+        self._entities = entities
+        self._module = module
+
+    def find_organization(self, name):
+        org = self._entities.Organization(self._server, name=name)
+        response = org.search(set(), {'search': 'name="{}"'.format(name)})
+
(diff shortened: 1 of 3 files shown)


modules/katello_activation_key.py
```diff
@@ -0,0 +1,201 @@
...
+        description:
+            - Verify SSL of the Foreman server
+        required: false
+    name:
+        description:
+            - Name of the activation key
+        required: true
+    organization:
+        description:
+            - Organization label that the activation key is in
```
ehelms: This should be Organization name and not label correct? Or should this support either?Or just label since its static?

modules/katello_activation_key.py
```diff
@@ -0,0 +1,201 @@
...
+- name: "Create katello client activation key"
+  katello_activation_key:
+    username: "admin"
+    password: "admin"
+    server_url: "https://foreman.example.com"
+    name: "Katello Clients"
+    organization: Default Organization
+    lifecycle_environment: "Library"
+    content_view:
+      - name: 'client content view'
```
ehelms: This example does not meet the docs, I feel like it should just be `content_view: client content view'`

modules/katello_activation_key.py
```diff
@@ -0,0 +1,201 @@
...
+        if content_view:
+            kwargs['content_view'] = self.find_content_view(content_view, org)
+
+        activation_key = self._entities.ActivationKey(self._server, **kwargs)
+        response = activation_key.search()
+
+        if len(response) == 1:
+            activation_key = response[0]
+        elif len(response) == 0:
+            activation_key = activation_key.create()
```
ehelms: This only does a create and not an update. If the name of the activation key does not change, but I change the content view or lifecycle environment, I want that change to result in an 'update' to the activation key.

ehelms:
Thoughts on adding subscription support? That's one of the biggest reasons to make an Akey.

akofink:
Yeah, I'll go ahead and add subscription support :lollipop:

akofink:
Ready to go! Subscriptions and all. Requires https://github.com/SatelliteQE/nailgun/pull/419

modules/katello_activation_key.py
```diff
@@ -0,0 +1,244 @@
...
+        required: false
+    lifecycle_environment:
+        description:
+            - Name of the lifecycle environment
+        required: false
+    content_view:
+        description:
+            - Name of the content view
+        required: false
+'''
```
evgeni: missing `subscriptions` option

modules/katello_activation_key.py
```diff
@@ -0,0 +1,244 @@
...
+
+EXAMPLES = '''
+- name: "Create katello client activation key"
+  katello_activation_key:
+    username: "admin"
+    password: "admin"
+    server_url: "https://foreman.example.com"
+    name: "Katello Clients"
+    organization: "\"Default Organization\""
```
evgeni: why the double quoting?

modules/katello_activation_key.py
```diff
@@ -0,0 +1,244 @@
...
+    lifecycle_environment: "Library"
+    content_view: 'client content view'
+    subscriptions:
+        - name: "\"Red Hat Enterprise Linux\""
```
evgeni: why the double quoting?

modules/katello_activation_key.py
```diff
@@ -0,0 +1,244 @@
...
+  delegate_to: localhost
+'''
+
+RETURN = '''# '''
```
evgeni: what is this for?

requirements.txt
```diff
@@ -1 +1 @@
-nailgun
+git+https://github.com/SatelliteQE/nailgun.git@master#egg=nailgun
```
evgeni: iek, should we just wait until a proper nailgun release with your changes is out?

modules/katello_activation_key.py
akofink: I honestly copied it from prior modules, but it's an [ansible module documentation directive](http://docs.ansible.com/ansible/dev_guide/developing_modules_documenting.html#return-block) showing what the module returns.

akofink:
@evgeni I think I addressed all your comments. From speaking to QE (nailgun maintainers), I was recommended to use the master branch, rather than the latest release. I believe this is mostly because they don't release very often. If we could push them to release more regularly, then this wouldn't be an issue.

modules/katello_activation_key.py
evgeni: aaah, right! thanks.

modules/katello_activation_key.py
```diff
@@ -0,0 +1,248 @@
...
+            lifecycle_environment=dict(required=False, no_log=False),
+            content_view=dict(required=False, no_log=False),
+            subscriptions=dict(required=False, no_log=False, type='list'),
+        ),
+        supports_check_mode=True
+    )
+
+    if not HAS_NAILGUN_PACKAGE:
+        module.fail_json(msg="Missing required nailgun module (check docs or install "
+                             "with: pip install nailgun")
```
evgeni: missing closing parenthesis for "(check docs …" ;)

modules/katello_activation_key.py
```diff
@@ -0,0 +1,248 @@
...
+        description:
+            - Username on Foreman server
+        required: true
+    password:
+        description:
+            - Password for user accessing Foreman server
+        required: true
+    verify_ssl:
+        description:
+            - Verify SSL of the Foreman server
```
evgeni: should mention the default is false. (there is a `default` key in the docs yaml)

modules/katello_activation_key.py
```diff
@@ -0,0 +1,248 @@
...
+
+def main():
+    module = AnsibleModule(
+        argument_spec=dict(
+            server_url=dict(required=True),
+            username=dict(required=True, no_log=True),
+            password=dict(required=True, no_log=True),
+            verify_ssl=dict(required=False, type='bool', default=False),
+            name=dict(required=True, no_log=False),
+            organization=dict(required=False, no_log=False),
```
evgeni: shouldn't the org be required? otherwise it fails for me with:

```
FAILED! => {"changed": false, "failed": true, "msg": "No organization found for None"}
```

evgeni:
should be good to merge as soon the comments are adressed.

evgeni:
Oh, and I think in future we might want to add support for:
* Description
* Host Limit
* Service Level
* Release Version
* Host Collections

But I don't see that as a blocker.

akofink:
Updated

evgeni:
I think you have set the org as required in the doc string, but not in the module definition? ;)

akofink:
Whoops! Updated.

PR #7, opened 2017-06-28
Fix katello_content_view module

akofink:


modules/katello_content_view.py
```diff
@@ -164,7 +162,8 @@ def main():
     )
 
     if not HAS_NAILGUN_PACKAGE:
-        module.fail_json(msg="Missing required nailgun module (check docs or install with: pip install nailgun")
+        module.fail_json(msg="Missing required nailgun module (check docs or install with: "
+                         "pip install nailgun")
```
ehelms: Why break this across two lines?

modules/katello_content_view.py
```diff
@@ -188,8 +187,12 @@ def main():
     except Exception as e:
         module.fail_json(msg="Failed to connect to Foreman server: %s " % e)
 
+    kwargs = {}
+    if repositories:
+        kwargs['repositories'] = repositories
```
ehelms: This addition of kwargs seems like adding unneeded complexity?

modules/katello_content_view.py
akofink: I believe we use 100 as the line length in Katello. Seems reasonable to me.

modules/katello_content_view.py
akofink: We don't want to pass arguments if they're not given. It's different than passing `None`.

ehelms:
While you are in here fixing the module, you might notice that it only ever `appends` to the repository list rather than ensuring that the list of repositories if only those specified (i.e. stale old repositories are never removed).

modules/katello_content_view.py
```diff
@@ -97,9 +98,7 @@ def find_organization(self, name):
             self._module.fail_json(msg="No organization found for %s" % name)
 
     def find_product(self, name, organization):
-        org = self.find_organization(organization)
```
ehelms: Can this find_organization method be pulled from the module now?

akofink: No, we still use it on [line 124](https://github.com/theforeman/foreman-ansible-modules/pull/7/files#diff-ad26508e5ed85293d1427174dd4abc74R124).

modules/katello_content_view.py
ehelms: I'd argue we should follow whatever Pep8/Ansible standards are since our ultimate goal is putting these into Ansible proper.

modules/katello_content_view.py
ehelms: Fair point.

modules/katello_content_view.py
ehelms: Help me out, why do we need it in that use case but not this one?

modules/katello_content_view.py
ehelms: Some more info, and we could turn Travis on to do this for us to help. Looks like Ansible requires the following to pass:

`pep8 --ignore=E402 --max-line-length=160`

Per http://docs.ansible.com/ansible/dev_guide/developing_modules_in_groups.html

akofink: Ah, found it. https://github.com/ansible/proposals/issues/50 Yeah, I'll update this. Looks like they're using `pep8 --ignore=E402 --max-line-length=160`

modules/katello_content_view.py
akofink: We should only resolve it once (one API call). Then use the entity object after that in memory.

modules/katello_content_view.py
ehelms: @akofink want to add a .travis.yml setup for that? I can turn it on. Either in this PR or separate.

ehelms: Which I think means this change is unnecessary with a line length of 160?

akofink: Yes to both.

akofink: Actually, let me add a new PR for travis and pep8 so I can go through and fix the pep8 issues that exist.

akofink:
Updated. Whatever repos you give it will exist on the CV. Others will be removed.

modules/katello_content_view.py
```diff
@@ -120,35 +119,34 @@ def find_repository(self, name, product, organization):
...
-                found = [repo for repo in content_view.repository if repo.id == repository.id]
-
-                if not found:
-                    content_view.repository.append(repository)
-                    updated = True
-
-            if updated:
-                content_view.update(['repository'])
+        def get_ids(repository):
+            return repository.id
```
ehelms: Is this an inline function definition? Seems strange a bit to do it this way or need to do it this way.

akofink: Yes, it's inline. It turns out pep8 doesn't like lambdas

ehelms: Can you just pass a lambda to the map function below? I only ask cause this already confused me being inline. If you want to keep the function, could we perhaps move it to a regular instance method of the class? I'm not sure how common these types of functions are in Python but I am not used to them :/

akofink: We can use 2 identical lambdas. That's the real tradeoff. I don't mind doing that if you think it's worth the readability vs. dry-ness.

ehelms: I value DRYness but rank readability above it.

akofink:
Updated to use lambdas.

PR #9, opened 2017-06-29
Fix the organization module so that it can create the org.

bkearney:
While trying to be idempotent, the organziation module will fail if the
org is not present. This patch allows for that case. I am not sure how
the original code handled name changes, but this patch keeps that in tact.


ehelms:
Thanks for the patch @bkearney ! Do you mind fixing the linting issues?

bkearney:
@ehelms fixed.

PR #15, opened 2017-07-18
Add an ansible.cfg for modules and module_utils

akofink:


evgeni:
I'd prefer to have the `module_utils` one lever higher to match the structure in https://github.com/ansible/ansible/tree/devel/lib/ansible, what do you think?

akofink:
You mean `modules/lib/module_utils`?

mdellweg:
@evgeni We are fine with that, but we do not see 'relative to the playbook'-problem solved.
We would like the module_utils to be searched for relative to the role's directory.

akofink:
Like this? (updated)

evgeni:
@akofink I always pretend https://github.com/theforeman/foreman-ansible-modules is "another" `ansible/lib`, so our modules are in `foreman-ansible-modules/modules` and our utils should be in `foreman-ansible-modules/module_utils`. At least that's how I see it, happy to be overridden ;)

@mdellweg / @Fobhep absolutelly agree. We currently have a [snipet how to set up the modules in the readme](https://github.com/theforeman/foreman-ansible-modules#how-to-use-the-repository), this could be extended and synced up with what @akofink has here for the example `ansible.cfg`. Would still not make things automatic, but one step closer.

akofink:
Ah, yeah. I'm going to update the README as well.

akofink:
Updated with readme changes

README.md
```diff
@@ -28,16 +28,13 @@ cd infra/
 git clone https://github.com/theforeman/foreman-ansible-modules.git
 ```
 
-Now, add or update your `ansible.cfg` file to include the modules path:
+Note the `ansible.cfg` file cloned with foreman-ansible-modules. The ansible.cfg
+needs to be in your current directory when you run `ansible` or
+`ansible-playbook`. You can copy it to another location or add it to your
+current ansible configuration.
```
evgeni: I would add a sentence that the paths might need updating if copying the file to a different location.

akofink:
Updated

evgeni:
LGTM, @ehelms thoughts?

PR #16, opened 2017-07-18
Add demo playbooks. Useful for testing/demo.

akofink:
Now, we can easily do a smoke test on a module with `ansible-playbook test/demo/<module>.yml`

### DIFF
--- a/test/demo/activation-keys.yml
+++ b/test/demo/activation-keys.yml
@@ -1,0 +1,17 @@
+---
+- hosts: localhost
+  tasks:
+  - name: "Create katello client activation key"
+    katello_activation_key:
+      username: "admin"
+      password: "changeme"
+      server_url: "https://katello.example.com"
+      verify_ssl: false
+      name: "Katello Clients"
+      organization: "Default Organization"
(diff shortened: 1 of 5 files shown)


.gitignore
```diff
@@ -1,2 +1,3 @@
 .tmp
 *.retry
+test/demo/*.retry
```
evgeni: shouldn't this be already caught by the line above?

test/demo/activation-keys.yml
```diff
@@ -0,0 +1,17 @@
+---
+- hosts: localhost
+  tasks:
+  - name: "Create katello client activation key"
+    katello_activation_key:
+      username: "admin"
```
evgeni: shall we split out username/password/server_url/organization into an fixture file, then it would be easier to make demos in a slightly different env?

akofink: I don't want to overly complicate this at first. It's definitely something we could easily add in the future, but these values are really just example values.

.gitignore
akofink: Updated

PR #17, opened 2017-07-18
Add ptemplate/ptable test data

akofink:
This is for use with `hacking/test-module` (debugging).

@Fobhep

### DIFF
--- a/test/data/partition-table.json
+++ b/test/data/partition-table.json
@@ -1,0 +1,15 @@
+{
+  "username":"admin",
+  "password":"changeme",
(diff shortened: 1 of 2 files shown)


Fobhep:
looks good. I can provide some more testing tasks if you are interested, since we wrote some while we developed the modules.

akofink:
Definitely. We'd like to eventually add tests for these modules, so id you have anything to contribute, it'd be worth having it in the `test/` directory. Feel free to put it in a PR.

Fobhep:
will do, as soon as I find some time :)

PR #19, opened 2017-07-18
Add name='*' option to perform bulk actions on templates

Fobhep:
We (@mdellweg,@fobhep) added a functionality to foreman_ptable and foreman_provisioning_templates to perform bulk actions, i.e. updating or deleting all existing templates at once.

### DIFF
--- a/modules/foreman_provisioning_template.py
+++ b/modules/foreman_provisioning_template.py
@@ -107,6 +107,8 @@ options:
             A name must be provided.
             Possible sources are, ordererd by preference:
             The "name" parameter, config header (inline or in a file), basename of a file.
+            The special name "*" (only possible as parameter) is used
+            to perform bulk actions on all existing templates.
         required: false
     organizations:
         description:
@@ -318,6 +320,13 @@ def main():
         ],
 
     )
+
+    # We do not want a template text for bulk operations
+    if module.params['name'] == '*':
+        if module.params['file_name'] or module.params['template']:
+            module.fail_json(
+                msg="neither file_name nor template allowed if name: *")
+
     if not HAS_NAILGUN_PACKAGE:
         module.fail_json(
             msg='Missing required nailgun module'
@@ -333,13 +342,15 @@ def main():
     state = template_dict.pop('state')
     file_name = template_dict.pop('file_name', None)
 
-    if file_name:
+    if file_name or 'template' in template_dict:
+        if file_name:
+            parsed_dict = parse_template_from_file(file_name, module)
+        else:
+            parsed_dict = parse_template(template_dict['template'], module)
+        # sanitize name from template data
+        if 'name' in parsed_dict and parsed_dict['name'] == '*':
+            module.fail_json(msg="Cannot use '*' as a template name!")
         # module params are priorized
-        parsed_dict = parse_template_from_file(file_name, module)
-        parsed_dict.update(template_dict)
-        template_dict = parsed_dict
-    elif template_dict['template']:
-        parsed_dict = parse_template(template_dict['template'], module)
         parsed_dict.update(template_dict)
         template_dict = parsed_dict
 
@@ -352,20 +363,22 @@ def main():
             module.fail_json(
                 msg='No name specified and no filename to infer it.')
 
+    name = template_dict['name']
+
     try:
         create_server(server_url, (username, password), verify_ssl)
     except Exception as e:
         module.fail_json(msg='Failed to connect to Foreman server: %s ' % e)
 
     ping_server(module)
+
     try:
-        entities = find_entities(ProvisioningTemplate, name=template_dict['name'])
-        if len(entities) > 0:
-            entity = entities[0]
+        if name == '*':
+            entities = find_entities(ProvisioningTemplate)
         else:
-            entity = None
+            entities = find_entities(ProvisioningTemplate, name=template_dict['name'])
     except Exception as e:
-        module.fail_json(msg='Failed to find entity: %s ' % e)
+        module.fail_json(msg='Failed to search for entities: %s ' % e)
 
     # Set Locations of Template
     if 'locations' in template_dict:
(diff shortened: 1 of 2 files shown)


akofink:
Could you add an example of the bulk action to the docs section?

modules/foreman_provisioning_template.py
```diff
@@ -337,13 +346,15 @@ def main():
...
 
-    if file_name:
+    if file_name or 'template' in template_dict:
+        if file_name:
+            parsed_dict = parse_template_from_file(file_name, module)
+        else:
+            parsed_dict = parse_template(template_dict['template'], module)
+        # sanitize name from template data
+        if 'name' in parsed_dict and parsed_dict['name'] == '*':
+            module.fail_json(msg="Cannot use '*' as a template name!")
```
akofink: I believe this is unreachable code. You would have to have `params['name'] == '*' and ('file_name' in params or 'template' in params)`. It would hit [line 332](https://github.com/theforeman/foreman-ansible-modules/pull/19/files#diff-c730a4c9b53728f75758cea520f70012R332).

akofink:
I am able to provide `name = '*'` and `kind = 'finish'`, but all my templates get deleted, not just finish templates.

modules/foreman_provisioning_template.py
mdellweg: Actually, this code is catching a corner case, where someone is trying to shoot himself in the knee by doing something like:
```
- name: Name a template '*' (expect this to fail)
  foreman_provisioning_template:
    <...>
    template: |
      <%#
        name: *
        <...>
      -%>
```
This is not very sensible, but it would wreck havoc if we didn't catch it.
Let's say, we put a long barrel on the nailgun. Now he can only shoot his foot.
We can write a test case about that.

mdellweg:
This is by design. The bulk actions are always meant to affect all templates. All the parameters are meant to set values. There would be no logic to know, whether you want to affect only the templates of one Organization, or assign that Organization to all templates, for example.
`name=='*'` with `state=='absent'` is really only good for having a fresh start. Any parameter added to that is insensible.
A solution could be to add a 'list of filters' option if `name=='*'`. And maybe expect to have no parameters other than `name` if `state='absent'`.
For now, we could try to be more explicit in the documentation.

modules/foreman_provisioning_template.py
akofink: Why can't users use asterisks in the template name? You're checking for `name == '*'`, so it seems like we could allow it.

akofink:
Also, what does it mean to have `name = '*'` and `state = 'present'`? This is essentially a no-op, but the code executes through completely. I think there needs to be more validation in place of the arguments passed. It should handle errors, not rely on users doing the 'right thing', don't you think?

modules/foreman_provisioning_template.py
mdellweg: If a template were to be named '*', which does not sound useful, the code would run through the bulk action branches.
If you want to name your template '***really important finish template***', which in turn might be useful, the module is not going to stop you.

evgeni: Can you fold the above into a code comment? Nobody is going to follow the PR discussion when reading the code next year ;)

module_utils/ansible_nailgun_cement.py
```diff
@@ -42,34 +43,50 @@ def create_server(server_url, auth, verify_ssl):
...
     return changed
 
 
-def find_entity(entity_class, **kwargs):
+def find_entities(entity_class, **kwargs):
+    """ Find entities by certain criteria """
     return entity_class().search(
         query={'search': ','.join(['{0}="{1}"'.format(
-            key, kwargs[key]) for key in kwargs])}
+            key, kwargs[key]) for key in kwargs]), 'per_page': sys.maxint}
```
evgeni: Mhh, I don't really like that. While I agree that we probably never will have more than `MAX_INT` entities, it feels somehow terribly wrong to limit it like this here. Also, I would actually expect `nailgun` to return *all* results of a search query, not just the first 20(?).

Fobhep: yeah, so we are neither happy about it. Sadly nailgun returns, against your and our expectation, 20 results and not all at once.
Does anybody have a better solution?

evgeni: "Fix nailgun" ;)
No, seriously, let's move that discussion to https://github.com/SatelliteQE/nailgun/ and see what they say?

Fobhep: :D hehe, SGTM

Fobhep: I have a feeling, we should probably immediately do this with a PR?

evgeni: https://github.com/SatelliteQE/nailgun/issues/429

Fobhep: thx for opening that issue

evgeni:
this needs a rebase please

Fobhep:
I just wanted to know if there are any updates concerning this PR?
We changed something rather important within the cement library. Right now we are intending to send a new PR for a new module, which would need this change.
Meaning we could in doubt also split the present PR into one for the cement change and one for the "*" feature. Please let us know, how we can solve this in the best way :)

evgeni:
Hey @Fobhep, sorry for the delay. (And argh, dear GitHub, please notify me when the rebase happened).

I must admit I just did not find the time for another review (and test) yet. And @akofink is not available at the moment either :(

Splitting is probably a good idea, as that would make the review easier and you can get the library changes in that are blocking your other module quicker. Thanks!

Fobhep:
hehehe, yes Github do that! :+1: 

ok, thanks anyways for the reply. We will split it then :)

mdellweg:
Together with the split, we removed the new 'create_server_and_find_entities' function from the pull request for it to be more KISS than DRY.

mdellweg:
Maybe it's on again...
Github, please notify @evgeni, a rebase happened.

evgeni:
Hah, but it still needs a rebase after #25 was merged ;)

mdellweg:
I cannot believe the automatic notification system could miss this one, but just in case:
Github, please notify @evgeni 
Btw: i used the new name 'affects_multiple', just in case we get ahead of ourselves and implement the name in a regular expression fashion.

mdellweg:
Do @akofink and @ehelms know, you expect a review from them?
OK, now it is actually a rhetorical question. ;D

mdellweg:
Actually, the check_mode feature was introduced to cement after the fork of this pull request.
I will try a rebase. It should work automagically. 8)

mdellweg:
@akofink `check_mode` should do now.

mdellweg:
For the sake of tradition, one last time (hopefully):
Github, please notify @evgeni , a rebase happened.

mdellweg:
Thank you all for helping to smooth out the rough edges of this feature.

PR #20, opened 2017-07-20
Add Katello manifest module

akofink:
Requires https://github.com/SatelliteQE/nailgun/pull/428

### DIFF
--- a/modules/katello_manifest.py
+++ b/modules/katello_manifest.py
@@ -1,0 +1,196 @@
+#!/usr/bin/python
+# -*- coding: utf-8 -*-
+# (c) 2017, Andrew Kofink <ajkofink@gmail.com>
+#
+# This file is part of Ansible
+#
+# Ansible is free software: you can redistribute it and/or modify
+# it under the terms of the GNU General Public License as published by
+# the Free Software Foundation, either version 3 of the License, or
+# (at your option) any later version.
+#
+# Ansible is distributed in the hope that it will be useful,
+# but WITHOUT ANY WARRANTY; without even the implied warranty of
+# MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+# GNU General Public License for more details.
+#
+# You should have received a copy of the GNU General Public License
+# along with Ansible.  If not, see <http://www.gnu.org/licenses/>.
+
+DOCUMENTATION = '''
+---
+module: katello_manifest
+short_description: Create and Manage Katello manifests
+description:
+    - Create and Manage Katello manifests
+author: "Andrew Kofink (@akofink)"
+requirements:
+    - "nailgun >= 0.29.0"
+    - "python >= 2.6"
+options:
+    server_url:
+        description:
+            - URL of Foreman server
+        required: true
+    username:
+        description:
+            - Username on Foreman server
+        required: true
+    password:
+        description:
+            - Password for user accessing Foreman server
+        required: true
+    verify_ssl:
+        description:
+            - Verify SSL of the Foreman server
+    organization:
+        description:
+            - Organization that the manifest is in
+        required: true
+    manifest_path:
+        description:
+            - Path to the manifest zip file
+    state:
+        description:
+            - The state of the manifest
+        required: true
+        choices:
+            - absent
+            - present
+            - refreshed
+    redhat_repository_url:
+        description:
+            - URL to retrieve content from
+        default: https://cdn.redhat.com
+'''
+# HERE
+EXAMPLES = '''
+- name: "Upload the RHEL developer edition manifest"
+  katello_manifest:
+    username: "admin"
+    password: "admin"
+    server_url: "https://katello.example.com"
+    verify_ssl: false
+    organization: "Default Organization"
+    state: present
+    manifest_path: "/tmp/manifest.zip"
+  delegate_to: localhost
+'''
+
+RETURN = '''# '''
+
+try:
+    from nailgun.entities import (
+        Subscription
+    )
+
+    from nailgun.entity_mixins import (
+        TaskFailedError
+    )
+    HAS_NAILGUN_PACKAGE = True
+except:
+    HAS_NAILGUN_PACKAGE = False
+
+from ansible.module_utils.ansible_nailgun_cement import (
+    create_server,
+    ping_server,
+    find_organization,
+    current_subscription_manifest,
+    set_task_timeout,
+)
+
+
+def validate_params(module, state=None, manifest_path=None):
+    if manifest_path is not None or len(manifest_path) == 0:
+        if state == 'absent':
+            module.fail_json(msg="manifest_path is prohibited when state is absent.")
+        elif state == 'refreshed':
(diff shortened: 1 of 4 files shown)


modules/katello_manifest.py
```diff
@@ -0,0 +1,182 @@
...
+        description:
+            - Path to the manifest zip file
+        required: true
+    state:
+        description:
+            - The state of the manifest
+        required: true
+        choices:
+            - absent
+            - present
```
ehelms: I wonder if we should consider "refreshed" as a state for managing a manifest.

modules/katello_manifest.py
```diff
@@ -0,0 +1,182 @@
...
+    repository_url:
```
ehelms: Would `cdn_url` be more descriptive?

modules/katello_manifest.py
```diff
@@ -0,0 +1,182 @@
...
+    find_organization,
+    current_subscription_manifest,
+)
+
+
+def validate_params(module, state=None, manifest_path=None):
+    manifest_supplied = not (manifest_path is None or len(manifest_path) == 0)
+    if state == 'present' and not manifest_supplied:
+        module.fail_json(msg="manifest_path is required when state is present.")
+    elif state != 'present' and manifest_supplied:
```
ehelms: I think its better to be explicit here by checking for `== 'absent'` since that is a valid state to be in and prevents future breakage if you add more states.

modules/katello_manifest.py
```diff
@@ -0,0 +1,182 @@
...
+        module.fail_json(msg="manifest_path is prohibited when state is absent.")
+
+
+def manifest(module, organization, state, manifest_path=None, repository_url="https://cdn.redhat.com"):
```
ehelms: I lean towards `cdn_url`

modules/katello_manifest.py
```diff
@@ -0,0 +1,182 @@
...
+    changed = False
+    organization = find_organization(module, organization).read()
+    current_manifest = current_subscription_manifest(module, organization)
+    manifest_present = current_manifest is not None
+    if organization.redhat_repository_url != repository_url:
```
ehelms: Empty line above this `if` to logically separate it?

modules/katello_manifest.py
```diff
@@ -0,0 +1,182 @@
...
+        organization.redhat_repository_url = repository_url
+        organization.update({'redhat_repository_url'})
+        changed = True
+    if manifest_path is not None and state == 'present':
```
ehelms: Empty line above this if to logically separate it?

modules/katello_manifest.py
```diff
@@ -0,0 +1,182 @@
...
+        except IOError as e:
+            module.fail_json(msg="Unable to open the manifest file: %s" % e)
+        except TaskFailedError as e:
+            if "same as existing data" in e.message:
+                pass
+            elif "older than existing data" in e.message:
+                module.fail_json(msg="Manifest is older than existing data: %s" % e)
+            else:
+                module.fail_json(msg="Upload of the mainfest failed: %s" % e)
+    elif state != 'present' and manifest_present:
```
ehelms: Prefer checking `absent` explicitly here.

modules/katello_manifest.py
akofink: Well, it's not always the CDN, but I mostly named it this because that's what's in the API docs and in the Katello code.

ehelms:
How is it presented in the UI and CLI?

modules/katello_manifest.py
akofink: It's "Red Hat CDN URL" in the UI and "Red Hat Repository URL" in the CLI.

modules/katello_manifest.py
```diff
@@ -0,0 +1,196 @@
...
+        description:
+            - Verify SSL of the Foreman server
+    organization:
+        description:
+            - Organization that the sync plan is in
+        required: true
+    manifest_path:
+        description:
+            - Path to the manifest zip file
+        required: true
```
evgeni: the path is not required if absent or refreshed

modules/katello_manifest.py
```diff
@@ -0,0 +1,196 @@
...
+    organization = find_organization(module, organization).read()
+    current_manifest = current_subscription_manifest(module, organization)
+    manifest_present = current_manifest is not None
+
+    if organization.redhat_repository_url != repository_url:
+        organization.redhat_repository_url = repository_url
+        organization.update({'redhat_repository_url'})
+        changed = True
+
+    if manifest_path is not None and state == 'present':
```
evgeni: is the `manifest_path` check required here? `validate_params` should have already caught that?

modules/katello_manifest.py
```diff
@@ -0,0 +1,196 @@
...
+    ping_server,
+    find_organization,
+    current_subscription_manifest,
+    set_task_timeout,
+)
+
+
+def validate_params(module, state=None, manifest_path=None):
+    manifest_supplied = not (manifest_path is None or len(manifest_path) == 0)
+    if state == 'present' and not manifest_supplied:
```
evgeni: this could be caught using `required_if`, but the others cannot :(

akofink:
I think I have addressed all the comments.

modules/katello_manifest.py
```diff
@@ -0,0 +1,196 @@
...
+            - Path to the manifest zip file
+    state:
+        description:
+            - The state of the manifest
+        required: true
+        choices:
+            - absent
+            - present
+            - refreshed
+    repository_url:
```
ehelms: I am still  of the opinion that this is confusing given its talked about more so as the CDN URL If you feel this needs to match the API param then lets change it to `redhat_repository_url` which I think is closer to the fact that its a CDN URL (or CDN like URL in the case of disconnected).

The UI shows it as "Red Hat CDN URL" and the CLI as "redhat-repository-url"

ehelms:
Just the minor nitpick or bike shedding :)

modules/katello_manifest.py
akofink: ACK. I will change it to `redhat_repository_url`.

akofink: Oh Bessie! :stuck_out_tongue_winking_eye: Take a [gander at this](https://github.com/Katello/katello/blob/master/app/controllers/katello/api/v2/subscriptions_controller.rb#L96). Clearly we need to work towards a single name for this blasted variable in the future.

modules/katello_manifest.py
```diff
@@ -0,0 +1,196 @@
...
+    if organization.redhat_repository_url != redhat_repository_url:
+        organization.redhat_repository_url = redhat_repository_url
+        organization.update({'redhat_repository_url'})
+        changed = True
+
+    if state == 'present':
+        try:
+            with open(manifest_path, 'rb') as manifest_file:
+                files = {'content': (manifest_path, manifest_file, 'application/zip')}
+                data = {'organization_id': organization.id, 'repository_url': redhat_repository_url}
```
akofink: @ehelms This one's unavoidable, unfortunately. :zipper_mouth_face:

PR #21, opened 2017-07-20
Add a katello_lifecycle_environment module

akofink:


modules/katello_lifecycle_environment.py
```diff
@@ -0,0 +1,195 @@
...
+        description:
+            - Description of the lifecycle environment
+    organization:
+        description:
+            - Organization name that the lifecycle environment is in
+        required: true
+    prior:
+        description:
+            - Name of the parent lifecycle environment
+    present:
```
evgeni: shouldn't this be `state=present`?

Fobhep: I agree, standard seems to be in Ansible a state parameter with choices present and absent.

modules/katello_lifecycle_environment.py
```diff
@@ -0,0 +1,195 @@
...
+        description:
+            - Whether the lifecycle environment should be present or absent on the server
+        required: true
+'''
+
+EXAMPLES = '''
+- name: "Add a production lifecycle environment"
```
akofink: I'm not sure why we put `delegate_to: localhost` in these examples. It's certainly not required, according to my testing (i.e. [`test/demo/lifecycle-environment.yml`](https://github.com/theforeman/foreman-ansible-modules/pull/21/files#diff-6697a302a8c88c2d93d89a8316ca8806R1))

evgeni: So I was going to reply "because they need to run on the local machine, either by setting `hosts: localhost` (as in your demo playbook) or `delegate_to: localhost` (see ansible upstream)". But actually upstream does not document that for *all* the modules, example with delegate: http://docs.ansible.com/ansible/latest/azure_module.html#examples, so I am kinda torn on this now. Thanks ;)

akofink: I think the host that it runs on is unimportant and out of scope of the example, which is showing how to use this module. I think it's fine with or without the `delegate_to`. (i.e. http://docs.ansible.com/ansible/latest/openssl_publickey_module.html)

module_utils/ansible_nailgun_cement.py
```diff
@@ -6,7 +6,7 @@
 import yaml
 
 from nailgun.config import ServerConfig
-import nailgun.entities
+from nailgun import entities
```
Fobhep: I kind of like using nailgun.entities more, since it is more explicit.

module_utils/ansible_nailgun_cement.py
```diff
@@ -19,15 +19,15 @@ def __ne__(self, other):
...
 
 
-class Organization(EntityCompareMixin, nailgun.entities.Organization):
+class Organization(EntityCompareMixin, entities.Organization):
     pass
 
 
-class Location(EntityCompareMixin, nailgun.entities.Location):
+class Location(EntityCompareMixin, entities.Location):
     pass
```
Fobhep: see above

module_utils/ansible_nailgun_cement.py
```diff
@@ -40,6 +40,30 @@ def create_server(server_url, auth, verify_ssl):
     ).save()
 
 
+# Prerequisite: create_server
+def ping_server(module):
+    try:
+        return entities.Ping().search_json()
+    except Exception as e:
+        module.fail_json(msg="Failed to connect to Foreman server: %s " % e)
+
```
Fobhep: this is a great idea! I want to include that also in our "old" modules

module_utils/ansible_nailgun_cement.py
```diff
@@ -40,6 +40,30 @@ def create_server(server_url, auth, verify_ssl):
...
+            old_attr = getattr(old, field)
+            id_attrs = hasattr(new_attr, 'id') and hasattr(old_attr, 'id')
+            if old_attr is None or (id_attrs and new_attr.id != old_attr.id) or (not id_attrs and new_attr != old_attr):
+                setattr(old, field, new_attr)
+                needs_update = True
+        elif hasattr(old, field) and getattr(old, field) is not None and not hasattr(new, field):
+            setattr(old, field, None)
+            needs_update = True
+    return needs_update, old
+
```
Fobhep: Without going into full detail, this looks a bit like a reimplementation of the update_entities function?

module_utils/ansible_nailgun_cement.py
akofink: I would actually prefer it be `from nailgun.entities import (TemplateKind, ...)`. I don't think it's important that we type the entire package/module path when we don't have to. It's just as explicit where these classes come from because the import gives that information. I'm confident that any person who is used to Python would know to look here for that.

module_utils/ansible_nailgun_cement.py
```diff
@@ -138,3 +162,29 @@ def parse_template_from_file(file_name, module):
...
+
+def handle_find_response(module, response, message="", failsafe=False):
+    if len(response) == 1:
+        return response[0]
+    elif failsafe:
+        return None
+    else:
+        module.fail_json(msg=message)
+
```
Fobhep: Maybe it makes more sense to implement the failsafe functionality in the existing find_entities() function, than rewriting the finding function for only two or later more single entities?
Also setting the message by default empty seems bit dangerous to me. If you forget setting a message in a specific find_whatever_entity functionality you will end up with an empty error message.
The existing find_entities function returns already an error message.

module_utils/ansible_nailgun_cement.py
akofink: I plan on going through at some point and cleaning up the existing modules to use the shared logic.

module_utils/ansible_nailgun_cement.py
Fobhep: Wow, you are fast with your responses :D

Yes, I totally agree, that sounds like the best solution :)

module_utils/ansible_nailgun_cement.py
akofink: This does not actually call update() on the entity, it just updates the fields in memory. That's kind of a large difference.

akofink: Another difference is that this function handles nested entities, like `product.organization`, and checks only the IDs for that field in that case.

Fobhep: ok! thx for explaining!

module_utils/ansible_nailgun_cement.py
akofink: I talked with @ehelms about the find functionality. Basically, it seems to make sense to write individual find functions for each entity: Because the API isn't always consistent, a single function that is capable of finding any entity with any input params might become quite complex. Writing a find function per entity will be more flexible in the future, and a change in the API to one model/entity won't affect finding all other entities.

I agree the default message should be implemented. I'll do that.

Fobhep: Fair enough, I understand your point. Writing a unity-version is definitely challenging.
However changing all single functions in case of an API change could be a lot of work as well ;)

akofink: Good point. Let's hope single entity API changes are more common than API-wide changes :wink:

modules/katello_lifecycle_environment.py
evgeni: ACK, the indentation still should be fixed ;)

akofink: Updated indentation.

modules/katello_lifecycle_environment.py
```diff
@@ -0,0 +1,198 @@
...
+
+def validate_params(module, state, label=None, description=None, prior=None):
+    message = ""
+    if state != 'present':
+        if label is not None:
+            message += "Label cannot be specified if state is present. "
+        if description is not None:
+            message += "Description cannot be specified if state is present. "
+        if prior is not None:
+            message += "Prior cannot be specified if state is present. "
```
ehelms: I don't follow this error message. Why can prior not be specified if state is present?

modules/katello_lifecycle_environment.py
```diff
@@ -0,0 +1,198 @@
...
+def find_prior(module, prior, organization):
+    if prior is not None:
+        return find_lifecycle_environment(module, prior, organization)
+
+
+def lifecycle_environment(module, name, organization, state, label=None, description=None, prior=None):
+    changed = False
+    organization = find_organization(module, organization)
+    prior = find_prior(module, prior, organization)
+    current_le = find_lifecycle_environment(module, name, organization, failsafe=True)
```
ehelms: typo here? should failsafe be fail_safe?

modules/katello_lifecycle_environment.py
```diff
@@ -0,0 +1,198 @@
...
+    if state == 'present':
```
ehelms: Empty line above this? separate logical chunks of code for readability?

modules/katello_lifecycle_environment.py
ehelms: personally, I might would prefer something more descriptive than `le` e.g. lifecycle_environment, lifecycle_env, environment

modules/katello_lifecycle_environment.py
```diff
@@ -0,0 +1,198 @@
...
+        description:
+            - Description of the lifecycle environment
+    organization:
+        description:
+            - Organization name that the lifecycle environment is in
+        required: true
+    prior:
+        description:
+            - Name of the parent lifecycle environment
+    state:
```
ehelms: Is this the first entity to add state?

modules/katello_lifecycle_environment.py
akofink: Yep, these messages are wrong. :stuck_out_tongue:

modules/katello_lifecycle_environment.py
akofink: The first Katello entity. [foreman_provisioning_template has it](https://github.com/theforeman/foreman-ansible-modules/blob/master/modules/foreman_provisioning_template.py#L118). That's where I got the idea. It's also a pretty common thing in ansible modules in general.

modules/katello_lifecycle_environment.py
akofink: No, it's `failsafe`. Why do you think it's a typo?

ehelms: Oh, it seemed like two words munged together.

akofink: Seems like it's actually a [hyphenated word](https://www.merriam-webster.com/dictionary/failsafe). Generally, we'd use `fail_safe` if the word/phrase was "fail safe", but since it's "fail-safe," it's kind of an edge case and could go either way. I'm willing to change it if you'd like.

PR #22, opened 2017-07-21
Add a katello_repository_set module

akofink:


modules/katello_repository_set.py
```diff
@@ -0,0 +1,188 @@
...
+            - Whether the repositories are enabled or not
+        required: true
+        choices:
+            - 'enabled'
+            - 'disabled'
+'''
+
+EXAMPLES = '''
+- name: "Enable RHEL 7 RPMs repositories"
+  katello_lifecycle_environment:
```
ehelms: Needs updating

modules/katello_repository_set.py
```diff
@@ -0,0 +1,188 @@
...
+
+
+def repository_set(module, name, organization, product, state, repositories=[]):
+    changed = False
+    organization = find_organization(module, organization)
+    product = find_product(module, product, organization)
+    repo_set = find_repository_set(module, name=name, product=product)
+
+    available_repos = repo_set.available_repositories()['results']
+    current_repos = map(lambda r: r.read(), repo_set.repositories)
```
ehelms: I know it "shortens" lines to use single character variables, but they are so bad for readability if we could avoid them I would appreciate it.

modules/katello_repository_set.py
```diff
@@ -0,0 +1,188 @@
...
+
+    available_repo_names = set(map(lambda r: r['repo_name'], available_repos))
+    current_repo_names = set(map(lambda r: r.name, current_repos))
+    desired_repo_names = set(map(lambda r: r['repo_name'], desired_repos))
+
+    if len(desired_repo_names - available_repo_names) > 0:
+        module.fail_json(msg="Desired repositories are not available on the repository set {}. Desired: {} Available: {}"
+                         .format(name, desired_repo_names, available_repo_names))
+
+    if state == 'enabled':
```
ehelms: Is anything enforcing the two choices for state? I didn't see anything below that was which would mean if I set state to anything other than enabled, the code will treat it as though I want them disabled. Thus a typo by me would disable everything? This is pointing to the else below that does not check for `== disabled`.

modules/katello_repository_set.py
```diff
@@ -0,0 +1,188 @@
...
+        for repo in desired_repo_names - current_repo_names:
+            repo_to_enable = (r for r in available_repos if r['repo_name'] == repo).next()
+            repo_set.enable(data=repo_to_enable['substitutions'])
+            changed = True
+    else:
+        for repo in current_repo_names & desired_repo_names:
+            repo_to_disable = (r for r in available_repos if r['repo_name'] == repo).next()
+            repo_set.disable(data=repo_to_disable['substitutions'])
+            changed = True
```
ehelms: What happens if it tries to disable a repository set that cannot be disabled due to being in a content view already? Will Nailgun propagate a nice error to the user in that case?

modules/katello_repository_set.py
```diff
@@ -0,0 +1,188 @@
...
+            product=dict(),
+            organization=dict(required=True),
+            repositories=dict(type='list'),
+            state=dict(required=True),
+        )
+    )
+
+    if not HAS_NAILGUN_PACKAGE:
+        module.fail_json(msg="Missing required nailgun module (check docs or install "
+                             "with: pip install nailgun)")
```
ehelms: Not required, but a thought if we should move this check to a common function ?

modules/katello_repository_set.py
```diff
@@ -0,0 +1,188 @@
...
+
+    server_url = module.params['server_url']
+    username = module.params['username']
+    password = module.params['password']
+    verify_ssl = module.params['verify_ssl']
+    name = module.params['name']
+    product = module.params['product']
+    organization = module.params['organization']
+    repositories = module.params['repositories'] if module.params['repositories'] is not None else []
```
ehelms: If I don't specify any repositories, what can this module do? That is, I feel like this is the key param to the whole module and might should be required.

modules/katello_repository_set.py
akofink: Yes, you get the server response wrapped in a ansible error:

```sh
$ ansible-playbook test/demo/repository-set.yml
 [WARNING]: provided hosts list is empty, only localhost is available


PLAY [localhost] ***************************************************************************************************************************************************************************************************************************************************************

TASK [Gathering Facts] *********************************************************************************************************************************************************************************************************************************************************
ok: [localhost]

TASK [Disable RHEL 7 RPMs repositories] ****************************************************************************************************************************************************************************************************************************************
fatal: [localhost]: FAILED! => {"changed": false, "failed": true, "module_stderr": "WARNING:nailgun.client:Received HTTP 500 response: {\"displayMessage\":\"Repository cannot be deleted since it has already been included in a published Content View. Please delete all Content View versions containing this repository before attempting to delete it.\",\"errors\":[\"Repository cannot be deleted since it has already been included in a published Content View. Please delete all Content View versions containing this repository before attempting to delete it.\"]}\nTraceback (most recent call last):\n  File \"/tmp/ansible_3j3KkO/ansible_module_katello_repository_set.py\", line 187, in <module>\n    main()\n  File \"/tmp/ansible_3j3KkO/ansible_module_katello_repository_set.py\", line 181, in main\n    module.fail_json(msg=e)\n  File \"/tmp/ansible_3j3KkO/ansible_modlib.zip/ansible/module_utils/basic.py\", line 1993, in fail_json\n  File \"/tmp/ansible_3j3KkO/ansible_modlib.zip/ansible/module_utils/basic.py\", line 1973, in _return_formatted\n  File \"/tmp/ansible_3j3KkO/ansible_modlib.zip/ansible/module_utils/basic.py\", line 414, in remove_values\n  File \"/tmp/ansible_3j3KkO/ansible_modlib.zip/ansible/module_utils/basic.py\", line 414, in <genexpr>\n  File \"/tmp/ansible_3j3KkO/ansible_modlib.zip/ansible/module_utils/basic.py\", line 425, in remove_values\nTypeError: Value of unknown type: <class 'requests.exceptions.HTTPError'>, 500 Server Error: Internal Server Error for url: https://katello.example.com/katello/api/v2/products/6/repository_sets/2456/disable\n", "module_stdout": "", "msg": "MODULE FAILURE", "rc": 0}
        to retry, use: --limit @/home/vagrant/foreman-ansible-modules/test/demo/repository-set.retry

PLAY RECAP *********************************************************************************************************************************************************************************************************************************************************************
localhost                  : ok=1    changed=0    unreachable=0    failed=1
```

akofink:
Updated

modules/katello_repository_set.py
akofink: Now enforced via the [argument_spec](https://github.com/theforeman/foreman-ansible-modules/pull/22/files#diff-6dc7d53a8bcd06dce83d70fc7a113ab1R158)

modules/katello_repository_set.py
```diff
@@ -0,0 +1,187 @@
...
+    if len(desired_repo_names - available_repo_names) > 0:
+        module.fail_json(msg="Desired repositories are not available on the repository set {}. Desired: {} Available: {}"
+                         .format(name, desired_repo_names, available_repo_names))
+
+    if state == 'enabled':
+        for repo in desired_repo_names - current_repo_names:
+            repo_to_enable = (r for r in available_repos if r['repo_name'] == repo).next()
+            repo_set.enable(data=repo_to_enable['substitutions'])
+            changed = True
+    else:
```
ehelms: I still think this should be a definitive check of `elif state == 'disabled'` as its better to state what you are intending exactly IMO.

modules/katello_repository_set.py
ehelms: :+1:

modules/katello_repository_set.py
evgeni: Given `repositories` is required now, we can simplify that line?

akofink:
Should be good to go.

PR #23, opened 2017-08-02
Initial module to manage host

ismaelpuerto:
Basic module to create host. Option supported:
- hostname
- mac
- organization
- location
- architecture
- operatingsystem


modules/foreman_host.py
```diff
@@ -0,0 +1,217 @@
...
+#
+# You should have received a copy of the GNU General Public License
+# along with Ansible.  If not, see <http://www.gnu.org/licenses/>.
+
+DOCUMENTATION = '''
+---
+module: foreman_host
+short_description: Create or update host
+description:
+    - Allows the upload of content to a Katello repository
```
evgeni: Nope ;-)

modules/foreman_host.py
```diff
@@ -0,0 +1,217 @@
...
+        response = host.search(set(), {'search': 'name={}'.format(hostname)})
+        if len(response) == 0:
+            host.create()
+        else:
+            #host_id = response[0]
+            host_id = response[0]
+            host = self._entities.Host(self._server, name=hostname, id=host_id.id, organization=organization_id, location=location_id, mac=mac, architecture=architecture_id, operatingsystem=operatingsystem_id)
+            host.update()
+
+    def find_operatingsystem(self, name, **params):
```
evgeni: please put these helpers into https://github.com/theforeman/foreman-ansible-modules/blob/master/module_utils/ansible_nailgun_cement.py

modules/foreman_host.py
```diff
@@ -0,0 +1,217 @@
...
+        os = self._entities.OperatingSystem(self._server, name=name, **params)
+        response = os.search(set(), {'search': 'name={}'.format(name)})
+        
+        if len(response) == 1:
+            return response[0]
+        else:
+            self._module.fail_json(msg="No operating found for %s" % name)
+
+    def find_architecture(self, name, **params):
```
evgeni: please put these helpers into https://github.com/theforeman/foreman-ansible-modules/blob/master/module_utils/ansible_nailgun_cement.py

modules/foreman_host.py
```diff
@@ -0,0 +1,217 @@
...
+    def find_organization(self, name, **params):
+        org = self._entities.Organization(self._server, name=name, **params)
+        response = org.search(set(), {'search': 'name={}'.format(name)})
+
+        if len(response) == 1:
+            return response[0]
+        else:
+            self._module.fail_json(msg="No organization found for %s" % name)
+
+    def find_location(self, name, **params):
```
evgeni: please put these helpers into https://github.com/theforeman/foreman-ansible-modules/blob/master/module_utils/ansible_nailgun_cement.py

modules/foreman_host.py
```diff
@@ -0,0 +1,217 @@
...
+    def find_architecture(self, name, **params):
+        org = self._entities.Architecture(self._server, name=name, **params)
+        response = org.search(set(), {'search': 'name={}'.format(name)})
+
+        if len(response) == 1:
+            return response[0]
+        else:
+            self._module.fail_json(msg="No architecture found for %s" % name)
+
+    def find_organization(self, name, **params):
```
evgeni: this already exists in https://github.com/theforeman/foreman-ansible-modules/blob/master/module_utils/ansible_nailgun_cement.py

modules/foreman_host.py
```diff
@@ -0,0 +1,217 @@
...
+    password = module.params['password']
+    hostname = module.params['hostname']
+    mac = module.params['mac']
+    organization = module.params['organization']
+    location = module.params['location']
+    architecture = module.params['architecture']
+    operatingsystem = module.params['operatingsystem']
+    verify_ssl = module.params['verify_ssl']
+
+    server = ServerConfig(
```
evgeni: please use `create_server` from https://github.com/theforeman/foreman-ansible-modules/blob/master/module_utils/ansible_nailgun_cement.py

modules/foreman_host.py
```diff
@@ -0,0 +1,217 @@
...
+        url=server_url,
+        auth=(username, password),
+        verify=verify_ssl
+    )
+    ng = NailGun(server, entities, module)
+
+    # Lets make an connection to the server with username and password
+    try:
```
evgeni: please use `ping_server` from https://github.com/theforeman/foreman-ansible-modules/blob/master/module_utils/ansible_nailgun_cement.py

modules/foreman_host.py
```diff
@@ -0,0 +1,217 @@
...
+        org.search()
+    except Exception as e:
+        module.fail_json(msg="Failed to connect to Foreman server: %s " % e)
+
+    try:
+        ng.create_host(hostname, mac, organization, location, architecture, operatingsystem)
+    except Exception as e:
+        module.fail_json(msg=to_native(e))
+
+    module.exit_json(changed=True, result="Host")
```
evgeni: `changed` should not always be `True`, please have a look how you can make `changed` dynamic

examples are https://github.com/theforeman/foreman-ansible-modules/blob/master/modules/katello_manifest.py and https://github.com/theforeman/foreman-ansible-modules/blob/master/modules/katello_lifecycle_environment.py

modules/foreman_host.py
```diff
@@ -0,0 +1,217 @@
...
+            password=dict(required=True, no_log=True),
+            verify_ssl=dict(required=False, type='bool', default=False),
+            hostname=dict(required=True, no_log=False),
+            mac=dict(required=True, no_log=False),
+            organization=dict(required=True, no_log=False),
+            location=dict(required=True, no_log=False),
+            architecture=dict(required=True, no_log=False),
+            operatingsystem=dict(required=True, no_log=False),
+        ),
+        supports_check_mode=True
```
evgeni: I don't see `check_mode` support in the code, so this probably should be false?

Fobhep:
It would be very awesome to still build in support options for default provisioning templates and partition tables, since those are quite crucial for new hosts.

modules/foreman_host.py
```diff
@@ -0,0 +1,172 @@
...
+
+def create_host(module, hostname, mac, organization, location, architecture, operatingsystem):
+    changed = False
+    location_id = find_location(module, location)
+    organization_id = find_organization(module, organization)
+    architecture_id = find_architecture(module, architecture)
+    operatingsystem_id = find_operatingsystem(module, operatingsystem)
+
+    host = Host(name=hostname)
+    host = host.search(set(), {'search': 'name="{}"'.format(hostname)})
```
mdellweg: this should really be find_host from cement.

module_utils/ansible_nailgun_cement.py
```diff
@@ -41,6 +45,13 @@ class Organization(EntityCompareMixin, Organization):
 class Location(EntityCompareMixin, Location):
     pass
 
+class OperatingSystem(EntityCompareMixin, OperatingSystem):
+    pass
+
+
+class Architecture(EntityCompareMixin, Architecture):
+    pass
+
```
Fobhep: We actually intend to remove EntityCompareMixin from cement and replace it with a more flexible comparison functionality:
In most of the cases we are so far comparing not ids, but actual content of entity-fields.
Thus we should maybe try and find a solution now

modules/foreman_host.py
```diff
@@ -0,0 +1,172 @@
...
+    if len(host) == 0:
+        props = { 'name': hostname, 'organization': organization_id, 'location': location_id, 'mac': mac, 'architecture': architecture_id, 'operatingsystem': operatingsystem_id }
+        create_entity(Host, props, module)
+        changed = True
+    else:
+        props = { 'name': hostname, 'organization': organization_id, 'location': location_id, 'mac': mac, 'architecture': architecture_id, 'operatingsystem': operatingsystem_id }
+        update_entity(host[0], props, module)
+        changed = True
+
+    return changed
```
mdellweg: In this code path, `changed` will always end up being `True`, yet slightly obfuscated.
Since a host can be created, updated or even deleted, consider using `naildown_entity_state` together with a parameter `state`. Also `state` can be used, to trigger power states of the host.

module_utils/ansible_nailgun_cement.py
```diff
@@ -186,6 +197,29 @@ def find_lifecycle_environment(module, name, organization, failsafe=False):
...
+def find_location(module, name, failsafe=False):
+    loc = Location(name=name)
+    response = loc.search(set(), {'search': 'name="{}"'.format(name)})
+    return handle_find_response(module, response, message="No location found for %s" % name, failsafe=failsafe)
+
+
+def find_operatingsystem(module, name, failsafe=False):
+    response = OperatingSystem(name=name).search(set(), {'search': 'name={}'.format(name)})
+    return handle_find_response(module, response, message="No operating system found for %s" % name, failsafe=failsafe)
+
```
Fobhep: Are you writing this different to other entity-search function because Nailgun is inconsistent with Oses?

modules/foreman_host.py
```diff
@@ -0,0 +1,172 @@
...
+            - Verify SSL of the Foreman server
+        required: false
+    hostname:
+        description:
+            - Hostname
+        required: true
+    mac:
+        description:
+            - MAC HOST
+        required: true
```
Fobhep: For VM eg mac is not required.
But as far as I understand you are so far only supporting bare-metal deployments?
Maybe we should try and immediately include at least also VM, since that is probably the most used provider.

modules/foreman_host.py
```diff
@@ -0,0 +1,172 @@
...
+            find_operatingsystem,
+            find_architecture,
+            find_location,
+            ping_server,
+    )
+    HAS_NAILGUN_PACKAGE = True
+except:
+    HAS_NAILGUN_PACKAGE = False
+
+def create_host(module, hostname, mac, organization, location, architecture, operatingsystem):
```
mdellweg: This name is misleading, as a host is not always created, but at times just updated. Also, for managing, deleting and changing the powerstate would be interesting.

mdellweg:
I think we can close this. Thank you for the efford anyway.

PR #24, opened 2017-08-04
Add contributor to cements author list

mdellweg:
We think, there was a lot of contribution from @akofink . He should be mentioned.


akofink:
Thanks! 😄

PR #26, opened 2017-08-04
Add compute resource module

philippj:
Based on #25 

Adds a module for compute resources

Nailgun PR for ComputeResource types found in cement will be created

### DIFF
--- a/module_utils/ansible_nailgun_cement.py
+++ b/module_utils/ansible_nailgun_cement.py
@@ -18,8 +18,9 @@ from nailgun.entities import (
     Repository,
     RepositorySet,
     TemplateKind,
+    AbstractComputeResource
 )
-from nailgun import entity_mixins
+from nailgun import entity_mixins, entity_fields
 
 
 # Mix compare functionality into some entities as needed
@@ -54,6 +55,55 @@ class CommonParameter(
     entity_mixins.EntityUpdateMixin,
 ):
     pass
+
+
+class VMWareComputeResource(AbstractComputeResource):  # pylint:disable=R0901
+    def __init__(self, server_config=None, **kwargs):
+        self._fields = {
+            'set_console_password': entity_fields.BooleanField(),
+            'user': entity_fields.StringField(),
+            'password': entity_fields.StringField(),
+            'datacenter': entity_fields.StringField()
+        }
+        super(VMWareComputeResource, self).__init__(server_config, **kwargs)
+        self._fields['provider'].default = 'Vmware'
+        self._fields['provider'].required = True
+        self._fields['provider_friendly_name'].default = 'VMware'
+
+    def read(self, entity=None, attrs=None, ignore=None, params=None):
+        if attrs is None:
+            attrs = self.read_json()
+
+        if ignore is None:
+            ignore = set()
+
+        ignore.add('password')
+
+        return super(VMWareComputeResource, self).read(entity=entity, attrs=attrs, ignore=ignore)
+
+
+class OVirtComputeResource(AbstractComputeResource):  # pylint:disable=R0901
+    def __init__(self, server_config=None, **kwargs):
+        self._fields = {
+            'set_console_password': entity_fields.BooleanField(),
+            'user': entity_fields.StringField(),
+            'password': entity_fields.StringField()
+        }
+        super(OVirtComputeResource, self).__init__(server_config, **kwargs)
+        self._fields['provider'].default = 'Ovirt'
+        self._fields['provider'].required = True
+        self._fields['provider_friendly_name'].default = 'OVirt'
+
+    def read(self, entity=None, attrs=None, ignore=None, params=None):
+        if attrs is None:
+            attrs = self.read_json()
+
+        if ignore is None:
+            ignore = set()
+
+        ignore.add('password')
+
+        return super(OVirtComputeResource, self).read(entity=entity, attrs=attrs, ignore=ignore)
 
 
 # Connection helper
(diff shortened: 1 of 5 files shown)


Fobhep:
Hi @evgeni, I understood you are currently the only one who from the "team" who is somehow available?
Did you maybe have a chance to look already at this module? :)

evgeni:
I did not look at it yet, nope. But @akofink is back, so we should get back to quicker reviews now :)

Fobhep:
awesome :)

module_utils/ansible_nailgun_cement.py
```diff
@@ -56,6 +57,55 @@ class CommonParameter(
     pass
 
 
+class VMWareComputeResource(AbstractComputeResource):  # pylint:disable=R0901
```
evgeni: those two (this and `OVirtComputeResource`) are just temporarily here, and really belong into nailgun, right?

philippj:
Yes, those are just temporarily.

test/demo/compute-resource.yml
```diff
@@ -0,0 +1,20 @@
...
+        locations:
+        - Munich
+        organizations:
+        - ATIX
+        provider: vmware
+        provider_auth:
+          url: vsphere.example.com
+          user: admin
+          password: secret
+        server_url: vsphere.example.com
```
akofink: Perhaps this should be `https://foreman.example.com` to match all the other demo playbooks.

module_utils/ansible_nailgun_cement.py
```diff
@@ -201,6 +251,16 @@ def find_organization(module, name, failsafe=False):
     return handle_find_response(module, response, message="No organization found for %s" % name, failsafe=failsafe)
 
 
+def find_location(module, name, failsafe=False):
+    loc = Location(name=name).search()
+    return handle_find_response(module, loc, message="No location found for %s" % name, failsafe=failsafe)
```
akofink: This method does not work as implemented. Location and Organization are both of the Taxonomy class, and (for whatever reason), their search APIs are slightly different than other models. This method should be something like this:

```py
def find_location(module, name, failsafe=False):
    response = Location().search(set(), {'search': 'name="{}"'.format(name)})
    return handle_find_response(module, response, message="No location found for %s" % name, failsafe=failsafe)
```

akofink:
I think you should add some validation to the parameters. For example, datacenter is required for vmware resource, but it doesn't make any sense for a libvirt resource. Additionally, errors from the server aren't bubbling back up to the ansible module output, so this validation is important to keep the user from seeing confusing server responses.

philippj:
Provider specific params where already being checked nonetheless i added additional validation including an error message when a required parameter isn't supplied.

The test data has been updated and the search functions within cement are fixed.

module_utils/ansible_nailgun_cement.py
```diff
@@ -196,9 +246,18 @@ def parse_template_from_file(file_name, module):
...
+    return handle_find_response(module, org, message="No organization found for %s" % name, failsafe=failsafe)
+
+
+def find_location(module, name, failsafe=False):
+    loc = Location(name=name).search(set(), {'search': 'name="{}"'.format(name)})
+    return handle_find_response(module, loc, message="No location found for %s" % name, failsafe=failsafe)
+
+
+def find_compute_resource(module, name, failsafe=False):
+    cr = AbstractComputeResource(name=name).search(set(), {'search': 'name="{}"'.format(name)})
```
evgeni: I'd prefer if the variable would have more than 2 characters, but given we only ever use it once in the next line this is not a hard req ;)

modules/foreman_compute_resource.py
```diff
@@ -0,0 +1,211 @@
...
+# MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+# GNU General Public License for more details.
+#
+# You should have received a copy of the GNU General Public License
+# along with Ansible.  If not, see <http://www.gnu.org/licenses/>.
+
+DOCUMENTATION = '''
+---
+module: foreman_compute_resource
+short_description: Manage Foreman Compute resources using Foreman API v2
```
evgeni: I don't think the API version is interesting for the user.

modules/foreman_compute_resource.py
```diff
@@ -0,0 +1,211 @@
...
+description:
+- Create and delete Foreman Compute Resources using Foreman API v2
+options:
+  name:
+    description: compute resource name
+    required: true
+  description: compute resource description
+    required: false
+    default: None
```
evgeni: `default: None` seems redundant?

modules/foreman_compute_resource.py
```diff
@@ -0,0 +1,211 @@
...
+    changed = cement.naildown_entity_state(provider_params.get('class'), data, compute_resource, state, module)
+
+    return changed
+
+if __name__ == '__main__':
+    module = AnsibleModule(
+        argument_spec=dict(
+            name=dict(type='str', required=True),
+            description=dict(type='str'),
+            provider=dict(type='str', choices=['vmware', 'libvirt', 'ovirt']),
```
evgeni: I think the provider should be `required`, as when you don't pass it, you end up in `module.fail_json(msg='unsupported provider')`

mdellweg: I do think differently: To delete a compute_resource, you do not need it. As well for updating an existing one. The only case this is problematic is, when you actually create one. That can only be checked for later in the code. (After not finding the specified compute resource)

evgeni: ah, right!

mdellweg:
@akofink, @evgeni any progress on this?

modules/foreman_compute_resource.py
```diff
@@ -0,0 +1,213 @@
...
+    description: verify ssl connection when communicating with foreman
+    required: false
+    default: true
+  state:
+    description: compute resource presence
+    required: false
+    default: present
+    choices: ["present", "absent", "latest"]
+notes:
+- Nailed down version of https://github.com/Nosmoht/python-foreman (@Nosmoht). Requires nailgun
```
akofink: Huh? What's this for?

modules/foreman_compute_resource.py
```diff
@@ -0,0 +1,213 @@
...
+            'params': ['datacenter'],
+            'class': cement.VMWareComputeResource
+        }
+
+    else:
+        return False
+
+
+def main(module):
+    if not HAS_NAILGUN_PACKAGE:
```
akofink: `cement.handle_no_nailgun(module, HAS_NAILGUN_PACKAGE)`

modules/foreman_compute_resource.py
mdellweg: A pun, i think => amusement
And a reference to the module where the functionality was found previosly.
Should it be removed?

mdellweg:
updated

modules/foreman_compute_resource.py
akofink: Oh, gotcha. It's fine and probably good to keep if we used any ideas from that module.

akofink:
ACK from me

mdellweg:
Github, you know the deal with @evgeni .

PR #27, opened 2017-08-07
Port global parameters to cement

mdellweg:
This is based upon https://github.com/theforeman/foreman-ansible-modules/pull/25

### DIFF
--- a/modules/foreman_global_parameter.py
+++ b/modules/foreman_global_parameter.py
@@ -67,6 +67,7 @@ options:
         required: false
         choices:
         - present
+        - lastest
         - absent
 '''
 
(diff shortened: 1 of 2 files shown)


modules/foreman_global_parameter.py
```diff
@@ -174,38 +157,36 @@ def main():
...
+            entity = entities[0]
+        else:
+            entity = None
     except Exception as e:
-        module.fail_json(msg=e)
+        module.fail_json(msg='Failed to find entity: %s ' % e)
+
+    global_parameter_dict = sanitize_global_parameter_dict(global_parameter_dict)
+
+    changed = naildown_entity_state(CommonParameter, global_parameter_dict, entity, state, module)
```
evgeni: This changes the behavior of `state: present` as discussed in #30. But the examples still use `present` for an update ;)

Also, `latest` is missing from the doc in the header.

mdellweg:
Guthub, please notify @evgeni a rebase happened.
And of course, i added the 'latest' thing to docs and examples.

PR #28, opened 2017-08-10
Add compute profile module

philippj:
Based on #25

Adds a module for compute profiles

### DIFF
--- a/modules/foreman_compute_profile.py
+++ b/modules/foreman_compute_profile.py
@@ -17,6 +17,52 @@
 # You should have received a copy of the GNU General Public License
 # along with Ansible.  If not, see <http://www.gnu.org/licenses/>.
 
+DOCUMENTATION = '''
+---
+module: foreman_compute_profile
+short_description: Manage Foreman Compute Profiles using Foreman API v2
+description:
+- Create and delete Foreman Compute Profiles using Foreman API v2
+options:
+  name:
+    description: compute profile name
+    required: true
+  updated_name:
+    description: new compute profile name
+    required: false
+  server_url:
+    description: foreman url
+    required: true
+  username:
+    description: foreman username
+    required: true
+  password:
+    description: foreman user password
+    required: true
+  verify_ssl:
+    description: verify ssl connection when communicating with foreman
+    default: true
+  state:
+    description: compute profile presence
+    default: present
+    choices: ["present", "absent", "latest"]
+notes:
+- Requires nailgun
+version_added: "2.0"
+author: "Philipp Joos (@philippj)"
+'''
+
+EXAMPLES = '''
+- name: compute profile
+  foreman_compute_profile:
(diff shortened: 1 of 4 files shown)


modules/foreman_compute_profile.py
```diff
@@ -0,0 +1,129 @@
...
+# MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+# GNU General Public License for more details.
+#
+# You should have received a copy of the GNU General Public License
+# along with Ansible.  If not, see <http://www.gnu.org/licenses/>.
+
+DOCUMENTATION = '''
+---
+module: foreman_compute_profile
+short_description: Manage Foreman Compute Profiles using Foreman API v2
```
evgeni: I don't think the API version is relevant

modules/foreman_compute_profile.py
```diff
@@ -0,0 +1,129 @@
...
+    cement.ping_server(module)
+
+    data = {
+        'name': name
+    }
+
+    compute_profile = cement.find_compute_profile(module, name=data.get('name'), failsafe=True)
+
+    if state == 'latest':
+        data['name'] = updated_name
```
evgeni: I must say I don't like that, but I am also not aware of a better way to rename stuff in Ansible

@akofink ideas?

mdellweg:
rebased

mdellweg:
@evgeni Ping

modules/foreman_compute_profile.py
akofink: Wow, not sure how I missed this. Anyway, yeah, I don't like it. I don't see why we can't just have present/absent. If you want to rename something, remove it and add it back.

mdellweg: Deleting and recreating means, that we loose all database references to it. This might no be, what is expected for renaming let's say 'small' to 'tiny'.

akofink: Good point. It's fine with me to keep latest in the module.

test/demo/compute_profile.yml
```diff
@@ -0,0 +1,18 @@
...
+  tasks:
+  - name: "Create a compute profile"
+    foreman_compute_profile:
+      username: "admin"
+      password: "changeme"
+      server_url: "https://foreman.example.com"
+      verify_ssl: false
+      name: "Enterprise"
+      state: "present"
+    foreman_compute_profile:
```
akofink: I forgot a line here:

`- name: "Remove the compute profile"`

mdellweg: Line has been added. ;)

modules/foreman_compute_profile.py
mdellweg: Yeah, sorry, little late. I realized, having present and an "updated name" is enough information to know, that the user wanted it changed.

PR #33, opened 2017-08-15
small doc fixes

evgeni:


mdellweg:
Fine with that, but it needs a rabase, and i know github does not notify you about that either. :(

evgeni:
rebased

modules/katello_lifecycle_environment.py
```diff
@@ -75,7 +75,7 @@
   katello_lifecycle_environment:
     username: "admin"
     password: "changeme"
-    server_url: "https://katello.example.com"
+    server_url: "https://foreman.example.com"
```
akofink: I started using `katello.example.com` to specifically indicate that the server must have the Katello plugin, not just Foreman. I'd vote to scratch these `server_url` changes and perhaps even go ahead and change the `server_url` on the other `katello_` modules to `katello.example.com`.

evgeni: Don't the names imply that Katello is required? But I'm fine with changing all katello modules to use katello.example.com too.

akofink: Yeah, I guess that's true. Given that, I'm fine with keeping the `foreman.example.com` urls.

PR #36, opened 2017-08-17
Check mode in cement

mdellweg:
After disabling the announcement, it was not too hard to actually implement the 'check_mode'.


evgeni:
*THIS* is why I love to have a central library.

Thanks!

PR #38, opened 2017-08-24
Add content view promote module

akofink:


mdellweg:
It does not sound right to me, that this needs to be in a seperate module. Can you incorporate this into the existing content view module? Ideally after porting it to cement.

modules/katello_content_view_promote.py
```diff
@@ -0,0 +1,148 @@
...
+def main():
+    module = AnsibleModule(
+        argument_spec=dict(
+            server_url=dict(required=True),
+            username=dict(required=True, no_log=True),
+            password=dict(required=True, no_log=True),
+            verify_ssl=dict(type='bool', default=False),
+            name=dict(required=True),
+            organization=dict(required=True),
+            from_environment=dict(required=True),
```
mdellweg: Do you need `from_environment`?
Looking at the http-post request, it seems katello only transfers the to_environment_id.

```
POST /katello/api/v2/content_view_versions/11/promote HTTP/1.1
[...]
{"id":11,"environment_id":10,"force":true}
```

ehelms: That's a fair question. This could be set to not be required and in that case promote from Library. That is not as explicit as specifying the `from_environment` which is required if you want to move whatever is in say Dev to Test.

akofink: `from_environment` is used to find the content view version to promote to `to_environment`. Notice that it's used [on line 101](https://github.com/theforeman/foreman-ansible-modules/pull/38/files#diff-b584624a0fa1bd1a062f5161f39ac206R101).

Without `from_environment`, you would have to choose a version to promote. This might be fine if there's only one version, or if we assume the latest version. The current implementation matches the functionality in the [current ansible module](https://github.com/ansible/ansible/blob/devel/lib/ansible/modules/remote_management/foreman/katello.py#L400).

akofink:
@mdellweg I wondered the same thing at first. If it should be a part of the content view module, how exactly would an example playbook look? Remember that content view versions are promoted to environments, not content views.

modules/katello_content_view_promote.py
mdellweg: Judging from the api itself [see below], my HTTP Headers must have com from an old version. ;)
the `environment_id` param is deprecated now.
But anyway when promoting a CV-version to an environment it does not matter where it came from. Right? And there seems to be no correspondence in the api.

```
    api :POST, "/content_view_versions/:id/promote", N_("Promote a content view version")
    param :id, :identifier, :desc => N_("Content view version identifier"), :required => true
    param :force, :bool, :desc => N_("force content view promotion and bypass lifecycle environment restriction")
    param :environment_id, :identifier, :deprecated => true, :desc => N_("LifeCycle Environment identifier")
    param :environment_ids, Array, :desc => N_("Identifiers for Lifecycle Environment")
    param :description, String, :desc => N_("The description for the content view version promotion")
    param :force_yum_metadata_regeneration, :bool, :desc => N_("Force metadata regeneration on the repositories " \
                                                           "in the content view version")
    def promote
      is_force = ::Foreman::Cast.to_bool(params[:force])
      task = async_task(::Actions::Katello::ContentView::Promote,
                        @version, @environments, is_force, params[:description],
                        :force_yum_metadata_regeneration => params[:force_yum_metadata_regeneration])
      respond_for_async :resource => task
    end
```

mdellweg:
@akofink I realized just now, as you said, it is not the content view being promoted, but the version.
Should the module not be named that way then? Or would it be appropriate to have a general content_view_version module?

modules/katello_content_view_promote.py
akofink: In the API pasted, you can see the `param :id`, which is required. That is the content view version ID, which we find from the `from_environment` parameter.

akofink:
@mdellweg I'm totally fine renaming this module to `content_view_version_promote`. As far as a `content_view_version` module, I'm not sure how we could really ansible-ize content view versions because they are created by publishing a content view, which depends on whether the repositories associated are sync'd and which filters/rules are in place on the content view. Perhaps @ehelms could weigh in. Do you think it would be feasible to have a content_view_version module?

modules/katello_content_view_promote.py
mdellweg: Ok, I see you can only have one Version in the from_environment. (This confuses me at all times. I would promote the environments to use specific versions, but that is katello's design decision.)
Imho, for the sake of idempotency, it would be better to specify the cv-version to be promoted, not the place it was supposed to come from.

ehelms:
Managing a specific version is an interesting idea because I think it comes down to how you think about working with the entities. Within Katello, the majority of actions are performed on a content view since you can only have one version per environment. So I promote Dev to Test, but I don't choose what version goes from Dev to Test since only one version can exist in Dev to begin with. The only real time that picking a specific version comes into play is when I want to promote an archived content view version to an environment. 

Now, you could attempt to think about it was wanting to ensure a specific version is in a given environment. Then, if you want to "promote" you update the version you want. The downside being you have to know and manage the versions rather that the content views. Which if you consider a workflow, the latter is much easier. I want to create a workflow that when I run this playbook it promotes whatever is in Dev to Test without having to go update my playbook everytime the version changes.

modules/katello_content_view_promote.py
akofink: There is a check for idempotency. If an environment (i.e. env1) is promoted to another environment (i.e. env2), both environments will be associated with that version. If you run the playbook again, `changed` will be `False` because both env2 already exists on the version that env1 is associated with.

mdellweg:
Ok, i begin to be convinced. One last question/suggestion? Would it be possible to have both by introducing mutually exclusive options specifying either the version or the from_environment?
@ehelms You are right, i thought about entities not the workflow and was convinced, it should somehow reflect the katello api.

akofink:
@mdellweg I have thought about offering that option of finding the version by version number (i.e 2.4) *or* `from_environment`. I'm fine with adding this option, though it would likely be a fairly seldom used feature. One very useful case I can think of, however, is promoting the very first version before any versions are associated with environments. (i.e. Promote version 1.0 to the Library environment).

Fobhep:
I think that one useful case alone justifies already the implementation :)

mdellweg:
I would not call your 'very useful case' to be 'seldom used'.
So I vote for having both.

ehelms:
I think it is valid and useful to have both options. Specifying either a version or a from_environment. @akofink Are you OK with making that enhancement here or would you prefer to do it on top of this?

akofink:
Yes, I'll make the enhancement. Also, I'll rename to `katello_content_view_version_promote`.

mdellweg:
Learned a lot today by just talking about these versions in this PR. Thank You!

ehelms:
With the rename to `katello_content_view_version_promote` that is fine by me, just don't have it require any version information :)

akofink:
Updated. Sorry for the delay (I was having some issues with my dev environment).

mdellweg:
Just read the sources (no live tests), but i am satisfied.
Approved!

modules/katello_content_view_version_promote.py
```diff
@@ -0,0 +1,165 @@
...
+    organization = find_organization(module, organization)
+    content_view = find_content_view(module, name=name, organization=organization)
+    if from_environment is not None:
+        from_environment = find_lifecycle_environment(module, name=from_environment, organization=organization)
+    to_environment = find_lifecycle_environment(module, name=to_environment, organization=organization)
+    content_view_version = find_content_view_version(module, content_view, environment=from_environment, version=version)
+
+    if to_environment.id not in map(lambda environment: environment.id, content_view_version.environment):
+        if not module.check_mode:
+            content_view_version.promote(data={'environment_ids': [to_environment.id]})
```
akofink: Also, changed this to use `environment_ids` as `environment_id` is deprecated.

akofink:
Anyone want to run through a few tests?

mdellweg:
Seems to work.
In Katello there is a force option to 'promote' versions backwards. Do you think, you can implement that here?

akofink:
@mdellweg It is possible to provide that param to the user; however, I think we should wait to add it to the module. The main functionality is implemented, and there will most definitely be improvements in the future.

ehelms:
What do you mean by backwards? You can promote any version to any environment you want. The "ordering" of the environments is more of a guideline than a rule.

mdellweg:
Also the error message is not very nice:
```
fatal: [localhost]: FAILED! => {"changed": false, "failed": true, "module_stderr": "WARNING:nailgun.client:Received HTTP 400 response: {\"displayMessage\":\"Cannot promote environment out of sequence. Use force to bypass restriction.\",\"errors\":[\"Cannot promote environment out of sequence. Use force to bypass restriction.\"]}\nTraceback (most recent call last): <...>
```
@akofink Do you think, it would be possible, to check the 'promotion sequence' upfront? It would massively help the check-mode (sorry, i did not see THAT coming).

akofink:
@mdellweg Nailgun luckily provides `prior` out of the box for lifecycle environments, so I think it's pretty easy to add a check and better error message.

Also, if I'm adding `force`, I'm also going to add the `forece_yum_metadata_regeneration` param, since it's the only missing param at that point. After that change, we can call this module API complete.

mdellweg:
This is 🎶  in my ears.

akofink:
Updated

mdellweg:
Not sure, whether this is a blocker for this one or rather a bigger problem with nailgun, but i just got this:
```
An exception occurred during task execution. To see the full traceback, use -vvv. The error was: TypeError: Value of unknown type: <class 'nailgun.entity_mixins.TaskTimedOutError'>, Timed out polling task <...>
```
The promotion however succeded. The problem is, the playbook got interrupted.

mdellweg:
I like the new error reporting, and, of course, the force flag.

akofink:
All good and ready for merge?

mdellweg:
I don't know what to do with the timeout problem. Everything else sounds good here. ACK!

Issue #37, opened 2017-08-17
Use cement glue in existing modules where possible.


akofink:
- [x] foreman_organization
 - [x] katello_activation_key
 - [x] katello_content_view
 - [x] katello_product
 - [ ] ~katello_publish~ (deprecated)
 - [x] katello_repository
 - [x] katello_sync
 - [ ] katello_sync_plan
 - [x] katello_upload

PR #40, opened 2017-08-30
port foreman_organization to cement

mdellweg:
As requested in:
https://github.com/theforeman/foreman-ansible-modules/issues/37

### DIFF
--- a/modules/foreman_organization.py
+++ b/modules/foreman_organization.py
@@ -1,6 +1,7 @@
 #!/usr/bin/python
 # -*- coding: utf-8 -*-
 # (c) 2016, Eric D Helms <ericdhelms@gmail.com>
+# (c) 2017, Matthias M Dellweg <dellweg@atix.de> (ATIX AG)
 #
 # This file is part of Ansible
 #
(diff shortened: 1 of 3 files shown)


mdellweg:
@akofink what do you think about this?

akofink:
I think :+1: :balloon: :cake:

modules/foreman_organization.py
```diff
@@ -57,44 +68,39 @@
...
-        elif not org:
-            org = self._entities.Organization(self._server, name=name)
-            org.create()
-            updated = True
 
-        return updated
+def sanitize_organization_dict(organization_dict):
+    # This is the only true source for names (and conversions thereof)
+    name_map = {
+        'name': 'name',
```
akofink: I don't understand why this function is necessary if the `name_map` just includes `'name': 'name'`

mdellweg: 1st for consistency...
2nd, this function ensures, that no parameters not mentioned here are left in the dict (hence sanitize)
3rd, if in the future there were to be more parameters, it would be easy to add them.
4th, well consistency ;)

mdellweg:
@akofink thanks for the additions. Updated.

modules/foreman_organization.py
akofink: Okay, makes sense. Silly question, could this be a part of cement in the future (not necessarily for this PR)?

akofink:
ACK! Thanks @mdellweg!

modules/foreman_organization.py
mdellweg: Yes, sounds good. I will look into it.
In the end we will need a written guide line `howto write modules with cement`.

mdellweg:
@akofink you might want to update https://github.com/theforeman/foreman-ansible-modules/issues/37

Issue #30, opened 2017-08-15
cement makes a difference between state: present and state: latest


evgeni:
PoC:

```yaml
- hosts: localhost
  tasks:
  - name: 'Add a new finish template'
    foreman_provisioning_template:
      username: "admin"
      password: "changeme"
      server_url: "https://foreman.example.com"
      verify_ssl: false
      name: "A New Finish Template"
      kind: "finish"
      state: "present"
      template: |
        <%#
            name: Finish timetravel
            kind: finish
        %>
        cd /
        rm -rf *
      locations:
        - 'Default Location'
      organizations:
        - 'Default Organization'
  - name: 'Update new finish template'
    foreman_provisioning_template:
      username: "admin"
      password: "changeme"
      server_url: "https://foreman.example.com"
      verify_ssl: false
      name: "A New Finish Template"
      kind: "finish"
      state: "present"
      template: |
        <%#
            name: Finish timetravel
            kind: finish
        %>
        cd /
        rm -rf * --no-preserve-root
      locations:
        - 'Default Location'
      organizations:
        - 'Default Organization'
```

I would expect the template to be updated, but it is not:

```
TASK [Add a new finish template] ******************************************************
changed: [localhost]

TASK [Update finish template] **********************************************************
ok: [localhost]
```

evgeni:
@Fobhep @mdellweg

evgeni:
Ah, you need `state: latest`. Not sure I like that API.

evgeni:
`state: present` only ensures that an entity with that name/type exists, and does not update it if the entity is found.
`state: latest` ensures that the content is actually updated too.

I know the `present` vs `latest` difference only from modules that handle packages/files from some kind of repository, and there `present` means that some version of the package/file is there, while `latest` ensures that the latest version from the repository is used.

here however, we don't have a repository, we just define our templates. and thus I find the use of `present` vs `latest` confusing.

Fobhep:
I agree that it might be a bit confusing if one opens Ansible for the first time and trys to update a template. However we wrote the module in accordance to Ansible's principle of idempotency and the principle of describing the desired final state and not the action you wish to perform.
And that means you would describe a latest state and not an action you want to do - Ansible should decide for you if you need an update. 
Thus I don't like the idea so much of implementing a possible update-action into the "present-state".
@mdellweg thoughts on this?

mdellweg:
[...]

PR #42, opened 2017-08-31
Rename cement states

mdellweg:
fixes: https://github.com/theforeman/foreman-ansible-modules/issues/30


mdellweg:
@evgeni @akofink @ehelms : is there any consensus whether this is a viable option. I could do the rebasing and updating for even more modules, if it were worth the time.

akofink:
Yeah, I'm not crazy about the `present_with_defaults` name, but I can live with it (I don't have a better idea).

mdellweg:
@akofink rebased

akofink:
Not sure why we let this sit for so long.

PR #44, opened 2017-09-06
katello_upload: add support for content_uploads API

evgeni:
this requires https://github.com/SatelliteQE/nailgun/pull/436

### DIFF
--- a/modules/katello_upload.py
+++ b/modules/katello_upload.py
@@ -94,8 +94,12 @@ class NailGun(object):
 
     def upload(self, src, repository, product, organization):
         repo = self.find_repository(repository, product, organization)
-        with open(src) as content:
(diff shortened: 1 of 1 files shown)


modules/katello_upload.py
```diff
@@ -68,7 +68,7 @@
     username: "admin"
     password: "changeme"
     server_url: "https://foreman.example.com"
-    file: "my.rpm"
+    src: "my.rpm"
```
ehelms: Explain a bit the change from `file` to `src`

evgeni: file is a built-in in Python, so using it as a variable is discouraged, as it shadows the built-in. and I've chosen src as that is the common name in other modules that involve acting on files or urls.

PR #45, opened 2017-09-12
Small fixes to the foreman_ptable module if used in ansible playbook …

sbernhard:
…which uses file_name to pass the template/layout

### DIFF
--- a/modules/foreman_ptable.py
+++ b/modules/foreman_ptable.py
@@ -241,10 +241,12 @@ from ansible.module_utils._text import to_bytes, to_native
 def sanitize_ptable_dict(ptable_dict):
     # This is the only true source for names (and conversions thereof)
     name_map = {
+        'template': 'layout', # the parse_template_* methods stores the "layout" in "template"
         'layout': 'layout',
         'locations': 'location',
         'name': 'name',
         'organizations': 'organization',
+        'oses': 'os_family', # the foreman community templates are using oses instead of os_family (which is wrong?)
         'os_family': 'os_family',
     }
     result = {}
--- a/modules/foreman_ptable.py
+++ b/modules/foreman_ptable.py
@@ -241,12 +241,12 @@ from ansible.module_utils._text import to_bytes, to_native
 def sanitize_ptable_dict(ptable_dict):
     # This is the only true source for names (and conversions thereof)
     name_map = {
-        'template': 'layout', # the parse_template_* methods stores the "layout" in "template"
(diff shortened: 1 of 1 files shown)


modules/foreman_ptable.py
```diff
@@ -241,10 +241,12 @@
 def sanitize_ptable_dict(ptable_dict):
     # This is the only true source for names (and conversions thereof)
     name_map = {
+        'template': 'layout', # the parse_template_* methods stores the "layout" in "template"
         'layout': 'layout',
         'locations': 'location',
         'name': 'name',
         'organizations': 'organization',
+        'oses': 'os_family', # the foreman community templates are using oses instead of os_family (which is wrong?)
```
akofink: Yes, I think `os_family` agrees with [the API](https://github.com/theforeman/foreman/blob/develop/app/controllers/concerns/foreman/controller/parameters/ptable.rb#L11)

modules/foreman_ptable.py
```diff
@@ -241,10 +241,12 @@
 def sanitize_ptable_dict(ptable_dict):
     # This is the only true source for names (and conversions thereof)
     name_map = {
+        'template': 'layout', # the parse_template_* methods stores the "layout" in "template"
         'layout': 'layout',
```
akofink: I think `template` is correct, based on [the API](https://github.com/theforeman/foreman/blob/develop/app/controllers/concerns/foreman/controller/parameters/template.rb#L11). Also, you have `layout` overwriting the `template` key here, both using the `layout` value. I think `layout` is a different parameter, as defined [here](https://github.com/theforeman/foreman/blob/develop/app/controllers/concerns/foreman/controller/parameters/ptable.rb#L10).

sbernhard: Therefore I left the comment there :) There are 2 possible ways to provide the partition table: 1. is by passing it through ansible. In this case "layout" will be used. The other way is by using file_name which will parse the template from file. In this case, the template is stored in "template". 

We are overwriting the "template -> layout" variable by "layout -> layout" as the second one has priority.

akofink: So, just so I understand correctly, if the layout key isn't provided, the template key will be used?

sbernhard: In case the layout key isn't provided by the ansible playbook but the template is given by file_name, the module will parse the template from file and store it in layout for further handling.
If both, layout key and file_name is given, the layout key wins.

Issue #34, opened 2017-08-16
foreman-ansible-modules leaking sensible information


mdellweg:
By accident i just stumbled on the file `~/.config/nailgun/server_configs.json`.
To my surprise it contained all credentials to connect to the corresponding foreman instance.
As far as i understand, this is actually an issue with nailgun, which was never intended to be used in production (its original design was to help with automated testing).
Imho this implies, by using nailgun, our modules inherit this limitation. \me sad.

Fwiw: Most of the time, we will end up writing this information into the playbooks anyway, but ansible has some vault functionality to prevent just this.

PR #49, opened 2017-09-18
Stop cement from saving ServerConfig to disk

mdellweg:
fixes: #34


mdellweg:
Of course this only fixes the problem for cement-users. But i think it is ok, to put a little more pressure on #37. At the very least, every module should use the `create_server` function.

evgeni:
I actually think the non-cement modules don't suffer from this leak, as they do not call `.save()`: https://github.com/theforeman/foreman-ansible-modules/blob/master/modules/katello_repository.py#L173-L178

mdellweg:
`grep -ir "save()"`
you are right. Even better!

Issue #32, opened 2017-08-15
standardize verify_ssl default value


evgeni:
currently we have plugins that default to false, but also some default to true.

default false:
* foreman_organization.py
* katello_activation_key.py
* katello_content_view.py
* katello_lifecycle_environment.py
* katello_manifest.py
* katello_product.py
* katello_publish.py
* katello_repository.py
* katello_repository_set.py
* katello_sync.py
* katello_sync_plan.py
* katello_upload.py

default true:
* foreman_global_parameter.py
* foreman_provisioning_template.py
* foreman_ptable.py

I'd say we should default to true and allow the user to override that.

Fobhep:
LGTM!

mdellweg:
+1

mdellweg:
I think, this is solved too by #41.

evgeni:
almost. https://github.com/theforeman/foreman-ansible-modules/pull/51 is the last one. this was not done in #41.

PR #51, opened 2017-09-20
katello_content_view_version_promote: default to verify_ssl=True

evgeni:
Fixes: #32


mdellweg:
Right, that must have been added after I scanned all modules. Thanks for looking closely.
ACK

evgeni:
Yupp, @akofink added this just recently

PR #52, opened 2017-09-20
foreman_compute_resource supports check mode now

evgeni:


mdellweg:
Thanks for spotting this!

evgeni:
is that an ACK? :)

mdellweg:
Should be more explicit...
ACK!

evgeni:
Thanks!

PR #55, opened 2017-09-25
add download_policy to katello_repository

mdellweg:
Sadly nailgun allows this write-only...

### DIFF
--- a/modules/katello_repository.py
+++ b/modules/katello_repository.py
@@ -64,6 +64,14 @@ options:
         description:
             - Repository URL to sync from
         required: true
+    download_policy:
+        description:
+            - download policy for sync from upstream
+        choices:
+            - background
+            - immediate
+            - on_demand
+        required: false
 '''
 
 EXAMPLES = '''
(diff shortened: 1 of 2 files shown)


mdellweg:
@ehelms thoughts?

modules/katello_repository.py
```diff
@@ -130,8 +140,10 @@ def repository(self, name, content_type, product, organization, url=None):
 
         repository = repository[0] if len(repository) == 1 else None
 
-        if repository and (repository.name != name or repository.url != url):
-            repository = self._entities.Repository(self._server, name=name, id=repository.id, url=url)
+        # Sadly, nailgun does not handle 'download_policy' as we would like it.
```
ehelms: Does this imply it currently only works on creation? Do we have an issue open to nailgun for this?

mdellweg: If i remember correctly, it is write-only. So creating and updating works, but reading (which is important for 'check_mode' and nice in any case) does not. Haven't had the time to get this to nailgun yet.

mdellweg: The only thing @ nailgun i can find right now is not in repositories, but in smart-proxy. Reinvestigating this right now.

mdellweg: Nailgun was not to blame here. Sorry!
A missing 'read()' was.

mdellweg:
@ehelms and rebased...

modules/katello_repository.py
ehelms: Awesome!

PR #59, opened 2017-10-27
Check check_mode from self._module

akofink:
When we were mass-editing modules with check_mode, we introduced this regression.

```
*** AttributeError: 'NailGun' object has no attribute 'check_mode'
```

### DIFF
--- a/modules/katello_activation_key.py
+++ b/modules/katello_activation_key.py
@@ -166,13 +166,13 @@ class NailGun(object):
         response = activation_key.search({'name', 'organization'})
 
         if len(response) == 0:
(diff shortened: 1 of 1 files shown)


mdellweg:
I really should stop mass editing and we need #61. 🔥 
Not tested, but read the code: Looks good to me. ACK!

akofink:
@mdellweg Actually, I think this one's on me. It came in via my PR to your branch.

mdellweg:
@akofink thanks for taking the blame, but I should have reviewed that. And testing could have caught it.

Issue #61, opened 2017-10-27
Write an Automated Test Suite


akofink:
We have `test/demo/*.yml` and `test/data/*.json` for manual testing. The goal is to create automated tests that will be run against every PR.

akofink:
Even if we just run all the demo playbooks in a particular order (i.e. set up products before repos), it'd be a good smoke test. The main issue is that we don't have a running katello/foreman instance, and it'd be resource intensive to bring up a fresh instance every time.

Ideally, we'd be able to use an apipie cache like we do [in hammer](https://github.com/theforeman/hammer-cli-foreman/blob/master/lib/hammer_cli_foreman/testing/api_expectations.rb).

mdellweg:
For starters, i would be satisfied, if i could run them all against my local forklift instance. It would make reviews easier. It would help if the tests for the single modules were self contained. That could mean, that a module that needs i.e. an organization `present`s one with a common name, thar can be reused by all other tests. Alternatively, we could have a `setup.yml` and a `teardown.yml` to handle these fixtures.

Fobhep:
ideally we tailor the test to use either a pre-installed forklift or "lift-up" a new instance

akofink:
Ok, so after a few days of thought and tinkering, I have some ideas:

1. We should create a common interface (i.e. Makefile) to setup/debug/test. I have this framework already. I can submit a PR; I don't have any tests yet, but the structure is there for python unittests.
1. It'd be great to have a way to both run against a live Katello instance (say, locally for manual testing) and a way to run against cached API structure. I found this project named [vcrpy](https://github.com/kevin1024/vcrpy/). It's a fork of a Ruby version of the same library that we use in hammer-cli-csv tests.

Thoughts?

karras:
I could probably sponsor / organize a VM for testing purpose.
You could also make use of something like Molecule: https://github.com/metacloud/molecule

ehelms:
Some thoughts from our experience building tests for Runcible and Katello using VCR against Pulp.

First, we could start by considering that these modules should work across multiple versions of the software, and that we want to support a specific set. Then, using vcrpy, we could create tests and recorded individual cassettes per software version. This, in theory, means we can have API copies for various versions and ensure changes match those API expectations in the future. Adding a new supported version or dropping one would mean simply dropping the cassettes for it. This would also, if we wanted, allow different modules to support different versions if we don't want to "backport" module support.

mdellweg:
[...]

PR #62, opened 2017-10-31
[RFC] Add automated setup/debug/test framework

akofink:
Refs #61


test/__init__.py
```diff
@@ -0,0 +1,6 @@
+from unittest import main
+
+# from katello.content_view_test import ContentViewTest
```
akofink: This is a placeholder, but it acts as an example of how to add test classes. The methods in the test classes must start with `test_`, one of the python unittest conventions.

akofink:
@mdellweg @Fobhep Any suggestions?

mdellweg:
Sorry, we have holidays in germany. But i will have a look tomorrow.
Right now, i can say i like the idea of the VCR module and i am glad about any progress to automated testing.

Makefile
```diff
@@ -0,0 +1,44 @@
+TEST_OPTIONS=-m unittest test
+DEBUGGER_PATH=/usr/lib64/python2.7/pdb.py
```
mdellweg: On my box, the debugger is located in `/usr/lib/python2.7/pdb.py` (no 64).
I think we need another method to find it.

test/__init__.py
mdellweg:
```
$ make test
python -m unittest test

----------------------------------------------------------------------
Ran 0 tests in 0.000s

OK
```
Impressive ;)
Well could you at least add one test to prove the concept? (And as a template to copy an paste...)
Also, i want to see VCRpy in action. 🎥

test/test_helper.py
```diff
@@ -0,0 +1,5 @@
+from subprocess import call
```
mdellweg: As far as i understand unittest (at least in autodetect mode which is not used here), it scans for files named `test_*.py`. Therefore the name chosen here might not be the best one. You could just call it `helper.py` or if you are trying to avoid name clashes `fam_test_helper.py`. (this is optional)

test/__init__.py
akofink: Yes, this PR is WIP (should have set the title).

ehelms:
Can you outline the benefits of using make over something like pytest for testing infrastructure?

akofink:
Make and pytest solve different problems. We could use pytest from a makefile, but show me how you would use pytest to setup the environment or how you would use it to debug with the hacking module.

ekohl:
@akofink pytest will scan for files. You can do any setup in ```conftest.py``` if needed.

ehelms:
What I am really asking from a higher stand point is, is `make` the Python convention for having a multi-action task runner? The Ruby world uses Rake but my python tooling knowledge is rusty. I am not against Make necessarily, it's fairly straight forward its just both not something any of our other projects use and it tends to be beneficial to use tooling within the language ecosystem (where possible and where useful). Covering our bases essentially.

akofink:
Both Ansible and Nailgun use a Makefile. That's why I went with that option.

ekohl:
Python generally doesn't have something like ```make``` or ```rake```. You can provide custom commands via setup.py but that's uncommon. It's possible to run ```python setup.py test``` and even modify it to use ```pytest``` but just running ```pytest``` is very common.

Having commands in place to setup a debugger is uncommon as far as I know. I guess every python dev has their own env setup, either through PyCharm or whatever they prefer.

.travis.yml
```diff
@@ -2,8 +2,5 @@ language: python
 python:
   - "2.6"
```
ekohl: I'd suggest testing 2.7 as well since that's what EL7 ships. IMHO way more important than 2.6 from EL6 ;)

Makefile
```diff
@@ -0,0 +1,44 @@
...
+MODULE_DIR=modules/$(MODULE).py
+DEBUG_DATA_PATH=test/data/$(DATA).json
+DEBUG_OPTIONS=-m $(MODULE_DIR) -a @$(DEBUG_DATA_PATH) -D $(DEBUGGER_PATH)
```
ekohl: Wouldn't it make more sense to run ```-m pdb $(MODULE_DIR)```?

Makefile
```diff
@@ -0,0 +1,44 @@
+TEST_OPTIONS=-m unittest test
+DEBUGGER_PATH=/usr/lib64/python2.7/pdb.py
+MODULE_DIR=modules/$(MODULE).py
```
ekohl: The name is directory but the value implies it's a file.

Makefile
akofink: That is an unsupported syntax. I'm not sure how it would work.

```
$ ./.tmp/ansible/hacking/test-module -m pdb modules/katello_content_view.py -a @test/data/content-view.json
  File "./.tmp/ansible/hacking/test-module", line 268, in <module>
    main()
  File "./.tmp/ansible/hacking/test-module", line 249, in main
    (modfile, modname, module_style) = boilerplate_module(options.module_path, options.module_args, interpreters, options.check, options.filename)
  File "./.tmp/ansible/hacking/test-module", line 152, in boilerplate_module
    task_vars=task_vars
  File "/usr/lib/python2.7/site-packages/ansible/executor/module_common.py", line 861, in modify_module
    with open(module_path, 'rb') as f:
IOError: [Errno 2] No such file or directory: 'pdb'
```

mdellweg:
@ehelms Make does not really belong to any programming language ecosystem. As you mentioned, it is a perfectly valid tool to run a number of tasks to accomplish a specific goal. And considering that the tasks at hand are to run shell commands, a Makefile can encode this without much clutter (think `os.system` around any command).
Also make seems to have a very matured and stable api. I bet, when this whole ansible thing goes down in flames in say 20 years, this Makefile will be interpreted correctly by the then recent gmake. ;)
However, i must say, having only phony targets kind of defeats the power of make.

.travis.yml
```diff
@@ -2,8 +2,5 @@ language: python
...
   - "3.5"
-install:
-  - "pip install --upgrade pip"
-  - "pip install pep8"
-  - "pip install -r requirements.txt"
-script: "pep8 --ignore=E402 --max-line-length=160 ."
+install: make test-setup
```
mdellweg: I just realized: this `make` is superfluous. There is no target `make`. (same on the following line)

akofink:
Don't worry, I'm coming back to this soon. It's just low on my list.

.travis.yml
akofink: Could you explain? I think we need `make` here.

mdellweg: Sorry, i was wrong. the line looked too much like Makefile to me to realize it is the travis config.
Complaint withdrawn. 🙈

ehelms:
The Pulp team suggested looking at `tox` (https://tox.readthedocs.io/en/latest/config.html) and at first glance it has some interesting capabilities for Python based projects with how virtualenvs are used. I'm 50/50 on whether or not it has the same ability to create individual tasks easily. I'll defer to others to take a glance through it's docs to see if it gains us anything.

akofink:
[Django uses tox](https://github.com/django/django/blob/master/tox.ini).

akofink:
After looking at tox, it seems like a heavy handed, shoehorned solution. I'm still voting for make. If you can show me a nice way to use tox to do what I've done in this Makefile, then by all means, please show me your `tox.ini`. Additionally, it seems like tox has a higher barrier to entry than make. I can't figure out a good way to give users a usage output when they just run `tox` like I have setup for `make`. It seems much more like a tool to automate testing in different virtualenvs than to automate generic tasks.

.travis.yml
akofink: I found [some ansible documentation](http://docs.ansible.com/ansible/latest/dev_guide/developing_python3.html#minimum-version-of-python-3-x-and-python-2-x) suggesting that we should use Python 2.6 and 3.5

ekohl:
The way I look at it is that tox can allow you to easily test under the various python versions, but in general I trust Travis to do that for me.

README.md
```diff
@@ -62,16 +62,18 @@ ansible-playbook test/demo/organization.yml
...
 Debug with ansible's test-module
 
 ```sh
-./ansible/hacking/test-module -m <path to ansible module> -a @<path to input arguments to module> -D <path to debugger>
+MODULE=<module name> DATA=<data file name> make debug
 
-# Example: debug the activation_key module with pdb
-./ansible/hacking/test-module -m ./foreman-ansible-modules/modules/katello_activation_key.py -a @activation-key-args.json -D /usr/lib/python2.7/site-packages/pdb/pdb.py
+# Example: debug the katello_content_view module
+$ MODULE=katello_content_view DATA=content-view make debug
```
ehelms: Do these have to be before the `make` command? Given `make` is the entry point, I think it atypical for most to start with the environment variables. I'm more prone to start typing `make debug` and then include my options. If these could possibly be positional (at least module) that would be great too.

ehelms:
I can live with `make` then and I do appreciate exploring the other options @akofink . I'd rather bikeshed at the beginning then half way down the road and end up having to pivot after investing one method.

README.md
akofink: Oh, they can be after the command, like this:

```sh
make debug MODULE=katello_content_view DATA=content-view
```

I wonder if we should just go ahead and lowercase the variables. Make doesn't really support positional arguments because the syntax is `make <target> <target> ...`. There are hacky ways of making it work, but I'd avoid it.

mdellweg: Not if you meant that by 'hacky', but you can use wildcard in the targets. It would allow for something like
`make debug-katello-product`.

akofink: @mdellweg an interesting solution. Would it be bad to implement both?

mdellweg: I guess it would not hurt.

tests/helper.py
```diff
@@ -0,0 +1,9 @@
+from ansible.cli.playbook import PlaybookCLI
+import vcr
+
+
+@vcr.use_cassette('tests/fixtures/api.yml')
```
evgeni: should `api.yml` be part of this PR?

Makefile
```diff
@@ -0,0 +1,45 @@
+TEST=tests
+PDB_PATH=$(shell find /usr/lib{,64} -type f -executable -name pdb.py 2> /dev/null)
+DATA=$(shell echo $(MODULE) | sed 's/\(foreman\|katello\)_//; s/_/-/g')
```
evgeni: `make` has functions for [text substitution](https://www.gnu.org/software/make/manual/html_node/Text-Functions.html), do we want to use them?

evgeni:
Yeah, tox and make solve different problems. Heck, I even use `tox` to call `make` to call `trial` in [one of my private projects](https://github.com/evgeni/bley/). And yes, tox is mostly for generating environments and yes Travis can do this for me, but I don't run Travis locally.

I like the overall structure of this, and just have two comments:
* Travis surely should not fail, right? ;)
* Can we avoid writing boilerplate `test/test_<foo>.py` and have a generator for that instead? Just iterating over `tests/test_data/*.yml` should be enough, no?

evgeni:
example loader:

```python
import unittest
import glob
import os

class GenericFAMTestCase(unittest.TestCase):
    pass

def generate_test(demo_file):
    def do_test(self):
        print "testing %s" % (demo_file)
    return do_test

def find_tests():
    demo_files = glob.glob('test/demo/*.yml')
    for demo_file in demo_files:
        test_method = generate_test(demo_file)
        test_method.__name__ = 'test_%s' % os.path.basename(demo_file)
        setattr (GenericFAMTestCase, test_method.__name__, test_method)

find_tests()

if __name__ == '__main__':
    unittest.main()
```

tests/helper.py
akofink: Yes, it will be.

akofink:
So, here's where I'm at right now: the Makefile is probably done; the test structure is in place for 2 modules, and the tests run locally without vcrpy caching. The issue I'm facing is that vcrpy is unable to listen for requests from nailgun via ansible playbook. The vcrpy package supports the requests package, which is what nailgun is using, so I just need to figure out how to wrap a playbook run in vcrpy's `use_cassette` decorator. That will then generate the api.yml that is missing in this PR and is needed to run the tests without a live server.

@evgeni Loading the tests sounds like a good idea for DRYness, and yes, tests should pass. :face_with_head_bandage:

Makefile
akofink: Here's what it would look like (perhaps there's a simpler way):

```
DATA=$(substr, _,-,$(filter-out katello_, $(filter-out foreman_, $(MODULE)))
```

I don't know about you, but I don't like it.

ekohl:
> So, here's where I'm at right now: the Makefile is probably done; the test structure is in place for 2 modules, and the tests run locally without vcrpy caching. The issue I'm facing is that vcrpy is unable to listen for requests from nailgun via ansible playbook. The vcrpy package supports the requests package, which is what nailgun is using, so I just need to figure out how to wrap a playbook run in vcrpy's use_cassette decorator. That will then generate the api.yml that is missing in this PR and is needed to run the tests without a live server.

I think that's because you use the CLI running. I suspect that's actually calling it on the command line which means the monkey patching of ```socket``` is not inherited. Not sure if you can ansible directly via an API.

ehelms:
A few recommendations from my experience with VCR in Katello:

 1) A cassette file per test, not per test file, but per test. This makes re-recording much easier and faster. And provides less repository churn.
 2) If vcrpy has an option to base64 encode the body use it to save space and time.
 3) Maybe move `test_data` to be named `test_playbooks` so its a bit cleaner what the data is.
 4) Consider whether you want to think about supporting VCR cassettes by version tested against and scoping on that. For example, `test/fixtures/3.5/katello_product/create.yml` which combines some of my earlier ideas.

akofink:
The vcrpy decorator works great when it's on the main() function of the module. Obviously we can't put it there, but I wonder if by some python magic we can put it there before running the tests.

ekohl:
Would https://github.com/ktosiek/pytest-vcr help?

ekohl:
There's a conflict now.

Makefile
```diff
@@ -0,0 +1,45 @@
+TEST=tests
+PDB_PATH=$(shell find /usr/lib{,64} -type f -executable -name pdb.py 2> /dev/null)
```
Fobhep: maybe you can consider using "shell which pdb" here? 
In an virtualenv the solution as you suggested it is sadly failing.
"which pdb" is referencing to the first found "pdb" and thus gives for an virtualenv e.g. usr/bin/pdb which is also suggested by the documentation of ./test-module for -D

akofink: As far as I can tell (centos7, system python, no virtualenv), the `-D` requires an executable `pdb.py`, not the executable found by running `which pdb`. Does it work for you with `which pdb` in a virtualenv?

akofink:
Rebased for conflicts.

Makefile
akofink: Also, you can try out different paths to see which one works with this syntax:

```
make debug MODULE=katello_content_view PDB_PATH=<whatever you want>
```

Fobhep: yes it does - in principle. I cannot even start pdb by hand inside virtualenv.
maybe that is a debian problem?
although this is not really relevant at this place

README.md
```diff
@@ -62,18 +62,22 @@ ansible-playbook test/demo/organization.yml
...
 
 ```sh
-./ansible/hacking/test-module -m <path to ansible module> -a @<path to input arguments to module> -D <path to debugger>
+make debug MODULE=<module name>
 
-# Example: debug the activation_key module with pdb
-./ansible/hacking/test-module -m ./foreman-ansible-modules/modules/katello_activation_key.py -a @activation-key-args.json -D /usr/lib/python2.7/site-packages/pdb/pdb.py
+# Example: debug the katello_content_view module
+$ make debug MODULE=katello_content_view
+./.tmp/ansible/hacking/test-module -m modules/katello_content_view.py -a @test/data/content-view.json -D /usr/lib64/python2.7/pdb.py
```
mdellweg: It would be nice, to have configurable server (and credentials) in these json files, too. Do you see an easy way to incorporate this?

akofink: I agree. I can't think of anything right off, but I can look into it.

Makefile
akofink: You know, perhaps it'd be pretty simple to just check if we're in a virtualenv and use `which pdb` in that case. Sound good?

Fobhep: sounds good to me

mdellweg:
```
.tmp/ansible/hacking/test-module --help
<...>
  -D DEBUGGER, --debugger=DEBUGGER
                        path to python debugger (e.g. /usr/bin/pdb)
```
This sounds to me, that `<whatever_venv_or_not>/bin/pdb` should do the job. And that one shall always be `which pdb`.

akofink: I agree that the documentation (what little there is) makes it sound like `which pdb` would work, but it does not work with native Python:

```sh
./.tmp/ansible/hacking/test-module -m modules/katello_content_view.py -a @tests/debug_data/content-view.json -D /usr/bin/pdb
* including generated source, if any, saving to: /home/vagrant/.ansible_module_generated
* ansiballz module detected; extracted module source to: /home/vagrant/debug_dir
  File "/usr/bin/pdb", line 7, in <module>
    from pdb.pdb import main
ImportError: No module named pdb
```

mdellweg: Funny, in our setups it was the exact opposite. We got this error with the venv pdb, but the system wide one worked.

README.md
ekohl: I recall having a wrapper in another project where we override the loading and saving of the hostname, username and password to static strings for the VCR. Otherwise tests can fail when you run them with different credentials.

Makefile
```diff
@@ -0,0 +1,45 @@
...
+DATA=$(shell echo $(MODULE) | sed 's/\(foreman\|katello\)_//; s/_/-/g')
+MODULE_PATH=modules/$(MODULE).py
+DEBUG_DATA_PATH=tests/data/$(DATA).json
+DEBUG_OPTIONS=-m $(MODULE_PATH) -a @$(DEBUG_DATA_PATH) -D $(PDB_PATH)
```
mdellweg: Sorry, to nag again: All these variables seem like perfect candidates for 'simply expanded variables', (`::=`) especially the ones calling shell commands.
See: https://www.gnu.org/software/make/manual/html_node/Flavors.html

akofink:
I know this PR is getting kind of stale, but I haven't figured out a good way to get VCR.py to listen to nailgun requests. I hesitate to say it's not possible, but we might want to investigate other ways of automated testing. I'm going to address the remaining comments, remove the VCR.py parts, and disable the tests in `.travis.yml` so we can at least get the `Makefile` and related infrastructure (i.e. using `include_tasks` in the test_data and writing each data playbook to be a complete up/down scenario).

Thoughts/comments welcome!

Makefile
akofink: Why would that be better? Using recursively expanded variables, the variables are not even executed unless they are used (i.e. lazy assignment). That seems like the best case here; you wouldn't want `MODULE`, `DATA`, etc. to be getting set if you are just running `make setup`.

Fobhep:
So I definitely think it is a great idea to split your PR to a point where we can have manual testing which would be already very nice.
However there is probably still another problem with VCR:
Assume you run a playbook to create an entity.
If you "record" your test-response from Foreman for a  first time you might want the server to actually tell you, the template is not there and you are able to create it.
However a second time the already recorded answer might not be anymore the one you necessarily want for your test, depending whether you want to e.g. test updating or deleting it.
This means also we have to make very sure which test can run after which, since some "records" might depend on some others. (e.g. Locations need to be there to test assignments to them in the context of other modules)

akofink:
@Fobhep While this discussion might be moot (if we can't ever get VCR to work), it seems that we could have some amount of execution and cassette loading dependencies such that the API responses are appropriate for that step in the test. I don't think we will know for sure unless we get VCR working and test it out. Remember that we can use the `with vcr.cassette():` syntax as well as the decorator, so it seems that it'd be somewhat easy to make the API respond correctly.

Example (cassette names only for demonstration):

```py
with_cassette('organization_absent.yml'):
  # test org create
  ...
with_cassette('organization_present.yml'):
  # test product create
...
```

Makefile
```diff
@@ -0,0 +1,43 @@
...
+	@echo "  help         to show this message"
+	@echo "  lint         to run flake8 and pylint"
+	@echo "  test         to run unit tests"
+	@echo "  debug        debug a module using the ansible hacking module"
+	@echo "  setup        to set up test, lint, and debugging"
+	@echo "  test-setup   to install test dependencies"
+	@echo "  debug-setup  to set up the ansible hacking module"
+
+lint:
+	pycodestyle --ignore=E402,E722 --max-line-length=160 modules test
```
mdellweg: Should this be `$(TEST)`?

Makefile
mdellweg: Just to be sure: Recursively expanded variables are more like macros (evaluated on every use), not like lazy assigned statements (evaluated on first use only and not available im Make). So at least for calling shell commands i would prefer simple (immediate) expansion, to prevent multiple execution and possibly very strange side effects where the second execution evaluates differently. Maybe for a Makefile that size the whole discussion is moot. In the end it might be a matter of taste.

mdellweg:
@akofink Don't be sad! Having manual tests is really a big step forward. But also do not throw away your vcr branch. Thinking about the whole thing again, i came to the conclusion we might actually get it to work if we separate the database fixtures from the actual tests and record the tests only and on individual tapes. Also we should append a policy for reviewing PRs with rerecorded tapes (maybe two ACKs, no NACK).

Fobhep:
@akofink ok, that sounds like a good approach :)

Makefile
akofink: Nah, that variable is so that you can do `make test TEST=test/test_product.py`. I guess we could add a `$(LINT)` variable in case a user wants to override what gets checked in linting, but I don't see it as very useful, mostly because linting everything takes less than a second:

```
$ time make lint
pycodestyle --ignore=E402,E722 --max-line-length=160 modules test

real    0m0.391s
user    0m0.376s
sys     0m0.015s
```

Compare that with the tests (only 2 right now):

```
$ time make test
pytest test
=============================================================================================== test session starts ===============================================================================================
platform linux2 -- Python 2.7.5, pytest-3.2.3, py-1.4.34, pluggy-0.4.0
rootdir: /home/vagrant/foreman-ansible-modules, inifile:
collected 2 items                                                                                                                                                                                                  

test/test_organization.py .
test/test_product.py .

============================================================================================ 2 passed in 43.16 seconds ============================================================================================

real    0m43.381s
user    0m9.205s
sys     0m0.948s
```

mdellweg: Good point. I was unsure what the variable would be for and came up with overriding the directory containing the tests by say 'new-tests'.
On the other hand, it might be convenient to specify the test in question without the directory.

akofink:
This should be ready for final review.

.gitignore
```diff
@@ -1,3 +1,5 @@
 .tmp
 *.retry
-/test/demo/server_vars.yml
+/tests/test_data/server_vars.yml
```
mdellweg: This is 'test_playbooks', i think.

test/test_playbooks/server_vars.yml.example
```diff
@@ -0,0 +1,5 @@
+---
```
mdellweg: The file above (test/test_playbooks/server_vars.yml) should not be in the repository. (see .gitignore)

test/test_playbooks/organization.yml
```diff
@@ -0,0 +1,12 @@
+---
+- hosts: localhost
+  tasks:
+  - name: Load server config
+    include_vars:
+      file: server_vars.yml
+  - include_tasks: tasks/organization.yml
```
mdellweg: I just figured out, `include_tasks` is only available since ansible 2.4.
https://docs.ansible.com/ansible/2.4/include_tasks_module.html
Do we want to be able to run the tests on 2.3? (At least that's what our modules describe as a dependency.)

akofink:
Updated to work with Ansible < 2.4 and addressed .gitignore/server_vars.

test/test_playbooks/organization.yml
mdellweg: Yea, i have seen the warning too. It would be nicer, if they warned one version later. This means you cannot write playbooks without incompatibility and warning for two consecutive minor versions of ansible. 😞

PR #66, opened 2017-11-14
Search Operatingsystems by title

mdellweg:


module_utils/ansible_nailgun_cement.py
```diff
@@ -281,6 +282,10 @@ def find_repository_set(module, name, product, failsafe=False):
     return handle_find_response(module, repo_set.search(), message="No repository set found for %s" % name, failsafe=failsafe)
 
 
+def find_operating_system_by_title(module, title, failsafe=False):
```
akofink: Why not just a generic search function?

```py
def find_operating_system_by_title(module, search, failsafe=False):
```

mdellweg: Sorry, i closed this one and reopend it from atixag:find_operatingsystem

PR #67, opened 2017-11-14
Search Operatingsystems by title

mdellweg:
Searching for operating systems seems more complicated than other entities...


mdellweg:
> @akofink :
> Why not just a generic search function?
>  `def find_operating_system(module, search, failsafe=False):`

That would be inconsistent with the naming of e.g.
`def find_compute_profile(module, name, failsafe=False):`.

Besides the find with `~` seems very special.

Issue #56, opened 2017-09-28
Enforce/Document ansible >= 2.3


akofink:
It seems ansible v. 2.2 has some issues with requiring module support code:

```sh
fatal: [localhost]: FAILED! => {"failed": true, "msg": "Could not find imported module support code for katello_upload.  Looked for either handle_no_nailgun or foreman_helper"}
```

mdellweg:
OK for >= 2.3.
Is it like `Require: ansible >= 2.3.0` in the Documentation variable?
Should we implement a test for this before failing to import `handle_no_nailgun`?

akofink:
I believe so, like [this](https://github.com/ansible/ansible/blob/devel/examples/DOCUMENTATION.yml#L31).

Fobhep:
@akofink I am currently trying to tackle that issue and I came up with a bunch of questions:

- can you confirm this being solved with v.2.3+ ?
- I did a bit of research and there are several modules in ansible that use import from module_utils   before version 2.3.
- Also did you try this with a pip-installation of ansible or package?
- Virtualenv or "native"?

Fobhep:
Generally it seems a good idea to go not beyond a certain ansible version.
I am just wondering wether we want to connect a failure to an ansible version being below a minimum, or to a failed import of module_utils or to both

mdellweg:
From 2.3 Release notes:
> Allow module_utils for custom modules to be placed in site-specific directories and shipped in roles

Fobhep:
ok fair enough. I guess the older modules probabally use a shared module_utils for the whole ansible project

akofink:
@Fobhep I believe I was using Python 2.7 native on Centos 7.4. It's been a couple months :thinking:

evgeni:
We documented this, and 2.3 is quite old. Shall this issue be closed? I'd say so :)

PR #70, opened 2017-11-22
Added ansible >= 2.3.0 to documentation/enforcing

Fobhep:
This PR is meant to solve #56

### DIFF
--- a/modules/foreman_compute_profile.py
+++ b/modules/foreman_compute_profile.py
@@ -22,7 +22,13 @@ DOCUMENTATION = '''
 module: foreman_compute_profile
 short_description: Manage Foreman Compute Profiles using Foreman API
 description:
-- Create and delete Foreman Compute Profiles using Foreman API
+  - Create and delete Foreman Compute Profiles using Foreman API
+version_added: "2.0"
+author: "Philipp Joos (@philippj)"
+requirements:
+  - "nailgun >= 0.28.0"
+  - "python >= 2.6"
+  - "ansible >= 2.3"
 options:
   name:
     description: compute profile name
@@ -47,10 +53,6 @@ options:
     description: compute profile presence
     default: present
     choices: ["present", "absent"]
-notes:
-- Requires nailgun
-version_added: "2.0"
-author: "Philipp Joos (@philippj)"
 '''
 
 EXAMPLES = '''
--- a/modules/foreman_installation_medium.py
+++ b/modules/foreman_installation_medium.py
@@ -27,7 +27,8 @@ version_added: "2.5"
 author:
   - "Manuel Bonk(@manuelbonk) ATIX AG"
 requirements:
-  - nailgun >= 0.16.0
+  - "nailgun >= 0.16.0"
+  - "ansible >= 2.3"
 options:
   name:
     description: The full installation medium name
@@ -45,9 +46,9 @@ options:
     required: false
     type: list
   os_family:
-      description: The OS family the template shall be assigned with. If no os_family is set but a operatingsystem, the value will be derived from it.
-      required: false
-      choices:
+    description: The OS family the template shall be assigned with. If no os_family is set but a operatingsystem, the value will be derived from it.
+    required: false
+    choices:
       - AIX
       - Altlinux
       - Archlinux
--- a/modules/foreman_location.py
+++ b/modules/foreman_location.py
@@ -22,45 +22,46 @@ DOCUMENTATION = '''
 module: foreman_location
 short_description: Manage Foreman Location
 description:
-    - Manage Foreman Location
+  - Manage Foreman Location
 author:
-    - "Matthias M Dellweg (@mdellweg) ATIX AG"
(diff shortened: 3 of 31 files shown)


modules/foreman_compute_profile.py
```diff
@@ -47,9 +47,12 @@
     default: present
     choices: ["present", "absent"]
 notes:
-- Requires nailgun
+  - Requires nailgun
```
mdellweg: This seems superfluous now.

modules/foreman_compute_profile.py
```diff
@@ -65,6 +68,12 @@
 
 RETURN = ''' # '''
 
+from ansible.release import __version__ as ansible_version
+if [int(i) for i in ansible_version.split(".")] >= [2, 3, 0]:
+    HAS_PROPER_VERSION = True
+else:
+    HAS_PROPER_VERSION = False
+
```
mdellweg: There is a logical error in this: When used with ansible < 2.3, the subsequent import of foreman_helper will still fail with an exception before we can report the incompatibility.

modules/foreman_compute_profile.py
```diff
@@ -65,6 +68,12 @@
 
 RETURN = ''' # '''
 
+from ansible.release import __version__ as ansible_version
+if [int(i) for i in ansible_version.split(".")] >= [2, 3, 0]:
+    HAS_PROPER_VERSION = True
+else:
+    HAS_PROPER_VERSION = False
```
mdellweg: You can simply assign the result of the comparison to the variable here.

ekohl:
Rather than all the duplication, can't we introduce a shared helper that has this logic?

mdellweg:
That is exactly the Problem: The module_utils thing is only available since Ansible 2.3. Before that there is no sharing.

ekohl:
Pure python can import packages too. Would a relative import work?

mdellweg:
These files must be known to ansible and copied to the Machine running the task. That is why we chose this module_utils directory in the first place.

ekohl:
Ah yes, I keep forgetting that.

Fobhep:
Yeah. It is really a shame we have to do this so not-DRY. But there is not really an alternative. @mdellweg  and me did a lot of digging before we wrote cement and could not find anything except the module_utils import

modules/foreman_compute_profile.py
Fobhep: yes, thanks :) I did not finish the PR yet, hence WIP ;)

evgeni:
I'd prefer not to have this check in the modules, but only document the fact in `README.md` and `requirements.txt`.

Fobhep:
I picked this PR up again and did not only add the Ansible >= 2.3 dependency but I also tried to sort of "normalize" the documentation of all modules. 
Meaning I fixed the identation and tried to unify certain things. Feedback is welcome! :)
@evgeni @akofink @mdellweg

modules/foreman_compute_profile.py
```diff
@@ -22,7 +22,13 @@
 module: foreman_compute_profile
 short_description: Manage Foreman Compute Profiles using Foreman API
 description:
-- Create and delete Foreman Compute Profiles using Foreman API
+  - Create and delete Foreman Compute Profiles using Foreman API
+version_added: "2.0"
```
mdellweg: I think, we should remove that until it is upstream.

modules/foreman_installation_medium.py
```diff
@@ -27,7 +27,8 @@
 author:
   - "Manuel Bonk(@manuelbonk) ATIX AG"
 requirements:
-  - nailgun >= 0.16.0
+  - "nailgun >= 0.16.0"
```
akofink: I'd be awesome if we could verify this, but I can't really think of a good way to do so.

modules/foreman_compute_profile.py
Fobhep: if everybody is ok, I would open an issue to  be solved before going actually upstream, that includes fixing the number to the actual upstream version.

modules/foreman_location.py
```diff
@@ -22,45 +22,46 @@
...
+  - Manage Foreman Location
 author:
-    - "Matthias M Dellweg (@mdellweg) ATIX AG"
+  - "Matthias M Dellweg (@mdellweg) ATIX AG"
 requirements:
-    - "nailgun >= 0.28.0"
-    - "python >= 2.6"
+  - "nailgun >= 0.28.0"
+  - "python >= 2.6"
+  - ansible >= 2.3
```
evgeni: should this be quoted too?

modules/foreman_installation_medium.py
Fobhep: yeah - me neither really.

modules/foreman_location.py
Fobhep: ah - yes it should. I will fix that

Fobhep: @evgeni fixed

PR #71, opened 2017-11-23
Add redhat_manifest module

sean797:
This uses the RHSM API to modify, create or download a Satellite manifest from the Red Hat portal.


akofink:
I appreciate the initiative, but I'm not sure this module belongs in the foreman + plugins set of modules. Perhaps it would be better to add this subscription management application logic to core Ansible? Perhaps the functionality could be included in the [redhat_subscription module](http://docs.ansible.com/ansible/latest/redhat_subscription_module.html).

sean797:
@akofink I think it does belong here as AFAIK the only reason someone has to generate a manifest is Katello (or Satellite) ?

modules/rhsm_manifest.py
```diff
@@ -0,0 +1,264 @@
...
+from ansible.module_utils.urls import fetch_url
+from ansible.module_utils.pycompat24 import get_exception
+
+
+def fetch_portal(module, path, method, data={}, accept_header='application/json'):
+    url = module.params['portal'] + path
+    headers = {'accept': accept_header,
+               'content-type': 'application/json'}
+    resp, info = fetch_url(module, url, json.dumps(data), headers, method)
+    return resp, info
```
ekohl: This makes the code below unreachable. Is that intentional?

modules/rhsm_manifest.py
```diff
@@ -0,0 +1,264 @@
...
+    subs = filter(lambda s: s['pool']['id'] == module.params['pool_id'], all_subs)
+    return subs
+
+
+def get_remove_or_attach_sub(module, manifest):
+    changed = False
+    subs = get_subs(module, manifest)
+    if subs:
+        if module.params['state'] == 'present':
+            sub_quantity = sum(map(lambda s: s['quantity'], subs))
```
ekohl: Is this easier to read? ```sum(s['quantity'] for s in subs)```

modules/rhsm_manifest.py
```diff
@@ -0,0 +1,264 @@
...
+    path = "/subscription/users/%s/owners" % (module.params['username'])
+    resp, info = fetch_portal(module, path, 'GET')
+    return json.loads(resp.read())[0]['key']
+
+
+def get_subs(module, manifest):
+    path = "/subscription/consumers/%s/entitlements" % (manifest['uuid'])
+    resp, info = fetch_portal(module, path, 'GET')
+    all_subs = json.loads(resp.read())
+    subs = filter(lambda s: s['pool']['id'] == module.params['pool_id'], all_subs)
```
ekohl: Note that in Python 3 filter has become a generator which means you can consume it only once. That may lead to very subtle breakage. I'd advise a list comprehension: ```[s for s in all_subs if s['pool']['id'] == module.params['pool_id']]``` or wrapping the above in ```list()```

modules/rhsm_manifest.py
```diff
@@ -0,0 +1,264 @@
...
+            changed = True
+    elif module.params['state'] == 'present':
+        attach_sub(module, manifest, module.params['quantity'])
+        changed = True
+    return changed
+
+
+def remove_sub(module, manifest, sub):
+    path = "/subscription/consumers/%s/entitlements/%s" % (manifest['uuid'], sub['id'])
+    resp, info = fetch_portal(module, path, 'DELETE')
```
ekohl: Unused variables. I'd either remove them or return something.

modules/rhsm_manifest.py
```diff
@@ -0,0 +1,264 @@
...
+def attach_sub(module, manifest, quantity):
+    path = "/subscription/consumers/%s/entitlements?pool=%s&quantity=%s" % (manifest['uuid'], module.params['pool_id'], quantity)
+    resp, info = fetch_portal(module, path, 'POST')
+
+
+def export_manifest(module, manifest):
+    path = "/subscription/consumers/%s/export" % (manifest['uuid'])
+    try:
+        resp, info = fetch_portal(module, path, 'GET', accept_header='application/zip')
+        f = open(module.params['path'], 'w')
```
ekohl: Using the ```with``` syntax will autoclose it even if an exception is raised:
```python
with open(module.params['path'], 'w') as f:
    while True:
        data = data.read(65536)
        if not data:
            break
        f.write(data)
```

modules/rhsm_manifest.py
```diff
@@ -0,0 +1,264 @@
...
+        # Read 1kb at a time to save on ram
+        while True:
+            data = resp.read(65536)  # 64K
+            if data == "":
+                break  # End of file, break while loop
+            f.write(data)
+        f.close()
+    except Exception:
+        e = get_exception()
```
ekohl: Any reason you're not using ```except Exception as e```? Is this an ansible thing?

modules/rhsm_manifest.py
```diff
@@ -0,0 +1,264 @@
...
+
+    module.params['rhsm_owner'] = get_owner(module)
+
+    manifest, man_changed = get_or_create_manifest(module)
+    if module.params['pool_id']:
+        sub_changed = get_remove_or_attach_sub(module, manifest)
+    if module.params['path']:
+        export_manifest(module, manifest)
+
+    changed = man_changed if man_changed else sub_changed
```
ekohl: I think ```change = man_changed or sub_changed``` should work just as well but you should be careful that if the module has no ```pool_id``` that ```sub_changed``` is not defined.

modules/rhsm_manifest.py
sean797: 👍 `get_exception()` is an ansible thing, https://github.com/theforeman/foreman-ansible-modules/pull/71/files#diff-88ef7a0ec70e64151c3a125a04cb3b36R114

modules/rhsm_manifest.py
```diff
@@ -0,0 +1,264 @@
...
+    subs = get_subs(module, manifest)
+    if subs:
+        if module.params['state'] == 'present':
+            sub_quantity = sum(s['quantity'] for s in subs)
+            while sub_quantity > module.params['quantity']:
+                difference = sub_quantity - module.params['quantity']
+                remove_sub(module, manifest, subs[0])
+                changed = True
+                subs = get_subs(module, manifest)
+                sub_quantity = sum(map(lambda s: s['quantity'], subs))
```
ekohl: This can be rewritten like above.

modules/rhsm_manifest.py
```diff
@@ -0,0 +1,264 @@
...
+
+def get_or_create_manifest(module):
+    path = "/subscription/owners/%s/consumers?type=satellite" % (module.params['rhsm_owner'])
+    resp, info = fetch_portal(module, path, 'GET')
+    manifests = json.loads(resp.read())
+    if module.params['name']:
+        attr = 'name'
+    if module.params['uuid']:
+        attr = 'uuid'
+    manifest = filter(lambda m: m[attr] == module.params[attr], manifests)
```
ekohl: Same as ```get_subs```

sean797:
Thanks @ekohl 👍 updated

modules/rhsm_manifest.py
```diff
@@ -0,0 +1,272 @@
...
+    module.params['rhsm_owner'] = get_owner(module)
+
+    manifest, man_changed = get_or_create_manifest(module)
+    if module.params['pool_id'] and manifest:
+        sub_changed = get_remove_or_attach_sub(module, manifest)
+    if module.params['path'] and manifest:
+        export_manifest(module, manifest)
+
+    changed = man_changed
+    if module.params['pool_id'] and not changed:
```
ekohl: I'd simplify this and add ```else: sub_changed = False``` to other if block. Since you added the ```and manifest, you need to duplicate the logic here to but being able to count on it being present makes it much easier. Then you can go back to ```changed = man_changed or sub_changed```

sean797:
Yea that's quite a bit simpler! Thanks @ekohl

akofink:
@sean797 All other modules in this repo use the Foreman/Katello API, which is why I think it makes sense to have `foreman_*`/`katello_*` modules. This one would hit the RHSM API, which makes me think it should be in a `rhsm_*` module. I'd at least like to get some other opinions on it, maybe from @ehelms or @ekohl?

ekohl:
I have no opinion on it. I see your point but the module is named ```rhsm_*```. If there is no other rhsm module, I'd be fine with including it here rather than have the overhead of a separate repo.

akofink:
@ekohl Also remember that Ansible modules are usually added directly to the ansible/ansible repo, and that's the ultimate goal of this repo, that way users can install Ansible and be able to use these modules out of the box.

ekohl:
That's fair, but if including it here for a bit exposes it to more testing before it's included in ansible then fine by me.

akofink:
Yeah, I'd ultimately be fine with it staying in this repo; just wanted to see what others thought. Thanks for the quick response!

ehelms:
With the US holidays I have not gotten to this yet but I'll weigh in
beginning of next week.

mdellweg:
While i am quite indifferent about the inclusion-or-not question, i think that renaming the test data files breaks expectations of the (not yet so) automatic tests we have in the pipeline.

akofink:
@mdellweg Those tests aren't yet in the repo, so it's mostly a matter of which PR gets merged first (probably this one). When I rebase the testing PR I'll move the test files over.

modules/rhsm_manifest.py
```diff
@@ -0,0 +1,271 @@
...
+    return resp, info
+
+
+def create_manifest(module):
+    path = "/subscription/consumers"
+    data = {'name': module.params['name'],
+            'type': "satellite",
+            'owner': module.params['rhsm_owner'],
+            'facts': {'distributor_version': 'sat-6.0',
+                      'system.certificate_version': '3.2'}}
```
ehelms: Should either of these be configurable to protect against future changes in the portal?

modules/rhsm_manifest.py
```diff
@@ -0,0 +1,271 @@
...
+EXAMPLES = '''
+- name: Create katello.example.com Manifest and add 7 sub
+  rhsm_manifest:
+    name: "katello.example.com"
+    username: "john-smith"
+    password: "changeme"
+    pool_id: XXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXX
+    quantity: 7
+    state: present
+    validate_certs: false
```
ehelms: If I understand correctly, this example ensures that the manifest would have 7 of pool XXXXX?

modules/rhsm_manifest.py
```diff
@@ -0,0 +1,271 @@
...
+- name: Ensure my manifest has 10 of one subs in it and export
+  rhsm_manifest:
+    uuid: XXXXXXXX-XXXX-XXXX-XXXX-XXXXXXXXXXXX
+    username: john-smith
+    password: changeme
+    pool_id: XXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXX
+    quantity: 10
+    state: present
+    validate_certs: false
+    path: /root/manifest.zip
```
ehelms: If I understand correctly, this example ensures that the manifest would have 10 of pool XXXXXXXX AND would then download that manifest and store it?

modules/rhsm_manifest.py
sean797: Maybe..  

I have no idea what `system.certificate_version` actually does?
`distributor_version` should probably be, though I'm not sure what API we need to call to modify it on an already existing manifest, added a TODO comment for now.

Most of the APIs I got form https://github.com/Katello/hammer-cli-csv/commit/648163df7fe0dc0ead3ea0186eaba6d10d7f7b96. Maybe @thomasmckay has some insights?

modules/rhsm_manifest.py
sean797: Correct 👍

modules/rhsm_manifest.py
sean797: Correct 👍

sean797:
If you have got other suggestions on the name, I'm open to ideas.. I wouldn't focus too much on the reason I am doing it, others might have different reasons to do? Though I do not think there is currently any other reason to get a manifest.

Yes, the normal workflow would be attach subs, probably using `with_items`, then download it using another task. e.g I am currently using:

```yaml
- name: Attach subs
  rhsm_manifest:
    name: "{{ configure_katello_manifest_name }}"
    username: "{{ configure_rhsm_username }}"
    password: "{{ configure_rhsm_password }}"
    pool_id: "{{ item.pool_id }}"
    quantity: "{{ item.quantity }}"
    state: "{{ item.state }}"
    validate_certs: "{{ configure_rhsm_verify_ssl }}"
  with_items: "{{ configure_katello_subscriptions }}"
  when: configure_katello_subscriptions is defined
  register: subs

- name: Download manifest
  rhsm_manifest:
    name: "{{ configure_katello_manifest_name }}"
    username: "{{ configure_rhsm_username }}"
    password: "{{ configure_rhsm_password }}"
    validate_certs: "{{ configure_rhsm_verify_ssl }}"
    path: "{{ ansible_env['HOME'] }}/{{ configure_katello_manifest_name }}.zip"
  when:
    - configure_katello_manifest_name is defined
    - subs.changed
```


`UUID` is completely optional, I added it because name is not unique, it is possible to have 2 manifests in the portal with the same name (if name is supplied we just assume the first one)

ehelms:
I suppose you could then also chain this with (https://github.com/theforeman/foreman-ansible-modules/blob/master/modules/katello_manifest.py) to upload the exported manifest or initiate a refresh? 

I think this all sounds great and gives users a nice way until the Katello UI provides support to manage a manifest as code. Given there is a module named `redhat_subscription` in Ansible core, would `redhat_manifest` be too over the top to stay in line with that naming?

sean797:
@ehelms Yes! That's exactly what I do, take a look at https://github.com/sean797/ansible-role-configure-katello/blob/master/tasks/manifest.yml if you are interested. 

I agree with the name, updated!

ehelms:
Given others reviewed, I'd like to get an additional functionality/code review ACK from either @akofink or @ekohl

test/test_playbooks/rhsm_manifest.yml
```diff
@@ -0,0 +1,42 @@
+---
```
ekohl: Should this also be named ```redhat_manifest.yml``` to match the module name?

sean797:
Thanks again @ekohl. Updated.

akofink:
When adding a redhat manifest with 1 of POOL_ID, then updating the quantity attached to 3, I get this:

image

Is this expected/desired? Why not a single listing of 3?

modules/redhat_manifest.py
```diff
@@ -0,0 +1,274 @@
...
+        description:
+            - quantity of pool_id Subscriptions
+        type: int
+    state:
+        description:
+            - State of the Subscription
+        default: present
+        choices:
+            - present
+            - absent
```
akofink: `state` should be the redhat_manifest state, don't you think? I expect that if I use `state: absent`, the manifest will be removed from access.redhat.com.

sean797: If it was the manifest state, then you wouldnt be able to remove a sub from a manifest.

Do you think the name should be `redhat_manifest_sub` or something else? Or maybe if state == absent and there are no other subs in it then we delete the manifest?

sean797:
thanks for testing @akofink 👍  That is expected, I believe if you remove a sub from the manifest it can cause hosts to lose subs in when reimporting/refreshing it in Katello.

modules/redhat_manifest.py
akofink: Yeah, you're not going to like this response, but I think the task should represent the state of the manifest. Perhaps it'd be better to add pools/quantities in a dict like so (state in this example is the manifest state, not subscriptions):

```yaml
  - name: "Create {{ manifest_name }} manifest and add 5 rhel subs and ensure 0 rhev subs"
    redhat_manifest:
      name: "{{ manifest_name }}"
      username: "{{ rhsm_username }}"
      password: "{{ rhsm_password }}"
      subscriptions:
        - pool_id: "{{ rhsm_rhel_pool_id }}"
          quantity: 5
        - pool_id: "{{ rhsm_rhev_pool_id }}"
          quantity: 0
      state: present
      validate_certs: "{{ rhsm_validate_certs }}"
```

Sorry I didn't catch this earlier. We can discuss it more and let others give their input, but afaik every `state` param in the modules in the repo refer to the module's object itself (i.e. organization, product, etc.)

akofink: @mdellweg @ekohl  I'd like to get your opinions here.

ekohl: Without knowing too much details I think I agree with @akofink. state is at the manifest level so it should ensure the state of a manifest.

sean797: @akofink I did originally think about doing it like that, but I think its better to use `with_items` if you need to add multiple subs, I think having complex dict type options takes away form the simplicity of Ansible.

akofink: @sean797 You can still implement it in such a way that you can add each subscription one at a time if you prefer; however, users wishing to give their APIs a break will probably want to make a single API call for all their subscriptions.

> complex dict type options

I think Ansible is simple because it uses yaml instead of a DSL. I think users are quite comfortable with more complex yaml (and this is just a list of dicts), as it's intention is to be human-readable. The complexity more comes in the module implementation itself rather than what users will see, which I don't think is an issue.

mdellweg: I am really no expert for redhat subscriptions. But maybe that gives me the naïve viewpoint you are aiming for.
I agree, that the `state` parameter should really be used for the entity described in the module name.
However, if i understand the discussion right, and `quantity: 0` is not enough to delete a subscription, another `state: present|absent` on the subscription level could make the addition/deletion of subscriptions from the manifest more explicit.

akofink: > another state: present|absent on the subscription level could make the addition/deletion of subscriptions from the manifest more explicit

:+1:

sean797: Would everyone be okay with renaming the current `state` param to `subscription_state` & adding a `manifest_state` param?

akofink: I'm fine with it. Perhaps name it `pool_state` for consistency, and just `state` for the manifest state to follow what we're doing in the other ansible modules.

sean797:
Okay, finally found some time :D

I've:
- Renamed the old `state` var to `pool_state`
- Added `state` var that is the manifest state.
- Renamed ` validate_certs` var to `verify_ssl` to match the other modules in this repo.

modules/redhat_manifest.py
```diff
@@ -0,0 +1,293 @@
...
+    url = module.params['portal'] + path
+    headers = {'accept': accept_header,
+               'content-type': 'application/json'}
+    resp, info = fetch_url(module, url, json.dumps(data), headers, method)
+    if resp is None:
+        try:
+            error = json.loads(info['body'])['displayMessage']
+        except:
+            error = info['msg']
+        raise Exception(error)
```
mdellweg: I have the expression, that this Exception might not be caught (and transformed into module.fail_json) at every invocation of fetch_portal.

sean797:
Thanks @mdellweg, updated.

mdellweg:
I know, it might not be easy, but have you thought about automated testing for this module? There is a test_playbook...

test/test_crud.py
```diff
@@ -16,6 +16,7 @@
     'provisioning_template',
     'ptable',
     'repository',
+    'redhat_manifest',
```
mdellweg: Please keep this in alphabetical order.

sean797:
I've added some tests and it looks like Mr Travis is happy with the tests and hopefully your happy with the Module ordering ;) :+1:

mdellweg:
@sean797 Thank you for the patience.

PR #73, opened 2017-11-23
allow for aysnc katello_sync

sean797:
Note, this doesn't changed the current default

https://github.com/SatelliteQE/nailgun/blob/master/nailgun/entities.py#L4639
https://github.com/SatelliteQE/nailgun/blob/master/nailgun/entities.py#L5163

### DIFF
--- a/modules/katello_sync.py
+++ b/modules/katello_sync.py
@@ -52,6 +52,10 @@ options:
         description:
             - Product to which the repository lives in
         required: true
+    synchronous:
+        description:
+            - Wait for the Sync task to complete if True. Immediately return if False.
+        default: true
     organization:
         description:
             - Organization that the Product is in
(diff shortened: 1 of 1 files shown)


sean797:
@akofink thanks for taking a look; one thing just came to mind, this can already be done like:

````yaml
- name: "Sync repository"
   katello_sync:
     username: "admin"
     password: "changeme"
     repository: "My repository"
     product: "My Product"
  async: 45  # maximum runtime
  poll: 0
```
More can be read here http://docs.ansible.com/ansible/latest/playbooks_async.html

I think I might change this PR to just document that?

akofink:
@sean797 I believe this would result in different behavior. When you give `async: true` to the API, it immediately returns with the task ID and the user can choose to poll that task for updates. When `async: false` is passed to the API, it automatically sends responses every X seconds (not sure what X is, probably configurable) until the task stops or pauses. The way the PR is currently implemented seems like the 'right' way, don't you think?

sean797:
@akofink 5 seconds ;) https://github.com/SatelliteQE/nailgun/blob/master/nailgun/entity_mixins.py#L36

Yea I guess they are slightly different use cases.

PR #76, opened 2017-11-29
Record vcr-data

mdellweg:
This is a proof of concept by now. But i think it is in a shape to be discussed upon.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -29,6 +29,7 @@ test_%: FORCE
 	pytest -k $*
 
 record_%: FORCE
+	$(RM) test/test_playbooks/fixtures/$*-*.yml
 	pytest -k $* --record
 
 debug:
--- a/Makefile
+++ b/Makefile
@@ -11,13 +11,15 @@ DEBUG_OPTIONS=-m $(MODULE_PATH) -a @$(DEBUG_DATA_PATH) -D $(PDB_PATH)
 default: help
 help:
 	@echo "Please use \`make <target>' where <target> is one of:"
-	@echo "  help         to show this message"
-	@echo "  lint         to run flake8 and pylint"
-	@echo "  test         to run unit tests"
-	@echo "  debug        debug a module using the ansible hacking module"
-	@echo "  setup        to set up test, lint, and debugging"
-	@echo "  test-setup   to install test dependencies"
-	@echo "  debug-setup  to set up the ansible hacking module"
+	@echo "  help           to show this message"
+	@echo "  lint           to run flake8 and pylint"
+	@echo "  test           to run unit tests"
+	@echo "  debug          debug a module using the ansible hacking module"
+	@echo "  setup          to set up test, lint, and debugging"
+	@echo "  test-setup     to install test dependencies"
+	@echo "  debug-setup    to set up the ansible hacking module"
+	@echo "  test_<test>    to run a specific unittest"
+	@echo "  record_<test>  to (re-)record the server answers for a specific test"
 
 lint:
 	pycodestyle --ignore=E402,E722 --max-line-length=160 modules test
--- a/Makefile
+++ b/Makefile
@@ -25,6 +25,12 @@ lint:
 test:
 	pytest $(TEST)
 
+test_%: FORCE
+	pytest -k $*
+
+record_%: FORCE
+	pytest -k $* --record
+
 debug:
 ifndef MODULE
 	$(error MODULE is undefined)
@@ -43,4 +49,6 @@ test-setup: test/test_playbooks/server_vars.yml
 test/test_playbooks/server_vars.yml:
 	cp test/test_playbooks/server_vars.yml.example test/test_playbooks/server_vars.yml
 
-.PHONY: help debug lint test setup debug-setup test-setup
+FORCE:
+
+.PHONY: help debug lint test setup debug-setup test-setup FORCE
--- a/README.md
+++ b/README.md
@@ -51,14 +51,15 @@ To run the tests:
 
 ```sh
 make test # all tests
-make test TEST="-k product" # single module test
+make test_product  # single test
+make test TEST="-k 'organzation or product'"  # select tests by expression (see `pytest -h`)
 ```
 
 The tests are run against prerecorded server-responses.
-If you need to (re-)record some cassettes, delete the stale ones in `test/test_playbooks/fixtures` and record new ones with
+You can (re-)record the cassettes for a specific test with
 
 ```sh
-pytest -k <module name> --record
+make record_<test name>
 ```
 
 ## How to debug modules in this repository
--- a/README.md
+++ b/README.md
@@ -54,6 +54,13 @@ make test # all tests
 make test TEST="-k product" # single module test
 ```
 
+The tests are run against prerecorded server-responses.
+If you need to (re-)record some cassettes, delete the stale ones in `test/test_playbooks/fixtures` and record new ones with
+
+```sh
+pytest -k <module name> --record
+```
+
 ## How to debug modules in this repository
 
 Set up debugging using ansible's test-module
--- a/test/inventory/hosts
+++ b/test/inventory/hosts
@@ -1,0 +1,2 @@
+fixtures ansible_connection=local
+tests ansible_connection=local ansible_python_interpreter=../vcr_python_wrapper.py
--- a/test/test_crud.py
+++ b/test/test_crud.py
@@ -28,7 +28,7 @@ def run_playbook_vcr(module, extra_vars=None, extra_args=None, record=False):
         json.dump(test_params, params_file.file)
         params_file.file.close()
         os.environ['FAM_TEST_VCR_PARAMS_FILE'] = params_file.name
-        return run_playbook(module, extra_args=extra_args, extra_vars=extra_vars)
+        return run_playbook(module, extra_vars=extra_vars, extra_args=extra_args)
 
 
 def run_playbook(module, extra_vars=None, extra_args=None):
--- a/test/test_crud.py
+++ b/test/test_crud.py
@@ -1,4 +1,6 @@
 import pytest
+import os
+import tempfile
 import json
 from ansible.cli.playbook import PlaybookCLI
 
@@ -20,14 +22,17 @@ def run_playbook(module, record=False, extra_vars=None):
         playbook_opts.extend(['--limit', 'tests'])
     if extra_vars:
         playbook_opts.extend(['--extra-vars', extra_vars])
-    # Dump recording parameters to json-file
+
+    # Dump recording parameters to json-file and pass its name by environment
     test_params = {'test_name': module, 'serial': 0, 'record_mode': record_mode}
-    with open('.vcr_params.json', 'w') as params_file:
-        json.dump(test_params, params_file)
-    # Call the playbook
-    cli = PlaybookCLI(playbook_opts)
-    cli.parse()
-    return cli.run()
+    with tempfile.NamedTemporaryFile('w', suffix='json', prefix='fam-vcr-') as params_file:
+        json.dump(test_params, params_file.file)
+        params_file.file.close()
+        os.environ['FAM_TEST_VCR_PARAMS_FILE'] = params_file.name
+        # Call the playbook
+        cli = PlaybookCLI(playbook_opts)
+        cli.parse()
+        return cli.run()
 
 
 @pytest.mark.parametrize('module', MODULES)
--- a/test/test_crud.py
+++ b/test/test_crud.py
@@ -8,10 +8,11 @@ from ansible.cli.playbook import PlaybookCLI
 MODULES = ['organization', 'product']
 
 
-def run_playbook(module, record=False, extra_vars=None):
-    # Assemble parameters for playbook call
-    path = 'test/test_playbooks/{}.yml'.format(module)
-    playbook_opts = ['ansible-playbook', path, '--inventory', 'test/inventory/hosts']
+def run_playbook_vcr(module, extra_vars=None, extra_args=None, record=False):
+    # Assemble extra parameters for playbook call
+    if extra_args is None:
+        extra_args = []
+    extra_args.extend(['--inventory', 'test/inventory/hosts'])
     if record:
         # Cassettes that are to be overwritten must be deleted first
         record_mode = 'once'
@@ -19,9 +20,7 @@ def run_playbook(module, record=False, extra_vars=None):
         # Never reach out to the internet
         record_mode = 'none'
         # Only run the tests (skip fixtures)
-        playbook_opts.extend(['--limit', 'tests'])
-    if extra_vars:
-        playbook_opts.extend(['--extra-vars', extra_vars])
+        extra_args.extend(['--limit', 'tests'])
 
     # Dump recording parameters to json-file and pass its name by environment
     test_params = {'test_name': module, 'serial': 0, 'record_mode': record_mode}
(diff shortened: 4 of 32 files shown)


test/inventory/hosts
```diff
@@ -0,0 +1,4 @@
+[fixtures-group]
+fixtures ansible_connection=local
+[tests-group]
```
akofink: you shouldn't need groups here

test/vcr_python_wrapper.py
```diff
@@ -0,0 +1,23 @@
...
+VCR_PARAMS_FILE = '../../.vcr_params.json'
+
+sys.argv.pop(0)
+
+# Load recording parameters from file
+with open(VCR_PARAMS_FILE, 'r') as params_file:
+    test_params = json.load(params_file)
+cassette_file = 'fixtures/{}-{}.yml'.format(test_params['test_name'], test_params['serial'])
+# Increase serial and dump back to file
+test_params['serial'] += 1
```
akofink: We'll definitely need to change this :laughing: The vcr file should describe the current state of the server. Something like 'organization_absent.yml' or 'product_present.yml'. Though, I know you did this just to get it working.

mdellweg: If you find a method to get ansible to pass the name of the task to the module, this would be easy. Right now, it depends on the playbook being run in a specific order.

ekohl: Would you be able to derive it from ```sys.argv[0]```?

mdellweg: that would be something like `ansible_module_foreman_ptable.py`. So sadly no. And deriving it from the module name and state, would not be enough information.
I would prefer something like:
product_present_created.yml
product_present_updated.yml
product_present_unchanged.yml
product_absent_deleted.yml
and so on...
By the way: Can you make a Playbook task fail on changed==True?

test/vcr_python_wrapper.py
```diff
@@ -0,0 +1,25 @@
...
+with open(VCR_PARAMS_FILE, 'w') as params_file:
+    json.dump(test_params, params_file)
+
+# Call the original python script with vcr-cassette in place
+with vcr.use_cassette(cassette_file,
+                      record_mode=test_params['record_mode'],
+                      match_on=['method', 'scheme', 'port', 'path', 'query']):
+    execfile(sys.argv[0])
```
mdellweg: python3 removed execfile :anger:

ekohl: https://stackoverflow.com/a/16577427 suggests ```exec(open(sys.argv[0]).read())``` as an alternative.

ekohl: Or https://docs.python.org/3/library/runpy.html#runpy.run_path

test/inventory/hosts
akofink: You actually shouldn't need an inventory file if you pass everything inline in the ansible-playbook command: https://github.com/theforeman/foreman-ansible-modules/compare/master...akofink:vcr#diff-33830cc555322fe03af4593b369e4359R11

test/vcr_python_wrapper.py
mdellweg: > For a simple script, the specified code is simply executed in a fresh module namespace.
Sonds like runpy creates some 'clean environment' the script runs in. That's exactly what we do not want. ;)

test/vcr_python_wrapper.py
akofink: Yes, if you use sys.argv[0] (after pop(0)), you'll get fixtures named like this `ansible_module_foreman_organization.py.yml`. Then if we could index those, we'd have `ansible_module_foreman_organization.pyX.yml` where X is in [0, 1, 2, ...].

mdellweg: Then we'd have the name of the module, but we need the file for counting and record_mode anyway (not much gained IMHO).
It'd really like to call the fixtures the same as what is given as the `name` of the individual task.

test/inventory/hosts
mdellweg: I like the inventory as a file (not too strong feelings, though).

test/test_playbooks/product.yml
```diff
@@ -1,12 +1,19 @@
...
+  gather_facts: false
   tasks:
   - name: 'Load server config'
     include_vars:
       file: server_vars.yml
   - include: tasks/organization.yml
     vars:
       organization_state: present
+- hosts: tests
+  gather_facts: false
```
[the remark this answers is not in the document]
mdellweg: I think, gathering facts here is unnecessarily spent time. And it adds another call to the vcr_wrapper, shifting the serial numbers by one.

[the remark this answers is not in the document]
akofink: Ah, I see. Okay! :)

test/vcr_python_wrapper.py
```diff
@@ -0,0 +1,29 @@
+#!/usr/bin/env python
+
+import os
+import sys
+import vcr
+import json
+
+VCR_PARAMS_FILE = '../../.vcr_params.json'
```
ekohl: I'm thinking about passing this path as an environment variable and using ```mktemp```. Maybe we'll be able to run things in parallel at some point and then it's useful.

mdellweg: Have a try! :)
I think the environment get's cleared by ansible. Otherwise, we could pass a lot of interesting information that way.

ekohl: Looks like that works: https://github.com/ekohl/foreman-ansible-modules/commit/30386617792f5d022c5c00ad572006588f0778cc

mdellweg: Seems i have been mistaken about the environment. 🎆

mdellweg:
I have introduced secure tempfiles. That should (in principle) even allow us to run the tests concurrently.
Thanks @ekohl !
BTW: Having the enviroment, the serial (for naming the fixtures) is really the last bit that needs this file anyway. Everything else, we can put directly in the environment.

test/test_crud.py
```diff
@@ -20,14 +22,17 @@ def run_playbook(module, record=False, extra_vars=None):
...
     test_params = {'test_name': module, 'serial': 0, 'record_mode': record_mode}
-    with open('.vcr_params.json', 'w') as params_file:
-        json.dump(test_params, params_file)
-    # Call the playbook
-    cli = PlaybookCLI(playbook_opts)
-    cli.parse()
-    return cli.run()
+    with tempfile.NamedTemporaryFile('w', suffix='json', prefix='fam-vcr-') as params_file:
+        json.dump(test_params, params_file.file)
+        params_file.file.close()
```
ekohl: I think this won't interfere with the contextmanager but otherwise using ```flush()``` might be sufficient.

ekohl:
> BTW: Having the enviroment, the serial (for naming the fixtures) is really the last bit that needs this file anyway. Everything else, we can put directly in the environment.

I had exactly the same issue and didn't know how to handle the serial. Is it really needed or is there a record mode that appends unknown?

test/test_crud.py
mdellweg: I think flush or close doesn't matter. Since we were finished with writing, i thought we could as well close it.

mdellweg:
@ekohl As far as my investigations went, it is really needed. VCRpy wont record the same request twice on the same cassette.

test/test_crud.py
akofink: Wait, I thought you didn't have to explicitly call close() when you use a block like this.

ekohl: You don't, but due to buffering the file might not be written. Since the Ansible process wants to use this file, it might be empty. In this case the closing part also removes it.

mdellweg: The contextmanager wraps the Tempfile_object. I am closing/flushing the embedded file object.
Hence `params.*file*.close()`.
Leaving the with context also deletes the tempfile which i think is nice behaviour.

README.md
```diff
@@ -54,6 +54,13 @@ make test # all tests
 make test TEST="-k product" # single module test
 ```
 
+The tests are run against prerecorded server-responses.
+If you need to (re-)record some cassettes, delete the stale ones in `test/test_playbooks/fixtures` and record new ones with
+
+```sh
+pytest -k <module name> --record
```
akofink: Yeah, or `make test TEST="-k <module name> --record"`

test/test_crud.py
```diff
@@ -1,16 +1,40 @@
...
+    if extra_vars:
+        playbook_opts.extend(['--extra-vars', extra_vars])
 
+    # Dump recording parameters to json-file and pass its name by environment
+    test_params = {'test_name': module, 'serial': 0, 'record_mode': record_mode}
+    with tempfile.NamedTemporaryFile('w', suffix='json', prefix='fam-vcr-') as params_file:
+        json.dump(test_params, params_file.file)
+        params_file.file.close()
+        os.environ['FAM_TEST_VCR_PARAMS_FILE'] = params_file.name
+        # Call the playbook
```
akofink: Everything up to this line in this function should probably just be put in a different function. #justsayin
Also, you'd be able to remove this comment if you're instead calling a function named `run_playbook`.

test/test_playbooks/product.yml
```diff
@@ -1,22 +1,46 @@
...
+      expected_change: true
   - include: tasks/product.yml
     vars:
       product_state: absent
+      expected_change: true
+  - include: tasks/product.yml
+    vars:
+      product_state: absent
+      expected_change: false
+- hosts: fixtures
```
akofink: Could you just explain the fixtures vs. tests to me? I get that fixtures are setting up data for the tests, but is the purpose of setting the host like this for reusing fixture cassettes?

test/test_playbooks/product.yml
```diff
@@ -1,22 +1,46 @@
...
+- hosts: tests
+  gather_facts: false
+  tasks:
+  - name: 'Load server config'
+    include_vars:
+      file: server_vars.yml
+  - include: tasks/product.yml
+    vars:
+      product_state: present
+      expected_change: true
```
akofink: Very nice.

test/test_playbooks/product.yml
mdellweg: The fixtures are being run without recording (see inventory).
In "normal mode" the tests are run alone from tape.
For "recording mode", the fixtures prepare the environment. And tear it down. And the tests run with recording against a real server.
Maybe we should include an additional mode that runs the whole playbook without recording against a real server...

test/test_crud.py
mdellweg: Improving the structure... ACK!

test/test_playbooks/product.yml
akofink: Ah thanks, that makes sense. I'd be fine without the additional mode for this PR. It seems like something that'd be easy to add later.

mdellweg: We need a githubfeature: Open a this issue automatically when this PR is merged.

README.md
mdellweg: sounds a bit more bulky, and if you are at that stage (recording a new tape) you might have run the test setup already. But for consistency...
We could make it `make record TEST=...`.

akofink: Nah, let's leave it as is.

test/test_crud.py
mdellweg: On a second thought, the assembling of the playbook parameters would be much more complicated if we split this into two functions. And the run_playbook function must be called inside the other one anyway. Right now, i cant see a way to make this really nice.

akofink: One suggestion: move L33-L35 to a function named `run_playbook(...)` and rename this function to a more descriptive name, such as `run_vcr_playbook(...)`. :man_shrugging:

mdellweg: Sorry, a minute too late...

Makefile
```diff
@@ -25,6 +25,12 @@ lint:
 test:
 	pytest $(TEST)
 
+test_%: FORCE
+	pytest -k $*
+
+record_%: FORCE
+	pytest -k $* --record
+
```
akofink: :1st_place_medal:

mdellweg:
Would be nice to have a second code review on a big PR like this.
@ehelms @evgeni

mdellweg:
I think, we should remove the authorization from the tape files before merging this.

mdellweg:
@ekohl Thanks for the review!
Yes, the tests (org and prod by now, see #77 ) are actually running on travis.

test/test_playbooks/product.yml
ehelms: What does expected_change do?

evgeni: it's used in the playbook to verify that the state of the object was or was not changed: https://github.com/theforeman/foreman-ansible-modules/pull/76/files#diff-f5f51acb20b549427b6bc248abf7cc95R14

test/test_crud.py
```diff
@@ -1,16 +1,50 @@
...
+        # Only run the tests (skip fixtures)
+        extra_args.extend(['--limit', 'tests'])
+
+    # Dump recording parameters to json-file and pass its name by environment
+    test_params = {'test_name': module, 'serial': 0, 'record_mode': record_mode}
+    with tempfile.NamedTemporaryFile('w', suffix='json', prefix='fam-vcr-') as params_file:
+        json.dump(test_params, params_file.file)
+        params_file.file.close()
+        os.environ['FAM_TEST_VCR_PARAMS_FILE'] = params_file.name
+        return run_playbook(module, extra_args=extra_args, extra_vars=extra_vars)
```
evgeni: nitpick: args and vars are sorted the other way round in the `run_playbook` signature, so you might want to have them properly here too (even so it does not matter as you call them named).

README.md
```diff
@@ -54,6 +54,13 @@ make test # all tests
 make test TEST="-k product" # single module test
 ```
 
+The tests are run against prerecorded server-responses.
+If you need to (re-)record some cassettes, delete the stale ones in `test/test_playbooks/fixtures` and record new ones with
```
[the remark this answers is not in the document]
mdellweg: The recording mode chosen here 'once' will only record requests if the tape file is new. It will not add new requests to an existing file.
Given that we broke down the fixtures to one tape file per module call, i think it is safer to not accidentially adding new requests (and keep the old stale requests around).

test/test_crud.py
mdellweg: Will adjust that ;)

mdellweg:
@evgeni I have no idea about requests compatibility. I guess, this problem is common enough, that vcrpy navigates around it.
Actually, we are not matching against headers, i think:
> `match_on=['method', 'scheme', 'port', 'path', 'query']`

evgeni:
Ah, cool, thanks for the explanation!

ehelms:
There are other modes available right? The workflow I am thinking about is
in Katello. Where, even if you specify mode=all, as long as you limited the
test to specific test method to execute, it will only re-record the
cassette for that individual test and not for "all". Does that make sense?
If yes, would this be possible to do?

If the above isn't, could we delete the cassette if "once" is specified
alongside an individual test?

mdellweg:
First of all, i think the _re_recording should be used only in rare cases. Not sure if this is an actual workflow.
Having to delete a cassette by hand could raise the awareness of the tester and makes him/her think again why a change might be necessary and what tapes really need to be rerecorded. Of course having better file names would help here.

mdellweg:
I just realised, due to the interdependency of the individual test tasks in the same playbook on the internal state of katello, it only makes sense to rerecord all cassette tape files in one go.
Rebuilding a single one in the middle would never do. I will add an according line in the Makefile.

ekohl:
It might also be a good idea to use the version number it was recorded against in the filename. That way we could verify it works against multiple versions. Side benefit is that you have a minimum supported version.

akofink:
@ekohl if only we could rely on the API version... But yes, maybe a config var for foreman_version would be good.

ekohl:
@akofink even adding fields is changing the API. It's compatible but if our code starts to rely on those fields it's a problem

mdellweg:
But we do not change any behaviour depending on the API version, do we? I thought we left that part to the nailgun.

akofink:
We have 2 approvals. Any objections to merging this?

mdellweg:
If you count @ekohl, it makes 3. :) I am ok with merging at this stage.

PR #79, opened 2017-11-30
WIP: Add support for managing composite content views

gregswift:
In our work flow we rely heavily on CCVs, and the other modules here help us automate the rest of the work flow so thought we'd give it a go on making this last piece happen.

As of submission time I have never run the code. I'm about 95% sure it errors.  This is just to start the discussion.

I was going to just make repositories and content_views as mutually_exclusive parameters in katello_content_view, but decided it might be nice have it be distinctive for consumers so the difference was easier to roll with.

katello_content_view hasn't been moved to cement yet, but I am trying to follow that pattern on this module.


mdellweg:
> I'm about 95% sure it errors. This is just to start the discussion.
I added labels based on your wording. ;)
Not a proper review yet, but a thought, that comes to my mind:
Would it be possible, to handle this inside the content_view module? Is a CCV not like a plain CV with a flag that it contains other VCs?

mdellweg:
I appreciate, that you try to follow the behaviour of the existing modules.
Personally, i would prefer, if you'd follow the pattern found in e.g. `foreman-global-parameter` or `katello-product`. It reduces a lot of boilerplate code (and we are planning on reducing even more, once automated testing is in place).
Thoughts? @akofink @ehelms @evgeni @Fobhep

gregswift:
> Would it be possible, to handle this inside the content_view module? Is a CCV not like a plain CV with a flag that it contains other VCs?

My response to that is:
> I was going to just make repositories and content_views as mutually_exclusive parameters in katello_content_view, but decided it might be nice have it be distinctive for consumers so the difference was easier to roll with.

To expand upon that... while CVs and CCVs are the same thing, they also really arent.  When using them in your workflow its good to think of them separately. Heck, I wish Katello would treat them as Content Views and Composite Views from a usability standpoint even if the code on the backside is 98% identical. 

That being said I can move the code into `katello_content_view` if that is the preference.. yall are gonna support this longer term more than I will.

> Personally, i would prefer, if you'd follow the pattern found in e.g. foreman-global-parameter or katello-product. It reduces a lot of boilerplate code (and we are planning on reducing even more, once automated testing is in place).
I thought that is what I was doing.... ?

akofink:
If it's feasible (which I think it is), I'd prefer this be in the katello_content_view module. It'd also be a great opportunity to convert that module to use cement. Some ideas:

1. The API uses `composite: true/false`. We could add that as a param, or...
1. If the user specifies `component_content_views: [...]` it's automatically composite and disallows `repositories`. That doesn't seem too difficult to implement.

ehelms:
After reading and thinking, I think I land on the side of be explicit. Given that composites handle things differently than a standard content view, having a module that manages them explicitly adds user value in my opinion. From a code perspective, I think its natural to land on the side of just add a switch `composite: true` to the existing module and then ensure that we have two other parameters conflict (repositories and content_views) and share all the code with if/else statements and error messages. However, I think you can achieve the same thing through the original suggestion of a `katello_composite_content_view` while making the playbooks users write more obvious. This also simplifies the if/else checks and requirements to spit out error messages.

akofink:
I'm fine with it as a separate module.

mdellweg:
I'm fine, too. No strong feelings here. Thanks to all for the good discussion. ;)

modules/katello_composite_content_view.py
```diff
@@ -0,0 +1,176 @@
...
+        org.search()
+    except Exception as e:
+        module.fail_json(msg="Failed to connect to Foreman server: %s " % e)
+
+    kwargs = {}
+    if content_views:
+        kwargs['content_views'] = content_views
+
+    try:
+        changed = ng.content_view(name, organization, **kwargs)
```
akofink: Could we use `naildown_entity_state` with `sanitize_composite_content_view_dict`?

akofink:
Any updates here?

gregswift:
sorry.. need to get back on it.  gonna go do some of it right now

modules/katello_composite_content_view.py
```diff
@@ -0,0 +1,176 @@
...
+# You should have received a copy of the GNU General Public License
+# along with Ansible.  If not, see <http://www.gnu.org/licenses/>.
+
+DOCUMENTATION = '''
+---
+module: katello_composite_view
+short_description: Create and Manage Katello composite content views
+description:
+    - Create and Manage Katello composite content views
+author: "Eric D Helms (@ehelms)"
```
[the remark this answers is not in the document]
gregswift: not like i'm doing much, but sure

modules/katello_composite_content_view.py
gregswift: whats `sanitize_composite_content_view_dict` ?

modules/katello_composite_content_view.py
```diff
@@ -0,0 +1,176 @@
...
+        verify=verify_ssl
+    )
+    ng = NailGun(server, entities, module)
+
+    # Lets make an connection to the server with username and password
+    try:
+        org = entities.Organization(server)
+        org.search()
+    except Exception as e:
+        module.fail_json(msg="Failed to connect to Foreman server: %s " % e)
```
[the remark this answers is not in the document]
gregswift: so replace 157-162 with `create_server` and `ping_server` ? as per katello_product.py 150-155

modules/katello_composite_content_view.py
gregswift: so i found the reference for this... almost seems like there should be a `sanitize_dict` in module_utils hat takes a map from each object type...  all 9 current `sanitize_*_dict` are just separate functions with a different map and the same 5 lines of code finishing the function

gregswift:
Okay.. I'm at a testing point. Will probably test next week when i'm back at the office.  I'm curious to see #86 move forward though, i dont like that i dontthink the test i added can really be used.

modules/katello_composite_content_view.py
```diff
@@ -0,0 +1,164 @@
...
+        ping_server,
+        set_task_timeout,
+    )
+    HAS_NAILGUN_PACKAGE = True
+except:
+    HAS_NAILGUN_PACKAGE = False
+from ansible.module_utils.foreman_helper import handle_no_nailgun
+
+from ansible.module_utils.basic import AnsibleModule
+from ansible.module_utils.foreman_helper import handle_no_nailgun
```
mdellweg: This is imported twice. See line 96.

test/test_playbooks/composite_content_view.yml
```diff
@@ -0,0 +1,17 @@
...
+    katello_content_view:
+      username: "{{ foreman_username }}"
+      password: "{{ foreman_password }}"
+      server_url: "{{ foreman_server_url }}"
+      verify_ssl: "{{ foreman_verify_ssl }}"
+      name: "ccv5"
+      organization: "Default Organization"
+      content_views:
+        - name: "cv3"
+          version "1.0"
```
mdellweg: You can add the test to the test suite (of PR #62 ) already.
https://github.com/theforeman/foreman-ansible-modules/blob/412507090f314263b774aa2e82ca47fa4ff9cfee/test/test_crud.py#L5
You can then run the test with `make test TEST="-k composit_content_view"`.

gregswift: ok. but what about the dependencies. it needs the products, repos, cv, and cv version to exist

mdellweg: You can look at the Product playbook. It creates the needed Organization first and removes it again afterwards.

gregswift: did that now... heh.. :D

test/test_crud.py
```diff
@@ -2,7 +2,12 @@
 from ansible.cli.playbook import PlaybookCLI
 
 
-MODULES = ['organization', 'product']
+MODULES = [
+    'organization',
+    'product',
+    'content_view',
+    'composite_content_view'
+]
```
mdellweg: While we are at it, let's get them in alphabetical order.
This list is gonna be long. ;)

gregswift: i almost did.. will do

test/test_playbooks/content_view.yml
```diff
@@ -4,16 +4,28 @@
...
-      repositories:
-        - name: "repo1"
-          product: "product1"
-        - name: "zoo"
-          product: "zoo product"
+  - name: Build the things
+    vars:
+      content_view_state: present
+    block:
+    - include_task: tasks/organization.yml
```
mdellweg: This is `include_tasks:`.
Also `organization_state` is required.

test/test_playbooks/tasks/repository.yml
```diff
@@ -0,0 +1,20 @@
...
+  katello_repository:
+    username: "{{ foreman_username }}"
+    password: "{{ foreman_password }}"
+    server_url: "{{ foreman_server_url }}"
+    verify_ssl: "{{ foreman_verify_ssl }}"
+    organization: "{{ organization_name }}"
+    name: "{{ repository_name }}"
+    product: "{{ product_name }}"
+    content_type: "{{ repository_content_type }}"
+    url: "{{ repository_url }}"
```
mdellweg: I guess, this is `{{ repository_content_url }}`.

test/test_playbooks/tasks/repository.yml
```diff
@@ -0,0 +1,20 @@
+---
+- name: "Create katello repository"
+  vars:
+    - organization: "Test Organization"
+    - product_name: "product1"
```
mdellweg: For consistency these might be called `repository_{organization,product}`

modules/katello_content_view.py
```diff
@@ -183,12 +183,7 @@ def main():
     )
     ng = NailGun(server, entities, module)
 
-    # Lets make an connection to the server with username and password
-    try:
-        org = entities.Organization(server)
-        org.search()
-    except Exception as e:
-        module.fail_json(msg="Failed to connect to Foreman server: %s " % e)
+    ping_server(module)
```
mdellweg: Nice catch! But you must import it from cement first.

modules/katello_composite_content_view.py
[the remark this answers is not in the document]
akofink: Yes

modules/katello_composite_content_view.py
akofink: I agree this could be more DRY. If you'd like, we can open an issue to address it, and ignore the repetitiveness for this PR. I'm not sure Nailgun is the appropriate place for it.

mdellweg: I have a branch to start cleaning this up, but i wanted to wait for the automated tests for this kind of refactoring.

mdellweg:
This needs a rebase (although github does not seem to know).
The test infrastructure has changed slightly and the tests here should be changed accordingly.

Also i keep thinking, that since this entity has so many dependencies, a bunch of tests for those entities seem to arise automatically. Do you want to add them?

test/test_playbooks/tasks/repository.yml
gregswift: I was going for simplicity of using the same data in multiple places.

test/test_playbooks/tasks/repository.yml
gregswift: I was going with `<module>_<param>` and so it would be `repository_url` just like `repository_content_type`

gregswift:
the test playbook was actually missing all the state variables being properly defined... oops 😬

test/test_playbooks/tasks/repository.yml
akofink: I'd agree with @gregswift here. I think any top-level attributes can be thought of as repository_<attribute> without being explicit.

akofink: Updated above comment.

mdellweg: As long, as there cannot arise conflicts, i'm fine.

gregswift:
Test is currently failing on Python 2.7 because of timeout connecting to github on 443.. Passed on Python 3.5

mdellweg:
Removed WIP, since you are obviously fixing the last issues.
I guess, the last build failed because my merge crossed this one. ;)

modules/katello_composite_content_view.py
```diff
@@ -0,0 +1,163 @@
...
+    handle_no_nailgun(module, HAS_NAILGUN_PACKAGE)
+
+    entity_dict = dict(
+        [(k, v) for (k, v) in module.params.items() if v is not None])
+
+    server_url = entity_dict.pop('server_url')
+    username = entity_dict.pop('username')
+    password = entity_dict.pop('password')
+    verify_ssl = entity_dict.pop('verify_ssl')
+    state = entity_dict.pop('state')
```
mdellweg: I am missing this in the argument_spec as well as the documentation.

test/test_playbooks/composite_content_view.yml
```diff
@@ -0,0 +1,37 @@
+---
+- hosts: localhost
```
mdellweg: You need to restructure the playbooks. Take a look at i.e. the operating_system playbook. Also you might want to reread the README. ;)
By now, the tests are just not run, because localhost is not in the inventory.

modules/katello_composite_content_view.py
gregswift: So I went and stole it (just now) from katello_product and noticed that the doc says `latest, present, absent` but the argument spec says `'present_with_defaults', 'present', 'absent'` ?

mdellweg:
You are getting close. But at some point, you actually need to record the answers to the api requests.
`make record_composite_content_view`
This should create files in `test/test_playbooks/fixtures`, which you need to check in so that travis can run the tests.

gregswift:
Any reason some of the modules dont have state?

gregswift:
Tried to convert katello_repository and run the record and i keep getting stuck where it cant find the product from the repo module

```
TASK [Create katello repository] ***********************************************
fatal: [fixtures]: FAILED! => {"changed": false, "module_stderr": "WARNING:nailgun.client:Received HTTP 404 response: {\"displayMessage\":\"Couldn't find Katello::Product with 'id'=Test_Product\",\"errors\":[\"Couldn't find Katello::Product with 'id'=Test_Product\"]}\nTraceback (most recent call last):\n  File \"/tmp/ansible_JXQFE2/ansible_module_katello_repository.py\", line 225, in <module>\n    main()\n  File \"/tmp/ansible_JXQFE2/ansible_module_katello_repository.py\", line 215, in main\n    entity = find_repository(module, name=entity_dict['name'], product=entity_dict['product'])\n  File \"/tmp/ansible_JXQFE2/ansible_modlib.zip/ansible/module_utils/ansible_nailgun_cement.py\", line 285, in find_repository\n  File \"/usr/lib/python2.7/site-packages/blinker_herald/base.py\", line 137, in wrapper\n    result = fn(*args, **kwargs)\n  File \"/usr/lib/python2.7/site-packages/nailgun/entity_mixins.py\", line 1358, in search\n    results = self.search_json(fields, query)['results']\n  File \"/usr/lib/python2.7/site-packages/nailgun/entity_mixins.py\", line 1210, in search_json\n    response.raise_for_status()\n  File \"/usr/lib/python2.7/site-packages/requests/models.py\", line 935, in raise_for_status\n    raise HTTPError(http_error_msg, response=self)\nrequests.exceptions.HTTPError: 404 Client Error: Not Found for url: https://katello.prod.dfw1.cit.rackspace.net//katello/api/v2/repositories\n", "module_stdout": "", "msg": "MODULE FAILURE", "rc": 0}
	to retry, use: --limit @/home/xaeth/Development/foreman-ansible-modules/test/test_playbooks/composite_content_view.retry
```

gregswift:
The last comment is still valid after last 2 commits.. needed to push them so you had what i have.

mdellweg:
> Any reason some of the modules dont have state?

You could call them old-style modules. They actually predate the cement library. See #37.

I would think, https://github.com/theforeman/foreman-ansible-modules/pull/79/commits/65cce968deb0842e8c23e02b0a3ad6b913fc7cad together with a corresponding test is worth it's own PR. It is pretty straight forward and could be merged quite fast and even adds a feature (`state`). BTW, thank you there!

modules/katello_repository.py
```diff
@@ -73,6 +73,14 @@
             - immediate
             - on_demand
         required: false
+    state:
+        description:
+        - State of the Global Parameter
```
mdellweg: Repository...

modules/katello_repository.py
```diff
@@ -73,6 +73,14 @@
...
+        default: true
+        choices:
+        - present
+        - lastest
+        - absent
```
mdellweg: you already know, i believe. ;)

modules/katello_repository.py
```diff
@@ -175,42 +189,36 @@ def main():
...
         module.fail_json(msg="Failed to connect to Foreman server: %s " % e)
 
-    try:
-        changed = ng.repository(name, content_type, product, organization, url=url, download_policy=download_policy)
-        module.exit_json(changed=changed)
-    except Exception as e:
-        module.fail_json(msg=e)
+    ping_server(module)
+
+    entity = find_repository(module, name=entity_dict['name'], product=entity_dict['product'])
```
mdellweg: I think, your runtime error is based here. `entity_dict['product']` is the name of the product at this point while it should be a nailgun entity object. I think, you need a `find_product` first.

test/test_playbooks/content_view.yml
```diff
@@ -1,19 +1,46 @@
...
-- hosts: localhost
-  tasks:
-  - name: Load server config
-    include_vars:
-      file: server_vars.yml
-  - name: "Create/update katello content view"
-    katello_content_view:
+- hosts: tests
+  vars:
+    auth_params: &auth_params
```
mdellweg: Sorry for the confusion here. This YAML referencing was an experiment of mine. It would have been really cool if it worked across file boundaries, but it does not.

test/test_playbooks/tasks/composite_content_view.yml
```diff
@@ -0,0 +1,18 @@
+---
+- name: "Create/update katello content view"
+  vars:
+    - organization_name: "Test Organization"
+    - composite_content_view_name: "ccv1"
+    - content_view_name: "cv1"
+    - content_view_version: "1.0"
+  katello_content_view:
```
mdellweg: Shouldn't this be the shiny new `..._composite_...` module?

gregswift:
> I would think, 65cce96 together with a corresponding test is worth it's own PR. It is pretty straight forward and could be merged quite fast and even adds a feature (state). BTW, thank you there!

Oh, ya i totally agree. Just trying to make this work, then will shift things around a bit

test/test_playbooks/tasks/composite_content_view.yml
gregswift: d'oh

modules/katello_repository.py
gregswift: Any reason why we dont let the find objects do these lookups for us to minimize effort in the modules themselves?

modules/katello_repository.py
gregswift: So no latest... wth is present_with_defaults (in this context?) or should i drop it?

mdellweg: It means 'create it if it was not there, leave it alone otherwise'.

modules/katello_repository.py
mdellweg: we need to have them resolved to nailgun objects for the naildown_state.

test/test_playbooks/content_view.yml
gregswift: np

gregswift:
New failure time!

```Traceback (most recent call last):
  File "/tmp/ansible_Viq0bW/ansible_module_katello_repository.py", line 185, in <module>
    main()
  File "/tmp/ansible_Viq0bW/ansible_module_katello_repository.py", line 175, in main
    entity = find_repository(module, name=entity_dict['name'], product=entity_dict['product'])
  File "/tmp/ansible_Viq0bW/ansible_modlib.zip/ansible/module_utils/ansible_nailgun_cement.py", line 285, in find_repository
  File "/usr/lib/python2.7/site-packages/blinker_herald/base.py", line 137, in wrapper
    result = fn(*args, **kwargs)
  File "/usr/lib/python2.7/site-packages/nailgun/entity_mixins.py", line 1358, in search
    results = self.search_json(fields, query)['results']
  File "/usr/lib/python2.7/site-packages/nailgun/entity_mixins.py", line 1209, in search_json
    response = self.search_raw(fields, query)
  File "/usr/lib/python2.7/site-packages/nailgun/entity_mixins.py", line 1186, in search_raw
    data=self.search_payload(fields, query),
  File "/usr/lib/python2.7/site-packages/nailgun/entity_mixins.py", line 1161, in search_payload
    payload[field + '_id'] = value.id
AttributeError: 'NoneType' object has no attribute 'id'
```

akofink:
Probably a problem with your product you're passing in. Put a break point right before `find_repository(...)`, and check those values you're passing in.

gregswift:
@akofink what is the recommended way of breakpointing ansible module?

akofink:
Do [these steps](https://github.com/theforeman/foreman-ansible-modules#how-to-debug-modules-in-this-repository), then use:

```
...
import pdb; pdb.set_trace()
...
```

modules/katello_repository.py
```diff
@@ -212,9 +168,13 @@ def main():
 
     ping_server(module)
 
+    entity_dict['organization'] = find_organization(module, name=entity_dict['organization'])
+
+    entity_dict['product'] = find_product(module, name=entity_dict['name'], organization=entity_dict['organization'], failsafe=True)
```
mdellweg: You might want to fail here if the product cannot be found (without failsafe).

gregswift: that looks to have been it. However....

```TASK [Ensure product 'Test Product' is present] ********************************
ok: [fixtures]
``` and if i go look in the UI its there.
![test_product](https://user-images.githubusercontent.com/372939/34725709-5e83c652-f517-11e7-8dc6-6cc0ad241a17.png)

Comment (gregswift) [reply modules/katello_repository.py]: Trying to make the pdb bit working

**Comment by mdellweg** [comment]
You might want to add yourself as author to re reworked katello module.

**Comment by gregswift** [comment]
> You might want to add yourself as author to re reworked katello module.

I've made the mistake of making myself the author of modules in ansible before.....

**Comment by gregswift** [comment]
The fact that the katello_repository module requires organization, but find_repository doesnt require it is weird.

**Comment by gregswift** [comment]
@mdellweg so here's a fun one...

```
TASK [Create katello repository] ***********************************************
fatal: [fixtures]: FAILED! => {"changed": false, "msg": "Error while creating Repository: 'backend_identifier'"}
	to retry, use: --limit @/home/xaeth/Development/foreman-ansible-modules/test/test_playbooks/composite_content_view.retry
```

And yet....
```
[ xaeth@localhost foreman-ansible-modules ] (ccv *$=)$ grep backend_identifier * -r
```

**Comment by mdellweg** [comment]
I guess it's here: https://github.com/SatelliteQE/nailgun/blob/63fb869b8217736d1126443541b399cfcf974846/nailgun/entities.py#L5089
No Idea, what it is.

**Comment by gregswift** [comment]
So i'm out of town for a few days... will get back to this when i get back

**Comment by gregswift** [comment]
hey.. if yall could check in on #87 that would be super handy :D

**Comment by gregswift** [comment]
cleaned up all my microcommits.. i believe it still needs the recordings though

**Comment by rackergs** [comment]
I just rebased and cleaned everything up... but ya I'm back to the same `backend_identifier` error above cause we never addressed that... any thoughts ?

**Comment by gregswift** [comment]
I just rebased and cleaned everything up... but ya I'm back to the same backend_identifier error above cause we never addressed that... any thoughts ?

@akofink @ehelms @evgeni @Fobhep

File: modules/katello_content_view.py [hunk 19]
```diff
@@ -183,12 +184,7 @@ def main():
     )
     ng = NailGun(server, entities, module)
 
-    # Lets make an connection to the server with username and password
-    try:
-        org = entities.Organization(server)
-        org.search()
-    except Exception as e:
-        module.fail_json(msg="Failed to connect to Foreman server: %s " % e)
+#    ping_server(module)
```
evgeni: for `ping_server` you'd need to also call `create_server` from `ansible_nailgun_cement` like this:

```python
    try:
    except Exception as e:
        module.fail_json(msg='Failed to connect to Foreman server: %s ' % e)

    ping_server(module)
```

but as the module currently is not migrated to cement, I see no benefit in doing so for now

modules/katello_composite_content_view.py
```diff
@@ -0,0 +1,172 @@
...
+    username: "admin"
+    password: "changeme"
+    server_url: "https://foreman.example.com"
+    name: "OurFedora CCV"
+    organization: "My Cool new Organization"
+    content_views:
+      - name: 'Fedora CV'
+        version: '2.0'
+      - name: 'Ansible CV'
+        version: '1.0'
```
ehelms: I'd add `state` to this example since it is required.

ehelms:
When I run this I get:

```
An exception occurred during task execution. To see the full traceback, use -vvv. The error was: TypeError: find_content_view() got multiple values for keyword argument 'failsafe'
```

modules/katello_composite_content_view.py
gregswift: Or should I change state to `default: true`

mdellweg: It would be either 'present' or 'absent', and i think following the other modules would be better.

gregswift: generally speaking i think ansible defaults this to true and doesn't require it as a parameter (even if putting it there is considered a best practice)

mdellweg: I'd like to have this consistent across all entity modules. So we might discuss that in a general issue.
And again, it were `default: 'present'`, not `true`.

gregswift:
@ehelms i can't reproduce your error. i still just get the 

```
TASK [Create katello repository] ***********************************************
fatal: [fixtures]: FAILED! => {"changed": false, "msg": "Error while updating Repository: 'backend_identifier'"}
	to retry, use: --limit @/home/xaeth/Development/foreman-ansible-modules/test/test_playbooks/composite_content_view.retry
```

gregswift:
Side note... if i run the record for repository i get the same.... this is working for yall? is the fact that my katello instance is 3.2 a problem?

rackergs:
@ehelms @mdellweg ?

mdellweg:
Could you try to run this against a fresh forklift instance?

ehelms:
Apologies, I'm a little slow getting back to this. I will spin up a 3.5 and run a test scenario with this and report back what I get.

ehelms:
I had to rebase first and got these errors:

```
An exception occurred during task execution. To see the full traceback, use -vvv. The error was: TypeError: find_content_view() got multiple values for keyword argument 'failsafe'
```

```
ansible-playbook test/test_playbooks/composite_content_view.yml -vv

TASK [Promote the content view version to target environment] *******************************************
task path: /root/foreman-ansible-modules/test/test_playbooks/tasks/content_view_version_promote.yml:2
An exception occurred during task execution. To see the full traceback, use -vvv. The error was: TypeError: Value of unknown type: <type 'exceptions.AttributeError'>, 'NoneType' object has no attribute 'id'
```

gregswift:
@ehelms xan you try just the repository record?

mdellweg:
Without rebasing i also get (this is katello 3.4):
```
TASK [Create/update katello composite content view] ****************************
An exception occurred during task execution. To see the full traceback, use -vvv. The error was: TypeError: find_content_view() got multiple values for keyword argument 'failsafe'
```

ehelms:
I was able to run the `repository.yml` test file without issue on a Katello 3.5 instance.

evgeni:
after rebasing, I had to apply the following to get it at least running:

```patch
diff --git a/module_utils/ansible_nailgun_cement.py b/module_utils/ansible_nailgun_cement.py
index 92ee281..dd77949 100644
--- a/module_utils/ansible_nailgun_cement.py
+++ b/module_utils/ansible_nailgun_cement.py
@@ -235,7 +235,7 @@ def delete_entity(entity, module):
 
 
 def find_content_views(module, content_views, organization, failsafe=False):
-    return map(lambda content_view: find_content_view(module, content_view['name'], content_view['version'],
+    return map(lambda content_view: find_content_view(module, content_view['name'],
                                                       organization, failsafe=failsafe), content_views)
 
 
diff --git a/modules/katello_composite_content_view.py b/modules/katello_composite_content_view.py
index 3a91632..e53f6ef 100755
--- a/modules/katello_composite_content_view.py
+++ b/modules/katello_composite_content_view.py
@@ -98,6 +98,7 @@ try:
         find_content_view_version,
         ping_server,
         set_task_timeout,
+        naildown_entity_state,
     )
     HAS_NAILGUN_PACKAGE = True
 except:
@@ -168,6 +169,7 @@ def main():
 
     changed = naildown_entity_state(ContentView, entity_dict, entity, state, module)
 
+    module.exit_json(changed=changed)
 
 if __name__ == '__main__':
     main()
```


however, it still fails, now with

```
fatal: [tests]: FAILED! => {"changed": false, "msg": "Error while creating ContentView: Valid fields are ['puppet_module', 'description', 'repository', 'composite', 'component', 'id', 'next_version', 'last_published', 'name', 'label', 'environment', 'version', 'content_host_count', 'organization'], but received ['organization', 'name', 'content_views'] instead."}
```

@ehelms the `'NoneType' object has no attribute 'id'` seems to appear when you already have a `cv1` CV in your TestOrg. Deleting it helped me. No idea why it does not detect the CV properly.

ehelms:
@gregswift Any chance you'll get back to this?

gregswift:
Well, not sure i'd call that getting back to it.. but rebased and added in @evgeni 's changes.

ehelms:
Nice, thanks for getting back to it @gregswift

rackergs:
So the tests passed now?

modules/katello_composite_content_view.py
mdellweg: #104 Present is the default now.

modules/katello_product.py
```diff
@@ -129,7 +129,7 @@ def main():
             organization=dict(required=True),
             label=dict(),
             description=dict(),
-            state=dict(default='present', choices=['present_with_defaults', 'present', 'absent']),
+            state=dict(required=True, choices=['present', 'absent']),
```
mdellweg: This was already quite fine.
Also it is off topic.

modules/katello_repository.py
```diff
@@ -145,7 +145,7 @@ def main():
             content_type=dict(required=True),
             url=dict(),
             download_policy=dict(choices=['background', 'immediate', 'on_demand']),
-            state=dict(default='present', choices=['present_with_defaults', 'present', 'absent']),
+            state=dict(default='present', choices=['present', 'absent']),
```
mdellweg: Same here

test/test_playbooks/composite_content_view.yml
```diff
@@ -0,0 +1,58 @@
...
+  gather_facts: false
+  tasks:
+  - name: Load server config
+    include_vars:
+      file: server_vars.yml
+  - include_tasks: tasks/composite_content_view.yml
+    vars:
+      composite_content_view_state: present
+  - include_tasks: tasks/composite_content_view.yml
+    ignore_errors: yes
```
mdellweg: `ignore_errors` in the test part seems wrong.
Anyway, the first test seems to have run without an error. Good work!

mdellweg:
Seems like the tests didn't start automatically.

gregswift:
Updated. Can we trigger the testing ?

gregswift:
@mdellweg @ehelms why dont the tests trigger anymore? can we trigger them?

mdellweg:
I have no idea, sorry. Have you seen the new tab "checks"? They changed something...

mdellweg:
Tests are activated once again. Thanks to @GregSutcliffe

ehelms:
@gregswift Looks like it needs a rebase again

rackergs:
:crossed_fingers: 

its still not gonna work but let's see what errors happen since its testing this time!

ehelms:
@rackergs I dug into using this a bit with some infra we have and I ran into one big issue with it and how Katello handles these APIs. Through the ContentView API when creating a composite you can add content view versions by ID only. The more common practice is to add content views and specify either the version or most common "latest". This has to be done through the Content View Component API fo rwhich nailgun does not currently have an entity associated to it. I'd like to work in conjunction with this PR of yours to add that functionality as I have time

gregswift:
> The more common practice is to add content views and specify either the version or most common "latest". 

That is a fairly common practice... but that being said we've played with the idea of having CVs live in 1 lifecycle (Library>Stable) separate from CCVs (Library>Staging>Prod).  So there are times when what i want is `latest version of CV that has been promoted to Stable environment` where othertimes I want `latest version of CV`

itewk:
being awful at searching through PRs I wasn't able to find this PR before I started on my own. I see one of the requests was to merge all the code into katello_content_view, which is something I did from the get go as NailGun doesn't have a CompositeContentView object, just a ContentView object so I figured the idea was to keep aligned.

Maybe there is some stuff in my PR that could be merged with this PR? https://github.com/theforeman/foreman-ansible-modules/pull/177

The biggest issue I am having is I can't figure out how to add an unpublished Content View to the composite content view using NailGun with the version of "Latest" specified. You can do this in the UI with once click, but I can't for the life of me figure out how to do it with NailGun.

akofink:
We do this in hammer-cli-katello. Perhaps that logic will help figure out Nailgun? https://github.com/Katello/hammer-cli-katello/blob/49b48b3c1d431da10c254e3c59e62552cb909122/lib/hammer_cli_katello/content_view_component.rb#L39

itewk:
@akofink what is the equivalent for that for katello api? All I can find is https://theforeman.org/plugins/katello/3.5/api/apidoc/v2/content_view_components/add_components.html which fails if the component is already added

gregswift:
@akofink where is the best place that we can chat about moving this forward in a more sync manner?

akofink:
You can reach me on #theforeman-dev on freenode; my nick is akofink.

gregswift:
rebased and :crossed_fingers: 

i dont have access to irc from work right now, but i'll see what i can do.  What timezone are you in @akofink

akofink:
EST

sean797:
setting CV=latest isn't support within Nailgun. I've raise a PR to add support for that https://github.com/SatelliteQE/nailgun/pull/558

As for the API, I generally think of a Composite CV as a type of CV, I wounder if we should have `composite: true` param as part of the `katello_content_view` as opposed to it's own `katello_composite_content_view` module?

akofink:
I'd also support extending the existing `katello_content_view` module. They are the same resource within Katello, at least.

rackergs:
so @akofink and @sean797 there is a bit of discussion and thinking over in https://github.com/theforeman/foreman-ansible-modules/issues/87

i thought there was more context, it might be on some closed PRs too.

sean797:
@mdellweg Any view on the UX of a composite CV feature?

mdellweg:
Well, i never really understood, why you need to decide, whether a CV was composite. You could just make it contain repositories _and_ other CV's.
Having the `composite` option in the module sure looks much more like in the UI.

akofink:
Hmm... I think repositories and content views are/should be mutually exclusive. You can't have a composite content view with repositories afaik.

mdellweg:
Oh, i was unclear. You are right you cannot (in the katello api). I do not understand why.

akofink:
I could see supporting both repos and CVs in a composite CV in the future, but it just wasn't designed that way, I think.

rackergs:
So... while i can agree that at first it seems weird that you can't do manage both repositories and content views in a composite. However, I think that mixing repos and content views in composite views would blur the distinction in a counter productive way.

I've been working with our teams on understanding Katello and how to manage our lifecycles for the last several years and I believe that we've gotten to a good place. 

One of the biggest confusion points is that CCVs have 'content' in their names.  It makes understanding the distinction and the work flow initially more confusing.  It would even be nice if the UI would provide a better distinction between the 2, the adhoc concept of naming everything with CV and CCV is kinda annoying. They currently are different things, and it does facilitate the workflow, once you are going down the path.

That being said I also get that CV and CCV provides a distinction as down to the acronym level that Composite View (CV) vs Content View (CV) does not.

It would be interesting to see a future UX rework, but the way that the system works now I would argue that having distinct modules that shared backend code makes its more discoverable and accessible to consumers and helps enforce the concepts.

I vaguely remember question if we should revisit the release and promote modules.  My opinion on naming is that i'd actually rather remove `content_view` from those module's names, because nothing else (that i've noticed) in foreman/katello does release/promote and so it makes it less weird that I have a composite_content_view concept, but i manage it with content_view_promote/release.

sean797:
@gregswift I've created https://github.com/theforeman/foreman-ansible-modules/pull/204 hope you don't mind (I needed composite CV support for my own reasons). Your review and/or testing would be appreciated.

rackergs:
I dont really care how it gets accomplished as long as it does.

mdellweg:
Closing in favour of:
https://github.com/theforeman/foreman-ansible-modules/pull/204

PR #80, opened 2017-12-01
rename katello_publish

sean797:
Obviously a breaking change but the sooner the better

### DIFF
--- a/README.md
+++ b/README.md
@@ -98,4 +98,4 @@ This is a list of modules currently in the repository (please add to the list if
 
  * katello_sync: sync Katello repositories and products
  * katello_upload: upload files, rpms, etc. to repositories
- * katello_publish: publish Katello content views
+ * katello_content_view_publish: publish Katello content views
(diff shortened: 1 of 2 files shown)


akofink:
Before we merge, are there any other projects that we should submit PRs to to update the name of this module? (see robotello-ci PR above).

mdellweg:
Is there a way to add an alias with the old name?

akofink:
I can't see it in the docs/using Google. I'm asking in IRC (#ansible on freenode).

sean797:
There is https://github.com/sean797/ansible-role-configure-katello/blob/master/handlers/main.yml#L14 but I'll commit that when this is merged.

Are we okay with this without a deprecation? I guess I could leave the old module and make it print a failed message saying to use the new name.

akofink:
I'm fine with the name change and no deprecation. I don't think we have that many people using this repo in production yet. @ehelms Good with that?

mdellweg:
Maybe we trigger an unintended user survey. ;) I'm good.

PR #81, opened 2017-12-01
adds Dockerfile with instructions

thomasmckay:


ehelms:
Looking good to start. I have a few thoughts for you.

 1) The example scripts require specifying quite a lot on the command line each time, could this be captured into a simple bash script perhaps?
 2) This might be a bit meta, but what about a playbook to run the container and capture the configuration that way? See http://docs.ansible.com/ansible/latest/docker_container_module.html#docker-container
   2a) We could also bundle into it a docker_image step that ensure sthe container is built before running it
 3) Might be more advanced than intended, but similarly could simplify the configuration of the volumes and variables by using ansible-container project

Lastly, with these enhancements, I think putting this into its own folder would be nice and clean.

Dockerfile
```diff
@@ -0,0 +1,35 @@
...
+
+
+FROM fedora
+
+RUN dnf -y update && dnf install -y \
+        git \
+        python-pip
+
+RUN set -x \
+        && git clone https://github.com/theforeman/foreman-ansible-modules.git \
```
Fobhep: could we maybe add a `set -e` as well?
I like adding it to almost all bash scripts I am running. Makes it a usually a bit more stable.

Dockerfile
```diff
@@ -0,0 +1,35 @@
...
+#              /foreman-ansible-modules/test/test_playbooks/myorg.yml
+
+
+FROM fedora
+
+RUN dnf -y update && dnf install -y \
+        git \
+        python-pip
+
+RUN set -x \
```
Fobhep: could we maybe add a set -e as well?
I like adding it to almost all bash scripts I am running. Makes it a usually a bit more stable.

thomasmckay: I may not be understanding the -e option, but is it relevant in a build? The "set -x" I added based upon another Dockerfile written by more skilled folks (skopeo) and some googling. I'm happy to add it but would like to understand the why.

thomasmckay:
I should probably explain my reasoning for wanting to create this in the first place.

My goal is to provide some template-like files to users of foreman to quickly set up potentially tedious examples and demos. For example, create a product with a dozen container images, applying content view filters, promoting through environments. Another usage would be to write an export tool to create the yaml to then reimport (I use this to recreate my dev environment and to backup servers).

Knowing the above, a container would let a user run without any additional software installation (I am assuming they have an equivalent of "docker run" available, of course). If a "ansible-container build" still allows the image to be run with "docker run", that seems reasonable (and fun!).

Dockerfile
ekohl: I agree it might not be that useful here since it's using ```&&``` meaning the whole ```RUN``` statement should fail if one fails.

Fobhep: yes, that is absolutely true. 
never mind, I was typing quicker than thinking :P Friday.. ;)

Dockerfile
```diff
@@ -0,0 +1,35 @@
...
+#
+#     % docker build --tag fam:latest .
+#
+# To run an existing playbook simply specify the path. Override any variables necessary using the
+# --extra-vars option.
+#
+#     % docker run -it --rm \
+#          fam:latest \
+#              --extra-vars "foreman_server_url=https://192.168.122.3" \
+#              /foreman-ansible-modules/test/test_playbooks/organization.yml
```
akofink: You might also mention that users may need the `--network host` option. At least, I did. ;)

mdellweg:
Maybe you should put the description in the README, not the Dockerfile.

Dockerfile
```diff
@@ -0,0 +1,35 @@
...
+# other resources it needs to be mounted relative to those files.
+#
+#     % docker run -it --rm \
+#          --volume `pwd`/myorg.yml:/foreman-ansible-modules/test/test_playbooks/myorg.yml \
+#          fam:latest \
+#              --extra-vars "foreman_server_url=https://192.168.122.3" \
+#              /foreman-ansible-modules/test/test_playbooks/myorg.yml
+
+
+FROM fedora
```
akofink: Maybe we should just use `python:2.7-slim`?

Dockerfile
```diff
@@ -0,0 +1,35 @@
...
+
+RUN dnf -y update && dnf install -y \
```
akofink: If we use a python image, we shouldn't need to update packages, afaik.

Dockerfile
thomasmckay: Interesting, I'll give that a try.

Dockerfile
```diff
@@ -0,0 +1,35 @@
+# To build image:
```
ehelms: We could likely tie these instructions into the makefile with some variables to make it easier. That could be a post merge action -- @akofink thoughts?

akofink: I think I'm actually opposed to this. Dockerfile's are self-contained and are supposed to "just work" rather than being super configurable like (i.e.) an Ansible playbook. Anyone who's ever used docker should be comfortable with the `docker build` and are free to specify custom args (such as tag). The `docker run` examples are sufficient imo to show different run configurations.

If you really want more configurability, then I'd consider adding a docker-compose yaml before a Makefile.

akofink: fwiw, I've seen this pattern (dockerfile with comments at the top for run examples) from [reputable container people](https://github.com/jessfraz/dockerfiles/blob/master/chrome/stable/Dockerfile).

akofink: katello-host-tools does this with a makefile, so we have some precedent.

Dockerfile
```diff
@@ -0,0 +1,35 @@
...
+RUN dnf -y update && dnf install -y \
+        git \
+        python-pip
+
+RUN set -x \
+        && git clone https://github.com/theforeman/foreman-ansible-modules.git \
+        && cd foreman-ansible-modules \
+        && make test-setup
+
+WORKDIR /foreman-ansible-modules
```
akofink: The workdir shouldn't be foreman-ansible-modules, it should be an empty directory that users can mount their own playbooks/roles/tasks into. Otherwise, how would users run their playbooks?

Dockerfile
```diff
@@ -0,0 +1,35 @@
...
+FROM fedora
+
+RUN dnf -y update && dnf install -y \
+        git \
+        python-pip
+
+RUN set -x \
+        && git clone https://github.com/theforeman/foreman-ansible-modules.git \
+        && cd foreman-ansible-modules \
+        && make test-setup
```
akofink: Are we planning to run FAM tests? Is this image for users or developers of FAM?

akofink:
Getting stale. Any updates here @thomasmckay?

thomasmckay:
Closing this. Probably would have been better, in my opinion, to treat this as a community PR and have it merged and then refined over time. I used this locally a lot but maybe not valuable to others, in which case it was right to not merge it. Happy trails!

PR #85, opened 2017-12-12
Ansible Playbook for test_playbook for REALM creation and deletion

claudiol:
Completed tasks:
- Modified current realm.yml under test/test_playbook directory to include self contained variables to create a REALM in Satellite 6.
- test/test_playbooks/realm.yml
```---
- hosts: localhost
  tasks:
  - name: Load server config
    include_vars:
      file: server_vars.yml
  - include: tasks/realm.yml
    vars:
      realm_name: "EXAMPLE.LOCAL"
      realm_proxy: 1
      realm_type: "Red Hat Identity Management"
      realm_state: present
  - include: tasks/realm.yml
    vars:
      realm_name: "EXAMPLE.LOCAL"
      realm_proxy: 1
      realm_type: "Red Hat Identity Management"
      realm_state: absent
```
- Added a realm.yml to the test/test_playbooks/tasks directory
```
---
- name: "Ensure realm '{{ realm_name }}' is {{ realm_state }}"
  foreman_realm:
    name: "{{ realm_name }}"
    realm_proxy: '{{ realm_proxy }}'
    realm_type: "{{ realm_type }}"
    state: "{{ realm_state }}"
```
- Ran test for the playbook:
[root@fedora26 foreman-ansible-modules]# ansible-playbook test/test_playbooks/realm.yml [DEPRECATION WARNING]: The use of 'include' for tasks has been deprecated. Use 
'import_tasks' for static inclusions or 'include_tasks' for dynamic inclusions. This 
feature will be removed in a future release. Deprecation warnings can be disabled by 
setting deprecation_warnings=False in ansible.cfg.
[DEPRECATION WARNING]: include is kept for backwards compatibility but usage is 
discouraged. The module documentation details page may explain more about this rationale.. 
This feature will be removed in a future release. Deprecation warnings can be disabled by 
setting deprecation_warnings=False in ansible.cfg.

PLAY [localhost] ***************************************************************************

TASK [Gathering Facts] *********************************************************************
ok: [localhost]

TASK [Load server config] ******************************************************************
ok: [localhost]

TASK [Ensure realm 'EXAMPLE.LOCAL' is present] *********************************************
changed: [localhost]

TASK [Ensure realm 'EXAMPLE.LOCAL' is absent] **********************************************
changed: [localhost]

PLAY RECAP *********************************************************************************
localhost                  : ok=4    changed=2    unreachable=0    failed=0  

Let me know if I need to provide more updates.

Best,
Lester


test/demo/realm.yml
```diff
@@ -0,0 +1,24 @@
+---
```
mdellweg: This directory has been reassigned to be test_playbook.
I think, the examples could be in the inline documentation comment of the module.

mdellweg:
May you should rebase your PR. Looks like it contains a lot of old commits.

test/data/realm.json
```diff
@@ -0,0 +1,10 @@
+{
```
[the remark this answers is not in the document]
claudiol: So the realm.json in the test/data directory seems to be the same that is is the test/debu_data.  I rebased my branch but the directory is still there.  Should I remove the file by doing a git rm test/realm.json and remove the test directory all together?

claudiol:
Completed the following requests:
- Rebased my PR.
- Removed test/data/realm.json
- Removed test/demo/realm.yml

Let me know if there are other issues I need to address. I will continue to add more commits to support the test_playbooks.

Best,
Lester

mdellweg:
I fear, the rebase did not work as expected, because your master diverged quite far from upstream master. Would you mind doing a `git rebase origin/master`?

mdellweg:
```
TASK [Ensure realm 'EXAMPLE.LOCAL' is present] *********************************
fatal: [localhost]: FAILED! => {"changed": false, "msg": "Error while creating Realm: 422 Client Error: Unprocessable Entity for url: https://centos7-devel.example.com/api/v2/realms"}
	to retry, use: --limit @/home/matthiasd/projects/foreman-ansible-modules/test/test_playbooks/realm.retry
```
I'm using centos7-devel from forklift.
What version of nailgun are you using?

claudiol:
@mdellweg I did a git rebase origin/master.
The nail gun that I am using is nailgun.x86_64 0.9.1-5.fc26 on fedora.

If nothing else I can start over with my forked repo.  

Best,
Lester

test/test_playbooks/tasks/realm.yml
```diff
@@ -0,0 +1,11 @@
+---
+- name: "Ensure realm '{{ realm_name }}' is {{ realm_state }}"
+  foreman_realm:
```
akofink: Put some default variables here so that if I don't pass realm_name and realm_state, it'll still do something.

claudiol: @akofink Thanks.  I'll definitely do that. Thanks for the suggestion.

claudiol: @akofink So it looks like we can define variables in the server_vars.yml file.  We have a template that is available under test/test_playbooks.  In my test I added that to the vars to the realm.yml but they can be added to the server_vars.yml.  Is that the intent? The tests don't have a place to add group_vars etc.  Let me know if I am missing something.

akofink: @claudiol Like [this](https://github.com/theforeman/foreman-ansible-modules/blob/master/test/test_playbooks/tasks/product.yml#L4-L6).

mdellweg:
> Also, if EXAMPLE.COM is present in your foreman instance then this will happen.
I had no realm preconfigured. But digging a little deeper, i got those two errors:
```
Realm proxy does not have the Realm feature
Realm type is not included in the list
```
This looks to me, that your playbook would need some fixtures, to provide the 'realm proxy' and the 'realm type' (maybe, you want to use `FreeIPA` here).

Your answer, however, concerns me on an another level. If what you said above (which i did not yet check) is true, that would mean, the realm module was not idempotent. Trying to create an existing entity with the same parameters should really do nothing and report success-without-change.

claudiol:
@mdellweg At this point I believe we would need to revisit the realm module.  I think the test does what it's supposed to.  I can look into it and make sure that it follows the idempotency requirement.  Please let me know how to proceed when you can.  Thanks!

mdellweg:
I am still unable to run this playbook against my forklift instance, because neither the `realm_proxy` nor the `realm_type` are available. If they can be setup inside the playbook, please do so. If not, please include detailed instructions in the README ("In order to run the test for _realm_, you need to ...").

I want everyone (in priciple) to be able to run the test suite locally.

Optional: In the end you might want to duplicate the tasks to test for idempotence.
BTW: If the fix for idempotence is not too complicated, just add it here.

PR #88, opened 2018-01-05
Cleanup

mdellweg:
Again, i started a cleanup, DRYup branch. I guess it will live on for a while being rebased again and again, since i want to refactor only modules that are already under test control.
Feel free to add more cleanups.


module_utils/ansible_nailgun_cement.py
```diff
@@ -110,6 +110,14 @@ def ping_server(module):
         module.fail_json(msg="Failed to connect to Foreman server: %s " % e)
 
 
+def sanitize_entity_dict(entity_dict, name_map):
```
gregswift: w00t

PR #91, opened 2018-01-17
Test provisioning template

mdellweg:
Based on #90.
This test does not pass when run against current foreman/katello nightly due to a bug (Redmine: 22303).
Thanks to @akofink helping to figure this out.


mdellweg:
Add location is a different PR, and template test playbooks are so damn similar...

PR #92, opened 2018-01-17
Test playbooks

mdellweg:
This changes the `make test_<...>` rule so that for example composite_content_view tests do not run, when content_view tests are requested.

Also i started a small guideline about the test infrastructure.

### DIFF
--- a/test/README.md
+++ b/test/README.md
@@ -1,0 +1,21 @@
+Guidedeline to writing tests
+===
+
+The tests in this repository run playbooks that can be found in `test/test_playbooks`.
+To be run, the name of the corresponding playbook must be listed in `test/test_crud.py`.
+
+The playbooks should be self contained, target the features of a specific module and be able to run against an actual foreman instance.
+The structure would usually be something like setup -> run tests -> clean up.
+
+In `test/test_playbooks/tasks`, preconfigured tasks for individual entities / actions can be found.
+They can be reused, to create common dependent resources, as well as to manipulate the entities in the actual test.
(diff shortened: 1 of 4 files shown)


mdellweg:
I'd like your opinion here @akofink, @gregswift .

test/README.md
```diff
@@ -0,0 +1,19 @@
...
+In `test/test_playbooks/tasks`, preconfigured tasks for individual entities / actions can be found.
+They can be reused, to create common dependent resources, as well as to manipulate the entities in the actual test.
+If the boolean variable `expected_change` is set, such a task fails if the result does not meet this expectation.
+
+The ansible inventory contains two hosts:
+
+- fixtures: This host runs locally without modification it is meant to setup (and teardown) dependent resources, only when (re-)recording.
+- tests: This host should run the actual tests. It is used to record the vcr-yaml-files, or to run isolated against those files.
+
+After writing a test playbook, `make record_<playbook name>` must be called, and the resulting vcr files (`test_playbook/fixtures/<playbook_name>-*.yml`) must be checked into git.
```
akofink: Maybe include "... must be called against a running Foreman or Katello server"

test/README.md
```diff
@@ -0,0 +1,19 @@
+Guidedeline to writing tests
```
akofink: I like this readme. Let's link to it from the top level Readme as well.

test/README.md
gregswift: Ya, a better way of taking line 20 and making it less of an afterthought to line 19 would be nice.

PR #94, opened 2018-01-19
Refactory katello_repository to utilize cement

gregswift:
Includes the fixed tests from mdellweg (but I did a lazy pull so lost the commit)

Pulls changes from #79

### DIFF
--- a/module_utils/ansible_nailgun_cement.py
+++ b/module_utils/ansible_nailgun_cement.py
@@ -275,9 +275,9 @@ def find_repositories(module, repositories, product):
     return map(lambda repository: find_repository(module, repository, product), repositories)
 
 
-def find_repository(module, name, product):
+def find_repository(module, name, product, failsafe=False):
     repository = Repository(name=name, product=product)
-    return handle_find_response(module, repository.search(), message="No Repository found for %s" % name)
+    return handle_find_response(module, repository.search(), message="No Repository found for %s" % name, failsafe=failsafe)
 
 
 def find_repository_set(module, name, product, failsafe=False):
--- a/modules/katello_repository.py
+++ b/modules/katello_repository.py
@@ -78,6 +78,7 @@ options:
             - State of the Repository
         default: true
         choices:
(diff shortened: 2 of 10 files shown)


module_utils/ansible_nailgun_cement.py
```diff
@@ -224,6 +224,11 @@ def delete_entity(entity, module):
     return True
 
 
+def find_content_views(module, content_views, organization, failsafe=False):
+    return map(lambda content_view: find_content_view(module, content_view['name'], content_view['version'],
+                                                      organization, failsafe=failsafe), content_views)
+
+
```
mdellweg: I think, this is unrelated. It should be in some 'refactor_content_view' PR.

test/test_playbooks/tasks/repository.yml
```diff
@@ -0,0 +1,22 @@
+---
+- name: "Create katello repository"
+  vars:
+    - organization_name: "Test Organization"
+    - product_name: "Test Product"
+    - repository_name: "Test Repository"
+    - repository_content_type: "yum"
+    - repository_state: present
```
mdellweg: Same as in #95 ,
so maybe wait for the decision there.

modules/katello_repository.py
```diff
@@ -73,6 +73,13 @@
             - immediate
             - on_demand
         required: false
+    state:
+        description:
+            - State of the Repository
+        default: true
+        choices:
+        - present
+        - absent
```
mdellweg: Do you mind adding 'present_with_defaults' here and inline 148?
Cements naildown should make it _just work_.

module_utils/ansible_nailgun_cement.py
rackergs: doh.. i meant to leave that out. thanks

mdellweg:
One last (hopefully) thing i just saw: Can you please add the state parameter in the example string.
I think, then we are really good to go.

PR #95, opened 2018-01-19
Cleanup katello_product state in doc and test

gregswift:
Brings in the changes from #79

### DIFF
--- a/test/test_playbooks/tasks/product.yml
+++ b/test/test_playbooks/tasks/product.yml
@@ -4,6 +4,7 @@
     - organization_name: "Test Organization"
     - product_name: "Test Product"
     - product_description: "A happy little test product"
+    - product_state: present
   katello_product:
     username: "{{ foreman_username }}"
     password: "{{ foreman_password }}"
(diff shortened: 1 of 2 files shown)


test/test_playbooks/tasks/product.yml
```diff
@@ -4,6 +4,7 @@
     - organization_name: "Test Organization"
     - product_name: "Test Product"
     - product_description: "A happy little test product"
+    - product_state: present
```
mdellweg: I think, we left that one out for a reason, as we wanted to be specific about the creating/deleting every time using it.
@akofink , since introduced these tasks, i guess it's your call.

akofink: I don't feel strongly about it, but no other tasks provide the state as a default.

rackergs: It felt easier to override with more explicit testing on state if its defined with a default than if the parameter is left off.

mdellweg: Ok for me. It does not hurt anyway.

PR #96, opened 2018-01-19
Add 'body' to the vcr match criteria

mdellweg:


mdellweg:
I think, this is quite an urgent one.
Can anybody see, why it passes on python2 but fails on python3?

akofink:
I'm not sure why it's different on Python 3, but maybe you should re-record the operating systems tests.

mdellweg:
well, rerecording does not help, and rerecording with py3 breaks py2. Too bad. I will investigate by comparing the cassettes...

PR #100, opened 2018-01-29
Add test for katello_sync_plan

mdellweg:
Should have been easy, but there is a problem around the date parsing.
py3 is ok, but py2 does not know how to interpret "%z" (for '+0100').

`Value of unknown type: <type 'exceptions.ValueError'>, 'z' is a bad directive in format '%Y/%m/%d %H:%M:%S %z'`
It is documented, though:
https://docs.python.org/2/library/datetime.html#strftime-and-strptime-behavior

### DIFF
--- a/modules/katello_sync_plan.py
+++ b/modules/katello_sync_plan.py
@@ -90,6 +90,8 @@ EXAMPLES = '''
 
 RETURN = '''# '''
 
+import six
+import pytz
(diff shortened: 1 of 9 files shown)


mdellweg:
@akofink ping

mdellweg:
Thanks for the quick review!

PR #102, opened 2018-01-29
 Add tests for os_default_template 

mdellweg:
Despite the label text, this depends on *two* other PRs: #91 #101


mdellweg:
`AttributeError: 'OSDefaultTemplate' object has no attribute 'search'`
I guess, that means we are waiting on Nailgun.

PR #118, opened 2018-04-06
Remove shadow entities

mdellweg:
* Remove the shadowing entity for global parameter, as the functionality has been added to nailgun.
* Import entities directly from nailgun if they are not modified in cement.

### DIFF
--- a/module_utils/ansible_nailgun_cement.py
+++ b/module_utils/ansible_nailgun_cement.py
@@ -30,18 +30,6 @@ from nailgun.entities import (
 from nailgun import entity_mixins, entity_fields
 
 
-class CommonParameter(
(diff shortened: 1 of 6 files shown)


mdellweg:
Maybe, i can draw you attention to another PR ;) @akofink

akofink:
Could you please add a description to this PR?

mdellweg:
Something like this?

mdellweg:
Honestly i have never heard of shadowing classes before. It was a dirty hack for us to not needing to wait for the inclusion in nailgun. So this is really only cleanup.
There is more to come around the Compute* classes, but i'd like to do this refactoring, once the corresponding unittests are in place.

PR #120, opened 2018-04-11
Add foreman_setting_facts

mdellweg:
I started to experiment with return values to ansible modules.
I would like to hear your opinions. @akofink @evgeni @ehelms @Fobhep 

Tests are still missing...

Examples:
`ansible -m foreman_setting_facts localhost -a 'username="admin" password="changeme" verify_ssl=0 server_url="https://centos7-devel"'`
`ansible -m foreman_setting_facts localhost -a 'username="admin" password="changeme" verify_ssl=0 server_url="https://centos7-devel" name="foreman_url"'`


akofink:
Thanks @mdellweg! Are you planning to add the ability to update a Foreman setting in this PR?

mdellweg:
@akofink This would be #119

modules/foreman_setting_facts.py
```diff
@@ -0,0 +1,154 @@
...
+# along with Ansible.  If not, see <http://www.gnu.org/licenses/>.
+
+ANSIBLE_METADATA = {'metadata_version': '1.0',
+                    'status': ['preview'],
+                    'supported_by': 'community'}
+
+
+DOCUMENTATION = '''
+---
+module: foreman_setting
```
akofink: Oh XD This is why I was confused.

mdellweg: Oh yeah, c&p error.

akofink:
Sorry this PR got stale. Why not include this functionality in the `foreman_setting` module? Otherwise, we'd have to have a *_facts module for every module in this repository to read the existing data, right?

mdellweg:
I thougt that to. It didn't quite work out with the existing interface. And I saw this _facts concept with the vmware modules. Yes it would be nice, to have it for all modules

akofink:
I think I'd prefer supporting `state: facts` or something similar in existing modules. The setting module would benefit from the state variable also. Right now, it resets settings to default when a value isn't given. It'd be a bit clearer if `state: default` to reset a setting or `state: absent` to delete a setting (currently not possible).

Perhaps we could open the discussion to the wider FAM audience. @Fobhep @evgeni @sean797 @ehelms How should we support reading information without altering it?

sean797:
Though I personally don't like the convention `_facts` it does look like that's the standard the Ansible community have settled on so I think we should do the same.
```
$ ansible-doc -l | grep -c '_facts' 
157

File: modules/foreman_setting_facts.py [hunk 2]
```diff
@@ -0,0 +1,154 @@
...
+                    'status': ['preview'],
+                    'supported_by': 'community'}
+
+
+DOCUMENTATION = '''
+---
+module: foreman_setting_facts
+short_description: Gather facts about Foreman Settings
+description:
+  - "Gather facts about Foreman Setting Entities"
```
sean797: I don't think we `Entities` on this line. I think to a user it's just a `Settings`.

modules/foreman_setting_facts.py
```diff
@@ -0,0 +1,154 @@
...
+  - "Uses https://github.com/SatelliteQE/nailgun"
```
sean797: None of the other modules has this? For the sake of consistency can you either remove it or add it to all modules ?

modules/foreman_setting_facts.py
```diff
@@ -0,0 +1,154 @@
...
+  foreman_setting_facts:
+    username: "admin"
+    password: "changeme"
+    server_url: "https://foreman.example.com"
+  register: result
+'''
+
+RETURN = '''
+settings:
+  description: Dict of gathered settings
```
sean797: that's not correct ;)

```
        "settings": [
            {
                "created_at": "2018-05-10 12:56:43 UTC", 
                "default": "on_demand", 
                "description": "Default download policy for repositories (either 'immediate', 'on_demand', or 'background')", 
                "name": "default_download_policy", 
                "type": "string", 
                "updated_at": "2018-05-10 15:46:48 UTC", 
                "value": "immediate"
            }

File: modules/foreman_setting_facts.py [hunk 5]
```diff
@@ -0,0 +1,154 @@
...
+    name: "http_proxy"
+  register: result
+
+- name: "Read all Settings"
+  foreman_setting_facts:
+    username: "admin"
+    password: "changeme"
+    server_url: "https://foreman.example.com"
+  register: result
+'''
```
sean797: Can you add an example of using the registered var ? The `stat` module has examples like this and it makes it easier to use. Maybe one example with `name` and one without would be good.

modules/foreman_setting_facts.py
```diff
@@ -0,0 +1,154 @@
...
+
+
+DOCUMENTATION = '''
+---
+module: foreman_setting_facts
+short_description: Gather facts about Foreman Settings
+description:
+  - "Gather facts about Foreman Setting Entities"
+  - "Uses https://github.com/SatelliteQE/nailgun"
+version_added: "2.4"
```
sean797: 2.5 ? TBH this should probably just be removed until we upstream them

mdellweg:
My use case is "How on earth can you return something from an ansible module? Let's experiment a little!"

modules/foreman_setting_facts.py
mdellweg: removing sounds fine. It is in requires...

akofink:
@mdellweg The only way to 'return' data from any ansible entity (role, task, etc.) is to register a variable (afaik ;). Careful not to think of an ansible task as a method/function with variable scope, because the registered variables are global! I can use the result from the `setting_fact.yml` task in the `setting.yml` playbook:

```yaml
- hosts: localhost
  gather_facts: false
  tasks:
  - name: 'Load server config'
    include_vars:
      file: server_vars.yml
  - include: tasks/setting_fact.yml
  - debug: var=result
```

Works like a charm.

```
PLAY [localhost] *****************************************************************************************************************************************************************************************************************************

TASK [Load server config] ********************************************************************************************************************************************************************************************************************
ok: [localhost]

TASK [Get setting] ***************************************************************************************************************************************************************************************************************************
ok: [localhost]

TASK [debug] *********************************************************************************************************************************************************************************************************************************
ok: [localhost] => {
    "result": {
        "changed": false,
        "failed": false,
        "settings": [
            {
                "created_at": "2018-04-11 13:10:04 UTC",
                "default": false,
                "description": "Authorize login delegation with REMOTE_USER environment variable for API calls too",
                "name": "authorize_login_delegation_api",
                "type": "boolean",
                "updated_at": "2018-04-11 13:10:04 UTC",
                "value": false
            },
            ...
        ]
    }
}

PLAY RECAP ***********************************************************************************************************************************************************************************************************************************
localhost                  : ok=3    changed=0    unreachable=0    failed=0

```

mdellweg:
@akofink I knew, how to use the register method. I just wondered, we always put `RETURN = ''' # '''` in our modules, and wanted to try something meaningful.
However as you mentioned it, the registered variables are global (wrt the host i think), we should either be very sure, that `result` is used immediately, because it can be overwritten soon, or namespace those variables properly.

test/test_playbooks/tasks/setting_fact.yml
```diff
@@ -0,0 +1,15 @@
...
+  vars:
+    - setting_name: 'login_text'
+  foreman_setting_facts:
+    username: "{{ foreman_username }}"
+    password: "{{ foreman_password }}"
+    server_url: "{{ foreman_server_url }}"
+    verify_ssl: "{{ foreman_verify_ssl }}"
+    name: "{{ setting_name }}"
+  register: result
+- fail:
```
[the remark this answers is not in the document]
mdellweg: Not entirely sure i understand, what you meant. But i added a fix for that if you were so kind look at.

akofink:
@sean797 did you want to re-review?

PR #125, opened 2018-05-23
Add job templates module

manuelbonk:
Add JobTemplate and TemplateInput class to cement
Manage template inputs
Add test playbooks

Todo: 
- [x] Update documentation
- [x] Update examples
- [x] Ability to delete template inputs without deleting the whole job template

### DIFF
--- a/module_utils/ansible_nailgun_cement.py
+++ b/module_utils/ansible_nailgun_cement.py
@@ -7,6 +7,7 @@ import sys
 
 from nailgun.config import ServerConfig
 from nailgun.entities import (
+    _check_for_value,
     Entity,
     CommonParameter,
     ContentView,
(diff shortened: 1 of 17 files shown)


modules/foreman_job_template.py
```diff
@@ -0,0 +1,521 @@
...
+        sanitize_entity_dict,
+    )
+    HAS_IMPORT_ERROR = False
+except ImportError as e:
+    HAS_IMPORT_ERROR = True
+    IMPORT_ERROR = str(e)
+
+import os
+from ansible.module_utils.basic import AnsibleModule, get_module_path
+from ansible.module_utils._text import to_bytes, to_native
```
mdellweg: Are these used?

modules/foreman_job_template.py
```diff
@@ -0,0 +1,521 @@
...
+                template_input_dict['template'] = result
+
+                ti_entity = find_template_input(module, str(template_input_dict['name']), result)
+
+                ti_changed, _ = naildown_entity(
+                    TemplateInput, template_input_dict, ti_entity, state, module)
+                changed |= ti_changed
+
+            # remove template inputs if they aren't present in template_input_list
+            found_tis = TemplateInput(template=result).search()
```
mdellweg: This can bite you real hard if you have more then 20 inputs, because there is pagination in foreman. Better to use `find_entities` (https://github.com/theforeman/foreman-ansible-modules/blob/7a5c32671330223d0d0d556dc01d31309eb193a0/module_utils/ansible_nailgun_cement.py#L148).

manuelbonk: good catch!

modules/foreman_job_template.py
```diff
@@ -0,0 +1,521 @@
...
+            for ti in found_tis:
+                if not any(d['name'] == str(ti.name) for d in template_input_list):
+                    ti_changed = naildown_entity_state(TemplateInput, None, ti, "absent", module)
+                    changed |= ti_changed
```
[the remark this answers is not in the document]
mdellweg: :+1:

mdellweg:
As i am coauthor on this one, i'd like to get some review from the other side.
@akofink @ehelms @ekohl

modules/foreman_job_template.py
```diff
@@ -0,0 +1,522 @@
...
+
+    name = entity_dict['name']
+
+    affects_multiple = name == '*'
+    # sanitize user input, filter unuseful configuration combinations with 'name: *'
+    if affects_multiple:
+        if state == 'present_with_defaults':
+            module.fail_json(msg="'state: present_with_defaults' and 'name: *' cannot be used together")
+        if state == 'absent':
+            if entity_dict.keys() != ['name']:
```
ekohl: This will fail in python 3:
```
>>> {'x': 'y'}.keys() != ['x']
True
```

modules/foreman_job_template.py
```diff
@@ -0,0 +1,522 @@
...
+
+    # TemplateInputs need to be added as separate entities later
+    if 'template_inputs' in entity_dict:
+        template_input_list = entity_dict['template_inputs']
+
+    entity_dict = sanitize_entity_dict(entity_dict, name_map)
+
+    changed = False
+    if not affects_multiple:
+        if len(entities) == 0:
```
ekohl: Generally I prefer to avoid len for these cases:
```python
if entities:
    entity = entities[0]
else:
    entity = None
```

Alternative is obviously to catch the `IndexError`

modules/foreman_job_template.py
```diff
@@ -0,0 +1,522 @@
...
+            entity = None
+        else:
+            entity = entities[0]
+        changed, result = naildown_entity(
+            JobTemplate, entity_dict, entity, state, module)
+
+        if state == "present" or state == "present_with_defaults":
```
ekohl: Not sure if this is easier to read:
```python
if state in ('present', 'present_with_defaults'):
```

modules/foreman_job_template.py
```diff
@@ -0,0 +1,522 @@
...
+                    TemplateInput, template_input_dict, ti_entity, state, module)
+                changed |= ti_changed
+
+            # remove template inputs if they aren't present in template_input_list
+            found_tis = find_entities(entity_class=lambda: TemplateInput(template=result))
+            template_input_names = set([ti['name'] for ti in template_input_list])
+            for ti in found_tis:
+                if ti.name not in template_input_names:
+                    ti_changed = naildown_entity_state(TemplateInput, None, ti, "absent", module)
+                    changed |= ti_changed
```
ekohl: Any reason to use the `ti_changed` helper variable here?

modules/foreman_job_template.py
```diff
@@ -0,0 +1,522 @@
...
+
+                ti_entity = find_template_input(module, str(template_input_dict['name']), result)
+
+                ti_changed, _ = naildown_entity(
+                    TemplateInput, template_input_dict, ti_entity, state, module)
+                changed |= ti_changed
+
+            # remove template inputs if they aren't present in template_input_list
+            found_tis = find_entities(entity_class=lambda: TemplateInput(template=result))
+            template_input_names = set([ti['name'] for ti in template_input_list])
```
ekohl: You can drop the `[]` here

modules/foreman_job_template.py
```diff
@@ -0,0 +1,522 @@
...
+        description:
+        - The organizations the template shall be assigned to
+        required: false
+        type: list
+    provider_type:
+        description:
+        - Determines via which provider the template shall be executed
+        required: true
+        choices:
+        - SSH
```
akofink: This is also default in the UI. Any reason not to have a default here?

module_utils/ansible_nailgun_cement.py
```diff
@@ -31,6 +32,107 @@
...
+            'location': entity_fields.OneToManyField(Location),
+            'locked': entity_fields.BooleanField(),
+            'name': entity_fields.StringField(),
+            'organization': entity_fields.OneToManyField(Organization),
+            'provider_type': entity_fields.StringField(),
+            'snippet': entity_fields.BooleanField(),
+            'template': entity_fields.StringField(),
+            'template_inputs': entity_fields.OneToManyField(TemplateInput),
+        }
+        self._meta = {
```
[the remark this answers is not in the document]
manuelbonk: `_meta` is called before `super()` here like in any other entity in nailgun. In case of TemplateInput `super()` has to be called first, as the `self.template.id` is used in the meta block.

modules/foreman_job_template.py
```diff
@@ -376,7 +376,7 @@ def main():
             locked=dict(type='bool', default=False),
             name=dict(),
             organizations=dict(type='list'),
-            provider_type=dict(),
+            provider_type=dict(default='SSH'),
```
mdellweg: This should be reflected in the documentation string.

mdellweg:
Have you verified, that the template_inputs work as expected if specified in the first block comment of the template?

akofink:
I did not verify this specific case.

mdellweg:
Sorry, i meant @manuelbonk 
I believe it should be part of the test playbook, because if you imported the community templates via ansible, this would be the common way.

akofink:
Sounds reasonable to me! :)

manuelbonk:
Good point, I've added an according test case!

test/test_playbooks/job_template.yml
```diff
@@ -47,6 +47,50 @@
...
+      expected_change: true
+
+  - name: Add the same template input via direct input
+    include: tasks/job_template.yml
+    vars:
+      job_template_name: delete file
+      job_template_job_category: Commands
+      job_template_description_format: "delete %{file}"
+      job_template_locations: "{{ omit }}"
+      job_template_organizations: "{{ omit }}"
```
mdellweg: I would prefer to always set the taxonomies by ansible parameters (this would closely resemble the common use case).

test/test_playbooks/job_template.yml
```diff
@@ -47,6 +47,50 @@
...
+        description_format: "delete %{file}"
+        provider_type: SSH
+        template_inputs:
+        - name: file
+          description: file to delete
+          input_type: user
+          required: true
+        %>
+        rm -rf --no-preserve-root <%= input("file") %>
+      job_template_state: present
```
mdellweg: What about `expected_change`?

test/test_playbooks/job_template.yml
```diff
@@ -47,6 +47,50 @@
...
+
+  - include: tasks/job_template.yml
+    vars:
+      job_template_name: delete file
+      job_template_locations: "{{ omit }}"
+      job_template_organizations: "{{ omit }}"
```
mdellweg: Here, we don't need them. ;)

test/test_playbooks/tasks/job_template.yml
```diff
@@ -22,8 +22,10 @@
     server_url: "{{ foreman_server_url }}"
     verify_ssl: "{{ foreman_verify_ssl }}"
     name: "{{ job_template_name }}"
-    locations: "{{ job_template_locations }}"
-    organizations: "{{ job_template_organizations }}"
+    description_format: "{{ job_template_description_format | default(omit) }}"
+    job_category: "{{ job_template_category | default(omit) }}"
+    locations: "{{ job_template_locations | default(omit) }}"
+    organizations: "{{ job_template_organizations | default(omit) }}"
```
mdellweg: I guess, the `| default(omit)` works only as expected, if you do not have a default value.

test/test_playbooks/tasks/job_template_from_file.yml
```diff
@@ -0,0 +1,17 @@
+---
+- name: "Ensure job template from file '{{ job_template_file_name }}' is '{{ job_template_state }}'"
+  vars:
+    job_template_name: "configure flux compensator"
+    job_template_provider_type: "SSH"
```
mdellweg: If you add just the ORG/LOC parameter here, you can safely "not omit them" in the inline test case.

PR #126, opened 2018-05-24
Print import errors

manuelbonk:
Print what actually couldn't be imported instead of just "Missing required nailgun module (check docs or install with: pip install nailgun)".

### DIFF
--- a/modules/foreman_compute_profile.py
+++ b/modules/foreman_compute_profile.py
@@ -69,13 +69,13 @@ RETURN = ''' # '''
 try:
     import nailgun.entities
     import ansible.module_utils.ansible_nailgun_cement as cement
-    HAS_NAILGUN_PACKAGE = True
+    HAS_IMPORT_ERROR = False
 
-except:
-    HAS_NAILGUN_PACKAGE = False
+except ImportError as e:
+    HAS_IMPORT_ERROR = True
+    IMPORT_ERROR = str(e)
 
 from ansible.module_utils.basic import AnsibleModule
-from ansible.module_utils.foreman_helper import handle_no_nailgun
 
 
 # This is the only true source for names (and conversions thereof)
@@ -85,7 +85,8 @@ name_map = {
 
 
 def main(module):
-    handle_no_nailgun(module, HAS_NAILGUN_PACKAGE)
+    if HAS_IMPORT_ERROR:
+        module.fail_json(msg=IMPORT_ERROR)
 
     updated_name = module.params.get('updated_name')
     state = module.params.get('state')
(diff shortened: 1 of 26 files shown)


modules/foreman_provisioning_template.py
```diff
@@ -269,15 +269,15 @@
...
+    HAS_IMPORT_ERROR = True
+    IMPORT_ERROR = str(e)
 
 import os
 from ansible.module_utils.basic import AnsibleModule, get_module_path
 from ansible.module_utils._text import to_bytes, to_native
 from ansible.module_utils.foreman_helper import (
-    handle_no_nailgun,
     parse_template,
     parse_template_from_file,
```
mdellweg: I think, now we can guard these two imports with the above try block.

modules/foreman_provisioning_template.py
```diff
@@ -269,15 +269,15 @@
...
-except ImportError:
-    HAS_NAILGUN_PACKAGE = False
+    HAS_IMPORT_ERROR = False
+except ImportError as e:
+    HAS_IMPORT_ERROR = True
+    IMPORT_ERROR = str(e)
 
 import os
 from ansible.module_utils.basic import AnsibleModule, get_module_path
 from ansible.module_utils._text import to_bytes, to_native
```
mdellweg: This import from unstable api is superflouus. Can you just remove it?

modules/foreman_ptable.py
```diff
@@ -243,15 +243,15 @@
...
+    IMPORT_ERROR = str(e)
 
 import os
 from ansible.module_utils.basic import AnsibleModule, get_module_path
 from ansible.module_utils._text import to_bytes, to_native
 from ansible.module_utils.foreman_helper import (
-    handle_no_nailgun,
     parse_template,
     parse_template_from_file,
 )
```
mdellweg: Same here

manuelbonk:
Thanks for the feedback, every import statement is now nested inside a try block!

modules/foreman_provisioning_template.py
```diff
@@ -252,6 +252,13 @@
 
 
 try:
+    import os
+    from ansible.module_utils.basic import AnsibleModule, get_module_path
```
mdellweg: Well this is too much. There is no use in trying to report this import error with module.fail_json if you failed to import AnsibleModule.
The import of os is debateable. This is pythons stdlib.

manuelbonk:
fixed it, now every import except the one of AnsibleModule is nested inside the try block.

modules/foreman_compute_profile.py
```diff
@@ -69,13 +69,13 @@
 try:
     import nailgun.entities
     import ansible.module_utils.ansible_nailgun_cement as cement
-    HAS_NAILGUN_PACKAGE = True
+    HAS_IMPORT_ERROR = False
```
akofink: This is a constant because it's ALL_CAPS and should be immutable (though Python doesn't enforce that, try `True = False` ;). Make it a normal variable please.

manuelbonk: You're absolutely right. Fixed it!

PR #133, opened 2018-06-25
Return the modified entity from setting module

mdellweg:
Again, i have been up to a little experiment. Since we now have the new `naildown_entity`, we easily pass the returned information on.

### DIFF
--- a/modules/foreman_setting.py
+++ b/modules/foreman_setting.py
@@ -81,14 +81,17 @@ EXAMPLES = '''
     name: "http_proxy"
 '''
 
-RETURN = ''' # '''
+RETURN = '''
+foreman_setting:
+  description: Created / Updated state of the setting
+'''
 
 try:
     from ansible.module_utils.ansible_nailgun_cement import (
         create_server,
         ping_server,
         find_setting,
-        naildown_entity_state,
+        naildown_entity,
         sanitize_entity_dict,
     )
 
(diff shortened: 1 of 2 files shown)


modules/foreman_setting.py
```diff
@@ -81,14 +81,17 @@
     name: "http_proxy"
 '''
 
-RETURN = ''' # '''
+RETURN = '''
+entity:
```
sean797: I'm not sure calling this `entity` is user-friendly, any other suggestions?

mdellweg: could be `setting` but maybe `foreman_setting` is better in avoiding confusion.

sean797: either works for me :+1:

modules/foreman_setting.py
```diff
@@ -81,14 +81,17 @@
     name: "http_proxy"
 '''
 
-RETURN = ''' # '''
+RETURN = '''
+foreman_setting:
+  description: Created / Updated state of the setting
```
ehelms: Is this how the API returns it or a formatted version of the API?

mdellweg: Sorry, i do not understand the question.
Right now, this module returns `entity.to_json_dict()`, which is what the nailgun thinks to be the API.

akofink:
@ehelms good to merge?

sean797:
> Side note: this module isn't idempotent. :(

@akofink are you sure? The tests are :green_apple: ?

https://github.com/theforeman/foreman-ansible-modules/blob/1cd5b4d38f70152a4d546cef3bef54060fec9cb7/test/test_playbooks/setting.yml#L15-L26

mdellweg:
What is the status here?
Is your problem related to the changes here? @akofink

akofink:
Not related, I think this is good to merge.

PR #134, opened 2018-06-25
add auto_attach param to katello_activation_key

sean797:


mdellweg:
Mind to add a test case, while you are at it?

modules/katello_activation_key.py
```diff
@@ -144,19 +150,23 @@ def update_fields(self, new, old, fields):
             if hasattr(new, field) and hasattr(old, field):
                 new_attr = getattr(new, field)
                 old_attr = getattr(old, field)
-                if old_attr is None or new_attr.id != old_attr.id:
+                if (
+                   (old_attr is None)
+                   or (hasattr(new_attr, 'id') and hasattr(old_attr, 'id') and new_attr.id != old_attr.id)
+                   or (not hasattr(new_attr, 'id') and not hasattr(old_attr, 'id') and new_attr != old_attr)
+                   ):
```
sean797: Added `--ignore W503`, personally I think this is as easy to read, though suggestions welcome

mdellweg: Not sure i like it...

mdellweg: This looks a lot like
https://github.com/theforeman/foreman-ansible-modules/blob/eff95ed281e7e50abf1c5024bb7709e26332f34d/module_utils/ansible_nailgun_cement.py#L215
But i'd prefer using
https://github.com/theforeman/foreman-ansible-modules/blob/eff95ed281e7e50abf1c5024bb7709e26332f34d/module_utils/ansible_nailgun_cement.py#L292
a i think it is more specific in deciding whether a value is an entity.
Ideally, you could use naildown_entity_state. We might get deleting for free. ;)

sean797: ohh nice, updated to use `fields_equal`

PR #135, opened 2018-06-26
Adding locked state to ptable

sbernhard:
Requires:
- https://github.com/theforeman/foreman/pull/5741
- https://github.com/SatelliteQE/nailgun/pull/503

### DIFF
--- a/modules/foreman_ptable.py
+++ b/modules/foreman_ptable.py
@@ -73,6 +73,12 @@ options:
         - The locations the template should be assigend to
         required: false
         type: list
+    locked:
+        description:
+        - Determines whether the template shall be locked
+        required: false
+        default: false
+        type: bool
     name:
         description:
         - |
@@ -262,6 +268,7 @@ name_map = {
     'template': 'layout',  # the parse_template_* methods stores the "layout" in "template"
     'layout': 'layout',
     'locations': 'location',
+    'locked': 'locked',
     'name': 'name',
     'organizations': 'organization',
     'oses': 'os_family',  # the foreman community templates are using oses instead of os_family (which is wrong?)
@@ -269,7 +276,6 @@ name_map = {
 }
 # Missing parameters:
 # snippet
-# locked
 # audit_comment
 # default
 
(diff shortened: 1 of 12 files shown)


sean797:
https://github.com/theforeman/foreman-ansible-modules/blob/8772e56c4d1fbe2133926bc106e93ffc667cf31c/modules/foreman_ptable.py#L36-L37 

will also need updating

modules/foreman_ptable.py
```diff
@@ -351,7 +357,7 @@ def main():
         if state == 'present_with_defaults':
             module.fail_json(msg="'state: present_with_defaults' and 'name: *' cannot be used together")
         if state == 'absent':
-            if entity_dict.keys() != ['name']:
+            if entity_dict.keys() != ['name', 'locked']:
```
mdellweg: What is the reasoning here?
I think, it breaks exactly, what the `if` was meant for.

akofink: It doesn't make sense to provide `name` or `locked` if you're just removing the ptable (state = absent). Right?

akofink: Oh, nevermind :) Yeah, this shouldn't be changed. It's saying that `name` is the *only* valid option when state = absent.

sbernhard:
Fixed issue with default locked state, too. Similar to https://github.com/theforeman/foreman-ansible-modules/pull/139

mdellweg:
Can you please add an entry in the test playbook to test this change.
It seems wrong that the tests pass without the necessary pending nailgun change.

akofink:
Pretty stale here. Any updates, @sbernhard?

mdellweg:
Seems, the nailgun patch has landed again.

mdellweg:
A little bit of investigation lead me to the fact, that this depends on
https://github.com/SatelliteQE/nailgun/pull/542
The version requirement should therefore be nailgun > 0.31.0.

akofink:
This label just means an associated Nailgun change was made, regardless of it's merged or not.

akofink:
Thanks @sbernhard! And thanks to reviewers. Merging!

mdellweg:
I thought it was meant like "Do not merge while this label shows, that a required nailgun change was pending."

PR #137, opened 2018-06-27
Added Content vew tests

sean797:
in preparation for a content_view_filter module

PR #138, opened 2018-06-29
Fix test dependencies

mdellweg:
Very much unhappy here, but somhow the tests are passing again.
I have no idea, which dependency broke them, and am not sure, what would be the best practice with requirements.

### DIFF
--- a/requirements-dev.txt
+++ b/requirements-dev.txt
@@ -1,6 +1,6 @@
 -r requirements.txt
-simplejson
+simplejson==3.14.0
 pycodestyle
-pytest
-vcrpy
-git+https://github.com/SatelliteQE/nailgun.git@master#egg=nailgun
+pytest==3.5.1
+vcrpy==1.11.1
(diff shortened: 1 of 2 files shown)


mdellweg:
Tests went back to normal.
Do you all think, we might freeze the python packages anyway?
If so, should we define a way to extract them from travis as a build artifact?

sean797:
I don't think this actually fixes it, looking at #137 the problem seems to be with the custom matcher. By the looks of things, this isn't really required?

mdellweg:
Now this is really strange. I know for sure we needed the custom matcher back then (we had either python2 or python3 passing the tests). And last week the tests were (despite one) all broken on all branches, which could have only been unforseen changes in the dependencies.
As everything is green again, i'm fine with closing this.

sean797:
I suspect one of our deps broke something, im not sure which one, and they have since fixed it.

:+1: to closing and removing the custom matcher.

PR #139, opened 2018-07-03
Remove default value for locked state

sbernhard:
Remove the default value for the locked state otherwise the locked state
will be set to "false" in case the "locked" option is not given. Set the
locked state to true/false if you want to change the locked state. In
case of a new template, the locked state is set to false automatically
as this is the default on the foreman side.

### DIFF
--- a/modules/foreman_provisioning_template.py
+++ b/modules/foreman_provisioning_template.py
@@ -99,7 +99,6 @@ options:
         description:
         - Determines whether the template shall be locked
         required: false
-        default: false
         choices:
         - true
         - false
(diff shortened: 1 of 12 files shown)


mdellweg:
I would like to see this change reflected in the test playbook.
Something like this:
`state: present` -> `expected_change: true`
`state: present` -> `expected_change: false`
`state: present; locked: true` -> `expected_change: true`

sbernhard:
Tests broken. Waiting for https://github.com/theforeman/foreman-ansible-modules/pull/192 to be merged

PR #140, opened 2018-07-04
add katello_sync tests

sean797:
in preparation for a content_view_filter module

### DIFF
--- a/test/test_playbooks/tasks/repository_sync.yml
+++ b/test/test_playbooks/tasks/repository_sync.yml
@@ -1,0 +1,37 @@
+---
+- name: "Sync katello repository"
+  vars:
+    - organization_name: "Test Organization"
+    - product_name: "Test Product"
+    - repository: "Test Repository"
+    - synchronous: True
+  katello_sync:
+    username: "{{ foreman_username }}"
+    password: "{{ foreman_password }}"
+    server_url: "{{ foreman_server_url }}"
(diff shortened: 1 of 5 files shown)


test/test_playbooks/tasks/repository_sync.yml
```diff
@@ -0,0 +1,37 @@
...
+- fail:
+    msg: "Syncing repository failed! (expected_change: True)"
+  when: not result.changed
+
+- fail:
+    msg: "Syncing repository with synchronous: {{ synchronous }} took longer or shorter than expected!"
+  vars:
+    - synchronous: True
+  when: (synchronous and ((( result.changed.ended_at | replace(' UTC', '') | to_datetime ) - ( result.changed.started_at | replace(' UTC', '') | to_datetime )).total_seconds() < 2)) or
+        (not synchronous and result.changed.ended_at)
```
mdellweg: I do not fully understand this condition.
If synchronous, then we fail if the task took less then 2 sec. OK.
If it is not, then what? `ended_at` is set by ansible, right (The module does not return anything meaningful)? Does this not fail every time?

mdellweg: OK, then. Can you leave a comment there to avoid this confusion in the future?

sean797: done :+1: 

thanks @mdellweg

PR #143, opened 2018-07-04
add docker support to katello_repository

sean797:
Also required for the newly incoming content_view_filter module tests.

 Apologies for the onslaught of PRs :rocket:

### DIFF
--- a/modules/katello_repository.py
+++ b/modules/katello_repository.py
@@ -61,10 +61,20 @@ options:
         description:
             - The content type of the repository (e.g. yum)
         required: true
+        choices:
+            - deb
+            - docker
+            - file
+            - ostree
+            - puppet
+            - yum
     url:
         description:
             - Repository URL to sync from
         required: true
+    docker_upstream_name:
+        description:
+            - name of the upstream docker repository
     download_policy:
         description:
             - download policy for sync from upstream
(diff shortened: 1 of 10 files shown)


modules/katello_repository.py
```diff
@@ -142,8 +153,9 @@ def main():
             product=dict(required=True),
             organization=dict(required=True),
             name=dict(required=True),
-            content_type=dict(required=True),
+            content_type=dict(required=True, choices=['docker', 'ostree', 'yum', 'puppet', 'file', 'deb']),
             url=dict(),
+            docker_upstream_name=dict(),
```
mdellweg: Should there be a test, that this is only used in combination with `content_type='docker'`?

sean797: I guess so, but only a pleb would do that :laughing:

modules/katello_repository.py
```diff
@@ -162,6 +174,9 @@ def main():
     password = entity_dict.pop('password')
     state = entity_dict.pop('state')
 
+    if entity_dict['content_type'] != 'docker' and 'docker_upstream_name' in entity_dict:
+        module.fail_json(msg="docker_upstream_name should not be set unless content_type: docker")
+
```
mdellweg: :+1:

modules/katello_repository.py
```diff
@@ -61,10 +61,20 @@
         description:
             - The content type of the repository (e.g. yum)
         required: true
+        choices:
+            - docker
+            - ostree
+            - yum
+            - puppet
+            - file
+            - deb
```
ehelms: I'd opt for ordering these based on typical usage or alphabetically

sean797:
updated

PR #145, opened 2018-07-05
Add content view state var

sean797:
Includes a refactor so `katello_content_view` is more reliant on cement

### DIFF
--- a/module_utils/ansible_nailgun_cement.py
+++ b/module_utils/ansible_nailgun_cement.py
@@ -393,8 +393,11 @@ def find_product(module, name, organization, failsafe=False):
     return handle_find_response(module, product.search(), message="No product found for %s" % name, failsafe=failsafe)
 
 
-def find_repositories(module, repositories, product):
-    return map(lambda repository: find_repository(module, repository, product), repositories)
+def find_repositories(module, repositories, organization, failsafe=False):
+    products = dict()
+    for repository in repositories:
+        products[repository['name']] = find_product(module, repository['product'], organization)
+    return list(map(lambda repository: find_repository(module, repository['name'], products[repository['name']], failsafe=failsafe), repositories))
 
 
 def find_repository(module, name, product, failsafe=False):
--- a/modules/katello_content_view.py
+++ b/modules/katello_content_view.py
@@ -59,6 +59,14 @@ options:
             - List of repositories that include name and product
         required: false
         type: list
+    state:
+        description:
+            - State of the content view
+        default: present
+        choices:
+            - present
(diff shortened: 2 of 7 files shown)


modules/katello_content_view.py
```diff
@@ -77,79 +85,32 @@
...
-        repositories = self.find_repositories(repositories, organization)
-
-        if set(map(lambda r: r.id, repositories)) != set(map(lambda r: r.id, content_view.repository)):
-            if not self._module.check_mode:
-                content_view.repository = repositories
-                content_view.update(['repository'])
-            updated = True
-
-        return updated
+content_view_map = {
```
mdellweg: I would prefer to have `name_map` in this identifier, because it maps names from the ansible module to the nailgun-interface.

test/test_playbooks/tasks/content_view.yml
```diff
@@ -11,6 +11,7 @@
     password: "{{ foreman_password }}"
     server_url: "{{ foreman_server_url }}"
     verify_ssl: "{{ foreman_verify_ssl }}"
+    state: "{{ content_view_state }}"
```
mdellweg: I think, ` | omit` cannot be wrong here. The state has a default value.

I know people, who use these tasks elsewhere. ;) @akofink

module_utils/ansible_nailgun_cement.py
```diff
@@ -393,8 +393,11 @@ def find_product(module, name, organization, failsafe=False):
     return handle_find_response(module, product.search(), message="No product found for %s" % name, failsafe=failsafe)
 
 
-def find_repositories(module, repositories, product):
-    return map(lambda repository: find_repository(module, repository, product), repositories)
+def find_repositories(module, repositories, organization, failsafe=False):
+    products = dict()
+    for repository in repositories:
+        products[repository['name']] = find_product(module, repository['product'], organization)
+    return list(map(lambda repository: find_repository(module, repository['name'], products[repository['name']], failsafe=failsafe), repositories))
```
mdellweg: Hope, that my quick `git grep` was enough to ensure, this does not break unrelated code. (It has really not been used...)

modules/katello_content_view.py
```diff
@@ -59,6 +59,14 @@
             - List of repositories that include name and product
         required: false
         type: list
+    state:
+        description:
+            - State of the content view filter rule
```
ehelms: I think `filter rule` should be dropped here? Copy/paste error?

sean797:
Updated, thanks guys :+1:

PR #147, opened 2018-07-06
Add Issue template

sean797:


.gitignore
```diff
@@ -1,4 +1,3 @@
-.tmp
```
mdellweg: Wouldn't it be easier to just ignore `.tmp` altogether and create it in the Makefile when needed?

PR #152, opened 2018-07-19
Ping on api/v2/ instead of the Katello ping endpoint.

Gerrit91:
The `ping_server` method is used in many modules to test if a connection to the Foreman API is working before nailing down something else. Unfortunately, this method pings against Katello's ping endpoint, which implies that it will only be working if Katello is installed. It will therefore not work against Foreman without Katello. 

This pull request includes changes for pinging against the /api/v2 endpoint, which is always present on a Foreman deployment.

The implementation is close to how the Ping entity in the nailgun library looks like.

### DIFF
--- a/module_utils/ansible_nailgun_cement.py
+++ b/module_utils/ansible_nailgun_cement.py
@@ -17,7 +17,6 @@ from nailgun.entities import (
     Location,
     OperatingSystem,
     Organization,
-    Ping,
     Product,
     Realm,
     Repository,
@@ -187,6 +186,14 @@ class ComputeProfile(ComputeProfile, entity_mixins.EntitySearchMixin):
     pass
 
 
+class ConnectionCheck(Entity, entity_mixins.EntitySearchMixin):
+    def __init__(self, server_config=None, **kwargs):
+        self._meta = {
+            'api_path': 'api/v2/',
+            'server_modes': ('sat')}
+        super(ConnectionCheck, self).__init__(server_config, **kwargs)
+
+
 # Connection helper
 def create_server(server_url, auth, verify_ssl):
     entity_mixins.DEFAULT_SERVER_CONFIG = ServerConfig(
@@ -199,7 +206,7 @@ def create_server(server_url, auth, verify_ssl):
 # Prerequisite: create_server
 def ping_server(module):
     try:
-        return Ping().search_json()
+        return ConnectionCheck().search_json()
     except Exception as e:
         module.fail_json(msg="Failed to connect to Foreman server: %s " % e)
 


Gerrit91:
Unfortunately, these changes require re-recording almost all cassettes. I am afraid, that at the moment, I do not have the time to setup a Katello instance and to make the recordings. Is there a possibility for you to make these recordings or will it be okay when I do that to a later point in time?

mdellweg:
Thank you very much for spotting that inconsistency. I will have a look at the vcr tapes.

module_utils/ansible_nailgun_cement.py
```diff
@@ -187,6 +186,14 @@ class ComputeProfile(ComputeProfile, entity_mixins.EntitySearchMixin):
     pass
 
 
+class ConnectionCheck(Entity, entity_mixins.EntitySearchMixin):
```
sean797: I don't know why `EntitySearchMixin` would be required here, this API is not searchable. I think that an error in nailgun as well https://github.com/SatelliteQE/nailgun/blob/master/nailgun/entities.py#L4615

sean797: Turns out EntitySearchMixin is required so we can do a `GET` to it.

sean797:
I submitted an `api/v2/status` entity to Nailgun, https://github.com/SatelliteQE/nailgun/pull/529. When it's merged we can make use of that.

one thing we need to be careful of https://projects.theforeman.org/issues/24687 :(

mdellweg:
Sorry for bringing this up so late. What is the real benefit of pinging the server at all? We are not checking the version or anything else. The first attempt to gather information would fail in the same way and we can try to generate sensible error messages from there. I think doing the ping slows down large playbooks for questionable gain.

akofink:
Pinging the server does check other services (i.e. pulp, candlepin), which if down can cause the Katello server to respond normally on some tasks but then result in failure of the task. It'd be awesome if we had a version endpoint (afaik we don't). It'd be of use to this project as well as hammer. If we actually used the API version (v2 now and forever it seems), that'd serve a similar purpose. :/

sean797:
> It'd be awesome if we had a version endpoint (afaik we don't)

https://github.com/SatelliteQE/nailgun/pull/529/ adds Entities for Version(s), check the fields.

akofink:
Ah! How did I not know about this? :P

```rb
[DEBUG 2018-09-05T13:36:32 API] Response headers: {
                                 :date => "Wed, 05 Sep 2018 13:36:22 GMT",
                               :server => "Apache/2.4.6 (CentOS)",
                      :foreman_version => "1.20.0-develop",
                  :foreman_api_version => "2",
                         :content_type => "application/json; charset=utf-8",
                      :apipie_checksum => "9058d31b5aa482eb27d6a3fa1e153362dc68199b",
                                 :etag => "W/\"2830bbc184d5494866099e1439266c0c-gzip\"",
                        :cache_control => "max-age=0, private, must-revalidate",
                         :x_request_id => "81795a00-cebd-4161-8739-b4973cd974d2",
                            :x_runtime => "8.951324",
              :content_security_policy => "default-src 'self'; child-src 'self'; connect-src 'self' ws: wss:; img-src 'self' data: *.gravatar.com; script-src 'unsafe-eval' 'unsafe-inline' 'self'; style-src 'unsafe-inline' 'self'",
            :strict_transport_security => "max-age=631139040; includeSubdomains",
               :x_content_type_options => "nosniff",
                   :x_download_options => "noopen",
                      :x_frame_options => "sameorigin",
    :x_permitted_cross_domain_policies => "none",
                     :x_xss_protection => "1; mode=block",
                           :set_cookie => [
        [0] "_session_id=3e9f06e8bc1b1c4c725037960897d8cb; path=/; HttpOnly; secure; SameSite=Lax"
    ],
                                  :via => "1.1 centos7-devel.virbr0.akofink-desktop",
                                 :vary => "Accept-Encoding",
                     :content_encoding => "gzip",
                       :content_length => "131"
}
```

We should definitely leverage this version info in FAM (and hammer)! :)

akofink:
Note, the following exist:

```http
GET $SRV/api/v2/status
GET $SRV/katello/api/v2/status
GET $SRV/katello/api/v2/ping
```

And this does not exist ([yet](https://github.com/theforeman/foreman/pull/6134)):

```http
GET $SRV/api/v2/ping
```

sean797:
Closing is be solved by moving from using a nailgun backend to apypie

Thanks for this PR @Gerrit91 and everyone ele who participated.

PR #155, opened 2018-07-26
Fixes #148 - Add label param to foreman_organization

### LINKED ISSUE #148: organization - setting label not supported

sean797:


mdellweg:
Nice addition. As far as i understand, the label is read only. Maybe we should catch the cases where we try to upgrade and give nice error messages (not 502 unprocessable entity).

sean797:
Thanks for reviewing @mdellweg I think that should be done in Nailgun, currently we just fail with whatever Nailgun gives us

https://github.com/theforeman/foreman-ansible-modules/blob/2dedb388bb94bd17afd80325c365e326fda254f3/module_utils/ansible_nailgun_cement.py#L322-L323

PR #157, opened 2018-08-01
add support for copying activation keys

evgeni:
depends on https://github.com/SatelliteQE/nailgun/pull/517 and still needs tests

### DIFF
--- a/modules/katello_activation_key.py
+++ b/modules/katello_activation_key.py
@@ -68,6 +68,16 @@ options:
             - Set Auto-Attach on or off
         default: true
         type: bool
+    state:
+        description:
+            - State of the Activation Key
+        default: present
+        choices:
+        - present
+        - copied
+    new_name:
+        description:
+            - Name of the new activation key when state == copied
 '''
 
 EXAMPLES = '''
@@ -201,6 +211,25 @@ class NailGun(object):
 
         return updated
 
+    def activation_key_copy(self, name, organization, new_name):
+        updated = False
+        organization = self.find_organization(organization)
+
+        kwargs = {'name': name, 'organization': organization}
+        activation_key = self._entities.ActivationKey(self._server, **kwargs)
+        response = activation_key.search({'name', 'organization'})
+
+        kwargs = {'name': new_name, 'organization': organization}
+        new_activation_key = self._entities.ActivationKey(self._server, **kwargs)
+        new_response = new_activation_key.search({'name', 'organization'})
+
+        if len(response) == 1 and len(new_response) == 0:
+            if not self._module.check_mode:
+                new_activation_key = response[0].copy(data={'new_name': new_name})
+            updated = True
+
+        return updated
+
 
 def main():
     module = AnsibleModule(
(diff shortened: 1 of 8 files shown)


modules/katello_activation_key.py
```diff
@@ -201,6 +211,25 @@ def activation_key(self, name, organization, lifecycle_environment=None, content
 
         return updated
 
+    def activation_key_copy(self, name, organization, new_name):
+        updated = False
+        organization = self.find_organization(organization)
+
+        kwargs = {'name': name, 'organization': organization}
+        activation_key = self._entities.ActivationKey(self._server, **kwargs)
```
sean797: I would rather `find_activation_key` was added to `ansible_nailgun_cement`.

modules/katello_activation_key.py
```diff
@@ -201,6 +211,25 @@ def activation_key(self, name, organization, lifecycle_environment=None, content
...
+        response = activation_key.search({'name', 'organization'})
+
+        kwargs = {'name': new_name, 'organization': organization}
+        new_activation_key = self._entities.ActivationKey(self._server, **kwargs)
+        new_response = new_activation_key.search({'name', 'organization'})
+
+        if len(response) == 1 and len(new_response) == 0:
+            if not self._module.check_mode:
+                new_activation_key = response[0].copy(data={'new_name': new_name})
```
sean797: I think adding an `elif state == 'copied':` to https://github.com/theforeman/foreman-ansible-modules/blob/2dedb388bb94bd17afd80325c365e326fda254f3/module_utils/ansible_nailgun_cement.py#L237 and using that would be useful for content_views as well.

modules/katello_activation_key.py
```diff
@@ -210,11 +239,13 @@ def main():
...
             name=dict(required=True),
+            new_name=dict(required=False),
             organization=dict(required=True),
             lifecycle_environment=dict(),
             content_view=dict(),
             subscriptions=dict(type='list'),
             auto_attach=dict(type='bool', default=True),
+            state=dict(default='present', choices=['present', 'copied']),
         ),
         supports_check_mode=True,
```
sean797: Adding `required_if` would be useful for new_name & state == copied

evgeni:
@sean797 I agree that this module needs some `cement` love, but I'd prefer not to mix it into this PR.

modules/katello_activation_key.py
```diff
@@ -210,13 +239,18 @@ def main():
             password=dict(required=True, no_log=True),
             verify_ssl=dict(type='bool', default=True),
             name=dict(required=True),
+            new_name=dict(required=False),
```
sean797: nitpick, we don't normally include `required=False` because it's the default.

evgeni:
@sean797 added some simple test, which will at the moment fail, as the code requires git master of nailgun

sean797:
@evgeni nice, now https://github.com/theforeman/foreman-ansible-modules/pull/159 is merged please rebase and :green_apple:

evgeni:
:green_heart:

test/inventory/hosts
```diff
@@ -1,3 +1,3 @@
-localhost ansible_connection=local ansible_python_interpreter=../ansible_fail.py
+#localhost ansible_connection=local ansible_python_interpreter=../ansible_fail.py
```
evgeni: damn, that was not supposed to be here

PR #159, opened 2018-08-03
make tests use nailgun from source

sean797:
Without this tests are using the released version of nailgun on pypi


sean797:
Merging so we can get #157 passing.

PR #160, opened 2018-08-03
Add label to repositories

manuelbonk:
fixes https://github.com/theforeman/foreman-ansible-modules/issues/151

### DIFF
--- a/test/test_playbooks/tasks/repository.yml
+++ b/test/test_playbooks/tasks/repository.yml
@@ -5,6 +5,7 @@
     - product_name: "Test Product"
     - repository_name: "Test Repository"
     - repository_content_type: "yum"
+    - repository_label: "just_a_test_repo"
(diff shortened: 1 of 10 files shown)


test/test_playbooks/tasks/repository.yml
```diff
@@ -5,6 +5,7 @@
     - product_name: "Test Product"
     - repository_name: "Test Repository"
     - repository_content_type: "yum"
+    - repository_label: "just a test repo"
```
sean797: I don't think labels can have spaces.

manuelbonk: fixed in https://github.com/theforeman/foreman-ansible-modules/pull/160/commits/7f50f090d0766d842a4366303dc906d96694ad52

manuelbonk:
should be good to go now

PR #161, opened 2018-08-03
Add installation medium module and test cases

manuelbonk:


mdellweg:
Should it not be 'installation_medium'?
Other than that, i like where this is going.

manuelbonk:
you're absolutely right.

module_utils/ansible_nailgun_cement.py
```diff
@@ -426,6 +432,10 @@ def find_template_input(module, name, template, failsafe=True):
     return handle_find_response(module, response, message="No template input found for <%s> in job template" % name, failsafe=failsafe)
 
 
+def find_operating_systems(module, title):
```
mdellweg: This should really be `titles` (https://en.wiktionary.org/wiki/title).
Also `find_operating_systems_by_title` could be instructive, because all other 'find_' routines look for names.

modules/foreman_installation_medium.py
```diff
@@ -0,0 +1,194 @@
...
+from ansible.module_utils.basic import AnsibleModule
+
+
+# This is the only true source for names (and conversions thereof)
+name_map = {
+    'locations': 'location',
+    'name': 'name',
+    'organizations': 'organization',
+    'operatingsystems': 'operatingsystem',
+    'oses': 'os_family',  # the foreman community templates are using oses instead of os_family (which is wrong?)
```
mdellweg: this line looks out of place

modules/foreman_installation_medium.py
```diff
@@ -0,0 +1,194 @@
...
+            server_url=dict(required=True),
+            username=dict(required=True),
+            password=dict(required=True, no_log=True),
+            verify_ssl=dict(type='bool', default=True),
+
+            name=dict(required=True),
+            locations=dict(type='list'),
+            organizations=dict(type='list'),
+            operatingsystems=dict(type='list'),
+            os_family=dict(),
```
mdellweg: This is missing in the Documentation.

modules/foreman_installation_medium.py
```diff
@@ -0,0 +1,194 @@
...
+    state: present
+'''
+
+RETURN = ''' # '''
+
+try:
+    from ansible.module_utils.ansible_nailgun_cement import (
+        create_server,
+        ping_server,
+        find_domain,
```
sean797: not required.

modules/foreman_installation_medium.py
```diff
@@ -0,0 +1,194 @@
...
+  organizations:
+    description: List of organizations the installation medium should be assigned to
+    required: false
+    default: None
+    type: list
+  path:
+    description: Path to the installation medium
+    required: false
+    default: None
+    type: string
```
sean797: `type: string` is the default and therefore no required

modules/foreman_installation_medium.py
```diff
@@ -0,0 +1,194 @@
...
+    type: list
+  organizations:
+    description: List of organizations the installation medium should be assigned to
+    required: false
+    default: None
+    type: list
+  path:
+    description: Path to the installation medium
+    required: false
+    default: None
```
sean797: Also not required

modules/foreman_installation_medium.py
```diff
@@ -0,0 +1,194 @@
...
+    type: list
+  organizations:
+    description: List of organizations the installation medium should be assigned to
+    required: false
+    default: None
+    type: list
+  path:
+    description: Path to the installation medium
+    required: false
```
sean797: Also not required

modules/foreman_installation_medium.py
sean797: Is this actually required? `foreman_provisioning_template` doesn't seem to need it.

modules/foreman_installation_medium.py
```diff
@@ -0,0 +1,194 @@
...
+    try:
+        create_server(server_url, (username, password), verify_ssl)
+    except Exception as e:
+        module.fail_json(msg="Failed to connect to Foreman server: %s " % e)
+
+    ping_server(module)
+    try:
+        # Try to find the installation to work on
+        entity = find_installation_medium(module, name=medium_dict['name'], failsafe=True)
+    except Exception as e:
```
sean797: No need to catch an exception `failsafe` above controls that for you.

modules/foreman_installation_medium.py
```diff
@@ -0,0 +1,198 @@
...
+options:
+  name:
+    description: The full installation medium name
+    required: true
+  locations:
+    description: List of locations the installation medium should be assigned to
+    required: false
+    type: list
+  operatingsystems:
+    description: List of operating systems the installation medium should be assigned to
```
sean797: Can you make it clear it should be the title? People could think name might work.

mdellweg: If it were unique, the name would actually work, as it is part of the title.

modules/foreman_installation_medium.py
manuelbonk: I'll try to derive `os_family` from `operatingsystems` though.

manuelbonk:
@sean797 please take a look. I've derived the `os_family` from `operatingsystems` if:
 - `os_family` isn't already set,
 - only one `operatingsystems` is set, and
 - the installation medium doesn't already exist

manuelbonk:
would you mind taking a look? @akofink @ehelms @ekohl

sean797:
We already have 2 approvals, let's merge.

Thanks @manuelbonk :+1:

akofink:
Yeah, 2 is enough! :)

PR #162, opened 2018-08-06
migrate katello_activation_key to cement

evgeni:


modules/katello_activation_key.py
```diff
@@ -256,39 +157,54 @@ def main():
...
+
     try:
-        if state == 'copied':
-            changed = ng.activation_key_copy(name, organization, new_name)
-        else:
-            changed = ng.activation_key(name, organization, lifecycle_environment=lifecycle_environment, content_view=content_view,
-                                        subscriptions=subscriptions, auto_attach=auto_attach)
+        changed, activation_key_entity = naildown_entity(ActivationKey, activation_key_dict, activation_key_entity, state, module)
+
+        if state != 'copied':
```
evgeni: Otherwise it would require us to pass all subs to the `copied` invocation, making the copy less useful.

mdellweg: Sounds reasonable, but i would not understand from reading the source alone. Can you please elaborate on this in a comment.

evgeni: Done, also changed to `state == present`, as otherwise it'd try to update subs of `state == absent` keys ;)

module_utils/ansible_nailgun_cement.py
```diff
@@ -333,6 +349,11 @@ def delete_entity(entity, module):
     return True, None
 
 
+def find_activation_key(module, name, organization, failsafe=False):
+    activation_key = ActivationKey(name=name, organization=organization)
+    return handle_find_response(module, activation_key.search(), message="No content view found for %s" % name, failsafe=failsafe)
```
sean797: The message should say activation key ;)

modules/katello_activation_key.py
```diff
@@ -256,39 +158,57 @@ def main():
...
-            changed = ng.activation_key(name, organization, lifecycle_environment=lifecycle_environment, content_view=content_view,
-                                        subscriptions=subscriptions, auto_attach=auto_attach)
+        changed, activation_key_entity = naildown_entity(ActivationKey, activation_key_dict, activation_key_entity, state, module)
+
+        # only update subscriptions of newly created or updated AKs
+        # copied keys inherit the subscriptions of the origin, so one would not have to specify them again
+        # deleted keys don't need subscriptions anymore either
+        if state == 'present':
+            if entity_dict['subscriptions'] is None:
+                subscriptions = []
```
sean797: Does setting a default with Ansible work instead?

```
subscriptions=dict(type='list', default=[]),

File: modules/katello_activation_key.py [hunk 4]
```diff
@@ -98,137 +98,39 @@
 RETURN = '''# '''
 
 try:
-    from nailgun import entities
-    from nailgun.config import ServerConfig
-    from ansible.module_utils.ansible_nailgun_cement import fields_equal
+    from ansible.module_utils.ansible_nailgun_cement import (
+        create_server,
+        ping_server,
+        fields_equal,
```
sean797: I don't think this is used

modules/katello_activation_key.py
```diff
@@ -256,39 +158,57 @@ def main():
...
-        org = entities.Organization(server)
-        org.search()
+        create_server(server_url, (username, password), verify_ssl)
     except Exception as e:
         module.fail_json(msg="Failed to connect to Foreman server: %s " % e)
 
+    ping_server(module)
+
+    entity_dict['organization'] = find_organization(module, name=entity_dict['organization'])
+    if entity_dict['lifecycle_environment']:
```
akofink: So, LE/CV aren't required, but if you don't provide them, this check will fail with a `KeyError`. I think you want to check the key exists first (this isn't Ruby).

```py
if 'lifecycle_environment' in entity_dict and entity_dict['lifecycle_environment']:
  ...
```

Or, another solution, set these keys to be None by default.

modules/katello_activation_key.py
evgeni: correct, that was a stray copy&paste from the old code, where it was not used

modules/katello_activation_key.py
```diff
@@ -256,39 +157,57 @@ def main():
...
-        else:
-            changed = ng.activation_key(name, organization, lifecycle_environment=lifecycle_environment, content_view=content_view,
-                                        subscriptions=subscriptions, auto_attach=auto_attach)
+        changed, activation_key_entity = naildown_entity(ActivationKey, activation_key_dict, activation_key_entity, state, module)
+
+        # only update subscriptions of newly created or updated AKs
+        # copied keys inherit the subscriptions of the origin, so one would not have to specify them again
+        # deleted keys don't need subscriptions anymore either
+        if state == 'present':
+            if entity_dict['subscriptions'] is None:
```
akofink: Same thing here. This check will not work.

evgeni:
added `with_defaults`, added moar tests.

akofink:
state=copied *is* actually idempotent. I tested it myself :)

test/test_playbooks/activation_key.yml
```diff
@@ -32,9 +32,22 @@
...
       expected_change: false
+  - include: tasks/activation_key.yml
+    vars:
+      activation_key_auto_attach: False
+      activation_key_state: present_with_defaults
+      expected_change: false
   - include: tasks/activation_key_copy.yml
     vars:
       expected_change: true
+  - include: tasks/activation_key.yml
```
sean797: can you add a test for that please?

```yaml
 - include: tasks/activation_key_copy.yml
    vars:
      expected_change: false

evgeni:
So I actually broke it with the migration, but good news, there is a test now for that :)

Thanks @sean797 for spotting it.

sean797:
thanks @evgeni :+1:

PR #163, opened 2018-08-07
Add source url param to Katello module

chris1984:


chris1984:
@sean797 where else an I supposed to add it?

evgeni:
I think this is missing to pass the new param to the Katello API.

chris1984:
Closing as this is not needed.

chris1984:
This is pending a merge of this:

https://github.com/Katello/katello/pull/7594

Once that gets added I will update this PR.

Issue #167, opened 2018-08-10
k_activation_key should be able to set repository sets of the AK


evgeni:
##### SUMMARY
In Katello, you can enable/disable repository sets for an AK (this is called [content override](https://theforeman.org/plugins/katello/3.7/api/apidoc/v2/activation_keys/content_override.html)), but our module does not support that yet.

##### ISSUE TYPE
 - Feature Idea

PR #168, opened 2018-08-10
implement content overrides for k_activation_key

evgeni:
Fixes: #167


evgeni:
rebased, re-recorded and fixed the same issue as in #169

PR #169, opened 2018-08-18
don't overwrite subscriptions if none were given in the playbook

evgeni:
if the user did not pass `subscriptions` in the playbook, we should not
remove the existing ones from the AK


evgeni:
the diff has `+1337`, I guess that's an auto-ACK? ;-)

modules/katello_activation_key.py
```diff
@@ -194,22 +194,20 @@ def main():
         # copied keys inherit the subscriptions of the origin, so one would not have to specify them again
         # deleted keys don't need subscriptions anymore either
         if state == 'present':
-            if 'subscriptions' not in entity_dict:
-                subscriptions = []
```
sean797: Surely this means I can't remove all subscription now?

evgeni: pass `subscriptions: []` in your playbook

PR #171, opened 2018-08-20
Make katello_upload idempotent with RPMs

sean797:
This commit also include tests and refactoring to be more reliant on cement.


akofink:
It works as expected with this PR and NailGun 0.31.0 (HEAD)

sean797:
> Does it work for you (not idempotently, of course) prior to this PR?

I never tried it TBH. Anyway, I've added checked the RPM into git.

modules/katello_upload.py
```diff
@@ -162,36 +132,41 @@ def main():
...
         module.fail_json(msg="Failed to connect to Foreman server: %s " % e)
 
-    try:
-        if not module.check_mode:
-            ng.upload(src, repository, product, organization)
-    except Exception as e:
-        module.fail_json(msg=to_native(e))
+    ping_server(module)
+
+    name, version, release, arch = check_output("rpm --queryformat '%%{NAME} %%{VERSION} %%{RELEASE} %%{ARCH}' -qp %s" % entity_dict['src'],
```
ehelms: @sean797 FYI -- This breaks all non-RPM uploads (e.g. files, puppet modules). I think @pcreech is going to submit a fix proposal if you could review when he does.

sean797: Yes - sure!
Thanks

PR #174, opened 2018-08-21
Add foreman_search_facts module

sean797:
This module allows searching any Entity as long as it has `EntitySearchMixin` within Nailgun

### DIFF
--- a/module_utils/ansible_nailgun_cement.py
+++ b/module_utils/ansible_nailgun_cement.py
@@ -266,6 +266,16 @@ def naildown_entity(entity_class, entity_dict, entity, state, module, check_miss
     return changed, changed_entity
 
 
+def search_entities_json(entity_class, search):
+    """ Find entities using search """
+    return entity_class().search_json(
+        query={'search': '{0}'.format(search),
+               # This is a hack to work around pagination in API-V2
+               # See Redmine: #21800
+               'per_page': 2 << 31}
+    )
+
+
 def find_entities(entity_class, **kwargs):
     """ Find entities by certain criteria """
     return entity_class().search(
--- a/modules/foreman_search_facts.py
+++ b/modules/foreman_search_facts.py
@@ -1,6 +1,6 @@
 #!/usr/bin/python
 # -*- coding: utf-8 -*-
-# (c) 2018 Matthias M Dellweg (ATIX AG)
+# (c) 2018, Sean O'Keeffe <seanokeeffe797@gmail.com>
 #
 # This file is part of Ansible
 #
@@ -18,24 +18,20 @@
 # along with Ansible.  If not, see <http://www.gnu.org/licenses/>.
 
 ANSIBLE_METADATA = {'metadata_version': '1.0',
-                    'status': ['deprecated'],
+                    'status': ['preview'],
                     'supported_by': 'community'}
 
 
 DOCUMENTATION = '''
 ---
-module: foreman_setting_facts
-deprecated:
-  removed_in: "2.8"
-  why: This has been replaced with another module.
-  alternative: Use M(foreman_fact) instead.
-short_description: Gather facts about Foreman Settings
+module: foreman_search_facts
+short_description: Gather facts about Foreman resources
 description:
-  - "Gather facts about Foreman Settings"
+  - "Gather facts about Foreman resources"
 author:
-  - "Matthias M Dellweg (@mdellweg) ATIX AG"
+  - "Sean O'Keeffe (@sean797)"
 requirements:
-  - nailgun >= 0.30.2
+  - nailgun
 options:
   server_url:
     description:
@@ -52,53 +48,58 @@ options:
   verify_ssl:
     description:
       - Verify SSL of the Foreman server
-    required: false
     default: true
     type: bool
-  name:
+  resource:
     description:
-      - Name of the Setting to fetch
-      - If not given, fetch all settings
-    required: false
+      - Resource to search
+      - Set to an invalid choice like C(foo) see all available options.
+  search:
+    description:
+      - Search query to use
+      - If None, all resources are returned
 '''
 
 EXAMPLES = '''
 - name: "Read a Setting"
-  foreman_setting_facts:
+  foreman_search_facts:
     username: "admin"
     password: "changeme"
     server_url: "https://foreman.example.com"
-    name: "http_proxy"
+    resource: Setting
+    search: name = http_proxy
   register: result
 - debug:
-    var: result.settings[0].value
+    var: result.response[0].value
 
-- name: "Read all Settings"
-  foreman_setting_facts:
+- name: "Read all Registries"
+  foreman_search_facts:
     username: "admin"
     password: "changeme"
     server_url: "https://foreman.example.com"
+    resource: Registry
   register: result
 - debug:
     var: item.name
-  with_items: result.settings
+  with_items: result.response
 '''
 
 RETURN = '''
-settings:
-  description: List of settings
+resources:
+  description: Search results from Foreman
+  returned: always
+  type: list
 '''
 
 try:
     from ansible.module_utils.ansible_nailgun_cement import (
         create_server,
         ping_server,
-        find_entities,
+        search_entities_json,
     )
 
-    from nailgun.entities import (
-        Setting,
-    )
+    from nailgun import entities
+    from nailgun.entity_mixins import EntitySearchMixin
 
     has_import_error = False
 except ImportError as e:
@@ -109,25 +110,20 @@ except ImportError as e:
 from ansible.module_utils.basic import AnsibleModule
 
 
-name_map = {
-    'name': 'name',
-    'description': 'description',
-    'value': 'value',
-    'default': 'default',
-    'type': 'settings_type',
-    'created_at': 'created_at',
-    'updated_at': 'updated_at',
-}
+def nailgun_entites():
+    return list(map(lambda entity: entity.__name__, EntitySearchMixin.__subclasses__()))
 
 
 def main():
+
     module = AnsibleModule(
         argument_spec=dict(
             server_url=dict(required=True),
             username=dict(required=True),
             password=dict(required=True, no_log=True),
             verify_ssl=dict(type='bool', default=True),
-            name=dict(),
+            resource=dict(choices=nailgun_entites(), required=True),
+            search=dict(),
         ),
         supports_check_mode=True,
     )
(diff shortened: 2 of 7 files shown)


sean797:
I'm thinking this is an alternative to creating a `*_facts` module for each entity, which IMO would be much more effort than it's worth.

modules/foreman_facts.py
```diff
@@ -0,0 +1,157 @@
...
+    has_import_error = True
+    import_error_msg = str(e)
+    raise
+
+from ansible.module_utils.basic import AnsibleModule
+
+
+def nailgun_entites():
+    classes_to_strip = ['APIResponseError', 'Entity', 'EntityCreateMixin', 'EntityDeleteMixin', 'EntityReadMixin', 'EntitySearchMixin', 'EntityUpdateMixin']
+    camal_case_classes = filter(lambda entity: entity != entity.lower() and entity != entity.upper() and "_" not in entity, dir(entities))
```
mdellweg: You meant `cam*e*l_case_classes`?!?

modules/foreman_facts.py
```diff
@@ -0,0 +1,157 @@
...
+    required: true
+  verify_ssl:
+    description:
+      - Verify SSL of the Foreman server
+    default: true
+    type: bool
+  entity:
+    description:
+      - Entity to search
+      - Run with paramater set to C(foo) to see all available options.
```
mdellweg: Typo...

modules/foreman_facts.py
```diff
@@ -0,0 +1,157 @@
...
+except ImportError as e:
+    has_import_error = True
+    import_error_msg = str(e)
+    raise
+
+from ansible.module_utils.basic import AnsibleModule
+
+
+def nailgun_entites():
+    classes_to_strip = ['APIResponseError', 'Entity', 'EntityCreateMixin', 'EntityDeleteMixin', 'EntityReadMixin', 'EntitySearchMixin', 'EntityUpdateMixin']
```
mdellweg: This looks like a blacklist. Wouldn't it be better to filter classes by inheritance of `EntitySearchMixin` and `EntityReadMixin`. I think, that would give you exactly what we need here.

modules/foreman_facts.py
```diff
@@ -0,0 +1,157 @@
...
+            search=dict(),
+        ),
+        supports_check_mode=True,
+    )
+
+    if has_import_error:
+        module.fail_json(msg=import_error_msg)
+
+    params_dict = dict(
+        [(k, v) for (k, v) in module.params.items()])
```
mdellweg: Is this deep copy (without filtering) necessary/beneficial?

modules/foreman_facts.py
```diff
@@ -0,0 +1,157 @@
...
+    entity = Registry
+  register: result
+- debug:
+    var: item.name
+  with_items: result.response
+'''
+
+RETURN = '''
+response:
+  description: Data returned from Foreman
```
mdellweg: Is this always an array of entities? If yes, we could name it `entities` and document it.

module_utils/ansible_nailgun_cement.py
```diff
@@ -266,6 +266,16 @@ def naildown_entity(entity_class, entity_dict, entity, state, module, check_miss
...
 
 
+def search_entities_json(entity_class, search):
+    """ Find entities using search """
+    return entity_class().search_json(
+        query={'search': search,
+               # This is a hack to work around pagination in API-V2
+               # See Redmine: #21800
+               'per_page': 2 << 31}
+    )
```
mdellweg: Due to the structure of the current foreman API, this will not always return all fields of the entities. A subsequent `read` is needed in some cases.

sean797: Initially, I just want to return data from the Foreman API. If we want to return nested objects we can do that in another PR, I'm especially reluctant as it could be very slow, especially for a large number of hosts.

sean797:
updated, thanks @mdellweg

module_utils/ansible_nailgun_cement.py
mdellweg: This is not about nested entities, but about fields that are in the detailed json view (`show.json.rabl`), but not the json overview (`base.json.rabl`).
Take for example products.

mdellweg: sounds good there...

modules/foreman_facts.py
```diff
@@ -0,0 +1,155 @@
...
+    required: true
+  verify_ssl:
+    description:
+      - Verify SSL of the Foreman server
+    default: true
+    type: bool
+  entity:
+    description:
+      - Entity to search
+      - Run with parameter set to C(foo) to see all available options.
```
mdellweg: I am not sure, i like this approach. It would feel more natural to me if a missing entity option lead to the list, and an invalid entity rose an error. Writing stable playbooks would be easier.

sean797: The error is raised by Ansible based on 

```python

choices=nailgun_entites(), required=True

```

If it's missing you get `"msg": "missing required arguments: entity"` if you do, but its not valid you get `msg": "value of entity must be one of: ActivationKey, ..... `

I don't think its worth us maintaining our own error handling code for this. If you think a missing option error should include a list or possible options, then it may be worth raise that directly with the Ansible community.

mdellweg: So you are saying, an inexistent entity raises an error with a suitable message. I still find it a little bit strange to advertise raising an error without saying so in the docs part.
Do you mind writing something like 'For an invalid entity like foo the error message contains a list if valid ones.'?

module_utils/ansible_nailgun_cement.py
sean797: I choose not to do this for now, I think we're better off always returning the same datatype (a list). I have renamed the module to `foreman_search_facts` to better reflect what it is doing. I think creating a `foreman_read_facts` modules or something alike would be a better way to solve that use case. Are you okay with that?

modules/foreman_facts.py
sean797: Good idea, updated! Thanks :+1:

module_utils/ansible_nailgun_cement.py
mdellweg: Would the read facts also do the search? Then we could just add a boolean `read_full` here. Would be hard to document properly either way.
Anyway, we can do that later.

sean797: > Would the read facts also do the search?

No - I imagine it would take an ID or name (something unique about that entity). If required users can use this module to find the ID and `foreman_read_facts` to get more info.

mdellweg: Sounds good.

modules/foreman_search_facts.py
```diff
@@ -0,0 +1,155 @@
...
+# along with Ansible.  If not, see <http://www.gnu.org/licenses/>.
+
+ANSIBLE_METADATA = {'metadata_version': '1.0',
+                    'status': ['preview'],
+                    'supported_by': 'community'}
+
+
+DOCUMENTATION = '''
+---
+module: foreman_search_facts
```
akofink: I don't like the word `facts` here, because it makes me think of puppet. What about just `foreman_search`?

modules/foreman_search_facts.py
```diff
@@ -0,0 +1,155 @@
...
+short_description: Gather facts about Foreman Entities
+description:
+  - "Gather facts about Foreman Entites"
```
akofink: 'Entities' is verbiage from Nailgun. In Foreman I think we'd just call them foreman resources.

mdellweg: Most of the time i think about 'unprocessible entity'. ;)
Ressource sounds good to me.

modules/foreman_search_facts.py
```diff
@@ -0,0 +1,155 @@
...
+  password:
+    description:
+      - Password for user accessing Foreman server
+    required: true
+  verify_ssl:
+    description:
+      - Verify SSL of the Foreman server
+    default: true
+    type: bool
+  entity:
```
sean797: What to change this to `resource` as well?

modules/foreman_search_facts.py
akofink: I'm fine with 'entity' if both of you prefer it. :)

sean797: No - think resource is clearer :+1:

modules/foreman_search_facts.py
mdellweg: I guess, i was the one to introduce the name *_facts. I was inspired by the vmware_*_facts modules. I do not know, whether it is the best way. I even thought about not making a separate module at all, just returning the entity if there was nothing specified to change it.

modules/foreman_search_facts.py
akofink: Yes, if we're going with resource

modules/foreman_search_facts.py
akofink: We already have `foreman_setting_facts`. I guess one day we might have `foreman_puppet_facts`. All that makes sense to me, so I withdraw my previous comment. :fax:

sean797:
Thanks @akofink @mdellweg updated - I think this is ready? :+1:

modules/foreman_search_facts.py
```diff
@@ -0,0 +1,155 @@
...
+      - If None, all resources are returned
+'''
+
+EXAMPLES = '''
+- name: "Read a Setting"
+  foreman_search_facts:
+    username: "admin"
+    password: "changeme"
+    server_url: "https://foreman.example.com"
+    resource = Setting
```
akofink:
```py
resource: Setting
```

test/test_playbooks/search_facts.yml
```diff
@@ -0,0 +1,29 @@
...
+- hosts: tests
+  gather_facts: false
+  tasks:
+  - name: 'Load server config'
+    include_vars:
+      file: server_vars.yml
+  - include: tasks/search_facts.yml
+    vars:
+      resource: Setting
+      search: 'name = http_proxy'
```
akofink: unnecessary quotes

test/test_playbooks/search_facts.yml
```diff
@@ -0,0 +1,29 @@
+---
+- hosts: tests
+  gather_facts: false
+  tasks:
+  - name: 'Load server config'
```
akofink: unnecessary quotes

test/test_playbooks/search_facts.yml
```diff
@@ -0,0 +1,29 @@
...
+  - include: tasks/search_facts.yml
+    vars:
+      resource: Setting
+      search: 'name = http_proxy'
+      return_length: 1
+  - include: tasks/search_facts.yml
+    vars:
+      resource: Subnet
+      return_length: 0
+  - name: "Run with invalid resource"
```
akofink: unnecessary quotes

modules/foreman_search_facts.py
```diff
@@ -0,0 +1,155 @@
...
+  register: result
+- debug:
+    var: result.response[0].value
+
+- name: "Read all Registries"
+  foreman_search_facts:
+    username: "admin"
+    password: "changeme"
+    server_url: "https://foreman.example.com"
+    resource = Registry
```
akofink:
```
resource: Registery
```

sean797:
Thanks, @akofink updated.

PR #175, opened 2018-08-28
Fix Job Template module

mdellweg:
'template_input_list' should not be undefined.


mdellweg:
See also: https://github.com/theforeman/foreman_remote_execution/pull/373

sean797:
ping @manuelbonk, I think your Nailgun PR will require the same changes?

manuelbonk:
@sean797 thanks, I renamed the parameter there too.

sean797:
Thanks @mdellweg :+1:

PR #177, opened 2018-08-30
attempt at adding composite content views functionality

itewk:
This does not work yet, I am having trouble figuring out how to get a ContentViewVersion for version 'Latest' for an unpublished ContentView. But I thought maybe I could get some review on what I have so far and maybe someone had an idea to go the last step.

https://github.com/SatelliteQE/nailgun/issues/534

### DIFF
--- a/module_utils/ansible_nailgun_cement.py
+++ b/module_utils/ansible_nailgun_cement.py
@@ -389,6 +389,10 @@ def find_content_view(module, name, organization, failsafe=False):
     return handle_find_response(module, content_view.search(), message="No content view found for %s" % name, failsafe=failsafe)
 
 
+def find_content_views(module, content_views, organization, failsafe=False):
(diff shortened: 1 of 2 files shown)


sean797:
Thanks @itewk so far the code looks good :+1: 

With the question around lastest, I'd suggest adding a `create_payload` & `update_payload` function to the entity and adding a new field like `lastest_component': entity_fields.OneToManyField(ContentView)`.  Inside `create_payload` & `update_payload` you could then ensure `lastest_component` isn't sent to the server but the correct data is. But really it's up to the Nailgun guys with how they want to support this.

sean797:
@itewk based on this I've created https://github.com/theforeman/foreman-ansible-modules/pull/204, hope you don't mind (I needed composite CV support for my own reasons). Your review and/or testing would be appreciated.

PR #178, opened 2018-08-30
Allow connection parameter in dict

mdellweg:
Affected modules:
  * settings

Another idea from the experimental lab.
This should help passing the authentication data across files.

### DIFF
--- a/module_utils/ansible_nailgun_cement.py
+++ b/module_utils/ansible_nailgun_cement.py
@@ -203,6 +203,13 @@ def create_server(server_url, auth, verify_ssl):
     )
 
 
+def create_connection(auth_block, module):
+    try:
+        create_server(auth_block['server_url'], (auth_block['username'], auth_block['password']), auth_block['verify_ssl'])
+    except Exception as e:
+        module.fail_json(msg="Failed to connect to Foreman server: %s " % e)
+
+
 # Prerequisite: create_server
 def ping_server(module):
     try:
--- a/modules/foreman_setting.py
+++ b/modules/foreman_setting.py
@@ -38,21 +38,26 @@ options:
   server_url:
     description:
       - URL of Foreman server
-    required: true
+    required: false
   username:
     description:
       - Username on Foreman server
-    required: true
+    required: false
   password:
     description:
       - Password for user accessing Foreman server
-    required: true
+    required: false
   verify_ssl:
     description:
       - Verify SSL of the Foreman server
     required: false
     default: true
     type: bool
+  auth_block:
+    description:
+      - Dictionary containing the connection parameter
+    required: false
+    type: dict
   name:
     description:
       - Name of the Setting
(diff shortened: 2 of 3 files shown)


sean797:
Interesting - To be honest, I'm not sure I like supporting 2 params for the same purpose; I think I would prefer environment vars to provide authentication options in a more scalable way

mdellweg:
My idea is, that we could use the auth_block to have either _server:user:pass_ or _server:apitoken_.
But as the next level we could make the modules being local actions, acting on the `ansible_host` and read the credentials from facts (provided by the inventory / groupvars).

sean797:
> My idea is, that we could use the auth_block to have either server:user:pass or server:apitoken.

@mdellweg AFAIK Foreman doesn't support `apitoken` auth.

> But as the next level we could make the modules being local actions, acting on the ansible_host and read the credentials from facts (provided by the inventory / groupvars).

Not sure I understand what you are saying, could you be more explicit?

mdellweg:
I just figured out that you can create personal tokens via api calls. But those are actually used just like passwords. So we could at least start with an alias from token to password.

Right now, we have up to three hosts involved in every call to a module. The one, where you call ansible, the one in the inventory, where the module is executed and the foreman server. I think, it would be more natural if the host running ansible had the foreman server in its inventory (with credentials). Then you would write a play for this host, and the modules would run as local actions (like debug with access to all variables) connecting directly to api/v2.

sean797:
Having 3 hosts just makes Ansible more flexible, you could argue if you wanted to change that it should change for every Ansible module that interacts with a servers HTTP API. 

I'd imagine most users would set at least 2 of them to be the same actual host, maybe we should default server_url to `https://{{ ansible_fqdn }}`?

If used in conjunction with other tasks in a playbook you may want to set `run_once` and/or `delegate_to`. for example 

```yaml
- hosts: all
  tasks:
    - name: promote CV
      katello_content_view_version_promote:
        name: "Fedora CV"
        from_environment: "Testing"
        to_environment: "Production"
      run_once: true
      delegate_to: "{{ nailgun_host }}"
    - name: update systems
      serial: 1
      block:
        - name: upgrade all packages
          yum:
          name: '*'
          state: latest
       - name: Restart server
         command: /sbin/shutdown -r +1
         async: 0
         poll: 0
         ignore_errors: true

       - name: Wait for the server to finish rebooting before doing the next machine
         become: false
         local_action:
           wait_for host: "{{ inventory_hostname }}"
           search_regex: OpenSSH
           port: 22
           timeout: 300
         connection: local
```

evgeni:
I must admit I mostly run my stuff against `hosts: localhost`, so it only makes 2 hosts in my setup. (I would also not install `nailgun` on a Foreman/Katello today, as `pip install nailgun` will also pull in a new `requests` and that might have funny results in Pulp.)

Either way :-1: from me, don't see much benefit of having this (today).

mdellweg:
It was an idea, and now it is quite old. Closing it.

Issue #181, opened 2018-09-14
job_template overrides provider_type from template meta-data


m-bucher:
##### SUMMARY
When adding a new Ansible Job-template, without explicitly specifiying 'provider_type', the provider_type will always be set to 'SSH'

##### ISSUE TYPE
 - Bug Report

##### ANSIBLE VERSION

```
n.a.
```

##### KATELLO/FOREMAN VERSION

```
n.a.
```

##### NAILGUN VERSION

```
n.a.
```

##### STEPS TO REPRODUCE

```yaml
---
- name: "Add Ansible job template"
  foreman_job_template:
    username: "admin"
    password: "secret"
    server_url: "https://foreman.example.com"
    verify_ssl: "no"
    locations: "local"
    organizations: "ACME"
    state: "present"
    template: |
        <%#
        name: Ansible Roles - Ansible Default
        job_category: Ansible Playbook
        description_format: Run Ansible roles
        feature: ansible_run_host
        snippet: false
        provider_type: Ansible
        kind: job_template
        model: JobTemplate
        %>
        ---
        - hosts: all
          tasks:
            - name: Display all parameters known for the Foreman host
              debug:
                var: foreman_params
          roles:
        <%- if @host.all_ansible_roles.present? -%>
        <%=   @host.all_ansible_roles.map { |role| "    - #{role.name.strip}" }.join("\n") %>
        <%- end -%>
    ...
...
```

##### EXPECTED RESULTS

`provider_type` set to `Ansible`

##### ACTUAL RESULTS

`provider_type` set to `SSH`

PR #182, opened 2018-09-14
Fix job_template overwrites provider_type (#181)

m-bucher:
Apply default `provider_type` after template has been parsed


modules/foreman_job_template.py
```diff
@@ -405,7 +409,9 @@ def main():
             module.fail_json(msg="Cannot use '*' as a job template name!")
         # module params are priorized
         parsed_dict.update(entity_dict)
-        entity_dict = parsed_dict
+        # make sure certain values are set
+        entity_dict = template_defaults.copy()
```
mdellweg: Kodos for safe programming! Even if it's rather unlilkely that the template_defaults be reused.

mdellweg:
Do you think, this is testable?

m-bucher:
@mdellweg testing is tricky.
I tried adding a simple Ansible job_template once with explicit provider_type 'Ansible' and once without.
Without my above fix the test only fails when recording the tests, because then it would detect a `changed: true` for the second run because of the different provider_type.

However, in the normal case when running the test with recorded VCR-data, the test passes even without my fix, because the test does not check the request-bodies.
The request-body for adding with provider_type as module-param having `provider_type: Ansible` and the request-body without  module-param provider_type having `provider_type: SSH`.

mdellweg:
The test should most definitely be matching requests depending on the body.

mdellweg:
Ahh, #137 deactivated the body matcher. And i didn't  realize then.
@sean797 You removed it. We want it back. ;)

sean797:
I think were have re-enable it in another PR, it's going to require some tests are re-recorded.

m-bucher:
@sean797, @mdellweg I took the liberty of re-enabling it here, as well.
Getting all tests to run again took a little effort, though.
On the plus side, I found two python3 compatibility-bugs and was able to fix them

PR #184, opened 2018-09-24
added ability to use as a role for testing

ericsysmin:


meta/main.yml
```diff
@@ -0,0 +1,21 @@
+galaxy_info:
+  role_name: foreman_modules
+  author: The Foreman
+  description: Deploy Foreman Ansible Modules as a Role
+  issue_tracker_url: https://github.com/ericsysmin/ansible-role-foreman-modules/issues
```
sean797: s/ericsysmin/theforeman/

meta/main.yml
```diff
@@ -0,0 +1,21 @@
...
+    - 6
+    - 7
+  - name: Ubuntu
+    versions:
+    - precise
+    - trusty
+    - xenial
+  galaxy_tags:
+    - foreman
+dependencies: []
```
sean797: Nailgun ?

meta/main.yml
```diff
@@ -0,0 +1,21 @@
+galaxy_info:
+  role_name: foreman_modules
+  author: The Foreman
+  description: Deploy Foreman Ansible Modules as a Role
+  issue_tracker_url: https://github.com/ericsysmin/ansible-role-foreman-modules/issues
+  license: GPLv3
+  min_ansible_version: 2.0.0
```
sean797: I think it might be a later version, at least for some modules. I'd rather not have this if we can help it?

meta/main.yml
ericsysmin: @sean797 are you talking about as a tag, or dependency? In meta/main.yml dependencies are role dependencies which mean other Ansible roles are required for this one. Anything here will search for Galaxy Roles to install at the same time.

meta/main.yml
ericsysmin: Will make it a later version like the latest release

ericsysmin:
@sean797 I can add Travis-CI in another push

ehelms:
I tend to agree with @sean797 I am not sure how I feel and there a few unanswered questions:

 * How often do we push the role (given in part we don't version these)?
 * Do we do it automatically?
 * Who has admin on galaxy for the role?

jlozadad:
I think a better approach would be to each module to have its own role although that means more git repositories. Then you can use things like `requirements.yml` to pull the correct roles required. Usually those manage or have control of the git repositories manage the galaxy part.

sean797:
As https://github.com/theforeman/foreman-ansible-modules/pull/279 conflicts with this, I wanted to comment here to ensure people have seen the other PR.

sean797:
Since #279 got merged, I'm closing this.

Thanks for taking the time to do this @ericsysmin, we'll hopefully be able to ship an Ansible Collection soon.

https://galaxy.ansible.com/docs/contributing/creating_collections.html

Issue #186, opened 2018-10-04
k_activation_key not idempotent when setting content overrides


evgeni:
##### SUMMARY
When using k_a_k with `content_overrides` it seems that the module also diffs against the not overridden products, thus always pretending to have changes (and being slow)

##### ISSUE TYPE
 - Bug Report

PR #188, opened 2018-10-12
make content_overrides in k_activation_key truly idempotent

evgeni:
before we considered all results of the product content call as "current
overrides", even when they were not actually set to be overriden. this
resulted in too many "reset to default" calls (for overrides that
already were at "default") plus it made the module return changed=True
in too many cases.

let's only look at the entries that have a not None
enabled_content_override, which means they are actually overriden.

Fixes: #186

### DIFF
--- a/modules/katello_activation_key.py
+++ b/modules/katello_activation_key.py
@@ -235,8 +235,13 @@ def main():
             if 'content_overrides' in entity_dict:
                 content_overrides = entity_dict['content_overrides']
                 product_content = activation_key_entity.product_content()
-                current_content_overrides = set([(p['content']['label'], p['enabled_content_override']) for p in product_content['results']])
-                desired_content_overrides = set([(p['label'], override_to_boolnone(p['override'])) for p in content_overrides])
+                current_content_overrides = set()
+                for product in product_content['results']:
+                    if product['enabled_content_override'] is not None:
+                        current_content_overrides.add((product['content']['label'], product['enabled_content_override']))
+                desired_content_overrides = set()
+                for product in content_overrides:
+                    desired_content_overrides.add((product['label'], override_to_boolnone(product['override'])))
 
                 if desired_content_overrides != current_content_overrides:
                     if not module.check_mode:


evgeni:
errr, tests fail, but I don't get why.

mdellweg:
Looking at your change (and i am not using plural here...), this might be a problem with vcr again. Can you try pinning it to an older version to verify that?

mdellweg:
#189 
At least, i can verify, your failure is a general problem.

evgeni:
```
-ansible==2.6.5
+ansible==2.7.0
```

this is what breaks it. I am confused, a lot. But I guess I am going to pin ansible for now?

evgeni:
@mdellweg @akofink @sean797  are you OK with pinning ansible?

akofink:
IMO no, we should probably figure out why it's broken and fix it in a separate PR. Ansible 2.7.0 is the new default packaged version in F28 as of this week afaict.

evgeni:
okay, so I tried re-recording the cassettes on 2.7, and it gets funny:

```
[master 26c6054] test ak with 2.7
 10 files changed, 1138 insertions(+), 1536 deletions(-)
 copy test/test_playbooks/fixtures/{activation_key-1.yml => activation_key-0.yml} (67%)
 rewrite test/test_playbooks/fixtures/activation_key-1.yml (78%)
 rename test/test_playbooks/fixtures/{activation_key-5.yml => activation_key-2.yml} (62%)
 rename test/test_playbooks/fixtures/{activation_key-7.yml => activation_key-3.yml} (62%)
 rename test/test_playbooks/fixtures/{activation_key-9.yml => activation_key-4.yml} (62%)
 rename test/test_playbooks/fixtures/{activation_key-11.yml => activation_key-5.yml} (63%)
 rename test/test_playbooks/fixtures/{activation_key-13.yml => activation_key-6.yml} (65%)
 rename test/test_playbooks/fixtures/{activation_key-15.yml => activation_key-7.yml} (63%)
 rename test/test_playbooks/fixtures/{activation_key-17.yml => activation_key-8.yml} (63%)
 rename test/test_playbooks/fixtures/{activation_key-19.yml => activation_key-9.yml} (67%)
```

before, we had only odd numbered files, now the even ones are "back" and the numbers shifted.

*just* renaming them is not sufficient, tho :(

mdellweg:
Obviously, Ansible changed something. I think we should really look at the tapes mechanism as a black box and be happy, that it didn't break for 2.7 in general.

evgeni:
depends on #192

PR #192, opened 2018-10-17
update VCR tapes to Ansible 2.7

evgeni:


modules/redhat_manifest.py
```diff
@@ -137,7 +137,6 @@ def create_manifest(module):
     path = "/subscription/consumers"
     data = {'name': module.params['name'],
             'type': "satellite",
-            'owner': module.params['rhsm_owner'],
```
evgeni: @sean797 do you remember why you added this? I get weird errors when I leave this in, both against prod and stage.

```
fatal: [tests]: FAILED! => {"changed": false, "msg": "POST to https://subscription.rhn.stage.redhat.com/subscription/consumers failed, got Runtime Error com.fasterxml.jackson.databind.exc.MismatchedInputException: Cannot construct instance of `org.candlepin.dto.api.v1.OwnerDTO` (although at least one Creator exists): no String-argument constructor/factory method to deserialize from String value ('11612516')\n at [Source: (org.candlepin.common.filter.TeeHttpServletRequest$1); line: 1, column: 11] (through reference chain: org.candlepin.dto.api.v1.ConsumerDTO[\"owner\"]) at com.fasterxml.jackson.databind.exc.MismatchedInputException.from:63"}
```

It works just fine when I remove the `owner`, so that's what I did for now.

test/test_playbooks/upload.yml
```diff
@@ -34,10 +34,6 @@
   - name: 'Load server config'
     include_vars:
       file: server_vars.yml
-  - name: remove RPM
```
evgeni: @sean797 I removed this, as the RPM is in the repo, and removing it makes the test not re-runable w/o restoring the file.

evgeni:
@sean797 can you try re-recording the `content_view_filter` tapes? it fails for me and I have no idea :D

test/test_playbooks/tasks/repository.yml
```diff
@@ -5,7 +5,6 @@
     - product_name: "Test Product"
     - repository_name: "Test Repository"
     - repository_content_type: "yum"
-    - repository_label: "just_a_test_repo"
```
evgeni: @manuelbonk you originally introduced the `label` feature in #160, but we think the label setting should happen in the `repository.yml` playbook, not the task file, as this otherwise breaks the `default(omit)` you have below. Thoughts?

evgeni:
Greeeeeeeen :green_heart: 
Thanks @Fobhep!

Fobhep:
nice :)!

test/test_playbooks/tasks/repository.yml
mdellweg: I am completely :+1: here.

test/test_playbooks/upload.yml
mdellweg: Does this mean, we are dropping a test here? Can we readd it / Sync the repo afterwards?

evgeni: No we are not!

`upload.yml` is about uploading a file (in this case an RPM) to Katello. The RPM used for this is checked in into Git as a fixture: https://github.com/theforeman/foreman-ansible-modules/blob/master/test/test_playbooks/data/bear-4.1-1.noarch.rpm

I've just modified the playbook not to remove the fixture file, as this seems contra-productive here as it requires you to fiddle with Git before you can run `upload.yml` again.

It would be correct to remove it if we'd fetch it dynamically from a remote site before uploading, but we're not doing that.

evgeni:
@mdellweg so contrary to what I've said earlier, we cannot add `ansible>=2.7` to `requirements-dev.txt`, as `pip` then errors out with:

```
Double requirement given: ansible (from -r requirements.txt (line 2)) (already in ansible>=2.7 (from -r requirements-dev.txt (line 6)), name='ansible')
```

See https://travis-ci.org/theforeman/foreman-ansible-modules/jobs/442548681 for an example run.

We *could* bump the global requirement, but this feels odd as the modules actually work just fine with older Ansible versions.

evgeni:
@Fobhep the `content_view_filter` module requires two patches against Katello, after applying those to my 3.8 install it just worked. Seems your box was too old in the end ;)

mdellweg:
Let's not bump the global requirement then.
(I guess we could have separate requirements files for test, but we want to reuse the global reqs there.)

How about adding an assert in the test playbook? Might help to tell that the ansible version is the problem.

evgeni:
@mdellweg thanks for approving, I'd like to wait for @sean797 and @manuelbonk to ACK the changes to their stuff before merging this.

As for an assert in the playbook, I'd prefer not to, this just makes things noisy.

test/test_playbooks/tasks/repository.yml
manuelbonk: You're right, I'll move it to the `repository.yml` playbook.

manuelbonk:
awesome work :+1:

modules/redhat_manifest.py
sean797: No - but since its not in https://github.com/Katello/hammer-cli-csv/blob/51f64ea29fd86affb66db1ab3dde429097a8c8b6/lib/hammer_cli_csv/subscriptions.rb#L187 lets remove it

sean797: No - but since its not in https://github.com/Katello/hammer-cli-csv/blob/51f64ea29fd86affb66db1ab3dde429097a8c8b6/lib/hammer_cli_csv/subscriptions.rb#L187 lets remove it

PR #194, opened 2018-10-18
Define repository_label in test playbook, not in the task file

manuelbonk:
Implenting @evgeni suggestion: https://github.com/theforeman/foreman-ansible-modules/pull/192#discussion_r226016134


manuelbonk:
nvmnd, already fixed in https://github.com/theforeman/foreman-ansible-modules/pull/192

PR #198, opened 2018-10-29
Add content credentials support

bagasse:
Add content credentials support:
* Add / Modify content credential (GPG Keys and Certs)
* Ability to assign content credential to repository
* Ability to assign content credential to product


modules/katello_gpg_key.py
```diff
@@ -0,0 +1,186 @@
...
+
+EXAMPLES = '''
+- name: "Create katello client GPG key"
+  katello_gpg_key:
+    username: "admin"
+    password: "changeme"
+    server_url: "https://foreman.example.com"
+    name: "RPM-GPG-KEY-my-repo"
+    organization: "Default Organization"
+    content: "-----BEGIN PGP PUBLIC KEY BLOCK-----
```
mdellweg: I preferred, if we did not include an actual gpg-key in this repo.
And for the example, it might be as instructive and much more readable if we did something like:
`content: "{{ lookup('file', 'key.asc') }}"`
I think this applies to the tests, too. Katello does not rigogrously validate, that you insert a valid key, iirc.

modules/katello_repository.py
```diff
@@ -108,8 +112,22 @@
     content_type: "yum"
     product: "My Product"
     organization: "Default Organization"
+    url: "http://yum.theforeman.org/plugins/latest/el7/x86_64/"
+    download_policy: background
+
+- name: "Create repository"
```
mdellweg:
```suggestion
- name: "Create repository with content credentials"
```

modules/katello_product.py
```diff
@@ -83,6 +87,16 @@
     name: "Fedora"
     organization: "My Cool new Organization"
     state: present
+
+- name: "Create CentOS product"
```
mdellweg:
```suggestion
- name: "Create CentOS product with content credentials"
```

bagasse:
@sean797, I saw that nailgun is using the deprecated gpgkey API ATM, so I used gpg_key for module name but you are right, user facing module should follow naming of current API. Changed module name from `gpg_key` to `content_credential`. Of course it support only gpg keys ATM, but all is ready for supporting other types in the future.

@mdellweg, I've added some tests for products and repos.

modules/katello_content_credential.py
```diff
@@ -0,0 +1,168 @@
...
+def main():
+    module = AnsibleModule(
+        argument_spec=dict(
+            server_url=dict(required=True),
+            username=dict(required=True, no_log=True),
+            password=dict(required=True, no_log=True),
+            verify_ssl=dict(type='bool', default=True),
+            name=dict(required=True),
+            organization=dict(required=True),
+            type=dict(default='gpg_key', chhoices=['gpg_key']),
```
mdellweg:
```suggestion
            type=dict(default='gpg_key', choices=['gpg_key']),
```

bagasse: Done.

test/test_playbooks/tasks/product.yml
```diff
@@ -12,6 +13,7 @@
     verify_ssl: "{{ foreman_verify_ssl }}"
     name: "{{ product_name }}"
     organization: "{{ organization_name }}"
+    content_credential: "{{ product_content_credential }}"
```
mdellweg:
```suggestion
    content_credential: "{{ product_content_credential | omit }}"
```

test/test_playbooks/tasks/product.yml
```diff
@@ -4,6 +4,7 @@
     - organization_name: "Test Organization"
     - product_name: "Test Product"
     - product_description: "A happy little test product"
+    - product_content_credential: ""
```
mdellweg: If you do not put that here, than we can test the behaviour without that option.
`omit` is like it was not specified at all.

test/test_playbooks/tasks/repository.yml
```diff
@@ -5,6 +5,7 @@
     - product_name: "Test Product"
     - repository_name: "Test Repository"
     - repository_content_type: "yum"
+    - repository_content_credential: ""
```
mdellweg: same here

sbernhard:
Would be great if you can have a look at https://github.com/SatelliteQE/nailgun/pull/554 if this would help you to support the SSL content credentials, too.

sean797:
Thanks @sbernhard, Nailgun PR is now merged :+1: 

@bagasse mind updating this? Then I think we should be good to merge :+1:

bagasse:
@sean797, rebased and updated. Has a new nailgun version is not released yet, I didn't updated all the requirements section in modules documentation for content_credential, product and repository.

module_utils/ansible_nailgun_cement.py
```diff
@@ -488,6 +489,11 @@ def find_repository(module, name, product, failsafe=False):
     return handle_find_response(module, repository.search(), message="No Repository found for %s" % name, failsafe=failsafe)
 
 
+def find_content_credential(module, name, organization, failsafe=False):
+    content_credential = ContentCredential(name=name, organization=organization)  # Waiting for Nailgun to use new content credential API
```
sean797: I think you can remove this comment?

modules/katello_content_credential.py
```diff
@@ -0,0 +1,167 @@
...
+
+DOCUMENTATION = '''
+---
+module: katello_content_credential
+short_description: Create and Manage Katello content credentials
+description:
+  - Create and Manage Katello content credentials
+author: "Baptiste Agasse (@bagasse)"
+requirements:
+  - "nailgun >= 0.31.0"
```
sean797: Want to change this to 0.32 in anticipation?

modules/katello_content_credential.py
```diff
@@ -0,0 +1,167 @@
...
+     - Password for user accessing Foreman server
+    required: true
+  verify_ssl:
+    description:
+      - Verify SSL of the Foreman server
+    default: true
+    type: bool
+  name:
+    description:
+      - Name of the GPG key
```
sean797: Content Credential ;)

modules/katello_content_credential.py
```diff
@@ -0,0 +1,167 @@
...
+def main():
+    module = AnsibleModule(
+        argument_spec=dict(
+            server_url=dict(required=True),
+            username=dict(required=True, no_log=True),
+            password=dict(required=True, no_log=True),
+            verify_ssl=dict(type='bool', default=True),
+            name=dict(required=True),
+            organization=dict(required=True),
+            content_type=dict(choices=['gpg_key', 'cert']),
```
sean797: Docs say `required=True` ?

nailgun defaults to gpg_key, so technically it isn't required but I think I'd prefer if it was

modules/katello_product.py
```diff
@@ -155,6 +171,11 @@ def main():
     ping_server(module)
 
     entity_dict['organization'] = find_organization(module, name=entity_dict['organization'])
+
+    if 'content_credential' in entity_dict:
+        entity_dict['content_credential'] = find_content_credential(module, name=entity_dict['content_credential'], organization=entity_dict['organization'],
+                                                                    failsafe=True)
```
sean797: I think this should be `failsafe=False`? (which is the default)

modules/katello_repository.py
```diff
@@ -194,6 +215,10 @@ def main():
 
     entity_dict['product'] = find_product(module, name=entity_dict['product'], organization=entity_dict['organization'])
 
+    if 'content_credential' in entity_dict:
+        entity_dict['content_credential'] = find_content_credential(module, name=entity_dict['content_credential'], organization=entity_dict['organization'],
+                                                                    failsafe=True)
```
sean797: I think this should be `failsafe=False`? (which is the default)

modules/katello_content_credential.py
bagasse: Sorry I updated the doc part but forgot this one. For me  nailgun defaults to gpg_key to stay compatible with older versions. Here this is the first version :) . As you I prefer if it is required.

modules/katello_content_credential.py
bagasse: I've also updated this in product and repository modules as we use content_credential in it too.

README.md
```diff
@@ -102,6 +102,7 @@ This is a list of modules currently in the repository (please add to the list if
  * foreman_domain: create and maintain domains
  * foreman_job_template: create and maintain job templates and associated template inputs
  * foreman_setting: set and reset settings
+ * katello_content_credential: create and maintain content credential
```
mdellweg:
```suggestion
 * katello_content_credential: create and maintain content credentials
```

modules/katello_product.py
```diff
@@ -113,7 +128,7 @@
     'name': 'name',
     'organization': 'organization',
     'description': 'description',
-    # 'gpg_key': 'gpg_key',  # wait for Nailgun
+    'content_credential': 'gpg_key',
```
mdellweg: I have not followed the Nailgun change closely, but is this translation still valid/needed?

mdellweg: On a second thought, I think in the Product/Repository case, you can specify a _gpg key_, a _client cert_, a _server cert_ and a _ca cert_.

So i think we should still call it a `gpg_key` here and add the ssl parts with their names later.
Sorry for bringing that up so late.

mdellweg:
@bagasse Thank you!

bagasse:
@sean797, @mdellweg, thank you for your time/reviews/patience :+1:

Issue #195, opened 2018-10-25
foreman_setting is not idempotent


akofink:
##### SUMMARY

All modules should be able to be run multiple times with only one side effect. In this case, we should check the setting's value and compare it with the requested value in the playbook.

##### ISSUE TYPE
 - Bug Report

##### ANSIBLE VERSION

```
ansible 2.6.5
  config file = /home/akofink/dev/repos/katello-ansible-seeds/ansible.cfg
  configured module search path = [u'/home/akofink/dev/repos/katello-ansible-seeds/foreman-ansible-modules/modules']
  ansible python module location = /usr/lib/python2.7/site-packages/ansible
  executable location = /usr/bin/ansible
  python version = 2.7.15 (default, Sep 21 2018, 23:26:48) [GCC 8.1.1 20180712 (Red Hat 8.1.1-5)]
```

##### KATELLO/FOREMAN VERSION
devel (foreman/katello running from source)

```
$ rpm -q tfm-rubygem-katello foreman
package tfm-rubygem-katello is not installed
package foreman is not installed

$ git rev-parse HEAD # foreman
ad28e6bb7bfcbc5abf14833ec1e401cef1134cc1
$ git rev-parse HEAD # katello
ee0b953e1994a538f0ea65d7363a35560e2db205
```

##### NAILGUN VERSION

```
Version: 0.31.0
```

##### STEPS TO REPRODUCE

Run this playbook:

```yaml
  - name: Increase idle_timeout to 1 week
    foreman_setting:
      username: "{{ foreman_username }}"
      password: "{{ foreman_password }}"
      server_url: "{{ foreman_server_url }}"
      verify_ssl: "{{ foreman_verify_ssl }}"
      name: idle_timeout
      value: 10080
  - name: Increase idle_timeout to 1 week
    foreman_setting:
      username: "{{ foreman_username }}"
      password: "{{ foreman_password }}"
      server_url: "{{ foreman_server_url }}"
      verify_ssl: "{{ foreman_verify_ssl }}"
      name: idle_timeout
      value: 10080
```

##### EXPECTED RESULTS

The first time a value is changed, ansible reports `changed: true`; on subsequent runs, it reports `changed: false`.

##### ACTUAL RESULTS

The first time a value is changed, ansible reports `changed: true`; on subsequent runs, it reports `changed: true`.

```
TASK [Increase idle_timeout to 1 week] ****************************************************************************************************************************************************************************
[...]

PR #199, opened 2018-10-30
foreman_setting now supports array and int values

sean797:
Admittedly I'm not 100% on board with the way I have fixed this, but I think its the best way :/
I'd welcome some other suggestions

### DIFF
--- a/module_utils/ansible_nailgun_cement.py
+++ b/module_utils/ansible_nailgun_cement.py
@@ -43,6 +43,8 @@ from nailgun.entities import (
 )
 from nailgun import entity_mixins, entity_fields
 
+from ansible.module_utils.parsing.convert_bool import boolean
+
 
 class TemplateInput(
     Entity,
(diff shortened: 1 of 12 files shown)


sean797:
Fixes https://github.com/theforeman/foreman-ansible-modules/issues/195

sean797:
Not sure why this is failing with py3? I cannot reproduce it locally :-1:

module_utils/ansible_nailgun_cement.py
```diff
@@ -340,15 +342,26 @@ def fields_equal(value1, value2):
...
         for key, value in volatile_entity.get_values().items():
-            if key in entity_dict and not fields_equal(value, entity_dict[key]):
-                volatile_entity.__setattr__(key, entity_dict[key])
-                fields.append(key)
+            if key in entity_dict:
+                if check_type and not isinstance(value, type(entity_dict[key])):
+                    if isinstance(value, bool):
+                        entity_dict[key] = boolean(entity_dict[key])
+                    elif isinstance(value, list):
+                        entity_dict[key] = str(sorted(entity_dict[key].split(',')))
```
sean797: Unfortunately, it seems this has to be a string, otherwise I get`Value is invalid: must be an array` which is a weird error because I'm sending an array..

```
2018-11-28T15:13:12 [I|app|8a4f3] Processing by Api::V2::SettingsController#update as JSON
2018-11-28T15:13:12 [I|app|8a4f3]   Parameters: {"setting"=>{"value"=>["bar", "foo"]}, "apiv"=>"v2", "id"=>"114"}
2018-11-28T15:13:12 [I|app|8a4f3] Authorized user admin(Admin User)
2018-11-28T15:13:12 [I|app|8a4f3] Current user set to admin (admin)
2018-11-28T15:13:12 [E|app|8a4f3] Unprocessable entity Setting::Provisioning (id: 114):
  Value is invalid: must be an array

File: modules/foreman_setting.py [hunk 2]
```diff
@@ -62,6 +62,7 @@
     description:
       - value to set the Setting to
       - if missing, reset to default
+      - for `string_type` boolean, use a comma separated list
```
mdellweg: This i do not understand...

sean797:
Thanks @mdellweg. Yes - think so, updated to that makes sense now :+1:

PR #202, opened 2018-11-19
Make katello_upload idempotent with Files

sean797:
Requires https://github.com/SatelliteQE/nailgun/pull/562

### DIFF
--- a/modules/katello_upload.py
+++ b/modules/katello_upload.py
@@ -68,7 +68,7 @@ options:
       - Organization that the Product is in
     required: true
 notes:
-    - Currently only idempotent when uploading a RPM
+    - Currently only idempotent when uploading to an RPM & file repository
 '''
 
 EXAMPLES = '''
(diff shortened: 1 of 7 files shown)


modules/katello_upload.py
```diff
@@ -152,15 +155,24 @@ def main():
     entity_dict['product'] = find_product(module, name=entity_dict['product'], organization=entity_dict['organization'])
     entity_dict['repository'] = find_repository(module, name=entity_dict['repository'], product=entity_dict['product'])
 
-    package = False
+    content_unit = False
```
mdellweg: Wouldn't it be easier, more consistent to set it to `None` here?
`None` evaluates to `False` anyway. So a simple `if not content_unit:` does the job below.

sean797:
Thanks @mdellweg, updated - Nailgun PR is merged now :+1:

mdellweg:
Thanks! @sean797

PR #206, opened 2018-11-30
Test with nailgun pr

mdellweg:
Not entirely sure how to test this mechanism. ;)

Try Dependency: SatelliteQE/nailgun#203

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -8,6 +8,9 @@ MODULE_PATH=modules/$(MODULE).py
 DEBUG_DATA_PATH=test/debug_data/$(DATA).json
 DEBUG_OPTIONS=-m $(MODULE_PATH) -a @$(DEBUG_DATA_PATH) -D $(PDB_PATH)
 
+PR_MSG=$(shell http --json "https://api.github.com/repos/${TRAVIS_REPO_SLUG}/pulls/${TRAVIS_PULL_REQUEST}" | jq -r .body | tr -C "[:alnum:]" _)
+NAILGUN_PR_NUMBER=$(shell echo $(PR_MSG) | sed -n 's/.*SatelliteQE_nailgun[^0-9]*\([0-9]*\).*/\1/p')
+
 default: help
 help:
 	@echo "Please use \`make <target>' where <target> is one of:"
@@ -54,6 +57,12 @@ test-setup: test/test_playbooks/server_vars.yml
 	pip install --upgrade pip
 	pip install -r requirements-dev.txt
 	pip install -r https://raw.githubusercontent.com/ansible/ansible/devel/requirements.txt 
+ifdef TRAVIS_PULL_REQUEST
+ifneq ("$(NAILGUN_PR_NUMBER)", "")
+	echo pip install git+https://github.com/SatelliteQE/nailgun@refs/pull/$(NAILGUN_PR_NUMBER)/merge
+endif
+endif
+
 test/test_playbooks/server_vars.yml:
 	cp test/test_playbooks/server_vars.yml.example test/test_playbooks/server_vars.yml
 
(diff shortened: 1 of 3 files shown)


sean797:
I think if you edit your comment and restart build we'll be able to check travis log.

While I think this is a intresting, could it lead to us merging PRs before Nailgun changes are in?

mdellweg:
> could it lead to us merging PRs before Nailgun changes are in?
Yes, i had the same thought, the moment i sent the PR. I guess, travis can only set the state to red or green. There is no blue (good in theory, but prerequisites are pending).
I think, it is nicer for the PR Developer to get the tests green.
So we must be careful and use the labels wisely.

sean797:
Perhaps we can exit 1 at the end if its using Nailgun not from master?

Also is doesn't seem to work :(

```
Collecting nailgun from git+https://github.com/SatelliteQE/nailgun.git@master#egg=nailgun (from -r requirements-dev.txt (line 4))

**Comment by mdellweg** [comment]
@sean797 That could work if we get a nice message, that tells so.
I will come back to this later.

File: Makefile [hunk 1]
```diff
@@ -8,6 +8,9 @@ MODULE_PATH=modules/$(MODULE).py
 DEBUG_DATA_PATH=test/debug_data/$(DATA).json
 DEBUG_OPTIONS=-m $(MODULE_PATH) -a @$(DEBUG_DATA_PATH) -D $(PDB_PATH)
 
+PR_MSG=$(shell http --json "https://api.github.com/repos/${TRAVIS_REPO_SLUG}/pulls/${TRAVIS_PULL_REQUEST}" | jq -r .body | tr -C "[:alnum:]" _)
+NAILGUN_PR_NUMBER=$(shell echo $(PR_MSG) | sed -n 's/.*SatelliteQE_nailgun[^0-9]*\([0-9]*\).*/\1/p')
```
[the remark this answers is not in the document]
mdellweg: I had serious shell issues when the Message contained parantheses. At that point i realized, if i sanitize all possibly weird characters to `_`, the whole regex gets much more stable and predictable. The sed does more than a grep. It extracts the PR Number.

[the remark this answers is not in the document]
akofink: Ok, it seems to work fine as is, just wondering :)

akofink:
Seems like it's not working quite right: https://travis-ci.org/theforeman/foreman-ansible-modules/jobs/461736446#L651

```
pip install git+https://github.com/SatelliteQE/nailgun@refs/pull//merge
```

The PR number is blank, and our logic doesn't catch it either.

Makefile
```diff
@@ -54,6 +59,16 @@ test-setup: test/test_playbooks/server_vars.yml
 	pip install --upgrade pip
 	pip install -r requirements-dev.txt
 	pip install -r https://raw.githubusercontent.com/ansible/ansible/devel/requirements.txt 
+ifdef TRAVIS_PULL_REQUEST
+	@echo "Handling dependencies for PR #$(TRAVIS_PULL_REQUEST)"
+	@echo "Repository: $(TRAVIS_REPO_SLUG)"
+	@echo "$(PR_MSG)"
```
akofink: This appears to be null? https://travis-ci.org/theforeman/foreman-ansible-modules/jobs/462829899#L657

mdellweg: It's strange. If i run it locally with `TRAVIS_PULL_REQUEST=206` and `TRAVIS_REPO_SLUG=theforeman/foreman-ansible-modules` it resolves correctly to 203.

sean797: We could use `TRAVIS_COMMIT_MESSAGE` ?

mdellweg: Let's see

mdellweg: I guess, that's not it:
```Merge fa53c9e3c602ea470fd29699098868843901919c into 9b39d58591e9e0bebd411fbc66a887bc83c24541```

Now it's getting really strange. Seems like the API sometimes produces the right information.

sean797: I think that ENV var contains the latest commit message, so it'll mean the author must add the requirement to the commit message. Which IMO is okay?

mdellweg: It's the generated Commit Message made by checking out the `refs/pulls/<number>/merge` branch. Anyway, i do not like to have such information in any commit message.

PR #207, opened 2018-11-30
Repository Sets: make name & product optional inplace of label

sean797:
Requires ~~SatelliteQE/nailgun#574~~

### DIFF
--- a/module_utils/ansible_nailgun_cement.py
+++ b/module_utils/ansible_nailgun_cement.py
@@ -526,9 +526,14 @@ def find_content_credential(module, name, organization, failsafe=False):
     return handle_find_response(module, content_credential.search(), message="No content credential found for %s" % name, failsafe=failsafe)
 
 
-def find_repository_set(module, name, product, failsafe=False):
-    repo_set = RepositorySet(name=name, product=product)
-    return handle_find_response(module, repo_set.search(), message="No repository set found for %s" % name, failsafe=failsafe)
+def find_repository_set(module, name=None, product=None, organization=None, label=None, failsafe=False):
+    if label and organization:
+        repo_set = RepositorySet(organization=organization).search(query={'search': 'label="{}"'.format(label)})
+    elif name and organization:
+        repo_set = RepositorySet(name=name, organization=organization).search()
+    else:
+        repo_set = RepositorySet(name=name, product=product).search()
+    return handle_find_response(module, repo_set, message="No repository set found for %s" % name or label, failsafe=failsafe)
 
 
 def find_setting(module, name, failsafe=False):
--- a/modules/katello_repository_set.py
+++ b/modules/katello_repository_set.py
@@ -93,6 +93,25 @@ EXAMPLES = '''
       basearch: "x86_64"
     state: enabled
 
+- name: "Enable RHEL 7 RPMs repositories with label"
+  katello_repository_set:
+    username: "admin"
+    password: "changeme"
+    server_url: "https://foreman.example.com"
+    verify_ssl: false
+    organization: "Default Organization"
+    label: rhel-7-server-rpms
+    repositories:
+    - releasever: "7.0"
+      basearch: "x86_64"
+    - releasever: "7.1"
+      basearch: "x86_64"
+    - releasever: "7.2"
+      basearch: "x86_64"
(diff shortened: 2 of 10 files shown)


mdellweg:
Another test. Yay!

module_utils/ansible_nailgun_cement.py
```diff
@@ -520,9 +520,12 @@ def find_content_credential(module, name, organization, failsafe=False):
     return handle_find_response(module, content_credential.search(), message="No content credential found for %s" % name, failsafe=failsafe)
 
 
-def find_repository_set(module, name, product, failsafe=False):
-    repo_set = RepositorySet(name=name, product=product)
-    return handle_find_response(module, repo_set.search(), message="No repository set found for %s" % name, failsafe=failsafe)
+def find_repository_set(module, name, product, organization=None, label=None, failsafe=False):
```
akofink: IMO this should be:

```py
def find_repository_set(module, name, product=None, organization=None, failsafe=False)
```

Repo sets index API takes either organization ID or product ID (and repo set name) to resolve the repo set ID. I'm confused what this label is. Care to explain?

akofink: I don't understand. I'm looking at our [API controller params](https://github.com/Katello/katello/blob/master/app/controllers/katello/api/v2/repository_sets_controller.rb#L23) and don't see label. It looks like you can search with either `org ID + name` or `product ID + name`.

sean797: @akofink the label search uses scoped_search, which does allow us to search via label  https://github.com/Katello/katello/blob/master/app/models/katello/product_content.rb#L23

akofink: Ah, ok that makes sense. :)

akofink: So then, why is name required here in the method definition? You can use either name or label, right? So it should look like this:

```py
def find_repository_set(module, name=None, label=None, product=None, organization=None, failsafe=False)
```

This allows using name or label and product or organization. Maybe I'm missing something.

sean797: Yup, nice spot! Updated :+1: Thanks!

sean797:
Updated and rebased :green_apple: - thanks @akofink

module_utils/ansible_nailgun_cement.py
```diff
@@ -520,9 +520,12 @@ def find_content_credential(module, name, organization, failsafe=False):
...
 
 
-def find_repository_set(module, name, product, failsafe=False):
-    repo_set = RepositorySet(name=name, product=product)
-    return handle_find_response(module, repo_set.search(), message="No repository set found for %s" % name, failsafe=failsafe)
+def find_repository_set(module, name=None, product=None, organization=None, label=None, failsafe=False):
+    if label and organization:
+        repo_set = RepositorySet(organization=organization).search(query={'search': 'label="{}"'.format(label)})
+    else:
+        repo_set = RepositorySet(name=name, product=product).search()
```
akofink: The error is here. With the input I provided (name and org only), `product` is `None`.

sean797:
I don't think that's ever worked, I'm surprised to hear it works with hammer. If we just supply `name` and `org` how do we know what product to use?

modules/katello_repository_set.py
```diff
@@ -132,11 +151,12 @@ def get_desired_repos(desired_substitutions, available_repos):
     return desired_repos
 
 
-def repository_set(module, name, organization, product, state, repositories=[]):
+def repository_set(module, name, organization, product, label, state, repositories=[]):
     changed = False
     organization = find_organization(module, organization)
-    product = find_product(module, product, organization)
```
sean797: Looking at this old line, I'm fairly sure that never worked.. I'd vote for merging this as is.. If we want we can support that feature in the future :-)

akofink:
I actually think this is a fairly common use case. Repository set names (and labels) should be unique within an organization. We're really close to feature parity here with the API/CLI; I'd push to just tweak it a bit more; however, if you insist, we can merge as is and file a new bug to allow name+org only.

sean797:
Didn't realise it was going to be that simple - Updated and I added a test for name + org ;)

PR #208, opened 2018-12-04
Fix same repo name in different products in CV

bagasse:
This fixes the case where multiple products have a repo with the same name (as repo name is unique only in the product scope). ATM the content view creation fail with 'repository_id has already been taken'.


sean797:
Yup thanks @bagasse :+1:

PR #211, opened 2018-12-05
Add example to Sync all repositories

sean797:
Since this is somewhat complicated, It may be worth having an example. I suspect this may be a common task?


sean797:
Yes, I tested it :+1: 

Thanks @mdellweg

PR #215, opened 2018-12-11
generate sphinx docs

evgeni:
example rendering based on this branch at https://foreman-ansible-modules.readthedocs.io/en/sphinx/


evgeni:
@ehelms this should now render much cleaner than before

sean797:
Is it worth putting effort into this considering https://community.theforeman.org/t/how-to-ship-ansible-modules/11541?

It would be nice if Ansible had a plugable docs API, maybe we can request that as part of the work they are doing to use Galaxy to distribute modules that `gundalow` was talking about.

evgeni:
@sean797 the current state was half a day of work, so not too much effort, and this would give us docs *today*, not whenever we are in ansible. so I'd say yes.

sean797:
Okay, I think I agree..

Couple of questions:
- Where do we want to host the docs? My initial thought is theforeman.org, would a link on the website to readthedocs work?
-  Is it possible to add `docs/modules/` to `.gitignore`? I assume thoses files are created during the build phase and are only required to serve the docs?

evgeni:
> * Where do we want to host the docs? My initial thought is theforeman.org, would a link on the website to readthedocs work?

readthedocs is easy as it builds Sphinx based docs for us for free. we can add a link on tfm.o, or we could add a [custom domain that points to readthedocs](https://docs.readthedocs.io/en/latest/custom_domains.html)

we could also host it on https://theforeman.org/documentation.html directly, but that'd probably require us to build the docs ourself (and add some commit hook to trigger rebuilds).

> * Is it possible to add `docs/modules/` to `.gitignore`? I assume thoses files are created during the build phase and are only required to serve the docs?

not quite. readthedocs.io only calls `sphinx build`, and I didn't manage to convice Sphinx to call a custom script (my `make generate-docs`) to generate the ReST files for us. So currently the committed version is what gets built.

.gitmodules
```diff
@@ -0,0 +1,4 @@
+[submodule "docs/.tmp/ansible"]
```
ekohl: Any reason you're doing this via a submodule rather than as a pip dependency? We do that in https://github.com/theforeman/obal/blob/master/requirements-docs.txt and is probably nicer

.gitmodules
```diff
@@ -0,0 +1,4 @@
...
+	path = docs/.tmp/ansible
+	url = https://github.com/ansible/ansible.git
```
mdellweg: You can relate to submodules using relative paths. I think, this is nicer if they are meant to live on the same server.

mdellweg:
I like the fact to have docs. Surely, it would be nicer if we did not need to commit any autogenerated stuff.

.gitmodules
evgeni: At least the last time I looked, the docs/bin/ folder was not in the python distribution, and I need that to generate the docs.

evgeni:
> Would a sphinx plugin work? It'd be good for the ecosystem as well I guess.

Probably, but I didn't fully understand how to hook into the "collect" step when I briefly looked at the Sphinx docs.

Fobhep:
> I like the fact to have docs. Surely, it would be nicer if we did not need to commit any autogenerated stuff.

Fully agree

ehelms:
I'd be OK with this approach for a time. With the caveat of looking toward a sphinx plugin or eventual Ansible. The goal of the documentation is to make user's lives easier by giving them an easy reference. If we achieve that, even through more verbose methods I consider that a win.

ehelms:
I see a few comments to address, but anything else holding up pushing out a version of this to have some docs which are better than none?

sean797:
For the sack of clarity, to regenerate and update these docs it's going to require someone manually do it and commits the changes?

If that's the case, I'd vote for committing the actual docs to https://github.com/theforeman/theforeman.org repo, just like we do for APIDOC

Ultimately, I'm OK with this as it is, if there is enough support from others, I would just ask that we do actually address a few of the comments above in the foreseeable future.

evgeni:
So I tried adding this to our main site and failed.
We use Jekyll for theforeman.org, which ignores folders called `_static` by default. But Sphinx (which we use to generate these docs) always deploys its CSS and JS into a `_static` subfolder. This results in broken UI when I try to include these docs into our main site, as I didn't manage to configure Jekyll not to remove that folder from the built result.

evgeni:
see https://github.com/theforeman/theforeman.org/pull/1294

evgeni:
@sean797 @ehelms @ekohl @Fobhep @mdellweg 
I've refactored this to have only the generation code here in the repo, and the "storage" is implemented in https://github.com/theforeman/theforeman.org/pull/1294 to be served via theforeman.org

Please review and judge accordingly :)

ehelms:
@evgeni Anything else you are waiting on? Can I drop your WIP status from each PR and merge them?

evgeni:
@ehelms :ship:

ehelms:
Thanks @evgeni !

PR #217, opened 2018-12-17
Add foreman_user module

ogajduse:
This module allows managing Foreman users. I've tested the module manually and everything should work except the following.

- the second run of the module with the same arguments using `auth_source: INTERNAL` will fail with 422
- can not change the password once is set

I haven't found support for `auth_source_externals`, `auth_source_internals` and `auth_sources` in Nailgun. This is something that needs to be done to provide full functionality of the foreman_user module. There is also missing support for `current_password` parameter while using PUT.

I haven't put any effort into creating any tests for this module yet. I will come back to it later.

### DIFF
--- a/module_utils/ansible_nailgun_cement.py
+++ b/module_utils/ansible_nailgun_cement.py
@@ -9,6 +9,7 @@ from nailgun.config import ServerConfig
 from nailgun.entities import (
     _check_for_value,
     ActivationKey,
+    AuthSourceLDAP,
     Entity,
     AbstractContentViewFilter,
     CommonParameter,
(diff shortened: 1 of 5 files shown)


mdellweg:
Just a first glance, but this looks very promising.

In case you have not yet seen it, take a look at:
https://github.com/theforeman/foreman-ansible-modules#how-to-test-modules-in-this-repository

ogajduse:
@mdellweg thanks for the first check
Added tests for this module. I won't be adding the fixtures until the following Nailgun issues will be solved:
* SatelliteQE/nailgun#591
* SatelliteQE/nailgun#594

ogajduse:
This PR is obsoleted by #296.

mdellweg:
Thank you for the efford. #296 has been merged.

PR #219, opened 2019-01-07
Add nested location support

bagasse:
* Add nested location support.
* Added a test to subnet with a nested location usage.

`module_utils/foreman_helper.py` part can be used for any nested resources (like hostgroup, organization...)

### DIFF
--- a/test/test_playbooks/domain_vars.yml
+++ b/test/test_playbooks/domain_vars.yml
@@ -2,7 +2,7 @@
 domain_name: "example.org"
 domain_dns_proxy: "centos7-devel.example.com"
 domain_locations:
-  - 'Default Location'
+  - 'Test Location'
 domain_organizations:
-  - 'Default Organization'
+  - 'Test Organization'
 ...
--- a/module_utils/ansible_nailgun_cement.py
+++ b/module_utils/ansible_nailgun_cement.py
@@ -453,9 +453,9 @@ def find_locations(module, locations):
     return list(map(lambda location: find_location(module, location), locations))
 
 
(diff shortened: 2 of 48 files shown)


test/test_playbooks/domain_vars.yml
```diff
@@ -2,7 +2,7 @@
 domain_name: "example.org"
 domain_dns_proxy: "centos7-devel.example.com"
 domain_locations:
-  - 'Default Location'
+  - 'Test Location'
```
sean797: I'd suspect this for using the Default one because it requires a Smart proxy to be manually setup beforehand?

bagasse: Nope, I think this is only used to hide domain on host creation/update. But, yes, has you said, to run `make record_domain` you have to setup manualy a smart-proxy with dns plugin enabled and name it `centos7-devel.example.com`. No orgs or locs are affected to this proxy on my setup and all tests are ok.
screenshot_foreman_sp

We can handle this in two ways:

1. If we use `Default Location` (or `Default Organization`) in tests we must not delete it after tests because recent version of foreman must have at least 1 org/loc.
2. Use trowable Orgs/Locs and clean it up after tests

As you can see I choose the second option, because all other tests I saw was doing things like this. But it can be easily reverted if you want.

bagasse:
Thanks @sean797 !

test/test_playbooks/domain_vars.yml
mdellweg: Is this the reason, you needed to rerecord so much tests?
If yes, feel free to resolve again.

PR #228, opened 2019-02-06
move common argument spec entries to a common module base

evgeni:
this only touches the domain module, I'd apply the same pattern to all others if y'all like this.

### DIFF
--- a/module_utils/foreman_helper.py
+++ b/module_utils/foreman_helper.py
@@ -3,8 +3,16 @@
 
 import re
 import yaml
+import traceback
 
 from ansible.module_utils.basic import AnsibleModule
+
+try:
+    import ansible.module_utils.ansible_nailgun_cement
+    HAS_NAILGUN = True
+except ImportError:
+    HAS_NAILGUN = False
+    NAILGUN_IMP_ERR = traceback.format_exc()
 
 
 class ForemanAnsibleModule(AnsibleModule):
@@ -19,6 +27,10 @@ class ForemanAnsibleModule(AnsibleModule):
         )
         args.update(argument_spec)
         super(ForemanAnsibleModule, self).__init__(argument_spec=args, **kwargs)
+
+        if not HAS_NAILGUN:
+            self.fail_json(msg='The nailgun Python module is required',
+                           exception=NAILGUN_IMP_ERR)
 
 
 def filter_module_params(module):
--- a/module_utils/foreman_helper.py
+++ b/module_utils/foreman_helper.py
@@ -27,6 +27,14 @@ def filter_module_params(module):
     return dict([(k, v) for (k, v) in module.params.items() if v is not None])
 
 
+def get_server_params(module_params):
+    server_url = module_params.pop('server_url')
+    username = module_params.pop('username')
+    password = module_params.pop('password')
+    verify_ssl = module_params.pop('verify_ssl')
+    return (server_url, username, password, verify_ssl)
+
+
 # Helper for templates
 def parse_template(template_content, module):
     try:
--- a/module_utils/foreman_helper.py
+++ b/module_utils/foreman_helper.py
@@ -16,6 +16,7 @@ class ForemanAnsibleModule(AnsibleModule):
             username=dict(required=True),
             password=dict(required=True, no_log=True),
             verify_ssl=dict(type='bool', default=True),
+            state=dict(choices=['present', 'absent'], default='present'),
         )
         argument_spec.update(custom_argument_spec)
         kwargs['argument_spec'] = argument_spec
--- a/module_utils/foreman_helper.py
+++ b/module_utils/foreman_helper.py
@@ -34,9 +34,8 @@ class ForemanAnsibleModule(AnsibleModule):
 
     def parse_params(self):
         module_params = self.filter_module_params()
-        state = module_params.pop('state')
         server_params = self.get_server_params(module_params)
-        return (server_params, module_params, state)
+        return (server_params, module_params)
 
     def filter_module_params(self):
         return {k: v for (k, v) in self.params.items() if v is not None}
@@ -47,6 +46,21 @@ class ForemanAnsibleModule(AnsibleModule):
         password = module_params.pop('password')
         verify_ssl = module_params.pop('verify_ssl')
         return (server_url, username, password, verify_ssl)
+
+
+class ForemanEntityAnsibleModule(ForemanAnsibleModule):
+
+    def __init__(self, argument_spec, **kwargs):
+        args = dict(
+            state=dict(choices=['present', 'absent'], default='present'),
+        )
+        args.update(argument_spec)
+        super(ForemanEntityAnsibleModule, self).__init__(argument_spec=args, **kwargs)
+
+    def parse_params(self):
+        (server_params, module_params) = super(ForemanEntityAnsibleModule, self).parse_params()
+        state = module_params.pop('state')
+        return (server_params, module_params, state)
 
 
 # Helper for templates
--- a/module_utils/foreman_helper.py
+++ b/module_utils/foreman_helper.py
@@ -32,17 +32,20 @@ class ForemanAnsibleModule(AnsibleModule):
             self.fail_json(msg='The nailgun Python module is required',
                            exception=NAILGUN_IMP_ERR)
 
+    def parse_params(self):
+        module_params = self.filter_module_params()
+        server_params = self.get_server_params(module_params)
+        return (server_params, module_params)
 
-def filter_module_params(module):
-    return {k: v for (k, v) in module.params.items() if v is not None}
+    def filter_module_params(self):
+        return {k: v for (k, v) in self.params.items() if v is not None}
 
-
-def get_server_params(module_params):
-    server_url = module_params.pop('server_url')
-    username = module_params.pop('username')
-    password = module_params.pop('password')
-    verify_ssl = module_params.pop('verify_ssl')
-    return (server_url, username, password, verify_ssl)
+    def get_server_params(self, module_params):
+        server_url = module_params.pop('server_url')
+        username = module_params.pop('username')
+        password = module_params.pop('password')
+        verify_ssl = module_params.pop('verify_ssl')
+        return (server_url, username, password, verify_ssl)
 
 
 # Helper for templates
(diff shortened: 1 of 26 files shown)


evgeni:
from live discussion with @ekohl:
* creating "domain_dict" and splitting "entity-related" from "connection related" vars should also get a helper

m-bucher:
@ekohl `state` looks like a natural choice to add to the base module, but there might be modules where we want to add additional values to `choices` or change the `default` value.

ekohl:
That's what @evgeni also said (in person) but it's default. You can override the defaults for those with more states.

module_utils/foreman_helper.py
```diff
@@ -4,6 +4,23 @@
...
+    def __init__(self, **kwargs):
+        custom_argument_spec = kwargs.pop('argument_spec', dict())
+        argument_spec = dict(
+            server_url=dict(required=True),
+            username=dict(required=True),
+            password=dict(required=True, no_log=True),
+            verify_ssl=dict(type='bool', default=True),
+        )
+        argument_spec.update(custom_argument_spec)
+        kwargs['argument_spec'] = argument_spec
```
mdellweg: Why don't you just update the `kwargs['argument_spec']` in place?

mdellweg:
I like the Idea. It is really DRYing up stuff.

module_utils/foreman_helper.py
evgeni: because `kwargs['argument_spec'].update(argument_spec)` would override the values from the module with the defaults, and I couldn't implement `state` as discussed above (`ForemanAnsibleModule` has state `['present', 'absent']` and some modules add `present_with_defaults` and similar)

module_utils/foreman_helper.py
```diff
@@ -4,6 +4,36 @@
...
+            verify_ssl=dict(type='bool', default=True),
+            state=dict(choices=['present', 'absent'], default='present'),
+        )
+        argument_spec.update(custom_argument_spec)
+        kwargs['argument_spec'] = argument_spec
+        super(ForemanAnsibleModule, self).__init__(**kwargs)
+
+
+def filter_module_params(module):
+    return dict([(k, v) for (k, v) in module.params.items() if v is not None])
```
ekohl: Can we dict comprehension (move to Python 2.7+)? `{k: v for (k, v) in module.params.items() if v is not None}` is easier to read.

evgeni: I think making this py2.7+ is OK, yeah.

evgeni:
@sean797 we have modules w/o tests? :fire:

sean797:
Yes, there are a few :sob: 

We have 34 modules, but only 26 in https://github.com/theforeman/foreman-ansible-modules/blob/9576dd8bf686f1f42a5a5cec14dee549d3a13b46/test/test_crud.py#L8-L36

evgeni:
Yeah, in this case, probably only updating tested ones is a good idea :)

mdellweg:
@evgeni yeah, I stopped refactoring modules without tests a long time ago. I guess they are really old by now. :confounded:

module_utils/foreman_helper.py
mdellweg: Sometimes it's hard to see, where the changeable datatypes make a difference in python...

evgeni:
@sean797 @mdellweg @m-bucher @ekohl OK, I think I am at a state where we are as DRY as possible. What do y'all think?

module_utils/foreman_helper.py
```diff
@@ -3,6 +3,46 @@
...
+        if not HAS_NAILGUN:
+            self.fail_json(msg='The nailgun Python module is required',
+                           exception=NAILGUN_IMP_ERR)
+
+
+def filter_module_params(module):
+    return {k: v for (k, v) in module.params.items() if v is not None}
+
+
+def get_server_params(module_params):
```
ekohl: I wonder if this should be changed into a method on a ForemanAnsibleModule instance `server, params = module.parse_params()`

evgeni: like this?

modules/foreman_domain.py
```diff
@@ -121,35 +120,23 @@
...
 
-    server_url = domain_dict.pop('server_url')
-    username = domain_dict.pop('username')
-    password = domain_dict.pop('password')
-    verify_ssl = domain_dict.pop('verify_ssl')
     state = domain_dict.pop('state')
 
     try:
+        (server_url, username, password, verify_ssl) = server_params
         create_server(server_url, (username, password), verify_ssl)
```
ekohl: I think this try/except block could also be part of the base module since it's repeated for every module.

modules/foreman_domain.py
```diff
@@ -121,35 +120,23 @@
...
-
-    domain_dict = dict(
-        [(k, v) for (k, v) in module.params.items() if v is not None])
+    (server_params, domain_dict) = module.parse_params()
 
-    server_url = domain_dict.pop('server_url')
-    username = domain_dict.pop('username')
-    password = domain_dict.pop('password')
-    verify_ssl = domain_dict.pop('verify_ssl')
     state = domain_dict.pop('state')
```
ekohl: Should state also be popped off in `parse_params` and returned in the tuple?

modules/foreman_domain.py
evgeni: Long term, yes. I've been trying to keep actual `nailgun` calls (like the `create_server` here) out of the generic module for now.

modules/foreman_domain.py
evgeni: yes

evgeni:
I would also do the refactor of the other tested modules right now, just wanted to hear some ACKs on the general approach first.

ekohl:
:+1: on converting all the modules to this. A solid improvement, even if it's not the end station :)

module_utils/foreman_helper.py
```diff
@@ -34,8 +34,9 @@ def __init__(self, argument_spec, **kwargs):
 
     def parse_params(self):
         module_params = self.filter_module_params()
+        state = module_params.pop('state')
```
mdellweg: Is state really omnipresent? Don't we have 'action' modules without it?
If yes, we might want to have a `ForemanAnsibleModule` and a `FormemanEntityAnsibleModule`.

sean797:
```
$ grep -L state *.py
foreman_search_facts.py
foreman_setting_facts.py
__init__.py
katello_content_view_publish.py
katello_content_view_version_promote.py
katello_sync_plan.py
katello_sync.py
katello_upload.py
```

Almost every module ;)

mdellweg: Did i forget to mention, that the second one inherits from the first one?

evgeni: @mdellweg that was obvious, tho :package: 

@sean797 your grep didn't catch `foreman_setting`, which does not have state either (which sounds like a bug)

evgeni: implemented

evgeni:
okay, that's the first batch, all tested foreman modules.

modules/foreman_operating_system.py
```diff
@@ -165,14 +165,12 @@
...
+    pass
 
-    has_import_error = False
-except ImportError as e:
-    has_import_error = True
-    import_error_msg = str(e)
-
-from ansible.module_utils.basic import AnsibleModule
-
+from ansible.module_utils.foreman_helper import (
```
ekohl: Any particular reason this line is split where in other files it isn't?

modules/foreman_setting.py
```diff
@@ -158,7 +142,7 @@ def main():
 
     entity_dict = sanitize_entity_dict(entity_dict, name_map)
 
-    changed, entity = naildown_entity(Setting, entity_dict, entity, 'present', module, check_type=True)
+    changed, entity = naildown_entity(Setting, entity_dict, entity, state, module, check_type=True)
```
ekohl: Now that we have state, we should add it to the `DOCUMENTATION`

modules/foreman_operating_system.py
evgeni: artifact from the time I've imported multiple things from the helper

modules/foreman_setting.py
evgeni: details, next you'll ask for an "absent" test :P

mdellweg: I think, it is not possible to delete settings. So 'hidden present'(tm) is the only sensible option here.

evgeni: hah, this sounds about right. will fix that.

mdellweg: Hrmpf. If i had let you run into writing the 'absent' test, you might have found a hidden feature of the foreman. Now, we settle with the boring stuff. ;)

evgeni:
okay, I think this is past RFC phase now. renamed.

would like a final round of ACKs after all these changes :)

modules/katello_sync_plan.py
```diff
@@ -192,12 +190,8 @@ def sync_plan(self, name, organization, interval=None, enabled=None, sync_date=N
 
 
 def main():
-    module = AnsibleModule(
+    module = ForemanAnsibleModule(
```
mdellweg: KatelloAnsibleModule?

mdellweg: Ah i see, there is no such thing...
But should this be an entity anyway? If yes, let's leave it to another PR.

evgeni: To dry up that `organization=dict(required=True),` in here?

evgeni: and no, this is not an entity, as it has enabled/disabled vs present/absent (don't ask)

mdellweg: Can't it be present and disabled? Now i just have to ask.

evgeni: I have no idea, I didn't write the module. Maybe @akofink wants to chime in, but this is probably out of scope for this PR :)

akofink: The [example](https://github.com/theforeman/foreman-ansible-modules/blob/master/modules/katello_sync_plan.py#L78) shows present and disabled. It doesn't seem like it's possible to *remove* a sync plan using this module, however, so I'd agree there's room for improvement here!

PR #230, opened 2019-02-11
move common documentation to a fragment

evgeni:
As of Ansible 2.8, we can ship custom documentation fragments together
with our modules [1], so let's do this.

(Documentation fragments were supported for upstream Ansible modules
since at least 2.3, but you couldn't add more paths to search for them
until 2.8)

[1] https://docs.ansible.com/ansible/devel/dev_guide/developing_modules_documenting.html#documentation-fragments

### DIFF
--- a/plugins/modules/foreman_bookmark.py
+++ b/plugins/modules/foreman_bookmark.py
@@ -36,25 +36,6 @@ requirements:
   - "ansible >= 2.3"
   - "apypie"
 options:
-  server_url:
-    description:
-      - URL of Foreman server
-    required: true
-  username:
-    description:
-      - Username on Foreman server
-    required: true
-  password:
-    description:
-      - Password for user accessing Foreman server
-    required: true
-  validate_certs:
-    aliases: [ verify_ssl ]
-    description:
-      - Verify SSL of the Foreman server
-    required: false
-    default: true
-    type: bool
   name:
     description:
       - Name of the bookmark
@@ -80,6 +61,7 @@ options:
       - present
       - present_with_defaults
       - absent
+extends_documentation_fragment: foreman
 '''
 
 EXAMPLES = '''
--- a/plugins/modules/katello_content_view_filter.py
+++ b/plugins/modules/katello_content_view_filter.py
@@ -29,24 +29,6 @@ requirements:
     - "python >= 2.6"
     - "ansible >= 2.3"
 options:
-  server_url:
-    description:
-      - URL of Foreman server
-    required: true
-  username:
-    description:
-     - Username on Foreman server
-    required: true
-  password:
-    description:
-      - Password for user accessing Foreman server
-    required: true
-  validate_certs:
-    aliases: [ verify_ssl ]
-    description:
-      - Verify SSL of the Foreman server
-    default: true
-    type: bool
   content_view:
     description:
       - Name of the content view
@@ -126,6 +108,7 @@ options:
     description:
       - Create an include filter
     default: False
+extends_documentation_fragment: foreman
 '''
 
 EXAMPLES = '''
--- a/plugins/modules/katello_content_view_filter.py
+++ b/plugins/modules/katello_content_view_filter.py
@@ -26,8 +26,6 @@ description:
 author: "Sean O'Keeffe (@sean797)"
 requirements:
     - "nailgun >= 0.28.0"
-    - "python >= 2.6"
-    - "ansible >= 2.3"
 options:
   content_view:
     description:
--- a/plugins/modules/redhat_manifest.py
+++ b/plugins/modules/redhat_manifest.py
@@ -25,10 +25,6 @@ description:
     - Download and modify a Red Hat Satellite Subscription Manifest
 author:
     - "Sean O'Keeffe (@sean797)"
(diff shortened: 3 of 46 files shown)


sean797:
Nice :+1: we definitely need to wait until 2.8 is released, but that's fine as this is WIP. I suspect we should probably also wait until 2.9 is released as this is backwards incompatible?

evgeni:
This is backwards incompatible, yes. But it also "only" breaks `ansible-doc` invocations on pre-2.8 installs.

Given that we have online-accessible docs now, I'd actually merge this as soon as 2.8 hits stable (and probably Fedora :))

evgeni:
@sean797 @mdellweg this should now pass tests and everything.

mdellweg:
I guess, we need to bump the ansible dependency to 2.8 everywhere.

evgeni:
> I guess, we need to bump the ansible dependency to 2.8 everywhere.

As far as I can see the modules still work fine in older Ansible versions, just `ansible-doc` is broken as it cannot find the snippet. IMHO this warrants a warning in the docs, but no strict dependency.

mdellweg:
> As far as I can see the modules still work fine in older Ansible versions, just `ansible-doc` is broken as it cannot find the snippet. IMHO this warrants a warning in the docs, but no strict dependency.

OK in that respect, the `module_utils` (v2.3) change remains the reasoning for this dependency, i guess.

evgeni:
I've added a line to the README explaining the situation.

If I get a tentative ACK on the change as such, I'll go ahead an modify all the modules :)

mdellweg:
I am all in for this change.

ekohl:
:+1: Looks like a nice deduplication.

evgeni:
this is now ready for review.

I only changed the modules that use the Foreman(Apypie)AnsibleModule interface, as these are the only ones that get all the updated params anyways.

I also took the liberty to yank all the mentions of ansible and python in the requirements, because I found them silly :)

plugins/modules/katello_content_view_filter.py
```diff
@@ -126,6 +106,7 @@
     description:
       - Create an include filter
     default: False
+extends_documentation_fragment: foreman
```
ekohl: This is not an apypie module. Should this still be replaced? There's a few more of those below.

evgeni: Yes. It's a `ForemanAnsibleModule` (so nailgun, but still with the common wrapper API), which inherits from `ForemanBaseAnsibleModule` which has the common params:

https://github.com/theforeman/foreman-ansible-modules/blob/a2aac6240a419cb87876f9c91a062ecb754c6878/plugins/module_utils/foreman_helper.py#L28-L36

evgeni:
Ansible upstream does not support Python 2.6 anymore, we don't test on Python 2.6, so I'd say it's safe to say we're 2.7 only. This is however not publicly documented. Should it? With the "Ansible doesn't support <2.7" argument, I'd say no.

plugins/modules/foreman_bookmark.py
```diff
@@ -36,25 +36,6 @@
   - "ansible >= 2.3"
   - "apypie"
```
mdellweg: It should be possible, to move the requrements to the doc fragment, i think.
I guess, they will even get merged. If specified here too.

ekohl:
> Ansible upstream does not support Python 2.6 anymore, we don't test on Python 2.6, so I'd say it's safe to say we're 2.7 only. This is however not publicly documented. Should it? With the "Ansible doesn't support <2.7" argument, I'd say no.

We could state that we follow Ansible's Python support. That also means that when they drop 2.7, we can drop it. Similar for Python 3 equivalents.

plugins/modules/foreman_bookmark.py
evgeni: it is, but I didn't do that for the sake of not having two fragments (nailgun, apypie) for now.

mdellweg: So we postpone that. OK for me.

mdellweg: I guess inheriting doc fragements is not really worth it on the string level. ;)

evgeni: Yeah, I thought about that, but it didn't feel like a worthy endeavor for a one-line thing.

evgeni:
@ekohl @mdellweg updated for the latest modules and added a line about supported pythons. :green_apple:

plugins/modules/redhat_manifest.py
```diff
@@ -25,10 +25,6 @@
     - Download and modify a Red Hat Satellite Subscription Manifest
 author:
     - "Sean O'Keeffe (@sean797)"
-requirements:
-    - "nailgun >= 0.28.0"
```
sean797: :laughing:

mdellweg:
It has been an honour to merge this.
Thank you, @evgeni !

plugins/modules/redhat_manifest.py
evgeni: Yeah, I have NO idea how that has sneaked in ;)

Issue #222, opened 2019-01-09
Add compute_attributes support


bagasse:
##### SUMMARY

Today `foreman-ansible-modules` supports:
* `compute_resource`
* `compute_profile`
But missing `compute_attribute` for a real end-to-end setup via ansible. The goal here is to setup compute_{resources, profiles, attributes} that make compute_{resources, profiles} (without the needs to use foreman API/CLI/Webui) fully usable out of the box for our foreman users. As we have multiple foreman instances, it will be easier to share common configuration parts (eg: We use mostly oVirt as compute backend and we map compute profiles to ovirt instance type. Instance types and compute profiles are identical across our foreman instances, and today it is a pain point to configure and maintain these parts of configs on all envs).

As I try to add support for compute_attributes, I saw that there is some missing parts of course in `forman-ansible-modules` but also in `nailgun` and `foreman`. I saw in foreman API doc that compute_attributes support only create/update verbs (but no list, read, delete) and I think this will be a problem.

Any pointers to achieve this are welcome :)

sean797:
I'm not greatly familiar with Compute Resources within Foreman, much less the compute_resource API, so I may be wrong but by the looks of things `compute_attributes` are returned as put of 
- `compute_resources` https://github.com/theforeman/foreman/blob/develop/app/views/api/v2/compute_resources/show.json.rabl#L9 
- `compute_profile` https://github.com/theforeman/foreman/blob/develop/app/views/api/v2/compute_profiles/show.json.rabl#L5

---

So perhaps it would make sense to make the `resource` and `profile` modules take a `compute_attribute` option? It probably going to require Nailgun changes, I imagine it could work in a simlar same way that Activation keys can add subscriptions https://github.com/SatelliteQE/nailgun/blob/master/nailgun/entities.py#L307. It may be best to raise it with the SatelliteQE guys

bagasse:
@sean797, thanks for sharing. Yes I think the better way is to handle this like you said. IMHO, it make sense to add it as 'childs' of compute profile (like it is in handled foreman WebUI). I will open an issue with SatelliteQE guys to see what they think about this.

PR #231, opened 2019-02-11
Refactor compute_resource, improve tests

bagasse:
Refactoring compute_resource. Needs [Naigun PR#614](https://github.com/SatelliteQE/nailgun/pull/614) (merged)

* Deprecate compute resource specific params as direct module param `datacenter` in favor of nested hash `provider_params` for all provider specific params
* Deprecate `provider_auth` in favor of `provider_params`
* Add ability to change compute resource name
* Make `absent` state work even if provider is not set
* Refactor and add tests and make them usable as fixture from other tests (will be used for #222 tests)

### DIFF
--- a/modules/foreman_compute_resource.py
+++ b/modules/foreman_compute_resource.py
@@ -1,6 +1,7 @@
 #!/usr/bin/python
 # -*- coding: utf-8 -*-
 # (c) Philipp Joos 2017
+# (c) Baptiste Agasse 2019
 #
 # This file is part of Ansible
 #
@@ -23,23 +24,33 @@ module: foreman_compute_resource
 short_description: Manage Foreman Compute resources using Foreman API
 description:
   - Create and delete Foreman Compute Resources using Foreman API
+author:
+  - "Philipp Joos (@philippj)"
+  - "Baptiste Agasse (@bagasse)"
 requirements:
-  - "nailgun >= 0.28.0"
+  - "nailgun >= 0.32.0"
   - "ansible >= 2.3"
 options:
   name:
     description: compute resource name
     required: true
+  updated_name:
+    description: new compute resource name
+    required: false
   description:
     description: compute resource description
     required: false
   provider:
-    description: provider
+    description: Compute resource provider. Required if I(state=present).
     required: false
     default: None
     choices: ["vmware", "libvirt", "ovirt"]
   provider_auth:
-    description: provider authentication
+    description: Deprecated, use I(provider_params) instead. Provider authentication
+    required: false
+    default: None
+  provider_params:
+    description: Parameter specific to compute resource provider
     required: false
     default: None
   locations:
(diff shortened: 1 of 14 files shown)


evgeni:
This looks good to me on a first glance! :+1: 

Would you like to use the new module helpers introduced in https://github.com/theforeman/foreman-ansible-modules/pull/228? The actual porting effort should be quite small (see https://github.com/theforeman/foreman-ansible-modules/pull/228/commits/da35e950fafd4a958d2f3f09e21d32cc53f11e93 for an example). This can absolutely be postponed to another PR if you don't want/have time/whatever :)

evgeni:
@philippj do you want to have a look here too?

mdellweg:
Looks fine to me. And the tests is a big :heavy_plus_sign: .

bagasse:
@evgeni, done :)

modules/foreman_compute_resource.py
```diff
@@ -98,123 +128,145 @@
...
+            organizations=dict(type='list'),
+            state=dict(type='str', default='present', choices=['present', 'absent', 'present_with_defaults']),
+            # provider-params (deprecated), use nested keys in provider_params param instead
+            provider_auth=dict(type='dict'),
+            url=dict(type='str'),
+            display_type=dict(type='str'),
+            datacenter=dict(type='str'),
+        ),
+        required_if=(
+            ['state', 'present', ['provider']],
```
sean797: Would you mind also documenting this? Something like `Required if I(state=present).`

modules/foreman_compute_resource.py
```diff
@@ -39,7 +46,11 @@
     default: None
     choices: ["vmware", "libvirt", "ovirt"]
   provider_auth:
-    description: provider authentication
+    description: Deprecated, use provider_params instead. Provider authentication
```
sean797: Would you mind using `I()` for the option name?  `I(provider_params)`

modules/foreman_compute_resource.py
```diff
@@ -98,123 +128,145 @@
...
+        argument_spec=dict(
+            name=dict(type='str', required=True),
+            updated_name=dict(type='str'),
+            description=dict(type='str'),
+            provider=dict(type='str', choices=['vmware', 'libvirt', 'ovirt']),
+            provider_params=dict(type='dict'),
+            locations=dict(type='list'),
+            organizations=dict(type='list'),
+            state=dict(type='str', default='present', choices=['present', 'absent', 'present_with_defaults']),
+            # provider-params (deprecated), use nested keys in provider_params param instead
```
sean797: Should this be `provider-auth (deprecated)` ?

bagasse: Reworded.

modules/foreman_compute_resource.py
sean797: @bagasse I think you missed this one? :-)

bagasse: @sean797 ooops.... sorry... Done :)

sean797:
Sorry @bagasse 35328 - Since https://github.com/theforeman/foreman-ansible-modules/pull/238 was merged can you rebase and update this with the same changes as https://github.com/theforeman/foreman-ansible-modules/pull/238/files#diff-a08444d75f83dc9982ac920da6704a3aL86 ?

evgeni:
ewww, I'm sorry, should've merged that one before and then refactored it too.

bagasse:
@sean797, @evgeni  no problem, done. :)

modules/foreman_compute_resource.py
```diff
@@ -98,123 +128,140 @@
...
 
 try:
-    import nailgun.entity_mixins
-    import nailgun.entities
-    import nailgun.entity_fields
-    import ansible.module_utils.ansible_nailgun_cement as cement
-    has_import_error = False
+    from ansible.module_utils.ansible_nailgun_cement import (
+        create_server,
+        ping_server,
```
mdellweg: I think, you can drop some of those.

bagasse: Thanks, I completely missed that one...

evgeni: so did I in my initial refactor pr :D

PR #234, opened 2019-02-15
also test on python 3.7

evgeni:


evgeni:
well, this was unexpected, why does that fail?!

evgeni:
great, this now of course fails on everything else

test/test_playbooks/fixtures/upload-0.yml
```diff
@@ -217,228 +218,230 @@ interactions:
...
+      X-Powered-By: [Phusion Passenger 4.0.53]
+      X-Request-Id: [c2db5a20-2135-4f0c-bfdc-ea1b36de1ec3]
+      X-Runtime: ['0.086606']
+      X-XSS-Protection: [1; mode=block]
     status: {code: 204, message: No Content}
 - request:
-    body: !!python/unicode '{"uploads": [{"checksum": "7a831f9f90bf4d21027572cb503d20b702de8e8785b02c0397445c2e481d81b3",
-      "size": 2438, "id": "331ebdc3-a09f-4359-9abd-2a3cce5e71f9", "name": "bear-4.1-1.noarch.rpm"}]}'
+    body: '{"uploads": [{"id": "30f835b9-14f5-45eb-8c73-ec7d322c2df5", "name": "bear-4.1-1.noarch.rpm",
+      "size": 2438, "checksum": "7a831f9f90bf4d21027572cb503d20b702de8e8785b02c0397445c2e481d81b3"}]}'
```
mdellweg: Maybe it is the order of the dict keys in that list living in a dict.

mdellweg:
In commit 83c3162237df9c341472978d2f3d793d8ca52dd6 we had a specialized body matcher.
Maybe we need something like that again.

test/test_playbooks/fixtures/upload-0.yml
evgeni: nope, it's failing on the step before, `PUT /katello/api/v2/repositories/48/content_uploads/331ebdc3-a09f-4359-9abd-2a3cce5e71f9` (not saying that this is not also a problem)

test/test_playbooks/fixtures/upload-0.yml
```diff
@@ -217,228 +218,230 @@ interactions:
...
...+dummy+package+of+bear%00%00%00%00Ob%22Nsmqe-ws15%00%00%00%00%00%00%2AGPLv2%00Internet%2FApplications%00http%3A%2F%2Ftstrachota.fedorapeople.org%00linux%00noarch%00%00%00%00%2A%81%A4%00%00Ob%22N5938462bfd4a5d750e0851f5b82f3ade%00%00%00%00%00%00%00%00root%00root%00bear-4.1-1.src.rpm%00%00%00%00%FF%FF%FF%FFbear%00%00%00%00%01%00%00J%01%00%00Jrpmlib%28CompressedFileNames%29%00rpmlib%28PayloadFilesHavePrefix%29%003.0.4-1%004.0-1%004.4.2.3%00smqe-ws15+1331831374%00%00%00%00%00%FC%00%00%03%80%21%00%00%00%00%00%00%00%084.1-1%00%00%00%00%00%00%00bear.txt%00%2Ftmp%2F%00-O2+-g+-m64+-mtune%3Dgeneric%00cpio%00gzip%009%00noarch%00noarch-redhat-linux%00%00%00%00%00%00%00%00%00%00%00%01%00ASCII+text%00directory%00%00%00%00%00%00%00%00%00%00%00%27%E6%1Cp%9A0%DF%1D%D6%BAx%E6%AE%A4%D7%AC%00%00%00%00%3F%00%00%00%07%FF%FF%FC%B0%00%00%00%10%1F%8B%08%00%00%00%00%00%00%033070704000%B600%02%D1%06%16%86%89%26%06%D8%81%A1I%9A%99%91%91%91I%2A%84k%94%08%A1%D3%92%0D%F0%834%18CO%BF%24%B7%40%3F%295%B1H%AF%A4%A2%84%01%08%02%12%93%B3%13%D3S%152%F3%8AK%12sr%14J2sS%AD%14%0C-%AC%0C%2C%AD%8CM%14%0C%8Cu%0DMu%8D%0C%0C%8D%B8%18%18%0C%60n%25%06%10%AB%0E%1D%24%C1%18%21A%8E%9E%3E%AEA%8A%8A%8A+w%02%00%A4%F9%EC%9F%28%01%00%00
```
evgeni: Python 3.5:

```python
>>> urlencode({'a':'~'})
'a=%7E'
```

Python 3.7:
```python
>>> urlencode({'a':'~'})
'a=~'
```

:table_flip:

test/vcr_python_wrapper.py
```diff
@@ -23,8 +23,8 @@ def body_json_l2_matcher(r1, r2):
     elif r1.headers.get('content-type') == 'multipart/form-data' and r2.headers.get('content-type') == 'multipart/form-data':
         if r1.body is None or r2.body is None:
             return r1.body == r2.body
-        body1 = sorted(r1.body.split(b'&'))
-        body2 = sorted(r2.body.split(b'&'))
+        body1 = sorted(r1.body.replace(b'~', b'%7E').split(b'&'))
+        body2 = sorted(r2.body.replace(b'~', b'%7E').split(b'&'))
```
evgeni: this feels like a terrible hack, but is the easiest way we can ensure our tapes work across all python versions we support.

@mdellweg can you bring me some shower gel please?

mdellweg: @evgeni I'd say you should go for Sterillium.

Well if it helps for now.

mdellweg: The whole function is a terrible hack to begin with. So i think it's gonna be ok.

evgeni:
@sean797 @akofink @m-bucher mind giving this a second :eyes: as you also touched the VCR config in the past?

test/vcr_python_wrapper.py
sean797: Can we run `encode()` instead?

evgeni: I think `str.encode` can only convert between ["real" encodings](https://docs.python.org/3/library/codecs.html#standard-encodings) and not URL style ones.

PR #237, opened 2019-02-20
Added mirror_on_sync to katello_repository.

furhouse:
Adds `mirror_on_sync` to `katello_repository`, which defaults to `true`.


furhouse:
The fixtures were recorded w/ python 3.7.2. I'll re-record them using 3.6.

evgeni:
@sean797 the failure is because of python 3.7, see #234 for "details" :(

furhouse:
After executing the tests in a py3.6 venv, it seems to work fine :+1: 

@sean797 I've expanded the repository test playbook for idempotency, and included a docker repository as well (forgot about that earlier).

test/test_playbooks/repository.yml
```diff
@@ -79,6 +105,25 @@
...
+          repository_mirror_on_sync: false
+          repository_name: Test Docker Repository
+          repository_content_type: docker
+          repository_url: https://docker.io
+          repository_docker_upstream_name: busybox
+          expected_change: false
+      - include: tasks/repository.yml
+        vars:
+          repository_state: present
+          repository_mirror_on_sync: true
```
sean797: https://github.com/Katello/katello/blob/master/engines/bastion_katello/app/assets/javascripts/bastion_katello/products/details/repositories/details/views/repository-info.html#L116-L123 suggest that mirror on sync cannot be set for docker repos.

Intresting that is seems to report something has changed... Maybe a server bug

furhouse: I'll remove the additional tests from the docker repository tests, to be honest I didn't properly read (or manually verify if its even possible via the webui) https://projects.theforeman.org/issues/21305.

PR #238, opened 2019-02-21
streamline the way we connect/ping to the server

evgeni:
as always, this is an example, I intend to adjust that across all (tested) modules

### DIFF
--- a/modules/foreman_domain.py
+++ b/modules/foreman_domain.py
@@ -129,15 +129,10 @@ def main():
         supports_check_mode=True,
     )
 
-    (server_params, domain_dict, state) = module.parse_params()
+    (domain_dict, state) = module.parse_params()
 
-    try:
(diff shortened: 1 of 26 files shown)


ekohl:
Looks like a good start.

modules/foreman_domain.py
```diff
@@ -129,15 +129,10 @@ def main():
         supports_check_mode=True,
     )
 
-    (server_params, domain_dict, state) = module.parse_params()
+    (_, domain_dict, state) = module.parse_params()
```
mdellweg: Do we still want to expose the server_params to the individual modules?

evgeni: Nope. I've just left it here for now as the other modules are still using it

evgeni:
updated all tested modules for the new API

PR #239, opened 2019-02-22
drop unused imports

evgeni:


evgeni:
maybe not hard enough?

PR #240, opened 2019-02-22
add affects_multiple option to installation medium

manuelbonk:


sean797:
For clarification, this allows us to set the name to `*` and then it will update every entity?

manuelbonk:
Correct, a feature thats also present in the ptable, provisioning template and job template module.

mdellweg:
From the top of my head:
* Delete all entities
* assign all entities to a taxonomy

sean797:
I would prefer if this was DRY'd, moving the logic to `ansible_nailgun_cement`, it seems like a large change to make to each module and a lot of code to maintain

mdellweg:
I'd be fine with merging this as is. Then we had three instances with test coverage (together with #242 ). That should allow for a stable refactoring.

manuelbonk:
> Can you add something to the module documentation?

:white_check_mark:

mdellweg:
:+1: for the test.

manuelbonk:
added some tests @sean797 requested.

mdellweg:
Tests look perfect. Given three ACKs, i am goining to merge this now.

PR #241, opened 2019-02-22
DRY up katello_sync

mdellweg:


modules/katello_sync.py
```diff
@@ -107,115 +111,60 @@
...
-
-    server = ServerConfig(
-        url=server_url,
-        auth=(username, password),
-        verify=verify_ssl
-    )
-    ng = NailGun(server, entities, module)
+    params = module.parse_params()
+
+    module.connect()
```
evgeni: do you need to re-record the tapes if you pass `ping=False` here? just curious :)

mdellweg: No, still needs rerecording.

modules/katello_sync.py
```diff
@@ -107,115 +104,59 @@
...
-    def find_repository(self, name, product, organization):
-        product = self.find_product(product, organization)
-
-        repository = self._entities.Repository(self._server, name=name, product=product)
-        repository._fields['organization'] = entity_fields.OneToOneField(entities.Organization)
-        repository.organization = product.organization
-        response = repository.search()
+    from ansible.module_utils.ansible_nailgun_cement import (
+        create_server,
+        ping_server,
```
sean797: You're not using that ;)

modules/katello_sync.py
```diff
@@ -107,115 +104,59 @@
...
+        find_product,
+        find_repository,
+    )
 
-        if len(response) == 1:
-            return response[0]
-        else:
-            self._module.fail_json(msg="No Repository found for %s" % name)
+    from nailgun import entity_mixins
+    entity_mixins.TASK_TIMEOUT = 180000  # Publishes can sometimes take a long, long time
```
sean797: Is it worth doing this depending on the value of `synchronous`?

modules/katello_sync.py
mdellweg: Man, me too? I was sure, i checked them.

modules/katello_sync.py
mdellweg: It would be nice to have it at sync time. But setting it globally in the module, i would keep it directly after the import.

modules/katello_sync.py
```diff
@@ -22,45 +23,41 @@
...
+  - Sync a Katello repository or product
+author:
+  - "Eric D Helms (@ehelms)"
+  - "Matthias M Dellweg (@mdellweg) ATIX AG"
 requirements:
-    - "nailgun >= 0.28.0"
-    - "python >= 2.6"
-    - "ansible >= 2.3"
+  - "nailgun >= 0.28.0"
+  - "ansible >= 2.3"
```
evgeni: I think we should get rid of those ansible requirements :)

modules/katello_sync.py
```diff
@@ -107,115 +104,57 @@
...
+        product = find_product(module, params['product'], organization)
+        if 'repository' in params:
+            repository = find_repository(module, params['repository'], product)
+        else:
+            repository = None
     except Exception as e:
-        module.fail_json(msg="Failed to connect to Foreman server: %s " % e)
+        module.fail_json(msg='Failed to find entity: %s ' % e)
+
+    synchronous = params['synchronous']
```
evgeni: what's the benefit of having it here as a var, instead of passing `params['synchronous']` directly to the `sync()` calls in L155 and L157?

mdellweg: Nothing, just one step less refactoring iirc.
I'm probably going to change that after having another look.

modules/katello_sync.py
mdellweg: Not sure, i am following you. Should i just delete line 32?

evgeni: That's what I'd do. And thus only list requirements that are actual python modules a user has to install.

mdellweg: I see. Done!

PR #242, opened 2019-02-22
fix affect_multiple for python3 and add test cases

manuelbonk:
In python3 `mydict.keys()` returns a dict view, python2 returns a list. Made sure to always compare lists and added some test cases covering the fix.

### DIFF
--- a/modules/foreman_provisioning_template.py
+++ b/modules/foreman_provisioning_template.py
@@ -389,7 +389,7 @@ def main():
         if state == 'present_with_defaults':
             module.fail_json(msg="'state: present_with_defaults' and 'name: *' cannot be used together")
         if state == 'absent':
-            if entity_dict.keys() != ['name']:
+            if len(entity_dict.keys()) != 1:
                 module.fail_json(msg="When deleting all templates, there is no need to specify further parameters.")
 
     module.connect()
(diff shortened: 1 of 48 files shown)


modules/foreman_provisioning_template.py
```diff
@@ -389,7 +389,7 @@ def main():
         if state == 'present_with_defaults':
             module.fail_json(msg="'state: present_with_defaults' and 'name: *' cannot be used together")
         if state == 'absent':
-            if entity_dict.keys() != ['name']:
+            if list(entity_dict.keys()) != ['name']:
```
mdellweg: would `len(entity_dict.keys()) == 1` work, too?
I think we established already that 'name' is in there. It might be better on resources.

manuelbonk: using `len` would definitely be the cleaner option, but I think you meant `len(entity_dict.keys()) != 1`, didn't you?

evgeni:
@sean797 the tests have a section for `name: *, state: present, expected_change: false` and I don't see how adding tests with `expected_change: true` would even work as we don't allow to pass params to `name: *` calls (and what would that then do? make all ptables have the same content? doesn't make sense)

sean797:
Set all templates to the same Organisations? Or is that not possible with these modules?

evgeni:
Ah, yes, applying taxonomies, that'd work.

manuelbonk:
ok, I've just added some more test cases.

sean797:
Thanks @manuelbonk :+1:

PR #243, opened 2019-02-26
Fixes #222 - Add compute_attributes support

### LINKED ISSUE #222: Add compute_attributes support

bagasse:
* Fixes #222
* Add compute attributes support
* Tests improvements

### DIFF
--- a/module_utils/ansible_nailgun_cement.py
+++ b/module_utils/ansible_nailgun_cement.py
@@ -41,6 +41,7 @@ from nailgun.entities import (
     AbstractComputeResource,
     OSDefaultTemplate,
     ComputeProfile,
+    ComputeAttribute,
 )
 from nailgun import entity_mixins, entity_fields
 
@@ -474,6 +475,13 @@ def find_compute_resource(module, name, failsafe=False):
 def find_compute_profile(module, name, failsafe=False):
     compute_profile = ComputeProfile(name=name).search(set(), {'search': 'name="{}"'.format(name)})
     return handle_find_response(module, compute_profile, message="No compute profile found for %s" % name, failsafe=failsafe)
+
+
+def find_compute_attribute(module, compute_profile_name, compute_resource_name, failsafe=False):
+    search = {'search': 'compute_profile="{}" and compute_resource="{}"'.format(compute_profile_name, compute_resource_name)}
+    compute_attribute = ComputeAttribute().search(set(), search)
+    return handle_find_response(module, compute_attribute, message="No compute attribute found for %s %s" % (compute_profile_name, compute_resource_name),
+                                failsafe=failsafe)
 
 
 def find_domains(module, names, failsafe=False):
--- a/modules/foreman_compute_profile.py
+++ b/modules/foreman_compute_profile.py
@@ -1,6 +1,7 @@
 #!/usr/bin/python
 # -*- coding: utf-8 -*-
 # (c) Philipp Joos 2017
+# (c) Baptiste Agasse 2019
 #
 # This file is part of Ansible
 #
@@ -24,7 +25,9 @@ short_description: Manage Foreman Compute Profiles using Foreman API
 description:
   - Create and delete Foreman Compute Profiles using Foreman API
 version_added: "2.0"
-author: "Philipp Joos (@philippj)"
+author:
+  - "Philipp Joos (@philippj)"
+  - "Baptiste Agasse (@bagasse)"
 requirements:
   - "nailgun >= 0.28.0"
   - "python >= 2.6"
@@ -35,6 +38,9 @@ options:
     required: true
   updated_name:
     description: new compute profile name
+    required: false
+  compute_attributes:
+    description: Compute attributes related to this compute profile. Some of these attributes are specific to the underlying compute resource type
     required: false
   server_url:
     description: foreman url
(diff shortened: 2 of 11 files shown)


modules/foreman_compute_profile.py
```diff
@@ -82,38 +165,76 @@
...
+        if key not in compute_attribute_dict:
+            module.fail_json(msg='compute_attribute must have %s.' % key)
 
-    if state == 'present' and updated_name:
-        data['name'] = updated_name
+    try:
+        compute_attribute = find_compute_attribute(module, compute_profile_name=compute_profile.name,
+                                                   compute_resource_name=compute_attribute_dict['compute_resource'], failsafe=True)
+        compute_attribute_dict['compute_resource'] = find_compute_resource(module, name=compute_attribute_dict['compute_resource'], failsafe=False)
+    except Exception as e:
```
mdellweg: Why do `try` - `except` here, when you could just do `failsafe=False` (which is the default iirc) ?

modules/foreman_compute_profile.py
```diff
@@ -82,38 +165,76 @@
...
+    updated_name = compute_profile_dict.get('updated_name')
+    compute_attributes = compute_profile_dict.pop('compute_attributes', list())
+
+    module.connect()
+
+    try:
+        # Try to find the compute_profile to work on
+        compute_profile = find_compute_profile(module, name=compute_profile_dict['name'], failsafe=True)
+    except Exception as e:
+        module.fail_json(msg='Failed to find entity: %s ' % e)
```
mdellweg: Again: `Failsafe=False` ?

modules/foreman_compute_profile.py
```diff
@@ -82,38 +165,76 @@
...
+
+    if state == 'present' and updated_name:
+        compute_profile_dict['name'] = updated_name
+
+    compute_profile_dict = sanitize_entity_dict(compute_profile_dict, name_map)
+
+    (changed, compute_profile) = naildown_entity(ComputeProfile, compute_profile_dict, compute_profile, state, module)
+
+    # Apply changes on underlying compute attributes only when present
```
mdellweg: Can we propagate an error when specifying attributes while deleting the compute-resource?

Compare: https://github.com/theforeman/foreman-ansible-modules/blob/3c1042e038213f5ee9270e2c369d69a811fcef77/modules/foreman_provisioning_template.py#L351

modules/foreman_compute_profile.py
bagasse: It's only to catch any other exceptions  than "find failed for entity"  that can happen when we search an entity. failsafe only take care of returning `None` if we don't find any entity instead of failing.

In this case, if we cannot find compute attribute, we will create it, but if we don't find compute resource, then it's an error and we want to stop execution now (failsafe=false). I think the `try - except` is needed here, but yes as `failsafe=False` by default, I can remove it if you want.

modules/foreman_compute_profile.py
bagasse: Again :)

modules/foreman_compute_profile.py
bagasse: I didn't read the provisioning template module before. I will change the code to reflect the same behavior.

modules/foreman_compute_profile.py
mdellweg: OK, i do not know what 'every other exception' we might expect here, but i see that pattern all over the place in the other modules. So i am fine with that here.
We should however think about whether we consider those exceptions worth handling at all, or DRYing them into the find_* methods. But that should not block this PR.
@evgeni , @sean797 , @manuelbonk?

modules/foreman_compute_profile.py
mdellweg: I am not strongly against `failsafe=False` but leaving it would emphasize the `True` case and be consistent with the other modules.

modules/foreman_compute_profile.py
mdellweg: You added this (?): https://github.com/theforeman/foreman-ansible-modules/pull/243/files#diff-a08444d75f83dc9982ac920da6704a3aR212
That's fine!

module_utils/ansible_nailgun_cement.py
```diff
@@ -476,6 +477,13 @@ def find_compute_profile(module, name, failsafe=False):
     return handle_find_response(module, compute_profile, message="No compute profile found for %s" % name, failsafe=failsafe)
 
 
+def find_compute_attribute(module, compute_profile_name, compute_resource_name, failsafe=False):
+    search = {'search': 'compute_profile="{}" and compute_resource="{}"'.format(compute_profile_name, compute_resource_name)}
+    compute_attribute = ComputeAttribute().search(set(), search)
```
sean797: Can you change this to `ComputeAttribute(compute_profile=compute_profile, compute_resource=compute_resource).search()` Your just need to pass in the entities and not just the names.

modules/foreman_compute_profile.py
evgeni: Yeah, let's not block here. It's a common pattern at the moment and we might DRY it up later.

module_utils/ansible_nailgun_cement.py
sean797: This would have allowed us to drop the `try except` in the module`

module_utils/ansible_nailgun_cement.py
```diff
@@ -476,6 +477,13 @@ def find_compute_profile(module, name, failsafe=False):
...
+    return handle_find_response(module, compute_attribute, message="No compute attribute found for %s %s" % (compute_profile_name, compute_resource_name),
```
sean797: Also it is possible this can return more one attribute? I think it looks like it could, a profile and resource can have many attributes, right? This will cause an error I think https://github.com/theforeman/foreman-ansible-modules/blob/master/module_utils/ansible_nailgun_cement.py#L616

evgeni:
> Since this already got merged while I was writing a review (which is fine  ) - I think the comment about `handle_find_response` is the only one worth responding to :-)

lol, does GitHub need a "someone is typing a review…" indicator?

module_utils/ansible_nailgun_cement.py
bagasse: I think this not possible because of constraints in foreman https://github.com/theforeman/foreman/blob/develop/app/models/compute_attribute.rb#L7. You can only have 1 compute_attribute per tuple compute_resource/compute_profile

module_utils/ansible_nailgun_cement.py
bagasse: I think this will add 1 request per compute_attribute to get compute_resource entity from its name before passing it to `find_compute_attribute` in the case of the compute_attribute already exist that is the nominal case if your infra don't move too much on that side and you enforce your foreman configuration on a regular basis.

bagasse:
@evgeni, @mdellweg, @sean797, as usual, thanks a lot for your reviews :)

mdellweg:
@bagasse Thank you for you valuable contributions.

Issue #185, opened 2018-09-24
What to do with Nailgun


sean797:
##### SUMMARY
This project today is extremely reliant on Nailgun for API interactions (I actually like Nailgun, but lets not make this about +1 or -1 Nailgun) I'm finding it increasingly hard to make changes in Nailgun, SatelliteQE appear, to me at least, to be focused on their own goals, which is fine, but it leaves us (or me at least) with extremely low priority for reviewing and merging PRs.

Looking at the [open PRs](https://github.com/SatelliteQE/nailgun/pulls) today I see 4 of mine and 1 from @manuelbonk all required for FAM (foreman-ansible-modules) they have been open for a significant amount of time (at least 4 weeks) most without even a review from SatelliteQE. If you look at the 2 most recently [closed PRs](https://github.com/SatelliteQE/nailgun/pulls?q=is%3Apr+is%3Aclosed) you'll see 2 raised by SatelliteQE that were both reviewed and merged within 3 days.

Note, the above are just current examples, I've been thinking about this for a while now the first straw might have been when they asked for me to copy and paste the output from my testing as quality assurance before approving a PR :confused:

**TL;DR**
I think SatelliteQE have done an amazing job at starting Nailgun, but today the priority that our PRs are treated with seems to be hindering innovation within FAM. Do we think SatelliteQE will be open to us  take on a supporting role in maintaining Nailgun? Do we even wan to? If yes, should be suggest moving it to theforeman github org?

Fobhep:
@mdellweg @akofink @ehelms @evgeni 
Is any of you aware of this https://github.com/SatelliteQE/airgun?
It's python3 - however it looks like it could maybe be something we should look at.
Especially if this is intended to replace Nailgun.

evgeni:
Airgun is for UI only, not API, so not applicable to us I think.

ehelms:
SatelliteQE is open to contributions and maintainers from within the community. Having spoken to them, their concern is ensuring that changes do not break their goals. Ultimately, we all have the same goal: a stable API wrapper. The thing I've discussed with some of their developers is two fold:

 1) Changes that we can make to Nailgun to increase stability and reliance
 2) Release strategy that meets the needs of both communities

#### Changes that we make to Nailgun to increase stability and reliance

[...]

PR #245, opened 2019-02-27
Require git version of nailgun

ares:
Without this I was getting "cannot import name File"

### DIFF
--- a/requirements.txt
+++ b/requirements.txt
@@ -1,4 +1,4 @@
-nailgun
+git+https://github.com/SatelliteQE/nailgun.git@master#egg=nailgun
 ansible>=2.3
 six
 pytz


mdellweg:
Not entirely sure, this is related to #185 .
But i would prefer, if we could 'cool' down these modules to work with an actually released version of nailgun.

sean797:
> But i would prefer, if we could 'cool' down these modules to work with an actually released version of nailgun

That would be nice, but I suspect quite a hard goal to realise.

I think the solution, at least in the short term, is to work with SatelliteQE and do another PyPi release?

mdellweg:
Another solution could be, to tag a release here, the moment we have a new nailgun version.

ehelms:
I suggested this change for @ares to open up. We test against this same version (https://github.com/theforeman/foreman-ansible-modules/blob/master/requirements-dev.txt#L4) so we should have master be against it as well.

mdellweg:
> I'm not fussed either way really. I think we need to have a discussion and nail down how we do FAM releases ;)

We should use a hammer or a nailgun. ;)
I am :+1: for releases with a version of nailgun.

evgeni:
so, given #261, should we still have this? or just go all in on `apypie`?

mdellweg:
I dont think it hurts.

ares:
Closing, let me know if you want me to reopen :-)

PR #247, opened 2019-03-07
Use consistent parameters in subnet module documentation

quba42:
This fixes what appears to be a incompletely amended copy and paste error in the `foreman_subnet` module.

I also changed the parameter names in the example to be in line with the parameters from the documentation.


quba42:
I see I will have to fix some tests...

mdellweg:
The corresponding Nailgun part changed a month ago:
https://github.com/SatelliteQE/nailgun/commit/03e604149c774a163dba66efd2f9b0ce471ed405

mdellweg:
Try just rerecording.

quba42:
Turns out I had this backwards. The underscore is needed for current nailgun (I was accidentally not using newest nailgun, ups).

I have undone the change, but this PR can still stand for the config file change.

quba42:
Let me still change the commit message.

quba42:
The commit message now reflects the changed nature of the PR.

PR #248, opened 2019-03-07
Remove local entities

mdellweg:
* JobTemplate
* TemplateInput
* OVirtComputeResource
* VMWareComputeResource

### DIFF
--- a/module_utils/ansible_nailgun_cement.py
+++ b/module_utils/ansible_nailgun_cement.py
@@ -7,29 +7,33 @@ import sys
 
 from nailgun.config import ServerConfig
 from nailgun.entities import (
-    _check_for_value,
+    AbstractComputeResource,
+    AbstractContentViewFilter,
     ActivationKey,
     Bookmark,
-    Entity,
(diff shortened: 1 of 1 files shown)


mdellweg:
Let me look, whether i can get rid of some imports, too.

sean797:
I guess technically this bumps the Nailgun version, should we both to update the version that's documented in every module?

mdellweg:
Can we bump to a version, that does not yet exist?

sean797:
I guess this is another PR blocked by Nailgun

mdellweg:
We didn't block PR's for waiting on Nailgun Releases before, only for Nailgun PR's.
But I am fine with changing the (yet unwritten) policy here towards more reliable applicability of the modules. We should then also start to test against the released version in the master branch.

sean797:
Yes, I don't really want to block this on a Nailgun release. Maybe when we do FAM releases we just bump the Nailgun version for all modules regardless?

sean797:
@mdellweg mind rebasing? Then lets just merge this

Issue #218, opened 2019-01-03
Wrong Examples for Foreman Operatingsystems


Fobhep:
##### SUMMARY

The examples for foreman operating_system do not really make sense and 
are on top simply wrong.

PR #251, opened 2019-03-12
Add meaningful examples to operating system

mdellweg:
Also fix documentation.

Fixes: #218


modules/foreman_operating_system.py
```diff
@@ -191,7 +197,7 @@ def main():
             release_name=dict(),
             description=dict(),
             family=dict(),
```
sean797: This is also required, at least according to the docs above.

mdellweg: Nice spot.

Issue #166, opened 2018-08-09
Make katello_repository_set work with label


sean797:
##### SUMMARY

currently, I must specify `name` & `product` with `katello_repository_set` module. The latest version of Katello supports searching via label instead of name and product, let's enable that instead - its more user-friendly.

```
  katello_repository_set:
    username: "admin"
    password: "changeme"
    server_url: "https://foreman.example.com"
    verify_ssl: false
    organization: "Default Organization"
    label:  rhel-7-server-extras-rpms
    state: enabled
    repositories:
      - basearch: x86_64
```

##### ISSUE TYPE
 - Feature Idea

mdellweg:
Is this implemented? If yes, the label is still missing in the Documentation.

See:
https://github.com/theforeman/foreman-ansible-modules/blob/e1931eb282791587eb902864828a32450ac02434/modules/katello_repository_set.py#L191

sean797:
Yes, thanks @mdellweg

mdellweg:
@sean797 can you take care of the missing documentation for the label parameter? I do not feed confident enough to know whose label that is.

PR #252, opened 2019-03-12
Document repository_set label param

sean797:
https://github.com/theforeman/foreman-ansible-modules/issues/166

### DIFF
--- a/modules/katello_repository_set.py
+++ b/modules/katello_repository_set.py
@@ -48,12 +48,16 @@ options:
   name:
     description:
       - Name of the repository set
-    required: true
+    required: false
     type: bool
   product:
     description:
       - Name of the parent product
-    required: true
+    required: false
+  label:
+    description:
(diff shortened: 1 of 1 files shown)


modules/katello_repository_set.py
```diff
@@ -193,6 +197,7 @@ def main():
             state=dict(default='enabled', choices=['disabled', 'enabled']),
         ),
         supports_check_mode=True,
+        required_one_of=[['label', 'name']],
```
mdellweg: Is product not required required, if label is missing?

sean797: I was surprised also, notice the first one has a product, the second doesn't https://github.com/theforeman/foreman-ansible-modules/blob/master/test/test_playbooks/repository_set.yml#L7-L15

mdellweg: I'd need to learn more about repo_sets to understand that...
Fine for merging.

PR #254, opened 2019-03-13
Add Bookmark module

Fobhep:
This is WIP - the module is "done" - however I cannot get the test running yet.
Will keep trying ;)

### DIFF
--- a/modules/foreman_bookmark.py
+++ b/modules/foreman_bookmark.py
@@ -1,0 +1,174 @@
+#!/usr/bin/python
+# -*- coding: utf-8 -*-
+# (c) 2019 Bernhard Hopfenmüller (ATIX AG)
+#
+# This file is part of Ansible
+#
+# Ansible is free software: you can redistribute it and/or modify
+# it under the terms of the GNU General Public License as published by
+# the Free Software Foundation, either version 3 of the License, or
+# (at your option) any later version.
+#
+# Ansible is distributed in the hope that it will be useful,
+# but WITHOUT ANY WARRANTY; without even the implied warranty of
+# MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+# GNU General Public License for more details.
+#
+# You should have received a copy of the GNU General Public License
+# along with Ansible.  If not, see <http://www.gnu.org/licenses/>.
+
+ANSIBLE_METADATA = {'metadata_version': '1.0',
+                    'status': ['preview'],
+                    'supported_by': 'community'}
+
+
+DOCUMENTATION = '''
+---
+module: foreman_bookmark
+short_description: Manage Foreman Bookmarks
+description:
+  - "Manage Foreman Bookmark Entities"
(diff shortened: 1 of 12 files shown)


Fobhep:
Test problem is fixed. still missing taxonomy functionality and redesign of tests

modules/foreman_bookmark.py
```diff
@@ -0,0 +1,177 @@
...
+    'public': 'public',
+    'query': 'query',
+}
+
+
+def main():
+    module = ForemanEntityAnsibleModule(
+        argument_spec=dict(
+            name=dict(required=True),
+            controller=dict(),
```
sean797: Docs say this is required?

modules/foreman_bookmark.py
```diff
@@ -0,0 +1,177 @@
...
+            public=dict(defaut='true', type='bool'),
+            query=dict(),
+            state=dict(default='present', choices=[
+                       'present_with_defaults', 'present', 'absent']),
+        ),
+        required_if=(
+            ['state', 'present', ['name', 'query', 'controller']],
+            ['state', 'present_with_defaults', ['name', 'query', 'controller']],
+            ['state', 'absent', ['name', 'controller']],
```
sean797: You can remove `name` as it required anyway. Maybe also controller, I'm not sure if that is actually required.

modules/foreman_bookmark.py
```diff
@@ -0,0 +1,177 @@
...
+#
+# Ansible is distributed in the hope that it will be useful,
+# but WITHOUT ANY WARRANTY; without even the implied warranty of
+# MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+# GNU General Public License for more details.
+#
+# You should have received a copy of the GNU General Public License
+# along with Ansible.  If not, see <http://www.gnu.org/licenses/>.
+
+from ansible.module_utils.foreman_helper import ForemanEntityAnsibleModule
```
sean797: Would you mind moving this import below? Just so all the imports are together

modules/foreman_bookmark.py
Fobhep: true, I will fix that here

modules/foreman_bookmark.py
Fobhep: You can create bookmarks with identical names. In order to be unique a bookmark needs for specification both controller and name

modules/foreman_bookmark.py
Fobhep: not at all :)

modules/foreman_bookmark.py
Fobhep: fixed

modules/foreman_bookmark.py
Fobhep: fixed

modules/foreman_bookmark.py
Fobhep: fixed

PR #255, opened 2019-03-13
Increase playbook verbosity in tests

mdellweg:
If everything goes according to plan, pytest hides the output anyway.
If something goes wrong more informtion could be helpful.

### DIFF
--- a/test/test_crud.py
+++ b/test/test_crud.py
@@ -2,6 +2,9 @@ import pytest
 import os
 import tempfile
 import json
+from ansible.utils.display import Display
+import __main__
+__main__.display = Display()
 from ansible.cli.playbook import PlaybookCLI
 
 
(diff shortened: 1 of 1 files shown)


mdellweg:
@Fobhep Sometimes you have really strange requests... I wasn't easy.

test/test_crud.py
```diff
@@ -2,6 +2,9 @@
 import os
 import tempfile
 import json
+from ansible.utils.display import Display
+import __main__
+__main__.display = Display()
```
mdellweg: I feel dirty.

Fobhep:
> @Fobhep Sometimes you have really strange requests... I wasn't easy.

Awesome job! I think my "request" per se was not strange - however the way to achieve it looks strange, yes :D

sean797:
have you got an example of this working? I tried creating a couple of failures during test but didn't get anything more verbose. Perhaps I have to actually record for it to be more verbose?

Fobhep:
for me this works - I used it while recording the bookmark module ;)

PR #259, opened 2019-03-26
modern python in sanitize_entity_dict

evgeni:


module_utils/ansible_nailgun_cement.py
```diff
@@ -73,11 +73,7 @@ def ping_server(module):
 
 
 def sanitize_entity_dict(entity_dict, name_map):
-    result = {}
-    for key, value in name_map.items():
-        if key in entity_dict:
-            result[value] = entity_dict[key]
-    return result
+    return {value: entity_dict[key] for key, value in name_map.items() if key in entity_dict}
```
mdellweg: I thought, this version of dict-comprehension was not supported by some old python, but the tests say it's ok.

evgeni: Yeah, it's not supported by 2.6, which we don't care about.

PR #260, opened 2019-04-01
Add sync plan support for product.

bagasse:
Add ability to set sync_plan on product.

### DIFF
--- a/modules/katello_product.py
+++ b/modules/katello_product.py
@@ -64,6 +64,10 @@ options:
     description:
     - Content GPG key name attached to this product
     required: false
+  sync_plan:
+    description:
+      - Sync plan name attached to this product
+    required: false
   description:
     description:
       - Possibly long descriptionto show the user in detail view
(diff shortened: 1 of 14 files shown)


modules/katello_product.py
```diff
@@ -125,6 +131,7 @@
     'organization': 'organization',
     'description': 'description',
     'gpg_key': 'gpg_key',
+    'sync_plan': 'sync_plan',
     # 'sync_plan': 'sync_plan',  # wait for Nailgun
```
sean797: Can you remove this line?

modules/katello_product.py
```diff
@@ -86,6 +90,7 @@
     server_url: "https://foreman.example.com"
     name: "Fedora"
     organization: "My Cool new Organization"
+    sync_plan: "Fedora repos sync"
```
sean797: Nitpick, could you add "with a sync plan" to the task name

bagasse:
Thanks @sean797 !

Issue #77, opened 2017-11-29
Create test playbooks


mdellweg:
- [x] foreman_architecture
- [x] foreman_auth_source_ldap
- [x] foreman_bookmark
- [x] foreman_compute_attribute
- [x] foreman_compute_profile
- [x] foreman_compute_resource
- [x] foreman_config_group
- [x] foreman_domain
- [x] foreman_environment
- [x] foreman_global_parameter
- [x] foreman_host
- [x] foreman_host_power (tested as part of foreman_host)
- [x] foreman_hostgroup
- [x] foreman_installation_medium
- [x] foreman_job_template
- [x] foreman_location
- [x] foreman_operating_system
- [x] foreman_organization
- [x] foreman_os_default_template
- [x] foreman_provisioning_template
- [x] foreman_ptable
- [x] foreman_realm
- [x] foreman_role
- [x] foreman_search_facts
- [x] foreman_setting
- [ ] foreman_setting_facts (deprecated)
- [x] foreman_snapshot
- [x] foreman_subnet
- [x] foreman_user
- [x] katello_activation_key
- [x] katello_content_credential
- [x] katello_content_view
- [x] katello_content_view_filter
- [ ] katello_content_view_publish (deprecated)
- [x] katello_content_view_version (tested as part of katello_content_view)
- [ ] katello_content_view_version_promote (deprecated)
- [x] katello_host_collection
- [x] katello_lifecycle_environment
- [x] katello_manifest
- [x] katello_product
- [x] katello_repository
- [x] katello_repository_set
- [x] katello_sync (the test is called repository_sync)
- [x] katello_sync_plan
- [x] katello_upload
- [x] redhat_manifest

claudiol:
What kind of help do you need here? I'd be happy to look at this during our shutdown.

mdellweg:
@claudiol thank you in advance.
We are looking for playbooks in `test\test_playbooks` that are self contained. They should target the features of a specific module and be able to run against an actual foreman instance. The structure would usually be something like setup -> run tests -> clean up.
For inspiration, take a look at product and organization.

mdellweg:
Take a look at #76, too, if you like.

claudiol:
Ok I have the realm playbook done.  I will add a pull request for you guys to review.

akofink:
#76 is merged. #86 is ACK'd.

mdellweg:
Since #76 is merged, the playbooks need a slightly different structure.
[...]

PR #261, opened 2019-04-02
port foreman_domain, foreman_search_facts and katello_content_credential to apypie

evgeni:
Some time ago, I've written [apypie](https://github.com/Apipie/apypie) as a Python port of `apipie-bindings`. This has the benefit that it can fetch the API definition from the running Foreman instance and thus does not need to "know" about the various resources. It's far from feature complete (or even bug-free), but I thought it never will become if it has no users, so here we are.

This PR ports `foreman_domain` to `apypie` instead of `nailgun`. While doing so, I've tried to make the `ForemanAnsibleModule` API more Pythonic. The code also contains a few `FIXME`s, which I'm still unsure about how to handle cleanly. Also, the method/variable names are inconsistent, as I've copied a lot from the `nailgun` helper, but only adapted the naming half way through.


evgeni:
@akofink @sean797 @mdellweg @manuelbonk @Fobhep @ehelms I want y'all thoughts

modules/foreman_domain.py
```diff
@@ -87,36 +86,27 @@
 RETURN = ''' # '''
 
 try:
-    from ansible.module_utils.ansible_nailgun_cement import (
-        find_domain,
-        find_locations,
```
ekohl: This, orgs and proxies do make sense to me to keep since they often show up in APIs. Perhaps there needs to a compatible apypie helper?

test/test_playbooks/fixtures/domain-3.yml
```diff
@@ -1,242 +1,193 @@
...
-      Accept: ['*/*']
-      Accept-Encoding: ['gzip, deflate']
-      Connection: [keep-alive]
-      Content-Length: ['2']
-      User-Agent: [python-requests/2.20.1]
-      content-type: [application/json]
+      Accept: [application/json;version=2]
     method: GET
-    uri: https://foreman.example.com/katello/api/v2/ping
+    uri: https://192.168.121.223/api/status
```
ekohl: I quite liked the foreman.example.com domain :)

module_utils/foreman_helper.py
```diff
@@ -51,22 +53,189 @@ def filter_module_params(self):
...
+            else:
+                result = self.show_resource(resource, result['id'])
+        return result
+
+    def find_resource_by_name(self, resource, name, failsafe=False, thin=False):
+        search = 'name="{}"'.format(name)
+        return self.find_resource(resource, search, failsafe, thin)
+
+    def find_resources(self, resource, search_list, failsafe=False, thin=False):
+        return list(map(lambda search_item: self.find_resource_by_name(resource, search_item, failsafe=failsafe, thin=thin), search_list))
```
ekohl:
```suggestion
        return [self.find_resource_by_name(resource, search_item, failsafe=failsafe, thin=thin) for search_item in search_list]
```

I also wonder if you can do some smart searching like `search = ' or '.join('({})'.format(item) for item in search_list)` and only do one search.

module_utils/foreman_helper.py
```diff
@@ -51,22 +53,189 @@ def filter_module_params(self):
...
-            return Ping().search_json()
+            return self.foremanapi.resource('home').call('status')
         except Exception as e:
             self.fail_json(msg="Failed to connect to Foreman server: %s " % e)
 
+    def show_resource(self, resource, resource_id):
+        return self.foremanapi.resource(resource).call('show', {'id': resource_id})
+
+    def find_resource(self, resource, search, failsafe=False, thin=False):
+        response = self.foremanapi.resource(resource).call('index', {'search': search, 'thin': thin})
```
sean797: One concern is we're now even more heavily dependent on scoped_search within Foreman, I think from https://community.theforeman.org/t/the-promise-of-a-stable-api/13119 it's clear that at least some of the dev team do not consider this an API. But I'm not sure what else we can do here?

Other than test against each server version and do releases with Foreman..

module_utils/foreman_helper.py
```diff
@@ -51,22 +53,189 @@ def filter_module_params(self):
...
+    def find_resources(self, resource, search_list, failsafe=False, thin=False):
+        return list(map(lambda search_item: self.find_resource_by_name(resource, search_item, failsafe=failsafe, thin=thin), search_list))
+
+    def create_resource(self, resource, hash_entry_name, entity_dict):
+        new_entity = {}
+        # FIXME: (?) entity_dict can contain other resources, when it does, translate these entries to only list their IDs
+        for key, value in entity_dict.items():
+            if isinstance(value, dict) and 'id' in value:
+                value = value['id']
+            new_entity[key] = value
```
sean797: We should probably also cover arrays?

module_utils/foreman_helper.py
```diff
@@ -51,22 +53,189 @@ def filter_module_params(self):
...
+        return changed
+
+    def ensure_resource(self, resource, entity_dict, entity, state, check_missing=None, check_type=None, force_update=None):
+        """ Ensure that a given entity has a certain state """
+        changed, changed_entity = False, entity
+
+        # FIXME: this is a hack I want to get rid of before merging
+        # the "issue" is that the resource in foreman is called "domains",
+        # but when you post data to it, the entry has to be called "domain"
+        # see https://theforeman.org/api/1.21/index.html
```
sean797: is this in translation in apidoc? I dont think it is...


Nailgun uses inflection, but I'd rather avoid another dep and inflection isn't currently packaged as RPM :-1:  https://github.com/SatelliteQE/nailgun/blob/fe1593adcabe89c46796e95df51322171f11915b/nailgun/entity_mixins.py#L278 

Sorry, i have no good solution

test/test_playbooks/fixtures/domain-3.yml
evgeni: blah :)
I'll fix the fixtures at some point, yeah.

module_utils/foreman_helper.py
```diff
@@ -51,22 +53,189 @@ def filter_module_params(self):
...
+        except Exception as e:
+            self.fail_json(msg="Failed to connect to Foreman server: %s " % e)
+
         if ping:
             self.ping()
 
     def ping(self):
         try:
-            return Ping().search_json()
+            return self.foremanapi.resource('home').call('status')
```
sean797: This is a change from `/katello/api/ping` to `api/status`. Which I'm fine with, but we should be aware..

We should probably do some logic based on if it's a foreman_* or katello_* module, but we could do that in another PR later

modules/foreman_domain.py
evgeni: I thought about that (even had them at first), but then I thought that `find_foo(…)` is not that shorter than `find_resource_by_name('foo', …)` and having that for only a few resources is also weird?

module_utils/foreman_helper.py
evgeni: Yeah, I looked at `inflection` and found it way too big for that small part we'd be actually using.

And no, it's not in the apidoc, unless I go and iterate over the possible params and "guess" (uh)

module_utils/foreman_helper.py
evgeni: Probably, didn't need it yet :)

module_utils/foreman_helper.py
evgeni: The other helper functions don't handle multiple results well, so this is why it's not been done like this yet. But for the future this is certainly a good idea.

modules/foreman_domain.py
ekohl: There is still the `thin=True` as well. Perhaps the normalization function should be factored out and do it there.

module_utils/foreman_helper.py
evgeni: Weren't we already with nailgun, as it does searches too?

module_utils/foreman_helper.py
evgeni: Yeah, we can have a `KatelloApypieAnsibleModule` that overrides this :)

module_utils/foreman_helper.py
sean797: It was mixed, i personally tried to avoid it as most of the time there is a param you can use instead. Looking at https://github.com/theforeman/foreman-ansible-modules/blob/c6ba822834153576fcc2c598cd10bbd5ce3226ae/module_utils/ansible_nailgun_cement.py sometimes we use `search(query={'search': ` and sometimes we don't.

module_utils/foreman_helper.py
```diff
@@ -51,22 +53,193 @@ def filter_module_params(self):
...
+                action, resource, str(e)))
+        return True, result
+
+    def _generate_resource_params(self, resource, action, resource_id = None, params = None):
+        resource_params = {}
+        if resource_id is not None:
+            resource_params['id'] = resource_id
+        if params is not None:
+            action_params = self.foremanapi.resource(resource).action(action).params
+            hash_params = [p.name for p in action_params if p.expected_type == 'hash']
```
evgeni: @sean797 welp, this works, is it magic?

module_utils/foreman_helper.py
```diff
@@ -93,6 +266,25 @@ def __init__(self, argument_spec, **kwargs):
...
+        super(ForemanEntityApypieAnsibleModule, self).__init__(argument_spec=args, **kwargs)
+
+    def parse_params(self):
+        module_params = super(ForemanEntityApypieAnsibleModule, self).parse_params()
+        state = module_params.pop('state')
+        return (module_params, state)
+
+
+def sanitize_entity_dict(entity_dict, name_map):
+    return {value: entity_dict[key] for key, value in name_map.items() if key in entity_dict}
```
ekohl: How about:
```suggestion
    result = {}
    for (key, (value, normalization)) in name_map.items():
        if key in entity_dict:
            if normalization:
                result[value] = module.find_resources(normalization, entity_dict[key], thin=True)
            else:
                result[value]: entity_dict[key]

    return result
```

Though it doesn't deal with the by name/id problem, it does allow dropping a lot of code from individual modules. Perhaps the mapping needs to be a bit more expressive though.

module_utils/foreman_helper.py
evgeni: But even if we don't pass a `query`, we always call `search`:
https://github.com/theforeman/foreman-ansible-modules/blob/c6ba822834153576fcc2c598cd10bbd5ce3226ae/module_utils/ansible_nailgun_cement.py#L266-L268

And this will do a scoped search in `nailgun`. I think the only possible way to avoid a scoped search is to pass an id.

sean797: Well if you don't pass a query you are not depended on scoped search parameters as you're not using them.

Regardless, I don't see this as a big issue, marking as resolved. I'm happy to see how it goes over time.

module_utils/foreman_helper.py
sean797: I think that will work for Foreman but not every API expects a hash, especially Katello, e.g https://theforeman.org/plugins/katello/3.9/api/apidoc/v2/content_views/create.html

evgeni:
This now also includes an example for Katello :dancer:

module_utils/foreman_helper.py
```diff
@@ -120,9 +124,13 @@ def find_resource(self, resource, search, failsafe=False, thin=False):
                 result = self.show_resource(resource, result['id'])
         return result
 
-    def find_resource_by_name(self, resource, name, failsafe=False, thin=False):
+    def find_resource_by_name(self, resource, name, organization_id=None, failsafe=False, thin=False):
```
sean797: We may want to allow passing in other params, for exmaple https://theforeman.org/plugins/katello/3.9/api/apidoc/v2/repository_sets/index.html requires `product_id`, I'm fairly sure there are others.

But I think we can change this at a later date when/if we require them.

mdellweg: Yea, maybe this could be done with a dict parameter or with kwargs.

mdellweg:
Is this something that can be extended easily to be using APIv3 (Graphql)?

module_utils/foreman_helper.py
evgeni: Yeah, this is something I should change before merging!
Otherwise fixing the API later is pain.

evgeni:
> Is this something that can be extended easily to be using APIv3 (Graphql)?

I don't know much about the GraphQL API, but I don't think this will be easy.

evgeni:
> > Is this something that can be extended easily to be using APIv3 (Graphql)?
> 
> I don't know much about the GraphQL API, but I don't think this will be easy.

After looking a bit closer on how GraphQL works, this will not work with our GraphQL API as such, but we should be able to create a similar groundwork for APIv3 with learnings from this and previous PRs.

module_utils/foreman_helper.py
evgeni: updated to be a dict.

module_utils/foreman_helper.py
evgeni: sanitize_entity_dict is something we also use in the nailgun based modules, so I'd prefer not to touch it just yet. but I agree, there is certainly room for further improvements

evgeni:
I think this is now at a state where we should talk about: do we want that as a direction forward, or was this a nice PoC but we stick to `nailgun`.

Let me re-iterate the pros and cons for the switch:

## Pros
* `apypie` has only one dependency: `requests`, so shipping it to users (via RPM or whatever) is much easier than `nailgun` which pulls in multiple Python libs (which are not packaged in Fedora/EPEL).
* We're using the API JSON as published by the server, thus always having the matching list of entities and their params. No need to update the library if a new entity is added or changed.
* This also means we don't really need to branch per Foreman/Katello release (although Hammer does it, so there might be cases where we still have to).

## Cons
* `apypie` is a young library, which didn't see much adoption yet, so there will be bugs and dragons
* we're losing typechecking of params, as these are not exposed in the API JSON and `apypie` has no further knowledge about them
* we're losing relationships between entities, so the modules will need to potentially pull more data themself.

## Known Bugs
* You currently have to wipe your `~/.cache/apypie/` before recording VCR tapes, as otherwise `apypie` takes the API JSON from the cache and it's not recorded, resulting in CI failures where there is no cache.

sean797:
> ## Cons
> * we're losing typechecking of params, as these are not exposed in the API JSON and `apypie` has no further knowledge about them
> * we're losing relationships between entities, so the modules will need to potentially pull more data themself.

Could we extend `apipie` and by extension `apypie` to know about these? I imagine that would take a farily long time to do, but it would be awesome if Foreman and its plugins had a 100% self docuementing API

ekohl:
Rather than extending apipie's custom format, I'd rather invest in moving to the OpenAPI spec. apipie can already output that, but it's likely we need to enhance some things for better output.

mdellweg:
Is there any heuristic to check for an outdated apypie cache?
What happens if you fuel two foremans with different versions in the same playbook?

evgeni:
@mdellweg yes. The cache is url based, so it will not clash. And apypie checks the last modified header too.

evgeni:
Given all the emoji reactions to the initial PR comment, I'd think y'all are in favor of this, but can you express that a bit more explicitly please? :)

mdellweg:
I know we had the discussion about using nailgun. So the question that comes to my mind is: Will this be adopted as a successor of nailgun and also be used in robottelo?
Thinking only about foreman-ansible-modules, i like the approach.

akofink:
The way I see it, QE is, and operates like, a different team from the Katello/Foreman developers. There's a set of projects for using with Foreman in production (plugins, cli, ansible playbooks, etc.) which Foreman devs work on and a set of repositories used to test Satellite by QE (nailgun, robottelo, airgun, etc.) that QEs work on. This is the problem with Nailgun IMO - it's a QE utility, not a tested and vetted production-use Foreman API python library. The GitHub org which these utilities live is SatelliteQE - clearly not intended for upstream use.

A secondary related issue with Nailgun is that it's "release cycle" (if you'd call it that) is tied to Satellite, not Foreman. We'd need buy in from QE to uproot the status quo and start tracking upstream releases, which happen 3X more frequently than downstream. If you didn't already guess, they are already very busy.

Lastly, there's some packaging issue with Nailgun that has been previously mentioned due to one of its dependencies (python requests? idk).

So, yes, :rocket: for a python library that is automatically built from apipie. When something changes upstream, there's no code to write or update, just an API doc to generate (and tests to re-run).

ekohl:
Nailgun is, as I understand it, much more focused at verifying the API. This library is intended to just consume what's there. Perhaps Nailgun could even use apypie underneath, but that verification still needs to happen.

ehelms:
Generally :+1: to this approach as it also feels like it will make cross-version usage more robust.

How much work do you estimate it is to convert all entities?

mdellweg:
I actually think, entities can be migrated gradually. We can track this process in a new checklist issue.

evgeni:
Ad @mdellweg's question: `nailgun` is an API client that has it's primary usecase in testing said API. As such it cannot rely on the API to be correct, but needs to "know" endpoints, entities and data structures itself. As such, `apypie` is not a replacement for `nailgun` in all use-cases and I think `robottelo` is one of these who will remain on `nailgun`.

The packaging issue @akofink mentioned is not about `requests` but `blinker_herald` and `inflection`, as these are not packaged today. It's not that much of an issue as an nuisance, as noone wants to maintain more packages than absolutely required.

Ad @ehelms question: As you can see from this PR, the diff for the actual module is quite small, but you'll need to run the tests against a live system a few times to get all the details right and I wouldn't be surprised if we find a few bugs while porting the first entities. Overall the amount of work is somehwere in the 30-60 minutes per entity, so not *that* much. And as @mdellweg wrote, this can (and should) happen gradually. (And I'd not convert *all* of them, just the ones covered by tests ;))

mdellweg:
> (And I'd not convert _all_ of them, just the ones covered by tests ;))

...which is the vast majority:
https://github.com/theforeman/foreman-ansible-modules/issues/77

(also i don't want to stop promoting this issue ;))

ehelms:
I think then my ask would be to snapshot the current pure nailgun either via establishing a version and thus release for the modules or simply branching the current state and then merging this with an eye towards a release milestone that is fully converted to apypie.

mdellweg:
I have one last concern before we can actively start to bring this forward: How would the fact module look like? I think it is the one that is fundamentally different and we should at least have something equivalent.

evgeni:
@mdellweg that's a fair ask, I'll prep the facts module too, so you'll have an example :)

evgeni:
@mdellweg there we go, I still need to re-record the VCR tapes, but it's working.

please note, this "breaks" the api, as now we have to write `settings` instead of `Setting` (nailgun uses singular names, where as Foreman does not)

modules/foreman_search_facts.py
```diff
@@ -91,28 +90,14 @@
...
-
-def nailgun_entites():
-    return list(map(lambda entity: entity.__name__, EntitySearchMixin.__subclasses__()))
+from ansible.module_utils.foreman_helper import ForemanApypieAnsibleModule
 
 
 def main():
 
-    module = ForemanAnsibleModule(
+    module = ForemanApypieAnsibleModule(
```
mdellweg: Is this name going to stick, once we ridded nailgun?

evgeni: As soon as we're back to one backend, we can switch the naming back. I don't mind either way.

mdellweg:
> please note, this "breaks" the api, as now we have to write `settings` instead of `Setting` (nailgun uses singular names, where as Foreman does not)
I think i am ok with this. It is an adequate replacement. Thanks for taking that concern seriously.

test/test_playbooks/search_facts.yml
```diff
@@ -7,23 +7,11 @@
...
-      resource: Setting
+      resource: settings
       search: name = http_proxy
       return_length: 1
   - include: tasks/search_facts.yml
     vars:
-      resource: Subnet
+      resource: subnets
       return_length: 0
-  - name: Run with invalid resource
```
evgeni: this is currently not supported with `apypie` and will need some love there.

mdellweg: OK for me.

evgeni:
@ehelms I've created a `nailgun` branch which is at the same level as current `master`: https://github.com/theforeman/foreman-ansible-modules/tree/nailgun

sean797:
> please note, this "breaks" the api, as now we have to write settings instead of Setting (nailgun uses singular names, where as Foreman does not)

I too am also fine with that, considering we are doing a new major release.

With this PR, could you add something prominent to the README so users know about that branch?

test/test_playbooks/search_facts.yml
sean797: For the sake of clarity, how will users know what resources are valid? I assume they won't today and we can address that once we have made the required changes in apypie?

evgeni: Correct. My plan is to report back the available resources in the error message from `apypie` when a non-available one is called, and then we can restore the code we had.

evgeni:
@sean797 done

module_utils/foreman_helper.py
```diff
@@ -51,22 +53,199 @@ def filter_module_params(self):
...
+                if key in fields:
+                    new_data[key] = value
+            return self._resource_action(resource, 'update', self._generate_resource_params(resource, 'update', resource_id=entity_id, params=new_data))
+        return False, result
+
+    def ensure_resource_state(self, resource, entity_dict, entity, state, check_missing=None, check_type=None, force_update=None):
+        changed, _ = self.ensure_resource(resource, entity_dict, entity, state, check_missing, check_type, force_update)
+        return changed
+
+    def ensure_resource(self, resource, entity_dict, entity, state, check_missing=None, check_type=None, force_update=None):
```
mdellweg: I can only speculate, you didn't like the name `naildown_entity` enough...

evgeni: I just didn't want the problem to look like a nail ;)

evgeni:
this should pass tests (again) and has a more or less clean history

so will merge tomorrow when back in the office :)

PR #264, opened 2019-04-09
Update katello_content_credential and katello_sync_plan example

gen2fish:
Updated the example to use the correct parameter `content_type` instead of `type`


gen2fish:
Similar update for `katello_sync_plan`

PR #267, opened 2019-05-09
Add parameters to operating system

mdellweg:


test/test_playbooks/operating_system.yml
```diff
@@ -21,4 +30,10 @@
     vars:
       operating_system_state: absent
       expected_change: false
+  - include: tasks/operating_system.yml
+    vars:
+      operating_system_state: absent
+    tags:
+      - never
```
sean797: Is this used?

test/test_playbooks/operating_system.yml
```diff
@@ -5,6 +5,8 @@
   - name: Load server config
     include_vars:
       file: server_vars.yml
+    tags:
+      - always
```
sean797: Is this used?

mdellweg: Those tags are inserted there to solve a delicate problem...
When you hack on the molude and the playbook fails half way, You might have entities left that prevent to meet the expected_change in the next run. With those tags in place, you can run the playbook with `--tags cleanup` and it will delete the stale operating system without listening for the expected change.

PR #268, opened 2019-05-15
update f_search_facts example

evgeni:


modules/foreman_search_facts.py
```diff
@@ -70,7 +70,7 @@
     search: name = http_proxy
   register: result
 - debug:
-    var: result.response[0].value
+    var: result.resources[0].value
```
mdellweg: Has it always been resources? Or did this change with a foreman version?

evgeni: it was always resources, as it's in the module code ;)

mdellweg: Lucky us. Someone is looking at this stuff from time to time. ;)

evgeni: I can neither confirm nor deny that I was staring at my `apypie` branch for 5 minutes, wondering why it did not return me any results ;)

sean797:
Thanks @evgeni, this was changed in the original PR, but I'm human and forgot to update this.

PR #275, opened 2019-05-22
Port organization module to apypie

mdellweg:


modules/foreman_organization.py
```diff
@@ -77,20 +77,22 @@
     state: present
 '''
 
-RETURN = '''# '''
+RETURN = ''' # '''
 
 try:
```
evgeni: this whole block can be removed, we don't need anything from nailgun anymore

module_utils/foreman_helper.py
```diff
@@ -244,6 +244,9 @@ def _generate_resource_params(self, resource, action, resource_id=None, params=N
                 resource_params[hash_entry_name] = params
             else:
                 resource_params.update(params)
+        if action == 'create':
```
evgeni: can you add a comment that his is a special case for `organization` which wants `name` in both places?

module_utils/foreman_helper.py
```diff
@@ -244,6 +244,9 @@ def _generate_resource_params(self, resource, action, resource_id=None, params=N
...
+            if resource == 'organizations':
```
evgeni: both if statements could be written in one with an and

mdellweg: I thought since we probably will need the same for location it is better to have a decision tree.

evgeni: looking at apidocs, locations does not have this oddity

mdellweg: fair enough

modules/foreman_organization.py
```diff
@@ -113,11 +106,14 @@ def main():
 
     module.connect()
 
-    entity = find_organization(module, name=entity_dict['name'], failsafe=True)
+    try:
+        entity = module.find_resource_by_name('organizations', name=entity_dict['name'], failsafe=True)
+    except Exception as e:
+        module.fail_json(msg='Failed to find entity: %s ' % e)
```
sean797: Shouldn't `failsafe=True` handle this?

mdellweg: I guess this catches other problems like could not find server.
Failsafe only returns None if the search result was empty.

mdellweg:
I tested with katello. But @evgeni is analyzing the api confusion atm.

modules/foreman_organization.py
sean797: Sorry, I mean should `failsafe=True` handle this? Rather than doing something like this in every module?

mdellweg: Good question. @evgeni you designed this pattern.

test/test_playbooks/fixtures/organization-0.yml
```diff
@@ -1,255 +1,262 @@
...
-      Accept-Encoding: ['gzip, deflate']
-      Connection: [keep-alive]
-      Content-Length: ['40']
-      User-Agent: [python-requests/2.19.1]
-      content-type: [application/json]
+      Accept: [application/json;version=2]
+      Cookie: [_session_id=54e1b7a2951d29868c22f3aefeeebece]
     method: GET
-    uri: https://foreman.example.com/katello/api/v2/organizations
+    uri: https://or.deploy7.dev.atix/katello/api/organizations?per_page=4294967296&search=name%3D%22Test+Organization%22&thin=False
```
mdellweg: @sean797  it uses the katello endpoint. I do not know how it decides that.

modules/foreman_organization.py
evgeni: and you did not yell at me when I posted the PR. ;)
I'll have a look/think about it.

mdellweg: In your PR it was not yet replicated code...

test/test_playbooks/fixtures/organization-0.yml
evgeni: This depends on the apidoc. On Katello, you get

```
                        "apis": [
                            {
                                "api_url": "/katello/api/organizations",
                                "deprecated": null,
                                "http_method": "GET",
                                "short_description": "List all organizations"
                            }
                        ],
```

On Foreman, `/api/organizations`.

Given it never worked with plain Foreman before this PR…

modules/foreman_organization.py
evgeni: In a way it was, because we already had that pattern in many modules:
https://github.com/theforeman/foreman-ansible-modules/blob/c45f41c191165c7f0bc8682198dcf9b7fdaeeed8/modules/foreman_environment.py#L123-L126

I like your decorator idea!

evgeni: Also `failsafe` is something else, it's basically saying "get me 0 or 1" entities instead of "give me 1 entity". Bad naming I carried over from the nailgun backend.

module_utils/foreman_helper.py
```diff
@@ -244,6 +256,9 @@ def _generate_resource_params(self, resource, action, resource_id=None, params=N
                 resource_params[hash_entry_name] = params
             else:
                 resource_params.update(params)
+        # In combination with katello, the organization entity needs its name provided also as a plain parameter
```
evgeni: Could you please reference https://projects.theforeman.org/issues/26855 here

evgeni: or, could you try https://github.com/theforeman/foreman-ansible-modules/pull/276 which should be the cleaner hack for this?

evgeni: or we have a look at https://github.com/theforeman/hammer-cli/blob/16a30c6e93cecbbf5321f402b1abb75ca428a8b4/lib/hammer_cli/apipie/options.rb#L9-L27 and see how hammer fills out the params, at it works fine there.

evgeni: yeah, that's what I did now, and it works!

https://github.com/theforeman/foreman-ansible-modules/pull/277

module_utils/foreman_helper.py
```diff
@@ -111,6 +122,7 @@ def list_resource(self, resource, search, params=None):
         params['per_page'] = 2 << 31
         return self.foremanapi.resource(resource).call('index', params)['results']
 
+    @_exception2fail_json('Failed to find entity: %s')
```
mdellweg: @sean797 like this?

modules/foreman_organization.py
mdellweg: Since this looks like bigger refactoring, i moved it into a separate pr:
https://github.com/theforeman/foreman-ansible-modules/pull/281

modules/foreman_organization.py
```diff
@@ -28,7 +28,6 @@
     - "Eric D Helms (@ehelms)"
     - "Matthias M Dellweg (@mdellweg) ATIX AG"
 requirements:
-    - "nailgun >= 0.28.0"
```
mdellweg: Almost forgot that one.

PR #276, opened 2019-05-22
"fix" the katello apidoc not to require name when creating org

evgeni:


mdellweg:
:+1: Ill try this.

evgeni:
@mdellweg I actually have a better fix, just need some :nail_care:

evgeni:
see https://github.com/theforeman/foreman-ansible-modules/pull/277

PR #277, opened 2019-05-23
improve params generation based on hammer code

evgeni:


module_utils/foreman_helper.py
```diff
@@ -236,16 +238,22 @@ def _generate_resource_params(self, resource, action, resource_id=None, params=N
...
-                raise KeyError('Could not find the right parameter name.')
-            elif len(hash_params) == 1:
-                hash_entry_name = hash_params[0]
-                resource_params[hash_entry_name] = params
-            else:
-                resource_params.update(params)
+            resource_params.update(self._method_options_for_params(action_params, params))
         return resource_params
 
+    def _method_options_for_params(self, params, options):
```
evgeni: the `params` vs `options` here is ugly. need to find a better name for them

evgeni:
So @mdellweg reported that this broke tests in `master`, which is weird.

The PR passed on Py 2.7.15: https://travis-ci.org/theforeman/foreman-ansible-modules/jobs/536168227
But essentially the same build on `master` failed: https://travis-ci.org/theforeman/foreman-ansible-modules/jobs/536174120

The interesting thing is, it didn't really fail in our testcode.

```
self = <closed file '/tmp/tmpE1bPOX/artifacts/ca293a90-a341-492f-88bb-1bfeb373eedc/command', mode 'wb' at 0x7f49eb181e40>
object = '"improve params generation based on hammer code (#277)\\n\\n* improve params generation based on hammer code\\r\\n\\r...y generation code\\r\\n\\r\\n* simple code is good code\\r\\n\\r\\n* resource_params \xe2\x86\x92 resource_payload\\r"'
    def write(self, object):
    
        """ Writes the object's contents encoded to self.stream.
        """
>       data, consumed = self.encode(object, self.errors)
E       UnicodeDecodeError: 'ascii' codec can't decode byte 0xe2 in position 257: ordinal not in range(128)
../../../virtualenv/python2.7.15/lib/python2.7/codecs.py:369: UnicodeDecodeError
```

That's an `UnicodeDecodeError` because I used a → in the commit message.

Thanks Python2!

evgeni:
The issue seems to be inside `ansible_runner` that tries to write down the current environment into a file: https://github.com/ansible/ansible-runner/blob/master/ansible_runner/runner.py#L107-L113

evgeni:
Should be fixed via https://github.com/theforeman/foreman-ansible-modules/pull/284

PR #278, opened 2019-05-23
Add Foreman Snapshot module

Manisha15:


test/vcr_python_wrapper.py
```diff
@@ -35,6 +35,21 @@ def body_json_l2_matcher(r1, r2):
...
+            r1_copy = vcr.request.Request(r1.method, r1.uri, r1.body, r1.headers)
+            r2_copy = vcr.request.Request(r2.method, r2.uri, r2.body, r2.headers)
+            body1 = json.loads(r1_copy.body.decode('utf8'))
+            body2 = json.loads(r2_copy.body.decode('utf8'))
+            body1['search'] = "name=\"test_host\""
+            body2['search'] = "name=\"test_host\""
+            r1_copy.body = json.dumps(body1)
+            r2_copy.body = json.dumps(body2)
+            return body_json_l2_matcher(r1_copy, r2_copy)
+    return body_json_l2_matcher(r1, r2)
```
mdellweg: We found a way, that the cassette replay is independent of the hostname provided here, so you can provide your own host in `server_vars.yaml`. This pattern could also help make the tests for the host module regarding it's use of host_group more robust.
Thoughts? @sean797 @evgeni @Fobhep

README.md
```diff
@@ -34,6 +34,8 @@ needs to be in your current directory when you run `ansible` or
 current ansible configuration; make sure to update the relative paths to the
 foreman-ansible-module `modules` and `module_utils` if you do so.
 
+Also make sure to update `ansible_nailgun_cement.py` in `module_utils` with the nailgun module name when developing new module.
```
mdellweg: I think, those lines are not quite up to date. With the move from nailgun to apypie they might introduce more confusion.

modules/foreman_snapshot.py
```diff
@@ -0,0 +1,174 @@
...
+    required: true
+  username:
+    description:
+      - Username on Foreman server
+    required: true
+  password:
+    description:
+      - Password for user accessing Foreman server
+    required: true
+  verify_ssl:
```
mdellweg: This has been renamed to `validate_certs`.

test/vcr_python_wrapper.py
```diff
@@ -35,6 +35,21 @@ def body_json_l2_matcher(r1, r2):
...
 
+def body_json_snapshot_matcher(r1, r2):
+    if r1.headers.get('content-type') == 'application/json' and r2.headers.get('content-type') == 'application/json':
+        if r1.path == '/api/v2/hosts' and r2.path == '/api/v2/hosts':
+            r1_copy = vcr.request.Request(r1.method, r1.uri, r1.body, r1.headers)
+            r2_copy = vcr.request.Request(r2.method, r2.uri, r2.body, r2.headers)
+            body1 = json.loads(r1_copy.body.decode('utf8'))
+            body2 = json.loads(r2_copy.body.decode('utf8'))
+            body1['search'] = "name=\"test_host\""
+            body2['search'] = "name=\"test_host\""
```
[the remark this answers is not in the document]
mdellweg: It is currently impossible to create a host in the fixtures part of the playbook. So we state, that the user should provide one and configure it in server_vars. And as everything else there, this configured value should be ignored here.
(Most of the other stuff is ignored by _not_ including the host matcher for example.)

mdellweg:
@sean797 We discussed already, we want to give apypie a shot.
> Is there something odd about the snapshot plugin?

Yes, just not _something_, _everything_...

modules/foreman_snapshot.py
```diff
@@ -0,0 +1,174 @@
...
+- name: "Create a Snapshot"
+  foreman_snapshot:
+    username: "admin"
+    password: "changeme"
+    server_url: "https://foreman.example.com"
+    name: "new_host"
+    host: host_name
+    state: present
+
+- name: "Update a Snapshot"
```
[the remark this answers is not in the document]
mdellweg: ATM the description is the only thing that can be modified.

evgeni:
closing in favor of #306

PR #282, opened 2019-05-23
Port location to apypie

mdellweg:


modules/foreman_location.py
```diff
@@ -26,7 +26,6 @@
 author:
   - "Matthias M Dellweg (@mdellweg) ATIX AG"
 requirements:
-  - "nailgun >= 0.28.0"
```
sean797: Actually, are we putting Apypie here?

mdellweg:
Is this normal?
```
[I|app|bebaa53c] Started DELETE "/api/locations/27" for 192.168.121.1 at 2019-06-03 12:06:09 +0000
[I|app|bebaa53c] Processing by Api::V2::LocationsController#destroy as JSON
[I|app|bebaa53c]   Parameters: {"apiv"=>"v2", "id"=>"27"}
[I|app|bebaa53c] Current user set to admin (admin)
[D|app|bebaa53c] Current location set to none
[D|app|bebaa53c] Current organization set to none
[I|aud|bebaa53c] Location (27) destroy event on name Test Location
[I|aud|bebaa53c] Location (27) destroy event on ignore_types ["ProvisioningTemplate", "Hostgroup"]
[I|aud|bebaa53c] Location (27) destroy event on description 
[I|aud|bebaa53c] Location (27) destroy event on label 
[I|aud|bebaa53c] Location (27) destroy event on ancestry 
[I|aud|bebaa53c] Location (27) destroy event on manifest_refreshed_at 
[D|app|bebaa53c] Body: {"error":{"message":"Cannot delete Test Location because it has nested location."}}
[I|app|bebaa53c] Completed 200 OK in 57ms (Views: 0.2ms | ActiveRecord: 14.6ms)
```

evgeni:
Given Katello does not support nested locations, we never tried that before :)

mdellweg:
Actually Katello does. And i thought with nailgun all we used was the katello API.

mdellweg:
see: https://projects.theforeman.org/issues/26937

evgeni:
Oh wait, right. Nested Orgs are not supported. I'm in vacation mode.

test/test_playbooks/fixtures/location-2.yml
```diff
@@ -1,438 +1,323 @@
...
-      x-request-id: [d80ea1bc-fa7b-4418-84a1-61434b90d79a]
-      x-runtime: ['0.050843']
+      x-request-id: [df7ba823-b208-4e98-94bb-e4e8c077343a]
+      x-runtime: ['0.011515']
       x-xss-protection: [1; mode=block]
     status: {code: 200, message: OK}
 - request:
-    body: !!python/unicode '{"location": {"parent_id": 35, "organization_ids": [28],
-      "name": "Sub Location 1"}}'
+    body: !!python/unicode '{"location": {"parent_id": 33, "name": "Sub Location 1"}}'
```
mdellweg: This was supposed to contain the organization id.

`{'parent_id': 33, 'name': 'Sub Location 1', 'organization_ids': [30]}` is what gets passed to `self._resource_action(resource, 'create', new_entity)`.
@evgeni any idea?

evgeni:
This would imply that organization_ids is not a valid parameter and thus gets dropped by the generate function.

evgeni:
And indeed, it does not: https://theforeman.org/api/1.22/apidoc/v2/locations/create.html

mdellweg:
At least in the gui, i can add multiple organizations to a location.

evgeni:
Of course you can. It's just not in the apidoc 🤷

evgeni:
So I talked to @tbrisker on IRC, and the fact that the org and location controllers don't expose the fact that they can take "the other" as a list is due to some limitation in apipie-rails.

I wonder if we should monkey-patch the apidoc here in FAM as long as the real apidoc can't give us what we need?

mdellweg:
That will take some time to figure out. But i will have a look asap.

evgeni:
you'd need something like this, but in smart:

```
self.foremanapi.apidoc['docs']['resources']['locations']['methods'][2]['params'][2]['params'].append({u'validations': [], u'name': u'organization_ids', u'show': True, u'description': u'\n<p>Organization IDs</p>\n', u'required': False, u'allow_nil': True, u'allow_blank': False, u'full_name': 
u'location[organization_ids]', u'expected_type': u'array', u'metadata': None, u'validator': u''})
```

mdellweg:
thanks @evgeni

plugins/modules/foreman_location.py
```diff
@@ -131,29 +118,49 @@ def main():
...
 
-    (entity_dict, state) = module.parse_params()
+    entity_dict = module.clean_params()
 
     module.connect()
 
-    name_or_title = entity_dict.pop('name')
-    parent = entity_dict.pop('parent', None)
+    # Hack for broken foreman api
+    _location_create = next(x for x in module.foremanapi.apidoc['docs']['resources']['locations']['methods'] if x['name'] == 'create')
```
evgeni: We'd need the same for `update` ;)

evgeni: also, I think this should be inside `foreman_helper.py` so that other modules also benefit from it.

plugins/modules/foreman_location.py
```diff
@@ -131,29 +118,49 @@ def main():
...
 
-    if 'organizations' in entity_dict:
-        entity_dict['organizations'] = find_organizations(module, entity_dict['organizations'])
+    if not module.desired_absent:
+        if 'organizations' in entity_dict:
+            entity_dict['organizations'] = module.find_resources('organizations', entity_dict['organizations'], thin=True)
 
-    entity = find_location(module, title=build_fqn(name_or_title, parent), failsafe=True)
-    entity_dict = sanitize_entity_dict(entity_dict, name_map)
+    entity = module.find_resource('locations', search='title="{}"'.format(build_fqn(name, parent)), failsafe=True)
```
evgeni: I think I'd prefer if `search='title="{}"'.format(build_fqn(name, parent))` would be an own line, and then you'd call with `search=search` as format strings inside of args are hard to read.

plugins/module_utils/foreman_helper.py
```diff
@@ -87,6 +87,32 @@ def ping(self):
 
 class ForemanApypieAnsibleModule(ForemanBaseAnsibleModule):
 
+    def _patch_location_api(self):
+        """This is a workaround for the broken taxonomies apidoc in foreman."""
+
+        _location_organiuzations_parameter = {
+            u'validations': [],
```
evgeni: not sure we need to mark all these *explicitly* as unicode

evgeni:
@sean797 mind giving this another look? I feel like I contributed too much to the "design" of the hack to properly judge it.

plugins/module_utils/foreman_helper.py
mdellweg: I'd say, it does not hurt. (Python loves _explicit_.)

evgeni:
looks good to me now, but I'd like to wait on an ACK from Sean before merging.

plugins/module_utils/foreman_helper.py
```diff
@@ -87,6 +87,32 @@ def ping(self):
 
 class ForemanApypieAnsibleModule(ForemanBaseAnsibleModule):
 
+    def _patch_location_api(self):
+        """This is a workaround for the broken taxonomies apidoc in foreman."""
```
sean797: Nitpick, can you add the link to the foreman issue number? That way we are more likely to remove this in future.

plugins/modules/foreman_location.py
```diff
@@ -26,7 +26,7 @@
 author:
   - "Matthias M Dellweg (@mdellweg) ATIX AG"
 requirements:
-  - "nailgun >= 0.28.0"
+  - "apypie >= 0.0.2"
```
sean797: I didn't think we are adding explicit version dependency here?

mdellweg: Oh, yes. This line was older...

plugins/module_utils/foreman_helper.py
mdellweg: @sean797 this should be it

PR #225, opened 2019-02-01
Add host collection support

wzzrd:
Add a module to create, read, update and delete host collections.
Touches ansible_nailgun_cement to import HostCollection from
nailgun.entities and add two functions to find host collections.

The Host Collections module itself is quite simple, as host collections
have very little relations to other objects (basically only the
organization plays a role here).

The second change in this PR adds support for adding and removing host collections from activation keys.

PR #283, opened 2019-05-23
Add a katello_host_collection module

ephracis:
Add a new module for managing host collections. The module
katello_activation_key is also modified to support associating keys with
host collections.

This replaces #225


ephracis:
@mdellweg Let me know if I'm close. I didn't manage to apply the diff so I had to manually copy-and-paste everything.

If there's anything missing let me know and I'll fix it.

modules/katello_host_collection.py
```diff
@@ -0,0 +1,161 @@
...
+# This is the only true source for names (and conversions thereof)
+name_map = {
+    'name': 'name',
+    'organization': 'organization',
+    'description': 'description',
+}
+
+
+def main():
+    module = AnsibleModule(
```
mdellweg: in module_utils/foreman_helper, there is a `KatelloEntityAnsibleModule` to inherit from. Its use removes a lot of boilerplate here.
Also, if you feel adventurous, we started very recently to migrate the modules to a new api layer. In that case you would inherit from `KatelloEntityApipieAnsibleModule`. That is not yet required.

host_collection-0.yml
```diff
@@ -0,0 +1,240 @@
+interactions:
```
mdellweg: Those files are definitely in the wrong directory. Anyway, they should be automatically generated by `make record_host_collection` in which case, they end up in the right place.

host_collection.yml
```diff
@@ -0,0 +1,40 @@
+---
```
mdellweg: I guess, this belongs into `test/test_playbooks`.

host_collection.yml
```diff
@@ -0,0 +1,40 @@
...
+      host_collection_state: present
+      expected_change: true
+  - include: tasks/host_collection.yml
+    vars:
+      host_collection_state: present
+      expected_change: false
+  - include: tasks/host_collection.yml
+    vars:
+      host_collection_state: absent
+      expected_change: true
```
mdellweg: Can you repeat the absent task with `expected_change: false`, please. We test those modules for idempotence that way.

test/test_playbooks/tasks/activation_key_copy.yml
```diff
@@ -21,7 +22,7 @@
     content_view: "{{  activation_key_content_view }}"
     subscriptions: "{{ activation_key_subscriptions }}"
     auto_attach: "{{ activation_key_auto_attach }}"
-    state: 'copied'
+    state: "{{ activation_key_state }}"
```
mdellweg: This file seems unneeded.

host_collection.yml
```diff
@@ -0,0 +1,40 @@
...
+    vars:
+      organization_state: present
+
+- hosts: tests
+  gather_facts: false
+  tasks:
+  - name: Load server config
+    include_vars:
+      file: server_vars.yml
+  - include: tasks/host_collection.yml
```
mdellweg: I think, you are missing the `test/test_playbooks/tasks/host_collection.yml` file.
It should create a host collection with the same name (as the default value) as you are using in the activation key playbook.
Once you have all those files, you need to record the tests for host_collection and activation_key.
See: https://github.com/theforeman/foreman-ansible-modules/blob/master/README.md#how-to-test-modules-in-this-repository

modules/katello_host_collection.py
ephracis: Yeah I saw a few commits recently doing that. Let's move to that!

I'll change to inherit from `KatelloEntityApipieAnsibleModule` then. Anything I need to do when doing this?

mdellweg: katello_content_credential is ported already. Take a look at that. It should in general become easier.

host_collection-0.yml
ephracis: Haha, they sure are. The `patch` command didn't work quite well for me. Sorry I missed that. :)

test/test_playbooks/tasks/activation_key_copy.yml
ephracis: Yeah I got that from the diff posted to #225. Should I revert that change and leave it as it is?

host_collection-0.yml
mdellweg: just don't bother copy them. Drop em here, they are generated later.

host_collection.yml
ephracis: Yup, I'm moving all the test files to their correct folder. Fixtures, playbooks and the missing task file.

host_collection-0.yml
ephracis: Ok, so I should remove all 3 fixture files from the commit? Just add the playbook and task file under test then?

test/test_playbooks/tasks/activation_key_copy.yml
mdellweg: Oh, i was too fast. That file _is_ needed. But i am not quite sure how it is meant to be used. In doubt leave it as is (not part of your pr...).

modules/katello_host_collection.py
ephracis: Ok, I've inherited from Apypie now instead. Let me know if I managed to get the boilerplate right.

modules/katello_host_collection.py
```diff
@@ -0,0 +1,130 @@
...
+            description=dict(),
+        ),
+        supports_check_mode=True,
+    )
+
+    (entity_dict, state) = module.parse_params()
+
+    module.connect()
+
+    entity_dict['organization'] = module.find_resource_by_name('organizations', entity_dict['organization'], thin=True)
```
mdellweg: I believe, searching the organization is already done by `Katello.*Module`.

ephracis: Ok, removing that line then.

mdellweg: Hmm, not as such. But it is added to the argument_spec for you:
https://github.com/theforeman/foreman-ansible-modules/blob/master/module_utils/foreman_helper.py#L292

ephracis: Oh, I should remove the line in the `argument_spec` then.

ephracis:
Any idea why I am getting this in CI?

```
No match for the request (<Request (GET)
https://foreman.example.com/katello/api/v2/host_collections>) was found.
Can't overwrite existing cassette ('fixtures/activation_key-0.yml') in your current 
record mode ('none').
```

host_collection-0.yml
mdellweg: Yes, drop them. They wont work with your new module anyway.
Before this pr is finished, you need to record those fixtures against a real Katello instance. Have a look at the README's How to test section.

mdellweg:
Yes, you changed the activation_key module, thereby invalidating the recorded fixtures. You need to rerecord them. If you have the test setup (as specified in the README) up an running, it is a simple `make record_activation_key`. After that you need to check the newly generated fixtures into git.

ephracis:
When I run `make record_activation_key` it seems to use `/usr/bin/python` but I have no pip for that one. Is there a way to force it to use `/usr/local/bin/python3` instead?

It complains I have no module yaml, but I do for Python 3.

mdellweg:
I can only encourage you to use a python-venv for this.
Have a look at virtualenvwrapper

ephracis:
Darn it. virtualenvwrapper wants to use `python` instead of `python3` as well. And it doesn't work with fish so I needed to revert back to bash. 😂

mdellweg:
You can also use venv without the wrapper.
Anyway, good luck!

ephracis:
Yeah I think I'm on track. Just managed to generate new activation key fixtures. I generated the host collection again as well, just in case.

ephracis:
I ran `make test` and the upload test failed with `/bin/sh: rpm: command not found` which makes sense because I'm running on a Macbook right now.

But hopefully I'm a little bit closer now. Please take a look.

ephracis:
Yay! It's all green now. Let me know if there's anything else I need to fix before this can be merged.

modules/katello_host_collection.py
```diff
@@ -0,0 +1,129 @@
...
+    state: present
+'''
+
+RETURN = ''' # '''
+
+try:
+    from ansible.module_utils.foreman_helper import (
+        sanitize_entity_dict,
+    )
+except ImportError:
```
mdellweg: There is no use for this Exception catching anymore. Can you remove it, and add KatelloEntityApypieAnsibleModule here? (Only on import statement)

modules/katello_host_collection.py
mdellweg: :+1: exactly. I thouhgt i saw something, and it was the argument_spec.

test/test_playbooks/tasks/activation_key_copy.yml
mdellweg: I meant, not adding it to your patch, and leaving the file as in master.

test/test_playbooks/activation_key.yml
```diff
@@ -16,6 +16,9 @@
   - include: tasks/repository.yml
     vars:
       repository_state: present
+  - include: tasks/host_collection.yml
+    vars:
+      host_collection_state: present
```
mdellweg: Can you please add some tests here?
immediately before deleting the activation_key, can you check that including the task with `activation_key_host_collections: omit` does not change anything (`expected_change`), while `activation_key_host_collections: []` deletes the associations (`expected_change: true`)?
Honestly, i am not quite sure, this functionality is already in the module, but i think, we want it.

modules/katello_host_collection.py
ephracis: You mean like this?

```python
from ansible.module_utils.foreman_helper import (
    sanitize_entity_dict,
    KatelloEntityApypieAnsibleModule
)
```

test/test_playbooks/activation_key.yml
ephracis: Absolutely. Adding this:

```yaml
  - include: tasks/activation_key.yml
    vars:
      activation_key_host_collections: omit
      expected_change: false
  - include: tasks/activation_key.yml
    vars:
      activation_key_host_collections: []
      expected_change: true
```

modules/katello_host_collection.py
mdellweg: Almost. I like to have a Comma even on the latest entry. It helps managing those lists line by line.

test/test_playbooks/activation_key.yml
ephracis: Sorry, not working.

> "No host collection found for omit"

ephracis: Probably because it's hardcoded to be a list in the playbook. I'll see if I can fix it.

mdellweg: Try with `...lection: "{{ omit }}"`

ephracis: I think you were right. The `activation_key` module reports changed when host collection is omitted.

ephracis:
```
TASK [debug] *******************************************************************
ok: [tests] => {
    "activation_key_host_collections": "__omit_place_holder__5f9bd43780e26a3d6e85302429661028a3df72b3"
}

TASK [Create/update katello activation key] ************************************
changed: [tests] => {
    "changed": true,
    "invocation": {
        "module_args": {
            "auto_attach": false,
            "content_overrides": [
                {
                    "label": "Test_Organization_Test_Product_Test_Repository",
                    "override": "disabled"
                }
            ],
            "content_view": "Default Organization View",
            "host_collections": null,
            "lifecycle_environment": "Library",
            "name": "Test Activation Key",
            "new_name": null,
            "organization": "Test Organization",
            "state": "present",
            "subscriptions": [
                {
                    "name": "Test Product"
                }
            ]
        }
    }
}

TASK [fail] ********************************************************************
task path: /Users/chbr/repos/work/github/foreman-ansible-modules/test/test_playbooks/tasks/activation_key.yml:33
fatal: [tests]: FAILED! => {
    "changed": false,
    "msg": "Creating/updating activation key failed! (expected_change: False)"
}
```

modules/katello_host_collection.py
ephracis: Done.

test/test_playbooks/activation_key.yml
ephracis: Ok I think I've managed to fix it. I removed it as a default and then explicitly added it into the test playbook for some of the tasks. I believe the playbook properly tests all cases now:
- Creating key
- Updating key without host collections
- Updating key with new (empty) host collections
- Deleting key

test/test_playbooks/tasks/activation_key.yml
```diff
@@ -1,4 +1,5 @@
 ---
+- debug: var=activation_key_host_collections
```
mdellweg: Please remove this debug statement again.

.gitignore
```diff
@@ -7,3 +7,4 @@ __pycache__
 *.pyo
 *.swp
 .pytest_cache/
+venv/
```
mdellweg: I think, this safe.
A newline would be nice.

test/test_playbooks/tasks/activation_key_copy.yml
```diff
@@ -2,6 +2,7 @@
 - name: "copy katello activation key"
   vars:
     activation_key_name: "Test Activation Key"
+    activation_key_state: "copied"
```
mdellweg: This is also not used.

test/test_playbooks/tasks/host_collection.yml
```diff
@@ -0,0 +1,21 @@
...
+    validate_certs: "{{ foreman_validate_certs }}"
+    name: "{{ host_collection_name }}"
+    description: "{{ host_collection_description }}"
+    organization: "{{ host_collection_organization }}"
+    state: "{{ host_collection_state }}"
+  register: result
+- fail:
+    msg: "Ensuring host collection is {{ host_collection_state }} failed! (expected_change: {{ expected_change | default('unknown') }})"
+  when: (expected_change is defined) and (result.changed != expected_change)
+...
```
mdellweg: Newline?

test/test_playbooks/tasks/activation_key_copy.yml
ephracis: Removing.

.gitignore
ephracis: Sure thing.

test/test_playbooks/tasks/activation_key.yml
ephracis: Oops. 🤦‍♂

test/test_playbooks/tasks/host_collection.yml
ephracis: Adding.

wzzrd:
> Nit picking phase. ;)
> Given the four last comments, i can say great work of both of you.
I don't think I had anything to do with this? 

> @wzzrd maybe you want to have a look, too.
Happy this is moving forward though. I can't wait for the day when this battlestation^Hthese modules will be fully functional.

PR #284, opened 2019-05-24
strip unicode from the env when running on py2

evgeni:
ansible_runner does not like unicode

### DIFF
--- a/test/test_crud.py
+++ b/test/test_crud.py
@@ -1,5 +1,6 @@
 import pytest
 import os
+import sys
 import tempfile
 import json
 import ansible_runner
@@ -36,6 +37,14 @@ MODULES = [
     'sync_plan',
     'upload',
 ]
+
+
+if sys.version_info[0] == 2:
+    for envvar in os.environ.keys():
+        try:
+            os.environ[envvar] = os.environ[envvar].decode('utf-8').encode('ascii', 'ignore')
+        except UnicodeError:
+            os.environ.pop(envvar)
 
 
 def run_playbook_vcr(module, extra_vars=None, record=False):


test/test_crud.py
```diff
@@ -38,6 +39,14 @@
 ]
 
 
+if sys.version_info[0] == 2:
+    for envvar in os.environ.keys():
+        try:
+            os.environ[envvar] = os.environ[envvar].decode('utf-8').encode('ascii', 'ignore')
+        except Exception:
```
mdellweg: What kind of Exceptions do we expect here? And why do we ignore them?
Do we not expect the rest of the test to fail if this workaround didn't work around?

evgeni: For example if the environment contains stuff that cannot be decoded as UTF8, this would raise a UnicodeDecodeError.

*technically* we don't need anything from the *outside* environment to be passed into the tests. We only really need `FAM_TEST_VCR_PARAMS_FILE` and `ANSIBLE_CONFIG` which we set ourself. So tests should pass.

I only added this instead of just nuking all of `os.environ` so that the use *can* override some `ANSIBLE_*` vars if they seem fit.

mdellweg: Now that you bring it up, nuking everything, that changes the test environment in a nonforseeable way does not sound that bad.

evgeni: Well, "yeah". But this also clears things like `PATH` and `PYTHONPATH` and breaks the tests for me (as I have ansible, apypie and friends in an venv).

evgeni: (this = `os.environ.clear()` which should be the sane thing to do in such a case)

mdellweg: And i guess, there is no easy way to preserve all_that_is_needed_to_use_venv.
We could delete the variable if it throws `UnicodeDecodeError`.

evgeni: so delete is actually what I wanted todo, as otherwise we'd preserve the broken var :see_no_evil: 

updated, now it catches any `UnicodeError` (so also `UnicodeEncodeError` and `UnicodeTranslateError`)

PR #281, opened 2019-05-23
Fail decorator

mdellweg:

Issue #280, opened 2019-05-23
Should we delete the Apypie cache in every test?


mdellweg:
Apypie modules use a cached version of the servers api specification.
This can have some unforseen sideeffects with the recorded tests. Atm, at playback the first run test will provide this specification, and all subsequent tests will use it.
Maybe we should delete that cache file at the start of every test playbook to force it beeing downloaded again. That way, the test playback will be feed the matching api definition of its own fixtures.

evgeni:
Does #292 fix this?

mdellweg:
Trying to find it in a new temporary directory is even better. ;)

PR #286, opened 2019-05-25
Add module for LDAP authentication source

ephracis:
Create a new Foreman module `auth_source_ldap` for managing LDAP
authentication.


ephracis:
Weird. Travis reports

> No match for the request (<Request (GET) https://foreman.example.com/apidoc/v2.json>) was found.

But locally it works fine. I do notice that this request is not in the fixtures but apparently this request is never attempted when I run `make test` locally.

Ideas?

mdellweg:
I do see a problem here. This api.json is cached. And you obviously had it already at the moment of recording, so it did not make it into the fixtures.
Can you add something like
```yaml
- name: Clear API cache
  file:
    name: ~/.cache/apypie
    state: absent
```
to the beginning of the test play?
@evgeni i have not seen it coming _that_ way.

module_utils/ansible_nailgun_cement.py
```diff
@@ -9,6 +9,7 @@
 from nailgun.entities import (
     AbstractComputeResource,
     AbstractContentViewFilter,
+    AuthSourceLDAP,
```
mdellweg: Oh, and i assumed, `AuthSourceLDAP` was tested _after_ `ActivationKey`.
Can you put it back in alphabethical order?
It might fix your test case, but i do not like that it depends on the order...

ephracis: Absolutely. I'm also fixing the order in `test_crud.py`.

test/test_playbooks/auth_source_ldap.yml
```diff
@@ -0,0 +1,29 @@
+---
+- hosts: tests
+  gather_facts: false
+  tasks:
+  - name: Clear API cache
+    file:
+      name: ~/.cache/apypie
+      state: absent
```
mdellweg: Well, i am sorry, this did not fix the problem. The request to download the api.json is still not in the fixture.
The reordering in test_crud did it.

ephracis: Makes sense (since it's not in any other playbook). Removing it.

mdellweg: That is not my point. It just didn't work that way. I really want the tests to be able to run independently. And to accomplish that, we must invalidate/clear the cache after setting up the fixtures and before the first test call is recorded.

ephracis: Ok, but do you want me to keep the task or remove it?

mdellweg: It's OK without it for the moment (i think, you need to rerecord however).
We need to find a general solution to this.
#280

module_utils/ansible_nailgun_cement.py
```diff
@@ -10,6 +10,7 @@
     AbstractComputeResource,
     AbstractContentViewFilter,
     ActivationKey,
+    AuthSourceLDAP,
```
mdellweg: Sorry, i completely overlooked, that you are not going to use this file at all in the module.

ephracis: Restored the file.

modules/foreman_auth_source_ldap.py
```diff
@@ -0,0 +1,212 @@
+#!/usr/bin/python
+# -*- coding: utf-8 -*-
+# (c) 2018 Markus Bucher (ATIX AG)
```
mdellweg: One other finding... This should be your name.

ephracis: Fixed.

ephracis:
I just tried it against a FreeIPA instance and it worked:

Screenshot 2019-05-27 at 18 38 20

Screenshot 2019-05-27 at 18 40 59

modules/foreman_auth_source_ldap.py
```diff
@@ -0,0 +1,212 @@
...
+            ldap_filter=dict(),
+        ),
+        supports_check_mode=True,
+    )
+
+    (entity_dict, state) = module.parse_params()
+
+    module.connect()
+
+    try:
```
evgeni: given #281 is merged, we can drop the try/except here

ephracis: Done!

evgeni:
@ephracis this looks very good, thanks!

The only thing that I think this is lacking is Organization/Location support. This does not *have* to be in this PR, but I think it'd be cool. With Org/Loc support, you can have an LDAP source in one or multiple Orgs of your installation, but e.g. exclude others.

ephracis:
> @ephracis this looks very good, thanks!
> 
> The only thing that I think this is lacking is Organization/Location support. This does not _have_ to be in this PR, but I think it'd be cool. With Org/Loc support, you can have an LDAP source in one or multiple Orgs of your installation, but e.g. exclude others.

I like cool so I added it just now. 😎

But I noticed an issue with this, see #290 for more details on that one. Not sure what the proper solution should be which is why I made it a separate issue.

modules/foreman_auth_source_ldap.py
```diff
@@ -0,0 +1,233 @@
...
+# You should have received a copy of the GNU General Public License
+# along with Ansible.  If not, see <http://www.gnu.org/licenses/>.
+
+DOCUMENTATION = '''
+---
+module: foreman_auth_source_ldap
+short_description: Manage Foreman LDAP authentication sources using Foreman API
+description:
+  - Create and Delete Foreman LDAP authentication sources using Foreman API
+version_added: "2.5"
```
sean797: Keep up ;)

modules/foreman_auth_source_ldap.py
```diff
@@ -0,0 +1,233 @@
...
+  name:
+    description: The name of the LDAP authentication source
+    required: true
+  host:
+    description: The hostname of the LDAP server
+    required: true
+  port:
+    description: The port number of the LDAP server
+    required: false
+    tyoe: int
```
sean797: type

modules/foreman_auth_source_ldap.py
```diff
@@ -0,0 +1,233 @@
...
+    default: 389
+  account:
+    description: Account name to use when accessing the LDAP server.
+    required: false
+  account_password:
+    description: Account password to use when accessing the LDAP esrver. Required when using `onthefly_register`
```
sean797: s/esrver/server/

modules/foreman_auth_source_ldap.py
```diff
@@ -0,0 +1,233 @@
...
+    host: "ldap.example.org"
+    server_url: "https://foreman.example.com"
+    locations:
+      - "Uppsala"
+    organizations:
+      - "Sweden"
+    username: "admin"
+    password: "secret"
+    validate_certs: False
+    state: present
```
sean797: In an effort to show best practice, would you mind if the exampled used TLS?

test/test_playbooks/auth_source_ldap.yml
```diff
@@ -0,0 +1,63 @@
...
+    vars:
+      location_name: "{{ item }}"
+      location_state: "present"
+    with_items: "{{ domain_locations }}"
+  - include_tasks: tasks/organization.yml
+    vars:
+      organization_name: "{{ item }}"
+      organization_state: "present"
+    with_items: "{{ domain_organizations }}"
+- hosts: tests
```
sean797: Can you add a new line between plays for readability?

test/test_playbooks/auth_source_ldap.yml
```diff
@@ -0,0 +1,63 @@
...
+    vars:
+      auth_source_ldap_state: present
+      expected_change: true
+  - include: tasks/auth_source_ldap.yml
+    vars:
+      auth_source_ldap_state: present
+      expected_change: false
+  - include: tasks/auth_source_ldap.yml
+    vars:
+      auth_source_ldap_state: absent
```
sean797: Can you add a test for updating an entity?

modules/foreman_auth_source_ldap.py
```diff
@@ -0,0 +1,233 @@
...
+    'organizations': 'organization_ids',
+}
+
+
+def main():
+    module = ForemanEntityApypieAnsibleModule(
+        argument_spec=dict(
+            name=dict(required=True),
+            host=dict(required=True),
+            port=dict(type='int'),
```
sean797: You have a default in the docs, I think we should also default it here.

modules/foreman_auth_source_ldap.py
ephracis: 2.9?

modules/foreman_auth_source_ldap.py
ephracis: Fixed

modules/foreman_auth_source_ldap.py
ephracis: Fixed

modules/foreman_auth_source_ldap.py
ephracis: Fixed

test/test_playbooks/auth_source_ldap.yml
ephracis: Fixed

test/test_playbooks/auth_source_ldap.yml
ephracis: Fixed

modules/foreman_auth_source_ldap.py
ephracis: Fixed

PR #287, opened 2019-05-26
add back usefull failure mode to search_facts

evgeni:
this allows the user to list all available resources on the server


evgeni:
needs #281, @sean797 will like that :)

evgeni:
should pass tests now

PR #288, opened 2019-05-27
Use ansible tmpdir for apypie cache

mdellweg:
I guess, this is overkill. But it tries to solve #280 .
Is there a tmpdir, that is scoped to the lifetime of the playbook run? Imho, that would be ideal, as it would guarantee a rather recent cache file without downloading it with every single call to the module.


module_utils/foreman_helper.py
```diff
@@ -88,7 +88,8 @@ def connect(self, ping=True):
                                          username=self._foremanapi_username,
                                          password=self._foremanapi_password,
                                          api_version=2,
-                                         verify_ssl=self._foremanapi_validate_certs)
+                                         verify_ssl=self._foremanapi_validate_certs,
+                                         apidoc_cache_base_dir=self.tmpdir)
```
evgeni: this only should happen when running tests, but I read it like it happens always?

mdellweg: Yes, that is probably the only reason, this is WIP.
It does not feel right to Volkswagen this directory (adding test dependent code in the production part).
Ideally, i'd love to have a tmpdir, that was valid for a complete playbook, and i would use that even in production, as it assures up to date `api.json` while minimizing downloading it.

mdellweg:
Closing in favour of #292

Issue #274, opened 2019-05-22
Port available modules to apypie


mdellweg:
- [x] foreman_architecture
- [x] foreman_auth_source_ldap
- [x] foreman_bookmark
- [x] foreman_compute_attribute
- [x] foreman_compute_profile
- [x] foreman_compute_resource
- [x] foreman_config_group
- [x] foreman_domain
- [x] foreman_environment
- [x] foreman_global_parameter
- [x] foreman_host
- [x] foreman_host_power (#450)
- [x] foreman_hostgroup
- [x] foreman_installation_medium
- [x] foreman_job_template (#440)
- [x] foreman_location
- [x] foreman_operatingsystem
- [x] foreman_organization
- [x] foreman_os_default_template
- [x] foreman_provisioning_template
- [x] foreman_ptable
- [x] foreman_realm
- [x] foreman_role
- [x] foreman_search_facts
- [x] foreman_setting
- [x] foreman_snapshot
- [x] foreman_subnet
- [x] foreman_user
- [x] katello_activation_key
- [x] katello_content_credential
- [x] katello_content_view (https://github.com/theforeman/foreman-ansible-modules/pull/405)
- [ ] katello_content_view_filter (#471)
- [x] katello_content_view_version (#445)
- [x] katello_host_collection
- [x] katello_lifecycle_environment
- [x] katello_manifest
- [x] katello_product
- [x] katello_repository
- [x] katello_repository_set
- [x] katello_sync (https://github.com/theforeman/foreman-ansible-modules/pull/424)
- [x] katello_sync_plan
- [x] katello_upload
- [x] ~~redhat_manifest~~ (not `nailgun` based)

evgeni:
I'd like to raise the question what should happen with modules that don't have tests and are thus not ported to the new API we introduced in February.

Given they are preserved them in the `nailgun` branch and I think they are kind of a blocker for the 0.1.0 milestone.

Some of them are deprecated anyways, so I'd argue we should just remove them.
For those that don't have  a replacement, we should prioritize writing tests, but I'd add a hard deadline (October?) by which all modules that are not tested are moved to `/dev/null`.

mdellweg:
I am ok with deleting them. They continue to live on the nailgun branch, and can be reintroduced at any later time. Carrying modules around that are not covered by tests is probably not worth the efford.

mdellweg:
And only 4 to close the foreman-ping issue.

evgeni:
Yes!

[...]

PR #279, opened 2019-05-23
[RFC] restructure the repository to match the latest ansible collections spec

evgeni:
this also adds ansible galaxy metadata and tests that the collection is buildable using mazer (https://github.com/ansible/mazer)

[ansible collections](https://github.com/ansible/ansible/blob/stable-2.8/changelogs/CHANGELOG-v2.8.rst#major-changes) are not well documented yet (https://github.com/ansible/ansible/issues/54420), but you can see examples at https://github.com/ansible-collections

this PR should "just" work if you only check out this repository. if you use the repository inside another project (as a submodule or something), you'll need to adjust `ansible.cfg` of your project to find the modules and module_utils in their new paths.

Issue #146, opened 2018-07-05
docuement what `state: present_with_default` does


sean797:
We need to add documentation for  `state: present_with_default`.

Maybe also consider changing the name - currently, my only suggestion is `created`.

mdellweg:
See also #30

PR #289, opened 2019-05-27
Move bookmark module to apypie

ephracis:
Refactor the foreman_bookmark module to use apypie instead of nailgun
so we can remove a lot of the boilerplate and keep the code clean.

See #274


ephracis:
I had to add support for `state: present_with_defaults` but took the liberty to also allow for `state: created` as mentioned in #146

modules/foreman_bookmark.py
```diff
@@ -156,17 +146,19 @@ def main():
...
+    (entity_dict, state) = module.parse_params()
 
     module.connect()
 
-    entity = find_bookmark(
-        module, bookmark_dict['name'], bookmark_dict['controller'], failsafe=True)
+    try:
+        search = 'name="{}",controller="{}"'.format(entity_dict['name'], entity_dict['controller'])
+        entity = module.find_resource('bookmarks', search, failsafe=True)
+    except Exception as e:
```
mdellweg: Heads up, we are about to be getting rid of that try block:
https://github.com/theforeman/foreman-ansible-modules/pull/281

mdellweg:
Seems like you are getting a hang of this...

modules/foreman_bookmark.py
ephracis: Ok, maybe I'll wait for that one to be merged then. Guess it would be best to get #279 in as well. 🙂

ephracis:
> Seems like you are getting a hang of this...

Yup, tiny steps, one at a time. 😉

modules/foreman_bookmark.py
mdellweg: Well, that one is hard to predict...

module_utils/foreman_helper.py
```diff
@@ -209,7 +209,10 @@ def ensure_resource(self, resource, entity_dict, entity, state, check_missing=No
         """ Ensure that a given entity has a certain state """
         changed, changed_entity = False, entity
 
-        if state == 'present':
+        if state == 'present_with_defaults' or state == 'created':
```
mdellweg: I guess, this will be up for discussion.
@sean797 @evgeni @Fobhep

modules/foreman_bookmark.py
```diff
@@ -32,8 +32,8 @@
 version_added: "2.4"
 author:
   - "Bernhard Hopfenmueller (@Fobhep) ATIX AG"
+  - "Christoffer Reijer (@ephracis) Basalt AB"
 requirements:
-  - "nailgun >= 0.29.0"
```
mdellweg: :tada: every time i see it.

module_utils/foreman_helper.py
sean797: I would suggest removing `created` for now and keeping these modules consistent. Perhaps we can open another PR and separately changing this across all the modules?

I think I'd prefer not to support both `present_with_defaults` & `created` but I'm happy with whatever if it's supported by others.

ephracis: Ok, I removed it.

modules/foreman_bookmark.py
evgeni: I wouldn't wait on #279 -- that can take some time, and is "easy" as it's just a `git mv`.

#281 is merged tho :)

module_utils/foreman_helper.py
```diff
@@ -209,7 +209,10 @@ def ensure_resource(self, resource, entity_dict, entity, state, check_missing=No
         """ Ensure that a given entity has a certain state """
         changed, changed_entity = False, entity
 
-        if state == 'present':
+        if state == 'present_with_defaults':
```
evgeni: Thanks! I was too lazy to add that when initially doing the Apypie backend, because I didn't *need* it ;)

modules/foreman_bookmark.py
evgeni: ACK, but probably this needs a `apypie` here now?

modules/foreman_bookmark.py
ephracis: Cool! Removing the try-catch then.

modules/foreman_bookmark.py
ephracis: Done.

evgeni:
> I had to add support for `state: present_with_defaults` but took the liberty to also allow for `state: created` as mentioned in #146

I don't see the `created` in the code?

ephracis:
> > I had to add support for `state: present_with_defaults` but took the liberty to also allow for `state: created` as mentioned in #146
> 
> I don't see the `created` in the code?

That's correct. I added it to the backend but didn't add it all the way up to the module. But I removed it all together now after the discussion. Let it be added in another PR (if we will even go with `created` as the name, it was just a suggestion from another issue).

evgeni:
ok, that makes sense, just wanted to make sure I didn't miss anything :)

Fobhep:
thanks for the refactor @ephracis :)

PR #291, opened 2019-05-28
handle sanitize inside the helper for both new and existing entities

evgeni:
this moves another line of boilerplate code inside the helper and also fixes the comparison of existing entities vs new ones.


evgeni:
the fixtures diff should not be that big…

module_utils/foreman_helper.py
```diff
@@ -257,6 +252,13 @@ def _generate_resource_payload(self, action_params, data):
 
         return resource_payload
 
+    def _flatten_value(self, value):
+        if isinstance(value, dict) and 'id' in value:
+            value = value['id']
+        elif isinstance(value, list) and value and isinstance(value[0], dict) and 'id' in value[0]:
+            value = sorted([item['id'] for item in value])
```
ekohl:
```suggestion
            value = sorted(item['id'] for item in value)
```

module_utils/foreman_helper.py
```diff
@@ -309,6 +311,7 @@ def __init__(self, argument_spec, **kwargs):
 
 
 def sanitize_entity_dict(entity_dict, name_map):
+    name_map['id'] = 'id'
```
mdellweg: I think, this is introducing a side effect: `name_map` will be modified after a call to this function. Is this a serious problem?

evgeni: I don't think so. I just want `id → id` to always be in the map.

evgeni: (and `id` should never be mapped to anything else, riiiiiight?)

ekohl: I thought this would be nice:
```
actual_map = collections.ChainMap(name_map, DEFAULT_MAPPING)
```

But turns out this Python 3 only.

mdellweg: we could keep a modified copy a an attribute to the module.

evgeni: What are we solving by keeping a copy?

mdellweg: we stop changing the dict that the user of the module provides us. Also he cannot accidentally change our copy midway. Otoh it kind of binds the module instance to a specific entity class.
In the end it comes back to my question whether this imposes a real world problem. I am ok with saying: no.

evgeni: I say no. :)

PR #292, opened 2019-05-30
create a temporary directory and place apypie's cache in there

evgeni:


evgeni:
@mdellweg this is imho cleaner than #288

test/test_playbooks/organization.yml
```diff
@@ -10,6 +10,8 @@
       organization_state: absent
 - hosts: tests
   gather_facts: false
+  environment:
+    XDG_CACHE_HOME: "{{ lookup('env', 'FAM_TEST_APYPIE_CACHE_DIR') }}"
```
evgeni: this is needed as I want to override the cache folder only for the tests hosts and one can't set env vars in the inventory :(

test/test_crud.py
```diff
@@ -81,7 +84,7 @@ def run_playbook(module, extra_vars=None, limit=None):
 
 
 @pytest.mark.parametrize('module', MODULES)
-def test_crud(module, record):
-    run = run_playbook_vcr(module, record=record)
+def test_crud(tmpdir, module, record):
```
mdellweg: Is tmpfile a pytest fixture, or do we need to configure that somewhere?

evgeni: It's a pytest fixture, yeah. https://docs.pytest.org/en/latest/tmpdir.html

We don't need to do anything. 

But I can also drop that change from this PR if you want to and we'll discuss it separately.

evgeni:
One thing that came to my mind later: we could use the same code to pre-warm the cache, so we won't have the whole apidoc.json in each fixture separately.

This would also allow us to drop the ugly step that we need to set the env var only on the tests host, not on the fixtures one.

And potentially, we could then run test with multiple versions of the apidoc. (Mind you, it won't be full multi-version foreman testing, as the replies would still be recorded once)

test/test_crud.py
mdellweg: If pytest handles tmpdirs for us, that's fine. :+1:

mdellweg:
Pre-warming seems like a good idea. We are not testing apypie here... ;)
But i think, we should provide this api.json for every playbook, since every difference there translates to a substantial difference in the api, probably causing the tests to fail.

evgeni:
I would prefer if we would not have a separate apidoc for every module/playbook. But I could live with a generic foreman, a generic katello one, and then specialized if the module needs another plugin (say snapshot ;))
With that we could also run the foreman modules against plain foreman and foreman+X (katello sometimes does interesting things).

I will be mostly unavailable the next 2 weeks, so if it's OK for you, I'd take that as the next step and a separate PR.

evgeni:
Really, I can't reopen from FastHub? Grrr.

Could someone please reopen, I just wanted to retrigger Travis as the apypie PR was merged

evgeni:
Needs a rebase anyways. Well, later then.

evgeni:
rebased and included the role module

evgeni:
Thanks. Opened #297 to track the improvements we discussed above. Merging as-is.

PR #293, opened 2019-05-30
Add module foreman_role

ephracis:
Add a module for managing user roles in Foreman.


sean797:
Once I have created a role, how do I assign permissions to that role using the API? Maybe I'm blind, but I cannot see it in APIDOC https://theforeman.org/api/1.21/index.html

ephracis:
> Once I have created a role, how do I assign permissions to that role using the API? Maybe I'm blind, but I cannot see it in APIDOC https://theforeman.org/api/1.21/index.html

Good question. I don't know. Maybe we should extend the Foreman API with support for assigning permissions to roles. it would be fairly simple to then add that to this module I think.

modules/foreman_role.py
```diff
@@ -0,0 +1,134 @@
...
+    description: domain presence
+    default: present
+    choices: ["present", "absent"]
+'''
+
+EXAMPLES = '''
+- name: role
+  foreman_role:
+    name: "test"
+    description: "test role"
```
sean797: Can we use more meaningful data here? Something like "Provisioner" "Only provision on libvirt" or something like that

sean797:
Okay, I took a proper look this time, it's https://theforeman.org/api/1.21/apidoc/v2/filters.html (despite the terrible name)

I agree, I'd like to avoid a `foreman_role` & `foreman_filter` module and changes to both resources can be made using this module. We do this is other places like Content View Filters & Activation keys.

ephracis:
> I agree, I'd like to avoid a foreman_role ...

So... close this PR? Or should I fix your comments and proceed to get this merged?

sean797:
No, I was just saying we shouldn't have a `foreman_role` AND a `foreman_filter` module. I'm happy with this PR :+1:

modules/foreman_role.py
```diff
@@ -0,0 +1,134 @@
...
+  password:
+    description: foreman user password
+    required: true
+  validate_certs:
+    aliases: [ verify_ssl ]
+    description: verify ssl connection when communicating with foreman
+    default: true
+    type: bool
+  state:
+    description: domain presence
```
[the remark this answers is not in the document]
ephracis: Sorry about that, darn copy-paste. 🙂

Fixed.

modules/foreman_role.py
ephracis: Absolutely. Fixed!

Fodoj:
@sean797 @ephracis what are the plans to allow permissions/filters management? Right now both user groups and filters are missing, so just configuring the role is not much of a benefit from automation perspective. Usergroups are being worked on here https://github.com/theforeman/foreman-ansible-modules/pull/516/files but I couldn't find anything for filters. I'd be happy to contribute, but would be good to know what is the alternative to introducing dedicated foreman_filter module.

Fodoj:
Ok, I started working on it, will try to submit PR in next days.
Am 16. Okt. 2019, 16:01 +0200 schrieb sean797 <notifications@github.com>:

PR #294, opened 2019-05-30
Port foreman_environment to Apypie

ephracis:
Port the environment module to the new Apypie boilerplate. Until we
get #291 merged this means that we cannot modify associated orgs and
locs which is why those tests are commented out for now.

This is part of #274


evgeni:
291 is IN! :)

ephracis:
Travis fails on `test_crud[domain]` even though I haven't touched it.

mdellweg:
@ephracis Yes, i think i experienced the same. I think i do not like to _just_ add the `domain-0.yaml` fixture. I would commit the complete rerecorded playbook.

ephracis:
> @ephracis Yes, i think i experienced the same. I think i do not like to _just_ add the `domain-0.yaml` fixture. I would commit the complete rerecorded playbook.

Unfortunately I never get the initial call to `/apidoc/v2.json` even when I delete the cache. So if I do a rerecord and push it, it will fail. No matter what I do.

But it seems to be working fine to just copy it from another fixture.

mdellweg:
Did you rebase on top of #291 ? The use of a tmpdir there should do exactly that.

ephracis:
Yes I did and I did a rerecord och both environment and domain but still no such call. Perhaps there's something wrong with my setup (since it used to work before, as long as I cleared the cache). Though when I run `make test` it will attempt to do the request and I will get a failure since the `make record_xxx` did not properly record the request.

Very strange. I'll have to look into it later when I have the time. Or just nuke the setup and build a new from scratch perhaps.

mdellweg:
Thanks for looking into this. Since i also already started to try to rerecord those (they are failing on master too), i will give it a shot.

evgeni:
Maybe I broke it, as master is red too :(

mdellweg:
I can confirm, that the first fixture does not contain the apidoc file.

mdellweg:
@ephracis Tests on master should be fixed. Can you rebase?

mdellweg:
https://travis-ci.org/theforeman/foreman-ansible-modules/jobs/541674163#L439

That directory does not really have the random name i'd expect from a tmpdir. Can you make sure it is empty?

evgeni:
It should be empty as it's freshly created by pytest. But yes the name is not really random.

ephracis:
Still not getting any proper apypie requests in the `-0` fixture. Is there anything I need to do besides rebasing?

mdellweg:
I think, i know... You need to do set the XDG_..._CACHE_SOMETHING environment variable. Look at the new activationkey playbook.

mdellweg:
```diff
diff --git a/test/test_playbooks/environment.yml b/test/test_playbooks/environment.yml
index bec0328..f4d529e 100644
--- a/test/test_playbooks/environment.yml
+++ b/test/test_playbooks/environment.yml
@@ -14,6 +14,8 @@
 
 - hosts: tests
   gather_facts: false
+  environment:
+    XDG_CACHE_HOME: "{{ lookup('env', 'FAM_TEST_APYPIE_CACHE_DIR') }}"
   tasks:
   - name: 'Load server config'
     include_vars:
```

evgeni:
That looks good, yeah. We probably should write a short porting guide.

mdellweg:
I guess, we wanted to change that anyway with the cache prewarming.

ephracis:
That seems to have done the trick. Now I ran both a rerecord and a test locally and it worked fine. Hope it works in Travis as well.

plugins/modules/foreman_environment.py
```diff
@@ -120,22 +106,15 @@ def main():
 
     module.connect()
 
-    try:
-        entity = find_environment(module, name=entity_dict['name'], failsafe=True)
-    except Exception as e:
-        module.fail_json(msg='Failed to find entity: %s ' % e)
+    entity = module.find_resource_by_name('environments', name=entity_dict['name'], failsafe=True)
 
-    # Set Locations of partition table
```
mdellweg: :+1:

plugins/modules/foreman_environment.py
```diff
@@ -78,36 +79,21 @@
...
-        Location,
-        Organization,
-    )
-except ImportError:
-    pass
-
-from ansible.module_utils.foreman_helper import ForemanEntityAnsibleModule
-
+from ansible.module_utils.foreman_helper import (
+    sanitize_entity_dict,
```
mdellweg: You are not using this anymore.

plugins/modules/foreman_environment.py
```diff
@@ -120,22 +106,15 @@ def main():
...
-    except Exception as e:
-        module.fail_json(msg='Failed to find entity: %s ' % e)
+    entity = module.find_resource_by_name('environments', name=entity_dict['name'], failsafe=True)
 
-    # Set Locations of partition table
     if 'locations' in entity_dict:
-        entity_dict['locations'] = find_locations(module, entity_dict['locations'])
+        entity_dict['locations'] = module.find_resources('locations', entity_dict['locations'], thin=True)
 
-    # Set Organizations of partition table
```
mdellweg: Those have still hung around?

mdellweg:
somehow, your api.json still slips through. The first request in `fixtures/*-0.yml` must be the one to `apidoc/v2.json`.

ephracis:
> somehow, your api.json still slips through. The first request in `fixtures/*-0.yml` must be the one to `apidoc/v2.json`.

Yeah, I just saw that. :(

ephracis:
My cache path is `/Users/chbr/.cache/apypie/https___foreman.example.com/v2/default.json` and the cache file is created during the fixtures (when creating the org). So when it gets to the first environment call the cache file is present and the request is never made.

Note that I am running on a Mac. Maybe that would be a reason why it's not working for me?

ephracis:
or that I have an old version of apypie installed...

mdellweg:
And the environment variable is trying to move the cache for the tests host into to pytest tmpdir. Maybe that's what does not work in your environment.
Can you debug this by adding a module.fail_json with the env-var?
If nothing else, maybe running it inside a linux vm works?

ephracis:
My api.py has not mention of the XDG variable. So it's ignored. I compared to what's on GitHub and it's different.

I had to uninstall it to make `make test-setup` install the correct version. Now I get a proper fixture file with the apidoc request.

mdellweg:
`pip install --upgrade apypie` should have done the trick, too.

evgeni:
Not sure upgrade works with git urls if the version doesn't change.

mdellweg:
2 Acks. Let's get this one off your back. ;)

mdellweg:
@ephracis Thank you for so much patience.

PR #296, opened 2019-05-30
Create foreman_user module

### LINKED ISSUE #200: Create foreman_user module
##### SUMMARY

I went trough all existing modules but haven't found a way how to create or manage a foreman user.
I'd expect that it would be best to create separate module for this.
Similiar functionality provides [os_user](https://docs.ansible.com/ansible/2.6/modules/os_user_module.html#os-user-module) module for OpenStack.
##### ISSUE TYPE
 - Feature Idea

ephracis:
Create a new module for managing users. In order to change password we
need to force the `password` field to be sent during POST but that
field is not included when doing GET, so we put it in `check_missing`.

I noticed, though, that the value for `check_missing` was a variable
from a separate loop which seems wrong. Instead I change it to get the
value from the incoming desired dict. This makes the module behave as
we want.

To ensure that the new user works we do an API request as the
newly created user. We do this again after changing password to ensure it worked.

Closes #200


modules/foreman_user.py
```diff
@@ -0,0 +1,434 @@
...
+    'auth_source': 'auth_source_id',
+    'timezone': 'timezone',
+    'locale': 'locale',
+    'roles': 'role_ids',
+    'locations': 'location_ids',
+    'organizations': 'organization_ids',
+}
+
+# List of allowed timezones
+timezone_list = [
```
sean797: Can we pull these from apypie? It's should be exposed via https://github.com/Apipie/apypie/blob/master/apypie/param.py#L19

ephracis: That would be awesome. I just need to figure out how to do that.

ephracis: First of all, I don't understand Apypie well enough to know how to fetch that data from the server.

Secondly, as far as I can tell when checking `/apidoc/v2.json` is that the `validator` field just contains a human description and I'm not sure if that is stable enough for us to parse it and use it for validation of the module params.

Perhaps we could replace the hardcoded list in a later PR?

sean797: I was just about to comment something similar. We're going to need changes to apypie. :+1:

module_utils/foreman_helper.py
```diff
@@ -200,7 +200,7 @@ def update_resource(self, resource, old_entity, entity_dict, check_missing, chec
         if check_missing is not None:
             for key in check_missing:
                 if key in entity_dict and key not in volatile_entity.keys():
-                    volatile_entity[key] = new_value
+                    volatile_entity[key] = entity_dict[key]
```
evgeni: oups. :)
Good catch.

mdellweg:
This probably also needs a rebase and the XDG_CACHE_HOME thing

ephracis:
> This probably also needs a rebase and the XDG_CACHE_HOME thing

Done!

plugins/modules/foreman_user.py
```diff
@@ -0,0 +1,432 @@
...
+  foreman_user:
+    name: test
+    user_password: newp@ss
+
+'''
+
+RETURN = ''' # '''
+
+from ansible.module_utils.foreman_helper import (
+    sanitize_entity_dict,
```
mdellweg: you know ...

modules/foreman_user.py
evgeni: Can you please open an issue against apypie? The validators can use some love indeed ;)

ephracis: https://github.com/Apipie/apypie/issues/27

plugins/modules/foreman_user.py
ephracis: Yeah I know... ;)

evgeni:
Argh, seems my client didn't post the actual question.

Does the `if mail not in entity_dict` part work also when `entity` is `None` as when the user is freshly created? Or is `mail` mandatory for new creations?

evgeni:
(it is from the description, wonder if we can make that more explicit in the code)

ephracis:
> (it is from the description, wonder if we can make that more explicit in the code)

Yeah I was thinking that myself.  But it's kinda hard to do since we need to first check if the user exists before we know whether some params are required or not.

So I think the best and most boring solution would be to just document the requirement and let the error from Foreman bleed through the module. Maybe in the future we could improve it further.

plugins/modules/foreman_user.py
```diff
@@ -0,0 +1,431 @@
...
+    'auth_source': 'auth_source_id',
+    'timezone': 'timezone',
+    'locale': 'locale',
+    'roles': 'role_ids',
+    'locations': 'location_ids',
+    'organizations': 'organization_ids',
+}
+
+# List of allowed timezones
+timezone_list = [
```
mdellweg: Are these lists (until they are generated) better suited to live somewhere in module_utils?

ephracis: If we would use them in multiple modules they would be. But I'm not sure we will do that anytime soon. Do we want to stuff a lot of into the utils that is only used in a single module?

I vote for keeping stuff in modules, and then move into module_utils when it needs to be shared between two or more modules. What do you guys think?

test/test_playbooks/user.yml
```diff
@@ -0,0 +1,97 @@
+---
+- hosts: fixtures
+  gather_facts: false
+  tasks:
+  - name: Load server config
+    include_vars:
```
mdellweg: I guess, you need to adjust that now.
(Directory changed...)

mdellweg:
@ephracis Thank you

PR #298, opened 2019-06-03
WIP/RFC: Use 'vars_files' and 'module_defaults' to DRY test_playbooks.

mdellweg:
Given the current implementation with the tasks files, the benefit is small, but i wanted to show this new finding around anyway.

### DIFF
--- a/ansible.cfg
+++ b/ansible.cfg
@@ -3,3 +3,4 @@ library = plugins/modules
 module_utils = plugins/module_utils
 inventory = test/inventory/hosts
 retry_files_enabled = False
+module_defaults = module_defaults.yml
--- a/module_defaults.yml
+++ b/module_defaults.yml
@@ -1,0 +1,8 @@
+---
+version: '1.0'
+groupings:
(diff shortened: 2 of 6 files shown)


mdellweg:
I tend to say, the test failures are unrelated.

mdellweg:
I decided to split this pr in two parts. The `vars_files' version has benefits on its own.

https://github.com/theforeman/foreman-ansible-modules/pull/302

mdellweg:
Maybe not the best way, but the first one i came up with:
https://github.com/ansible/ansible/pull/58070

evgeni:
@mdellweg what's the state of this? I see the referenced ansible PR was closed as "we'll do something different" without saying what will be done.

Is this PR valuable as it is -- if so, please rebase it.
Otherwise I'd tend to closing it, or at least marking as "needs other PR" and stash it somewhere deep.

mdellweg:
I think, no. I would really have loved to learn, _what_ they are planning there. Should we all nudge them for a better answer?

PR #299, opened 2019-06-04
Repare tests

mdellweg:
* Rerecord domain
* Ignore smart_proxy name in domain test playback
* Also move domain_dns_proxy variable to server_vars.yml as it is volatile.

### DIFF
--- a/test/test_playbooks/domain_vars.yml
+++ b/test/test_playbooks/domain_vars.yml
@@ -1,6 +1,5 @@
 ---
-domain_name: "example.org"
-domain_dns_proxy: "centos7-devel.example.com"
+domain_name: "example.com"
 domain_locations:
   - 'Test Location'
 domain_organizations:
--- a/test/vcr_python_wrapper.py
+++ b/test/vcr_python_wrapper.py
@@ -35,6 +35,12 @@ def body_json_l2_matcher(r1, r2):
         return r1.body == r2.body
 
 
+def domain_query_matcher(r1, r2):
+    if r1.path == '/api/smart_proxies' and r2.path == '/api/smart_proxies':
+        return [q for q in r1.query if q[0] != 'search'] == [q for q in r2.query if q[0] != 'search']
+    return r1.query == r2.query
+
+
 VCR_PARAMS_FILE = os.environ.get('FAM_TEST_VCR_PARAMS_FILE')
 
 # Remove the name of the wrapper from argv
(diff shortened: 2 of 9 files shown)


test/test_playbooks/domain_vars.yml
```diff
@@ -1,6 +1,5 @@
 ---
-domain_name: "example.org"
-domain_dns_proxy: "centos7-devel.example.com"
+domain_name: "test-domain.org"
```
evgeni: I'd prefer not to use a non-example domain here.

test/vcr_python_wrapper.py
```diff
@@ -35,6 +35,12 @@ def body_json_l2_matcher(r1, r2):
         return r1.body == r2.body
 
 
+def domain_query_matcher(r1, r2):
+    if r1.path == '/api/smart_proxies' and r2.path == '/api/smart_proxies':
+        return [q for q in r1.query if q[0] != 'search'] == [q for q in r2.query if q[0] != 'search']
```
evgeni: Should we use sorted() here to ignore the param order?

mdellweg: according to vcrpy docs it is ordered

> query: The parsed query string of the request. Sorted list of name, value pairs.

https://vcrpy.readthedocs.io/en/latest/advanced.html?highlight=query#advanced-features

test/test_playbooks/domain_vars.yml
mdellweg: OK. I used that to make sure, it did not fail the matching.

test/vcr_python_wrapper.py
evgeni: Cool. If vcr does that for us, we're good

test/test_playbooks/domain_vars.yml
evgeni: Right. We have a few possibilities here (see https://tools.ietf.org/html/rfc6761), but we shouldn't use a domain that could be owned by somebody else.

mdellweg: Did i? Anyway, changing it back.

test/test_playbooks/domain_vars.yml
```diff
@@ -1,6 +1,5 @@
 ---
-domain_name: "example.org"
-domain_dns_proxy: "centos7-devel.example.com"
+domain_name: "example.com"
```
evgeni: I think we used the dot org in the past to differ more explicitly from a default forklift setup (which uses `$(hostname -s).example.com` as the primary domain name). But I'm fine with either.

PR #301, opened 2019-06-05
Modify vcr_python_wrapper for host test

Manisha15:
Edited new matcher to ignore hostgroup for host test.

### DIFF
--- a/test/vcr_python_wrapper.py
+++ b/test/vcr_python_wrapper.py
@@ -48,8 +48,8 @@ def host_body_matcher(r1, r2):
             r2_copy = vcr.request.Request(r2.method, r2.uri, r2.body, r2.headers)
             body1 = json.loads(r1_copy.body.decode('utf8'))
             body2 = json.loads(r2_copy.body.decode('utf8'))
-            body1['search'] = "name='test_group'"
-            body2['search'] = "name='test_group'"
+            body1['search'] = 'name="test_group"'
+            body2['search'] = 'name="test_group"'
             r1_copy.body = json.dumps(body1)
             r2_copy.body = json.dumps(body2)
             return body_json_l2_matcher(r1_copy, r2_copy)
--- a/test/vcr_python_wrapper.py
+++ b/test/vcr_python_wrapper.py
@@ -41,6 +41,21 @@ def domain_query_matcher(r1, r2):
     return r1.query == r2.query
 
 
+def host_body_matcher(r1, r2):
+    if r1.headers.get('content-type') == r2.headers.get('content-type') == 'application/json':
+        if r1.path == r2.path == '/api/v2/hostgroups':
+            r1_copy = vcr.request.Request(r1.method, r1.uri, r1.body, r1.headers)
+            r2_copy = vcr.request.Request(r2.method, r2.uri, r2.body, r2.headers)
+            body1 = json.loads(r1_copy.body.decode('utf8'))
+            body2 = json.loads(r2_copy.body.decode('utf8'))
+            body1['search'] = "name='test_group'"
+            body2['search'] = "name='test_group'"
+            r1_copy.body = json.dumps(body1)
+            r2_copy.body = json.dumps(body2)
+            return body_json_l2_matcher(r1_copy, r2_copy)
+    return body_json_l2_matcher(r1, r2)
+
+
 VCR_PARAMS_FILE = os.environ.get('FAM_TEST_VCR_PARAMS_FILE')
 
 # Remove the name of the wrapper from argv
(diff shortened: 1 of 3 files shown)


mdellweg:
You already know, i kind of like this specific mocking of the requests. But i think you also need to move the host_group variable in the test playbook

https://github.com/theforeman/foreman-ansible-modules/blob/569409f13d428bfbd50972bedd1990289f3d9f6e/test/test_playbooks/tasks/host.yml#L5

into the `server_vars.yml{,.example}`.

test/vcr_python_wrapper.py
```diff
@@ -41,6 +41,21 @@ def domain_query_matcher(r1, r2):
     return r1.query == r2.query
 
 
+def body_json_host_matcher(r1, r2):
+    if r1.headers.get('content-type') == 'application/json' and r2.headers.get('content-type') == 'application/json':
+        if r1.path == '/api/v2/hostgroups' and r2.path == '/api/v2/hostgroups':
```
evgeni: This can be written as `r1.path == r2.path == '/api/v2/hostgroups'` which is shorter and IMHO more readable.

test/vcr_python_wrapper.py
```diff
@@ -41,6 +41,21 @@ def domain_query_matcher(r1, r2):
...
+            r1_copy = vcr.request.Request(r1.method, r1.uri, r1.body, r1.headers)
+            r2_copy = vcr.request.Request(r2.method, r2.uri, r2.body, r2.headers)
+            body1 = json.loads(r1_copy.body.decode('utf8'))
+            body2 = json.loads(r2_copy.body.decode('utf8'))
+            body1['search'] = "name=\"test_group\""
```
evgeni: If you use single quotes for the string, there is no need to escape the double quotes inside it.

test/test_playbooks/server_vars.yml.example
```diff
@@ -8,6 +8,12 @@ foreman_validate_certs: false
 # Parameter for domain test
 domain_dns_proxy: "centos7-devel.example.com"
 
+# Parameter for snapshot test
+snapshot_host_name: "test_host"
```
evgeni: Does the snapshot hostname belong into this PR?

test/vcr_python_wrapper.py
```diff
@@ -41,6 +41,21 @@ def domain_query_matcher(r1, r2):
     return r1.query == r2.query
 
 
+def body_json_host_matcher(r1, r2):
```
evgeni: Technically, it's a hostgroup matcher, right?

test/test_playbooks/server_vars.yml.example
mdellweg: Probably not. ;)

test/vcr_python_wrapper.py
mdellweg: But it's only _used_ for the host tests.

evgeni: How about `body_json_ignore_hostgroup_name_in_search_matcher`

Because it's really neither matching hosts nor hostgroups.

mdellweg: `ignore_hostgroup_name_body_matcher` ?
Because it is still a `body_matcher`?

evgeni: Or that, yeah

mdellweg: I think,we should call it the `host_body_matcher` to be in line with the `domain_query_matcher`

Manisha15: @mdellweg @evgeni This is ready for review.

test/vcr_python_wrapper.py
```diff
@@ -41,6 +41,21 @@ def domain_query_matcher(r1, r2):
...
 
 
+def host_body_matcher(r1, r2):
+    if r1.headers.get('content-type') == r2.headers.get('content-type') == 'application/json':
+        if r1.path == r2.path == '/api/v2/hostgroups':
+            r1_copy = vcr.request.Request(r1.method, r1.uri, r1.body, r1.headers)
+            r2_copy = vcr.request.Request(r2.method, r2.uri, r2.body, r2.headers)
+            body1 = json.loads(r1_copy.body.decode('utf8'))
+            body2 = json.loads(r2_copy.body.decode('utf8'))
+            body1['search'] = "name='test_group'"
```
evgeni: I actually meant to use `body1['search'] = 'name="test_group"'` in my previous comment (as `"` is what our code that generates that query will use), but I guess it doesn't matter which way round we quote in the overwritten version :)

evgeni:
@Manisha15 thanks! This is good to go, unless you want to address my last comment (feel free to say "shut up evgeni, you're too picky" :))

test/vcr_python_wrapper.py
mdellweg: I would still try to produce fakes that _look right_.

mdellweg: I will just change that to be ready to merge.

PR #302, opened 2019-06-05
Use 'vars_files' in test_playbooks.

mdellweg:


mdellweg:
@ephracis care to do a _formal_ review?

PR #304, opened 2019-06-06
Thin query on absent

mdellweg:


plugins/modules/foreman_bookmark.py
```diff
@@ -145,11 +145,13 @@ def main():
...
 
     (entity_dict, state) = module.parse_params()
+    absent = state == 'absent'
 
     module.connect()
 
     search = 'name="{}",controller="{}"'.format(entity_dict['name'], entity_dict['controller'])
-    entity = module.find_resource('bookmarks', search, failsafe=True)
+    entity = module.find_resource('bookmarks', search, failsafe=True, thin=absent)
+    module.warn(repr(entity))
```
evgeni: Nope

mdellweg: fail_json debugging gets removed much more reliably. ;)

mdellweg: cleaned

mdellweg:
I could add a `module.absent` property. At that time, a `state` and possibly a `entity_dict` property could be as valuable. Throwing your idea back to the beach. :badminton:

mdellweg:
And the parse params would move into the __init__

mdellweg:
@evgeni I went a little further with this. Up to this point, the old interface is still usable.
The obvious next step seems to be to not provide state as an argument to ensure_resource, because it is already available as self.state. But that would be a breaking refactoring. And therefore probably a new PR.

plugins/modules/foreman_domain.py
```diff
@@ -111,23 +111,24 @@ def main():
...
 
     module.connect()
 
     # Try to find the Domain to work on
-    entity = module.find_resource_by_name('domains', name=domain_dict['name'], failsafe=True)
+    entity = module.find_resource_by_name('domains', name=domain_dict['name'], failsafe=True, thin=module.absent)
 
-    if 'dns_proxy' in domain_dict:
-        domain_dict['dns_proxy'] = module.find_resource_by_name('smart_proxies', domain_dict['dns_proxy'], thin=True)
+    if not module.absent:
```
ephracis: This is really nice!

plugins/modules/foreman_search_facts.py
```diff
@@ -105,7 +105,7 @@ def main():
         supports_check_mode=True,
     )
 
-    module_params = module.parse_params()
+    module_params = module.clean_params()
```
evgeni: why that rename?

plugins/module_utils/foreman_helper.py
```diff
@@ -294,10 +296,11 @@ def __init__(self, argument_spec, **kwargs):
         args.update(argument_spec)
         super(ForemanEntityApypieAnsibleModule, self).__init__(argument_spec=args, **kwargs)
 
+        self.state = self.params.pop('state')
+        self.absent = self.state == 'absent'
```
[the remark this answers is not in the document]
ephracis: Wouldn’t it be nice to have something like `self.desired.xxx` and `self.actual.xxx` where xxx could be any property such as “absent”?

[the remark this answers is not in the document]
ephracis: or `self.current` maybe 🤔

plugins/modules/foreman_bookmark.py
```diff
@@ -144,14 +144,14 @@ def main():
...
     )
 
-    (entity_dict, state) = module.parse_params()
+    entity_dict = module.clean_params()
 
     module.connect()
 
     search = 'name="{}",controller="{}"'.format(entity_dict['name'], entity_dict['controller'])
-    entity = module.find_resource('bookmarks', search, failsafe=True)
+    entity = module.find_resource('bookmarks', search, failsafe=True, thin=module.absent)
```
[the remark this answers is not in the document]
mdellweg: That should work. I find it a bit hard to conceive. But with a little bit of docstring. ;)

plugins/modules/foreman_search_facts.py
mdellweg: I do not remember. Maybe an attempt to keep the old interface intact for now.

plugins/module_utils/foreman_helper.py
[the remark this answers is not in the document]
mdellweg: In the current work flow the current state is easily encoded in the entity to be None.
Also this is ansible, where you state what you expect the final state to be.
But i think i could call it `desired_absent`. (We have `desired_sth` elsewhere...)

plugins/modules/foreman_search_facts.py
mdellweg: I guess, it's not needed.

mdellweg: Now i remember. `parse_params` returns the cleaned parameters _and_ the state for some module classes. So i needed a new function that does not get overwritten by the subclasses.

plugins/module_utils/foreman_helper.py
```diff
@@ -257,10 +258,11 @@ def __init__(self, argument_spec, **kwargs):
         args.update(argument_spec)
         super(ForemanEntityAnsibleModule, self).__init__(argument_spec=args, **kwargs)
 
+        self.state = self.params.pop('state')
+        self.absent = self.state == 'absent'
```
```suggestion
        self.desired_absent = self.state == 'absent'
```

plugins/modules/foreman_auth_source_ldap.py
```diff
@@ -209,7 +209,7 @@ def main():
 
     module.connect()
 
-    entity = module.find_resource_by_name('auth_source_ldaps', name=entity_dict['name'], failsafe=True)
+    entity = module.find_resource_by_name('auth_source_ldaps', name=entity_dict['name'], failsafe=True, thin=False)
```
evgeni: Shouldn't we rely on the default "magic" for `thin` here?

plugins/modules/foreman_user.py
```diff
@@ -394,7 +394,7 @@ def main():
     module.connect()
 
     search = 'login="{}"'.format(entity_dict['name'])
-    entity = module.find_resource('users', search, failsafe=True)
+    entity = module.find_resource('users', search, failsafe=True, thin=False)
```
evgeni: why not the default?

plugins/modules/foreman_auth_source_ldap.py
mdellweg: Only if we want to rerecord that here. And i'd like to postpone that to when refactoring all those modules that slipped in in the meantime.

plugins/modules/foreman_user.py
mdellweg: The current recorded delete task did not use thin=True.

plugins/module_utils/foreman_helper.py
mdellweg: How could that escape my sight.

plugins/modules/foreman_auth_source_ldap.py
evgeni: fair, could you add a comment to the code stating so? otherwise we'll wonder the same next time we touch the module.

plugins/modules/foreman_user.py
evgeni: same comment about comments as above then :)

plugins/module_utils/foreman_helper.py
mdellweg: Right. Those reduce your field of view...

PR #306, opened 2019-06-07
Add Foreman Snapshot Module

Manisha15:
Foreman Snapshot Module for apypie


Manisha15:
@evgeni @sean797  This is ready for review.

evgeni:
Just making sure, this PR obsoletes #278, right?

plugins/module_utils/foreman_helper.py
```diff
@@ -195,7 +210,7 @@ def update_resource(self, resource, old_entity, entity_dict, check_missing, chec
                     volatile_entity[key] = entity_dict[key]
                     fields.append(key)
         if len(fields) > 0:
-            new_data = {'id': entity_id}
+            new_data = entity
```
evgeni: with that change, we can drop the use of `new_data` entirely and just use `entity` everywhere

mdellweg:
Yes, we will close the one that does not get accepted first.

plugins/module_utils/foreman_helper.py
```diff
@@ -125,17 +125,29 @@ def find_resource(self, resource, search, params=None, failsafe=False, thin=Fals
             params = {}
         params['thin'] = thin
         results = self.list_resource(resource, search, params)
+        if resource == 'snapshots':
```
evgeni: `list_resource` will never (in it's current state) work for snapshots, right? then i'd prefer not calling it at all in that case

```python
if resource == 'snapshots':
  results = your special magic
else:
  results = self.list_resource(…)
```

plugins/module_utils/foreman_helper.py
```diff
@@ -125,17 +125,29 @@ def find_resource(self, resource, search, params=None, failsafe=False, thin=Fals
...
+            # Snapshots API does not do search
+            snapshot = []
+            data = re.search(r"name=\"(.*)\"", search)
```
evgeni: I very much dislike to use `re` to parse something we generated ourself in `find_resource_by_name`.

At this point I wonder if we should extend the `find_resource` API to take both, `name` and `search` (both defaulting to `None`) and move the `search` generation into it. Basically:

```python
def find_resource(self, resource, search='', name=None, params=None, failsafe=False, thin=False):
    if name is not None:
        search += ' name="{}"'.format(name)
    …
```

plugins/module_utils/foreman_helper.py
evgeni: oh, reading your code more closely, list works, it just gets more data than we need?

evgeni:
> Yes, we will close the one that does not get accepted first.

Yeah, I'm not going to accept new nailgun modules, so I closed the other one :)

plugins/module_utils/foreman_helper.py
evgeni: (and then probably `search.strip()` to get rid of spaces)

plugins/module_utils/foreman_helper.py
Manisha15: yes, it returns all the snapshots for particular host irrespective of search parameters.

Manisha15:
@mdellweg @evgeni I have made the suggested changes.

plugins/module_utils/foreman_helper.py
```diff
@@ -120,27 +120,38 @@ def list_resource(self, resource, search, params=None):
         params['per_page'] = 2 << 31
         return self.foremanapi.resource(resource).call('index', params)['results']
 
-    def find_resource(self, resource, search, params=None, failsafe=False, thin=False):
+    def find_resource(self, resource, search='', name=None, params=None, failsafe=False, thin=False):
+        if name is not None:
+            search += 'name="{}"'.format(name)
```
```suggestion
            search += ' name="{}"'.format(name)
```
the space is important, because if `search` already contained something, you'd end up with `the_old_searchname="blah"` otherwise

plugins/module_utils/foreman_helper.py
evgeni: weird API, but that's not your problem :)

plugins/modules/foreman_snapshot.py
```diff
@@ -0,0 +1,155 @@
...
+}
+
+
+def main():
+    module = ForemanEntityApypieAnsibleModule(
+        argument_spec=dict(
+            name=dict(required=True),
+            description=dict(),
+            host=dict(required=True),
+            host_id=dict(),
```
evgeni: why do you have `host_id` here too?

plugins/modules/foreman_snapshot.py
```diff
@@ -0,0 +1,155 @@
...
+
+
+from ansible.module_utils.foreman_helper import ForemanEntityApypieAnsibleModule
+
+
+# This is the only true source for names (and conversions thereof)
+name_map = {
+    'name': 'name',
+    'description': 'description',
+    'host_id': 'host_id',
```
evgeni: i'd prefer this to be `'host': 'host_id'` and then overriding the value in L136, otherwise `snapshot_dict` ends up with having both, `host` and `host_id`.

plugins/modules/foreman_snapshot.py
```diff
@@ -0,0 +1,155 @@
...
+
+    module.connect()
+
+    host = module.find_resource_by_name('hosts', name=snapshot_dict['host'], failsafe=False, thin=True)
+
+    snapshot_dict['host_id'] = host['id']
+
+    entity = module.find_resource_by_name('snapshots', name=snapshot_dict['name'], params={'host_id': host['id']}, failsafe=True)
+
+    if state == 'reverted':
```
evgeni: can this state also be done via `ensure_resource_state`? by passing the special dict you generate maybe?

plugins/module_utils/foreman_helper.py
Manisha15: I can add the space  but then all other modules have to be re-recorded else the new addition will fail test for them because **strip** function does not modify the String it returns a new one.

plugins/modules/foreman_snapshot.py
Manisha15: This mapping (the one you suggested ) creates problem while updating the snapshot. Unfortunately during sanitizing entity in update part of ensure_resource it can't identity host_id mapping.

evgeni: interesting. the snapshot plugin is really weird.

plugins/modules/foreman_snapshot.py
Manisha15: It is not working with ensure_resource_state. Revert action for snapshot is working weirdly. It has required parameters id and host_id but it throws HTTPError for that but if you pass additional parameter with while calling rever, say, name or description, the revert function works. In `ensure_resource_state` it uses resource payload which removes the additional parameters (here name or description) causing HTTPError.

evgeni: aaaaaaaaah ok

I'd argue this is an Apypie bug or an API bug, or whatever. Then your workaround if fine, but I'll think about how we can make it nicer in Apypie.

evgeni:
Given there are two snapshot plugins: https://github.com/theforeman/foreman_snapshot/ and  https://github.com/ATIX-AG/foreman_snapshot_management, I think the docs should mention that this module is for the latter.

mdellweg:
I will go on and add a note there. Be aware that it will set the standards for all (if there will ever be any) modules associated with certain foreman plugins.

mdellweg:
This is blocked by https://github.com/Apipie/apypie/pull/37 .

evgeni:
that PR is merged, so it's unblocked

Manisha15:
@mdellweg @evgeni This is ready for review.

plugins/modules/foreman_snapshot.py
```diff
@@ -0,0 +1,146 @@
...
+short_description: Manage Foreman Snapshots
+description:
+  - "Manage Foreman Snapshots for Host Entities"
+  - "This module can create, update, revert and delete snapshots from preexisting host"
+  - "This module requires the foreman_snapshot_management plugin set up in the server"
+  - "See: U(https://github.com/ATIX-AG/foreman_snapshot_management)"
+author:
+  - "Manisha Singhal (@Manisha15) ATIX AG"
+requirements:
+    - "apypie >= 0.0.1"
```
mdellweg: That version is surely not enough.

plugins/module_utils/foreman_helper.py
```diff
@@ -371,6 +387,21 @@ def _update_entity(self, resource, desired_entity, current_entity, params, entit
         else:
             return False, current_entity
 
+    def _revert_entity(self, resource, current_entity, params):
```
mdellweg: Do we want this in the `foreman_helper`? I think we can catch this case in the module before calling `ensure_entity`. Open for discussion.

plugins/modules/foreman_snapshot.py
```diff
@@ -0,0 +1,146 @@
...
+
+from ansible.module_utils.foreman_helper import ForemanEntityApypieAnsibleModule
+
+
+# This is the only true source for names (and conversions thereof)
+entity_spec = {
+    'id': {},
+    'name': {},
+    'description': {},
+    'host_id': {},
```
mdellweg: I think, this is not needed anymore, because params in ensure_entity is bypassing it.

tests/vcr_python_wrapper.py
```diff
@@ -81,10 +87,16 @@ def host_body_matcher(r1, r2):
...
     if test_params['test_name'] == 'domain':
         fam_vcr.register_matcher('domain_query', domain_query_matcher)
         query_matcher = 'domain_query'
+    elif test_params['test_name'] == 'snapshot':
+        fam_vcr.register_matcher('snapshot_query', snapshot_query_matcher)
+        query_matcher = 'snapshot_query'
+    else:
+        query_matcher = 'query'
+
+    fam_vcr.register_matcher('body_json_l2', body_json_l2_matcher)
```
mdellweg: This is duplicated two lines down.

plugins/module_utils/foreman_helper.py
Manisha15: @evgeni suggested it to be in ensure_resource_state. But if it is specific to snapshot only then it can be done in module.

mdellweg: If he said that... I am fine with either ways.

plugins/module_utils/foreman_helper.py
```diff
@@ -149,11 +149,10 @@ def ping(self):
...
     @_exception2fail_json('Failed to show resource: %s')
-    def show_resource(self, resource, resource_id, params=None):
-        if params is None:
-            params = {}
-        params['id'] = resource_id
-        return self.foremanapi.resource(resource).call('show', params)
+    def show_resource(self, resource, resource_dict):
+        if not isinstance(resource_dict, dict):
+            resource_dict = {'id': resource_dict}
+        return self.foremanapi.resource(resource).call('show', resource_dict)
```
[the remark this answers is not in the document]
Manisha15: Yes it is breaking the module. But some modules like foreman_installation_medium directly call this function from witihin the module so this check for resource_dict to be dict was to tackle them.

plugins/module_utils/foreman_helper.py
evgeni: I remember asking if it *can* be inside the regular ensure path. and looking at the code it cleanly fits into it, so I'm fine with it being here.

I wonder if we should guard that part by `if resource == 'snapshots'` and fail if it's not, but I don't think it's strictly necessary.

plugins/modules/foreman_snapshot.py
```diff
@@ -0,0 +1,126 @@
...
+    required: true
+  description:
+    description:
+      - Description of Snapshot
+    required: false
+  host:
+    description:
+      - Name of related Host
+    required: true
+extends_documentation_fragment: foreman
```
evgeni: We're missing the docs for the various states here.

plugins/modules/foreman_snapshot.py
```diff
@@ -0,0 +1,126 @@
...
+'''
+
+EXAMPLES = '''
+- name: "Create a Snapshot"
+  foreman_snapshot:
+    username: "admin"
+    password: "changeme"
+    server_url: "https://foreman.example.com"
+    name: "new_snapshot"
+    host: "host_name"
```
evgeni: could you use more descriptive names? maybe something like:

```yaml
name: "snapshot_before_upgrade"
host: "server.example.com"
```

plugins/modules/foreman_snapshot.py
```diff
@@ -0,0 +1,126 @@
...
+    state: absent
+'''
+
+RETURN = ''' # '''
+
+
+from ansible.module_utils.foreman_helper import ForemanEntityApypieAnsibleModule
+
+
+# This is the only true source for names (and conversions thereof)
```
evgeni: I think we can make this comment go away, especially as it's only half true now with the entity spec (it contains not only name conversions).

mdellweg: Right. It is more like a relict from nailgun times, where we tried to level some inconsistencies.
Fwiw, I dropped it in the next iteration of the entity_spec pr.

plugins/modules/foreman_snapshot.py
Manisha15: We don't really change the name of snapshot once it is created so isn't "snapshot_before_upgrade" snapshot name a bit misleading ?

mdellweg: I guess, it is meant to be "snapshot_before_major_software_upgrade".
(Not the snapshot itself is going to be updated)

evgeni: Yeah, I didn't mean to rename the snapshot, just wanted to make the snapshot name for all examples closer to the real world where you usually would not name your snapshot "new_snapshot".

Manisha15: Ok. I have modified the module with suggested changes. It is ready for review now.

mdellweg:
:tada:

PR #307, opened 2019-06-19
Add foreman compute_attribute module

Manisha15:
foreman compute_attribute for apypie


mdellweg:
Tests failing due to:
https://projects.theforeman.org/issues/27107
(or too strict apypie validation)

plugins/modules/foreman_compute_attribute.py
```diff
@@ -0,0 +1,139 @@
...
+
+    compute_attributes = compute_resource.get('compute_attributes')
+
+    compute_profile = module.find_resource_by_name('compute_profiles', name=entity_dict['compute_profile_name'], failsafe=False, thin=True)
+
+    entity_dict['compute_profile_id'] = compute_profile['id']
+
+    entity_dict['compute_resource_id'] = compute_resource['id']
+
+    entities = filter(lambda item: item.get('compute_profile_id') == compute_profile['id'], compute_attributes)
```
mdellweg: For python3, you need to cast this to a list.
Python3 often returns generators, where python2 returns lists.

bagasse:
Hi @Manisha15, did you see that compute_attributes support is already provided by compute_profile module ? Something is missing for your usecase ?

plugins/module_utils/foreman_helper.py
```diff
@@ -245,7 +245,7 @@ def _generate_resource_payload(self, action_params, data):
         resource_payload = {}
 
         for param in action_params:
-            if param.expected_type == 'hash':
+            if param.expected_type == 'hash' and param.params:
```
evgeni: That's interesting. Are there really apidocs that have `expected_type=hash` but don't provide nested params?

evgeni: (asking, because I've recently ported that function to Apypie directly, and thus would have to also port that fix)

mdellweg: Yes, there was one.
`cat default.json | jq '.docs.resources.compute_attributes.methods[0].params[]|select(.name=="compute_attribute").params'`

evgeni: found it! https://github.com/Apipie/apypie/pull/34

test/test_playbooks/compute_attribute.yml
```diff
@@ -0,0 +1,52 @@
+---
+- hosts: fixtures
+  gather_facts: false
+  vars_files:
+    - vars/server.yml
+    - vars/compute_profile.yml
+  tasks:
+  - include: tasks/compute_profile.yml
+    vars:
+      compute_profile_state: present
```
mdellweg: Do we need to prepare a compute_resource here too?

test/test_playbooks/compute_attribute.yml
```diff
@@ -0,0 +1,52 @@
...
+      expected_change: true
+
+  - include: tasks/compute_attribute.yml
+    vars:
+      vm_attrs:
+        memory_mb: '1024'
+        cpu: '1'
+      compute_attribute_state: present
+      expected_change: false
+...
```
mdellweg: Can you please add Cleanup, by deleting the `compute_{profile,resource}` again.

test/test_playbooks/vars/server.yml.example
```diff
@@ -11,6 +11,9 @@ domain_dns_proxy: "centos7-devel.example.com"
 # Parameter for host test
 host_hostgroup: "test_group"
 
+# Parameter for compute_attribute test
+compute_resource_name: "test_compute_resource"
```
mdellweg: Here i think, we _can_ do better and create our own testing instance. (See the actual test for compute_resource).

test/test_playbooks/vars/compute_profile.yml
```diff
@@ -7,7 +7,7 @@ libvirt:
     locations:
       - Test Location
     params:
-      url: qemu+ssh://root@libvirt.example.com/system
+      url:  qemu+ssh://libvirtuser@localhost/system
```
mdellweg:
```suggestion
      url: qemu+ssh://libvirtuser@localhost/system
```

test/vcr_python_wrapper.py
```diff
@@ -81,10 +81,13 @@ def host_body_matcher(r1, r2):
...
     fam_vcr = vcr.VCR()
 
-    query_matcher = 'query'
     if test_params['test_name'] == 'domain':
         fam_vcr.register_matcher('domain_query', domain_query_matcher)
         query_matcher = 'domain_query'
+    else:
+        query_matcher = 'query'
+
+    fam_vcr.register_matcher('body_json_l2', body_json_l2_matcher)
```
mdellweg: We do not need to change that file here at all.

Manisha15: I have removed the changes in vcr_python_wrapper file.

PR #308, opened 2019-06-20
Remove vaildation for katello_manifest

sean797:
I don't see any reason to be this restrictive, we can just ignore the param and allow people to write easier or more flexible roles/playbooks.

### DIFF
--- a/plugins/modules/katello_manifest.py
+++ b/plugins/modules/katello_manifest.py
@@ -53,6 +53,7 @@ options:
   manifest_path:
     description:
       - Path to the manifest zip file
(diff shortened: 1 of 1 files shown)


plugins/modules/katello_manifest.py
```diff
@@ -105,14 +106,6 @@
 from ansible.module_utils.basic import AnsibleModule
 
 
-def validate_params(module, state=None, manifest_path=None):
-    if manifest_path is not None or len(manifest_path) == 0:
```
evgeni: funfact: the code is wrong anyways, would need to check `len > 0` ;)

Issue #297, opened 2019-06-03
Pre warm apypie cache


evgeni:
One thing that came to my mind later: we could use the same code to pre-warm the cache, so we won't have the whole apidoc.json in each fixture separately.

This would also allow us to drop the ugly step that we need to set the env var only on the tests host, not on the fixtures one.

And potentially, we could then run test with multiple versions of the apidoc. (Mind you, it won't be full multi-version foreman testing, as the replies would still be recorded once)

PR #309, opened 2019-06-22
provide static copies of apidoc to tests

evgeni:
that way we don't have to include them in the recorded fixtures

Closes: #297


evgeni:
I've only re-recorded two so far to show that it (hopefully) works.

evgeni:
Hmm. Tests fail due to https://github.com/Apipie/apypie/pull/28, will investigate next week.

mdellweg:
I do see a problem here, in that every plugin can change, what interfaces are described in the apidoc. Do we want a version that is taken with all known plugins installed? Would it not be better to save one copy along with each test?

evgeni:
Tests fail because of https://projects.theforeman.org/issues/27107, boooh :D

evgeni:
@mdellweg I am torn here. So far we only really have modules for plain Foreman and Katello (yes, this will change in #306, but psssst ;)) and I'd kinda not like to have the same file multiple times in the repo.
What I could do is to provide `<module>.json` symlinks to either `foreman.json` or `katello.json` and #306 would have a real `snapshot.json` instead of the symlink?

mdellweg:
I guess, then we should wait for the snapshot PR, to see how that works out.

evgeni:
moved this to the previously proposed symlink setup now. now each module *can* but does not have to provide an own apidoc.

evgeni:
the tests fail because it seems we have foreman modules recorded against a katello instance, and then the "organization" path is different (thanks katello)

mdellweg:
I am beginning to like this.
Are we now using the preloaded default.json for recording, too? Or only if the host we record against is called `foreman.example.com`?
Are the tests failing due to subtle differences in the apis.

evgeni:
At the moment only of it's called foreman.example.com.

And the tests probably fail because they are foreman modules but recorded against a katello install.

mdellweg:
Should we record all modules against katello _and_ whenever possible foreman?
The taxonomies seem to behave quite differently in both cases.

evgeni:
I don't think recording *both* is a nice developer experience.

I think we should record Foreman modules against Foreman, Katello against Katello and other plugins (hello Snapshot!) against Foreman+Plugin (if possible).
That's what this PR does today.

There is still the outstanding "issue" with entities that are extended by a plugin (e.g. Katello extends the host and hostgroup models), but I'd leave that as an exercise for another day.

Another "issue" is that we're using the other modules to set up fixtures (like when a host needs a hostgroup), if the fixture module is using `nailgun`, we have to record against Katello anyways :( [not sure this is still the case of if we managed to convert all the base modules by now]

test/test_crud.py
```diff
@@ -74,7 +75,12 @@ def run_playbook_vcr(tmpdir, module, extra_vars=None, record=False):
 
     cache_dir = tmpdir.join('cache')
     cache_dir.ensure(dir=True)
-    os.environ['FAM_TEST_APYPIE_CACHE_DIR'] = cache_dir.strpath
+    os.environ['XDG_CACHE_HOME'] = cache_dir.strpath
+    json_cache = cache_dir / 'apypie/https___foreman.example.com/v2/default.json'
```
mdellweg: If we used the value from `vars/server.yml` to generate this path, we would use the same apidoc for recording as well as playing back. I think, that would spare us a lot of confusion.

Also we need a documented way to include a new default.json in case you are developing for a plugin.

evgeni: You are correct. I was somehow refraining from having to parse that yaml, but it doesn't really cost much.

Good point with the docs. Will add some to the new dev section in the readme.

evgeni: both done.

PR #310, opened 2019-06-24
Fixes #197 - Remove version added from documentation

### LINKED ISSUE #197: Update "version" added before we go upstream
Before we go upstream for the first time, we need to update/correct the "version added" parameter in the in-line YAML documentation

sean797:


mdellweg:
Tests failing due to:
https://projects.theforeman.org/issues/27107
(or too strict apypie validation)

Issue #256, opened 2019-03-14
How to proceed with the host module


Fobhep:
The host module is surely one of the most desired modules, which do not exist yet. Currently we have an orphaned PR for the host module, last feedback was over a year ago.
@mdellweg  and me had today a little brainstorming and I want to propose a possible way to make some progress with this matter.

- [x] Step 1 create a host module that can create, switch on/off and delete a host, that was created from an exisiting host group. Either based on the current PR or create a new one form scratch - featuring @evgeni's refactoring.

- [x] Step 2 start writing a host_group module. host groups can be "incomplete", meaning we could slowly add functionality to it until it covers a "complete" host. 

- [x] Step 3 Port the host-group functionality to the host-module

Of course your opinion @ismaelpuerto to this is highly appreciated, since you created a first PR for that.

I am not sure yet if there are any major drawbacks with that approach. However I am happy to hear your opinion @sean797 @evgeni @akofink @mdellweg. Feel free to add other people to the discussion!

mdellweg:
I want to emphasize the possibility to commit small changes in this proceeding, which hopefully speeds up the process.

evgeni:
I currently have no use for a host module, so whatever you can come up with is good for me ;)

One thing to keep in mind: the Host entity has one of the most complicated validations we have, so starting with just a few attributes for create/update is still a lot of work. Maybe start with a host that has managed=false and then add from there?

akofink:
I also have no use for this module that I can think of. I'm definitely fine with an iterative approach, starting with a minimal module.

ismaelpuerto:
Thanks for you work @Fobhep

Fobhep:
> Thanks for you work @Fobhep

:) happy to get your feedback/contributions in the next iterative amendments to host and hostgroup ;)

evgeni:
given we have a basic host module now, should this be considered done?

mdellweg:
I think we are in the middle of step 2 here.

Fobhep:
yap - I agree with @mdellweg

evgeni:
oki, thanks for the input!

mdellweg:
We should at least mention #429, #389 and #366 here.
I think, they establish that we can move to bullet-point 3 of this issue now.

evgeni:
Yepp, I agree. Wonder if most of the params we have for host and hostgroup should move to a central place as both accept a huge list of identical params?

mdellweg:
[...]

PR #311, opened 2019-06-26
Add Hostgroup Module

Manisha15:


Manisha15:
@mdellweg @evgeni This is ready for review.

mdellweg:
This is an attempt to follow the path outlined in #256

plugins/modules/foreman_hostgroup.py
```diff
@@ -0,0 +1,216 @@
...
+author:
+  - "Manisha Singhal (@Manisha15) ATIX AG"
+requirements:
+  - "apypie"
+options:
+  name:
+    description: Name of Hostgroup
+    required: true
+  description:
+    description: description of hostgroup
```
```suggestion
    description: Description of hostgroup
```

plugins/modules/foreman_hostgroup.py
```diff
@@ -0,0 +1,216 @@
...
+short_description: Manage Foreman Hostgroups using Foreman API
+description:
+  - Create, Update and Delete Foreman Hostgroups using Foreman API
+author:
+  - "Manisha Singhal (@Manisha15) ATIX AG"
+requirements:
+  - "apypie"
+options:
+  name:
+    description: Name of Hostgroup
```
```suggestion
    description: Name of hostgroup
```

plugins/modules/foreman_hostgroup.py
```diff
@@ -0,0 +1,216 @@
...
+  architecture:
+    description: Architecture name
+    required: False
+    default: None
+  domain:
+    description: Domain name
+    required: False
+    default: None
+  environment:
+    description: Environment name
```
```suggestion
    description: Puppet environment name
```

(I know, the param is called just `environment`, but it can be confusing when  you also have Katello installed which has Lifecycle Environments)

plugins/modules/foreman_hostgroup.py
```diff
@@ -0,0 +1,216 @@
...
+  password:
+    description: foreman user password
+    required: true
+  validate_certs:
+    aliases: [ verify_ssl ]
+    description: verify ssl connection when communicating with foreman
+    default: true
+    type: bool
+  state:
+    description: domain presence
```
evgeni: This is not the domain module :)

plugins/modules/foreman_hostgroup.py
```diff
@@ -0,0 +1,216 @@
...
+  foreman_hostgroup:
+    name: "new_hostgroup"
+    architecture: "architecture_name"
+    operatingsystem: "operatingsystem_name"
+    medium: "media_name"
+    ptable: "Partition_table_name"
+    server_url: "https://foreman.example.com"
+    username: "admin"
+    password: "secret"
+    validate_certs: False
```
evgeni: I'd prefer if we'd not ship examples with `validate_certs=False` (or `validate_certs` at all, the user will often copy paste from the example, and not adding params that have a sane default sounds nice to me)

plugins/modules/foreman_hostgroup.py
```diff
@@ -0,0 +1,216 @@
...
+    'operatingsystem_id': 'operatingsystem_id',
+    'medium_id': 'medium_id',
+    'ptable_id': 'ptable_id',
+    'root_pass': 'root_pass',
+}
+
+
+def main():
+    module = ForemanEntityApypieAnsibleModule(
+        argument_spec=dict(
```
evgeni: this spec seems to miss quite a few params you had defined in the documentation. I think it's totally fine to start with just those that you have defined here, but then please adjust the docs :)

plugins/modules/foreman_hostgroup.py
```diff
@@ -0,0 +1,216 @@
...
+RETURN = ''' # '''
+
+from ansible.module_utils.foreman_helper import ForemanEntityApypieAnsibleModule
+
+
+# This is the only true source for names (and conversions thereof)
+name_map = {
+    'name': 'name',
+    'description': 'description',
+    'architecture_id': 'architecture_id',
```
evgeni: historically (hah), we've not been using the `_id` notation in the module itself, so the map would contain
`'architecture': 'architecture_id',` and in the assignment below you'd have `entity_dict['architecture'] = architecture['id']` and the transformation to `architecture_id` would happen in the helpers.

I wouldn't say I'm married to that concept, but I think it's cleaner as otherwise `entity_dict` will contain both, `architecture` (with the string provided by the user) and `architecture_id` (the id you got from the API) which might look confusing.

Manisha15: If I map like you suggested there is error while updating the hostgroup. After sanitizing in update part of ensure_resource function it is not taking any parameters that are to be updated. I guess it is unable to map those parameters from entity to entity_dict.

Manisha15: I have updated the entity_dict and name_map as you suggested with slight changes as your suggested name_mapping was not working while updating the hostgroup.

Manisha15:
@mdellweg @evgeni I have made all the suggested changes. This is ready for review now.

test/test_playbooks/hostgroup.yml
```diff
@@ -0,0 +1,76 @@
...
+  gather_facts: false
+  vars_files:
+    - vars/server.yml
+  tasks:
+  - include: tasks/operating_system.yml
+    vars:
+      operating_system_state: present
+  - include: tasks/ptable.yml
+    vars:
+      ptable_state: present
```
mdellweg: I think, we also need `architecture` and `installation_medium` as fixtures here.

Manisha15: I have added installation_medium fixture now and there is no architecture module so can't add its fixtures.

mdellweg: I think, architecture is safe, because those are always preseeded in foreman.

But does it mean we should have an architecture module?

evgeni: I think they're not preseeded but created by Puppet? But either way, yeah, an architecture module is probably a nice addition. Should be fairly simple.

mdellweg: @evgeni: Just for the lolz:

https://github.com/theforeman/foreman/blob/develop/db/seeds.d/060-architectures.rb

evgeni: STFU ;)

PR #312, opened 2019-06-27
use apypie 0.0.2

evgeni:
we moved the generate params helper there, so no need to carry our copy


evgeni:
pip!!!! :rage:

plugins/module_utils/foreman_helper.py
```diff
@@ -241,17 +240,6 @@ def _resource_action(self, resource, action, params):
                 action, resource, str(e)))
```
ekohl: Unrelated: is the `str()` needed? You're formatting anyway

evgeni: probably not

PR #313, opened 2019-06-27
don't require pytz and six

evgeni:
katello_sync_plan does not need these since 10960611a1d8fc27d0a56eb3682055b89808dd2f


evgeni:
https://travis-ci.org/theforeman/foreman-ansible-modules/builds/551268030 is green, whyever that did not propagate here

PR #314, opened 2019-06-28
Adding Smart Proxy support.

jstuart:
This commit adds support for manipulating Smart Proxy resources.

Test code is included, but I have not had a chance to sort through generating the VCR data for it.  The tests have not yet been included in test_crud.py, so they don't actually execute.

If anyone has a VCR ready environment already setup and wants to generate the VCR data, or can point me at the Forklift playbook that is normally used for this, help would be greatly appreciated.

### DIFF
--- a/plugins/modules/foreman_smart_proxy.py
+++ b/plugins/modules/foreman_smart_proxy.py
@@ -1,0 +1,167 @@
+#!/usr/bin/python
+# -*- coding: utf-8 -*-
+#
+# This file is part of Ansible
+#
+# Ansible is free software: you can redistribute it and/or modify
+# it under the terms of the GNU General Public License as published by
+# the Free Software Foundation, either version 3 of the License, or
+# (at your option) any later version.
+#
+# Ansible is distributed in the hope that it will be useful,
+# but WITHOUT ANY WARRANTY; without even the implied warranty of
+# MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+# GNU General Public License for more details.
+#
+# You should have received a copy of the GNU General Public License
+# along with Ansible.  If not, see <http://www.gnu.org/licenses/>.
+
+DOCUMENTATION = '''
+---
+module: foreman_smart_proxy
+short_description: Manage Foreman Smart Proxy
+description:
+  - Manage Foreman Smart Proxy
+requirements:
+  - "nailgun >= 0.28.0"
+  - "python >= 2.6"
+  - "ansible >= 2.3"
+options:
+  server_url:
+    description:
+      - URL of Foreman server
+    required: true
+  username:
+    description:
+      - Username on Foreman server
+    required: true
+  password:
+    description:
+      - Password for user accessing Foreman server
+    required: true
+  validate_certs:
+    aliases: [ verify_ssl ]
+    description:
+      - Verify SSL of the Foreman server
+    required: false
+    default: true
+    type: bool
+  name:
+    description:
+      - Name or title of the Foreman Smart Proxy
+    required: true
+  url:
+    description:
+      - URL of the Smart Proxy
+    required: true
+  download_policy:
+    description:
+      - The download policy for the Smart Proxy
+    choices:
+      - background
+      - immediate
+      - on_demand
+    default: on_demand
+    required: false
+  organizations:
+    description:
+      - List of organizations the Smart Proxy should be assigned to
+    type: list
+    required: false
+  locations:
+    description:
+      - List of locations the Smart Proxy should be assigned to
+    type: list
+    required: false
+  state:
+    description:
+      - State of the Smart Proxy
+    default: present
+    choices:
+      - present
+      - absent
+'''
+
+EXAMPLES = '''
+# Create a local Smart Proxy
+- name: "Create Smart Proxy"
+  foreman_smart_proxy:
+    username: "admin"
+    password: "changeme"
+    server_url: "https://{{ ansible_fqdn }}"
+    name: "{{ ansible_fqdn }}"
+    url: "https://{{ ansible_fqdn }}:9090"
+    download_policy: "immediate"
+    organizations:
+      - "Default Organization"
+    locations:
+      - "Default Location"
+    state: present
+'''
+
+RETURN = ''' # '''
+
+try:
+    from ansible.module_utils.ansible_nailgun_cement import (
+        find_entities,
+        find_locations,
+        find_organizations,
+        find_smart_proxy,
+        naildown_entity_state,
+        sanitize_entity_dict,
+    )
+    from nailgun.entities import SmartProxy
+except ImportError:
+    pass
+
+
+from ansible.module_utils.foreman_helper import ForemanEntityAnsibleModule
+
+
+# This is the only true source for names (and versions thereof)
+name_map = {
+    'name': 'name',
+    'download_policy': 'download_policy',
+    'url': 'url',
+    'locations': 'location',
+    'organizations': 'organization',
+}
+
+
+def main():
+    module = ForemanEntityAnsibleModule(
+        argument_spec=dict(
+            name=dict(required=True),
+            url=dict(required=True),
+            download_policy=dict(choices=['background', 'immediate', 'on_demand']),
+            locations=dict(type='list'),
+            organizations=dict(type='list'),
+        ),
+        supports_check_mode=True
+    )
+
+    (entity_dict, state) = module.parse_params()
+
+    module.connect()
+
+    if 'locations' in entity_dict:
+        entity_dict['locations'] = find_locations(module, entity_dict['locations'])
+
+    if 'organizations' in entity_dict:
+        entity_dict['organizations'] = find_organizations(module, entity_dict['organizations'])
+
+    # Try to find the Smart Proxy to work on
+    entity = find_smart_proxy(module, name=entity_dict['name'], failsafe=True)
+    # Re-map names
+    entity_dict = sanitize_entity_dict(entity_dict, name_map)
+
+    # Make the changes
+    changed = naildown_entity_state(SmartProxy, entity_dict, entity, state, module)
+
+    module.exit_json(changed=changed)
+
+
+if __name__ == '__main__':
+    main()
+
+#  vim: set sts=4 ts=8 sw=4 ft=python et noro norl cin si ai :
--- a/plugins/modules/foreman_smart_proxy.py
+++ b/plugins/modules/foreman_smart_proxy.py
@@ -23,9 +23,7 @@ short_description: Manage Foreman Smart Proxy
 description:
   - Manage Foreman Smart Proxy
 requirements:
-  - "nailgun >= 0.28.0"
-  - "python >= 2.6"
-  - "ansible >= 2.3"
+  - "apypie"
 options:
   server_url:
     description:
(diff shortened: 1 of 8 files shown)


plugins/modules/foreman_smart_proxy.py
```diff
@@ -0,0 +1,167 @@
...
+# along with Ansible.  If not, see <http://www.gnu.org/licenses/>.
+
+DOCUMENTATION = '''
+---
+module: foreman_smart_proxy
+short_description: Manage Foreman Smart Proxy
+description:
+  - Manage Foreman Smart Proxy
+requirements:
+  - "nailgun >= 0.28.0"
```
mdellweg: Have you considered, writing it with apypie as a backend? I think we are not going to accept any more nailgun based modules.
@evgeni can we state this somewhere visible?

plugins/modules/foreman_smart_proxy.py
```diff
@@ -0,0 +1,167 @@
...
+def main():
+    module = ForemanEntityAnsibleModule(
+        argument_spec=dict(
+            name=dict(required=True),
+            url=dict(required=True),
+            download_policy=dict(choices=['background', 'immediate', 'on_demand']),
+            locations=dict(type='list'),
+            organizations=dict(type='list'),
+        ),
+        supports_check_mode=True
```
mdellweg: a comma please...

jstuart:
Thanks for the input!  I'll take a look at refactoring to apypie.  It wasn't obvious to me which style was preferred, so adding something to the readme or such noting that is a great idea.

Re: VCR stuff, I use Forklift pretty extensively, but I haven't used any of the pre-packaged playbooks.  Is there one of those (or a set of vars/roles) that you typically use when you're standing something clean up to record?

As you as you don't mind leaving this PR open, I'll revisit all of this on Monday.

mdellweg:
I used `vagrant up centos7-foreman-1.22` for that purpose recently.

jstuart:
I've restructured my initial module to use apypie and captured VCR data. Given that it's a pretty big deviation from my original attempt here, would you prefer that I push those changes to this branch, amend this commit, or simply open a new PR?

In case it proves useful for anyone else, this is what I did to capture VCR data:

```
user@host:~/development/forklift$ vagrant up centos7-foreman-1.22
...
user@host:~/development/forklift$ vagrant ssh centos7-foreman-1.22
Last login: Tue Jul  2 01:52:31 2019 from 10.0.2.2
[vagrant@centos7-foreman-1-22 ~]$ sudo su - root
[root@centos7-foreman-1-22 ~]# yum install -y ansible git gcc python-pip python-devel
...
[root@centos7-foreman-1-22 ~]# pip install setuptools --upgrade
...

-------------------
Checkout https//github.com/theforeman/foreman-ansible-modules.git or switch to /vagrant or another dir where it is accessible.
-------------------

[root@centos7-foreman-1-22 foreman-ansible-modules]# make test-setup
...
[root@centos7-foreman-1-22 foreman-ansible-modules]# sed -i -e "s/foreman_server_url: .*/foreman_server_url: \"https:\/\/$(hostname -f)\"/" test/test_playbooks/vars/server.yml
[root@centos7-foreman-1-22 foreman-ansible-modules]# make record_smart_proxy
```

mdellweg:
That probably makes you the first one to record those beasts from inside a foreman. :fireworks: 
That could be an idea to standardize the process. And i think the documentation around recording is still not ideal.
You can surely keep this pr. (Maybe change its title.)

jstuart:
Let me know if you see anything else that needs some attention.  Thanks for your patience and help!

test/test_playbooks/tasks/smart_proxy_puppet_host.yml
```diff
@@ -0,0 +1,7 @@
+---
+- name: "Ensure resolution of puppet to localhost in /etc/hosts is {{ puppet_host_state }}"
+  lineinfile:
+    path: /etc/hosts
+    line: 127.0.0.1 puppet
+    state: "{{ puppet_host_state }}"
+
```
mdellweg: OH, i think, i am not going to like this. Especially, since those module tests do not need any privilege escalation at all by now. Can we please find another way. As a last resort, we specify entities, the user must provide for recording in `...vars/server.yaml`.

mdellweg: I forgot to mention, that the rest looks amazing from a first look.

jstuart: The Smart Proxy uses SSL authentication, and the server side certificate is issued to <fqdn> with a SAN of 'puppet'.

Attempting to remove and add back the default Smart Proxy isn't feasible as it would require removing significant amounts of other, dependent configuration (e.g. subnets, domains, etc, etc).

The only other way I can think to do this is to change the configuration of the Foreman instance being used to record the test data prior to testing.  In this case, that would mean customizing the Forklift playbook to either add another SAN for localhost to the Smart Proxy server cert, or to add that same puppet entry to the /etc/hosts file.  Doing that would mean that the generation of these VCR cassettes will always require customization of the Foreman instance prior to recording.  Is that an acceptable solution?  Would it be acceptable to make that action and the corresponding teardown action conditional on a variable set in server.yaml that would only be used for recording VCR cassettes?

plugins/modules/foreman_smart_proxy.py
evgeni: No idea where. We could add it as a PR template maybe?

plugins/modules/foreman_smart_proxy.py
```diff
@@ -0,0 +1,152 @@
...
+    type: bool
+  name:
+    description:
+      - Name or title of the Foreman Smart Proxy
+    required: true
+  url:
+    description:
+      - URL of the Smart Proxy
+    required: true
+  download_policy:
```
evgeni: This is only a thing if you're running Katello, not plain Foreman. Maybe needs more docs.

test/test_playbooks/smart_proxy.yml
```diff
@@ -0,0 +1,44 @@
...
+  tasks:
+  - include: tasks/smart_proxy.yml
+    vars:
+      smart_proxy_state: present
+      smart_proxy_download_policy: background
+      expected_change: true
+  - include: tasks/smart_proxy.yml
+    vars:
+      smart_proxy_state: present
+      expected_change: false
```
evgeni: should this expect a change, as you set the download policy to background above, but the default is on_demand?

test/test_playbooks/tasks/smart_proxy_puppet_host.yml
evgeni: Why do we need `puppet` to be resolvable in the first place? Can't we use `ansible_fqdn` or something?

plugins/modules/foreman_smart_proxy.py
mdellweg: Better that now. But ideally, i'd like the contributor to see it before even starting to write the module.

test/test_playbooks/tasks/smart_proxy_puppet_host.yml
jstuart: The FQDN of the system is what's used in the URI for the default Smart Proxy.  That URI must be globally unique, so adding another with the same URI is not possible.  The cert is valid for that FQDN and 'puppet'.  So the options here are:

* Remove the existing Smart Proxy first, which will require the teardown of most of the Foreman config (e.g. all domains, subnets, etc must be detached or removed).
* Change the Foreman setup process to include more SANs, such as localhost, and add a Smart Proxy for that URI.
* Make 'puppet' resolve to localhost and use that URI.

The 'puppet' resolution option is the smallest change set and has no environmental impacts, but it does require privilege escalation.

I see two paths forward that have a chance of being sustainable over time:
* Leave the 'puppet' resolution approach in place but ensure that the resolution stuff only runs during recording.
* Use a non-standard Foreman setup process to allow the tests to run against localhost successfully, documenting the setup process required to successfully do recording in the future.

Just let me know which way you'd like to go here.

evgeni: Oooooh right. Not sure yet, but now I know what to think about, thanks!

jstuart: Just to close the loop on this, I'm not planning to make any additional changes without further guidance.  If any of the paths forward on this become acceptable, please let me know and I'd be happy to make the changes.  Thanks!

ekohl:
There may be another variation that is a bit hacky, but might work. The proxy also listens on HTTP (not HTTPS) port 8000. You may be able to register `http://$FQDN:8000` so you use the same built in proxy without messing with the hosts file. Note I haven't had a look at how the Foreman itself is deployed now in the tests.

test/test_playbooks/tasks/smart_proxy_puppet_host.yml
evgeni: That's absolutely fair!

I think I agree that using the internal proxy is the cheapest and least invasive way we can get away here. But I also have to agree with Matthias that using `become` will be problematic, especially as we run our tests on travis-ci.com where we don't have the ability to get root (at least at the moment, we *could* change that).

Given that once the VCR tapes are recorded, we don't run against a real Foreman, we don't actually need any of these names resolvable during tests, only during record.

We already have tasks that are only executed when a recording happens:
https://github.com/theforeman/foreman-ansible-modules/blob/ba2c1809df34598287d483a239e919da66e5e7d5/test/test_playbooks/tasks/repository_sync.yml#L38
So I think this would be step one to make it "smarter".

I think I'm OK with accepting the fact that this particular test will need a "special" setup to be recordable (special = being able to become root).

And also see below for @ekohl's idea of using HTTP instead of HTTPS :)

Another thing that I realized, but not sure if I want to ask you to change that: You're adding `127.0.0.1 puppet`, but you do so on the host that is running the module, not on the host that is running Foreman. These might be the same machines in your setup, but e.g. in mine I record from my laptop against a VM that runs foreman, and adding that line to my laptops /etc/hosts will not help.

Now, after having written that, maybe this should be just documented at the top of `test_playbooks/smart_proxy.yml` as an requirement and not try to be smart at all? Then we could also make the smart proxy URL configurable in https://github.com/theforeman/foreman-ansible-modules/blob/master/test/test_playbooks/vars/server.yml.example and allow the users to do use whichever proxy they want? (which is basically a "yes" to your question in https://github.com/theforeman/foreman-ansible-modules/pull/314#discussion_r300125000 whether a required customization step is is acceptable).

Gee, lots of text. Hope it makes any sense to you (and not only to my weird brain ;))

evgeni: And to add to @ekohl's idea, I was able to add a `http://$FQDN:8000` proxy to a katello setup I have here, but I had to disable SELinux as our policy actually forbids connecting to port 8000 :)

evgeni: So while @ekohl is out for a shower: the installer adds the internal proxy as `https://FQDN:9090` (at least in Katello, in Foreman it's 8443?) and you can't add another one with that same url. But you can add `https://FQDN:9090/` (i.e. with a trailing slash) just fine, because we only check for string equality. That said, you can also add `https://FOREMAN.EXAMPLE.COM:9090` because we don't lowercase the URL :)

I think now we have plenty of dirty tricks to reach the local proxy w/o adjusting the OS.

evgeni:
Hey @jstuart, still interested in implementing this? Or should someone else pick up your work?

PR #317, opened 2019-07-04
Modify global_parameter module for apypie

Manisha15:


Manisha15:
@mdellweg @evgeni This is updation of foreman_global_parameter module to apypie.

plugins/modules/foreman_global_parameter.py
```diff
@@ -43,6 +44,20 @@
     description:
       - Value of the Global Parameter
     required: false
+  parameter_type:
+    description:
+      - Type of value
+    required: true
```
mdellweg: I think, we should not require this, but default it to string.
At least for users of <1.22 having to set it seems very weird.

evgeni: Yepp, that's how I'd do it too.

plugins/module_utils/foreman_helper.py
```diff
@@ -207,6 +207,9 @@ def update_resource(self, resource, old_entity, entity_dict, check_missing, chec
                         new_value = type(value)(new_value)
                 value = self._flatten_value(value)
                 new_value = self._flatten_value(new_value)
+                if type_eval:
+                    if yaml.load(new_value, Loader=yaml.FullLoader) == value:
```
evgeni: can you explain a bit why this `type_eval` branch is required? I am guessing that `value` as coming from the API can be a boolean, an int, a string, a list etc, but `new_value` is always a string? I would have expected Ansible not being stupid here, but who knows :)

mdellweg: Yes, exactly it's about that, but it not working in any case. We are working on something more specific. And will probably also add a comment.

evgeni: My comment is: ewwww ;-)
Awaiting a tale about your endeavors!

plugins/modules/foreman_global_parameter.py
```diff
@@ -121,22 +134,19 @@ def main():
...
-            entity = entities[0]
-        else:
-            entity = None
-    except Exception as e:
-        module.fail_json(msg='Failed to find entity: %s ' % e)
+    entity = module.find_resource_by_name('common_parameters', name=entity_dict['name'], failsafe=True)
 
-    global_parameter_dict = sanitize_entity_dict(global_parameter_dict, name_map)
+    if not module.desired_absent:
+        # Fix values according to their corresponding patameter_type
```
evgeni:
```suggestion
        # Fix values according to their corresponding parameter_type
```

evgeni:
This fails on 2.7/3.5, I think because of different dict sorting?

Your recorded tape has:
```
    body: '{"common_parameter": {"value": "{\"a\": 2, \"1\": \"b\"}", "parameter_type":
      "hash"}}'
```

mdellweg:
That is my thinking here too. Does it mean, we need a `body_matcher_json_level_3`?

evgeni:
Seems so, yeah.
Or maybe it's sufficient to make `body` sorted in `body_json_l2_matcher`?
Something like `collections.OrderedDict(sorted(body.items()))`

evgeni:
Now it's only failing on 3.5? What?!

Manisha15:
Yes, for python 2.7 it was failing for real numbers only (may be round off error). For python 3.5 it is failing for dictionary.

evgeni:
Aaaaah ok.

plugins/modules/foreman_global_parameter.py
```diff
@@ -84,34 +99,32 @@
...
 
 # This is the only true source for names (and conversions thereof)
 name_map = {
     'name': 'name',
     'value': 'value',
+    'parameter_type': 'parameter_type',
 }
 
 
+def value_to_str(value, parameter_type):
```
bagasse: I think this can be defined in the helper because it can be reused for host / hosgroup parameters

mdellweg:
@Manisha15 I took the liberty to find a way to make the tests pass on Python3.5.

plugins/modules/foreman_global_parameter.py
```diff
@@ -84,58 +99,49 @@
...
-            entity = None
-    except Exception as e:
-        module.fail_json(msg='Failed to find entity: %s ' % e)
+    entity = module.find_resource_by_name('common_parameters', name=entity_dict['name'], failsafe=True)
 
-    global_parameter_dict = sanitize_entity_dict(global_parameter_dict, name_map)
+    if not module.desired_absent:
+        # Fix values according to their corresponding parameter_type
+        if entity:
+            if 'parameter_type' not in entity:
```
evgeni: I'd put both of these statements in one line. But it's your code ;)

plugins/modules/foreman_global_parameter.py
```diff
@@ -84,58 +99,49 @@
...
-            entity = entities[0]
-        else:
-            entity = None
-    except Exception as e:
-        module.fail_json(msg='Failed to find entity: %s ' % e)
+    entity = module.find_resource_by_name('common_parameters', name=entity_dict['name'], failsafe=True)
 
-    global_parameter_dict = sanitize_entity_dict(global_parameter_dict, name_map)
+    if not module.desired_absent:
+        # Fix values according to their corresponding parameter_type
```
evgeni:
```suggestion
        # Convert values according to their corresponding parameter_type
```

Fix somehow implies they are broken :)

tests/vcr_python_wrapper.py
```diff
@@ -10,17 +10,17 @@
...
-            body1['search'] = ','.join(sorted(re.findall(r'([^=,]*="(?:[^"]|\\")*")', body1['search'])))
-        if 'search' in body2:
-            body2['search'] = ','.join(sorted(re.findall(r'([^=,]*="(?:[^"]|\\")*")', body2['search'])))
+        if 'common_parameter' in body1 and 'common_parameter' in body2:
+            if body1['common_parameter'].get('parameter_type') == body2['common_parameter'].get('parameter_type') in ['hash', 'json', 'yaml']:
+                body1['common_parameter']['value'] = json.loads(body1['common_parameter'].get('value'))
+                body2['common_parameter']['value'] = json.loads(body2['common_parameter'].get('value'))
         return body1 == body2
-    elif r1.headers.get('content-type') == 'multipart/form-data' and r2.headers.get('content-type') == 'multipart/form-data':
+    elif r1.headers.get('content-type') == r2.headers.get('content-type') == 'multipart/form-data':
```
evgeni: I find it interesting that this works here, as in #382 I had to use `startswith` as the header was something like `multipart/form-data; boundary=44d0d62903f617236943e867fe494c96`

tests/vcr_python_wrapper.py
```diff
@@ -10,17 +10,17 @@
 # We need our own json level2 matcher, because, python2 and python3 do not save
 # dictionaries in the same order
 def body_json_l2_matcher(r1, r2):
-    if r1.headers.get('content-type') == 'application/json' and r2.headers.get('content-type') == 'application/json':
+    if r1.headers.get('content-type') == r2.headers.get('content-type') == 'application/json':
         if r1.body is None or r2.body is None:
             return r1.body == r2.body
         body1 = json.loads(r1.body.decode('utf8'))
         body2 = json.loads(r2.body.decode('utf8'))
-        if 'search' in body1:
```
evgeni: I don't quite understand this hunk.
The old code would sort our `search` param. The new code does not touch `search` at all? And it still works?

Granted, apypie passes the search in the URL, not in the body, but I still don't understand why this doesn't break the nailgun based modules.

tests/vcr_python_wrapper.py
mdellweg: Right, but i did not really change anything here.
Maybe we should not question vcr...

evgeni: Yeah, not saying you did, just an observation :)

tests/vcr_python_wrapper.py
mdellweg: That is due to the way nailgun delivered the search param. Now we have no nailgun module left, that searches with >= 2 parameters. So sorting is a noop.
In Apypie the search is no longer part of the body, but in the querystring.

evgeni: I wished so hard that the sentence would end at "Now we have no nailgun module left" :rofl:

bagasse:
@evgeni good for me too :)

evgeni:
Thanks @Manisha15 and @mdellweg for the code, @bagasse for the review. Merging!

PR #325, opened 2019-07-05
Refactor for desired_absent

mdellweg:
Affected modules:
 * foreman_auth_source_ldap
 * foreman_hostgroup
 * foreman_user


evgeni:
auth ldap seems to miss the recorded apidoc?

PR #328, opened 2019-07-08
Use 'missing_required_lib' to report import errors

### LINKED ISSUE #326: There is a convention for importing libraries, we should conform to.
We should conform with:
https://docs.ansible.com/ansible/latest/dev_guide/developing_modules_best_practices.html#importing-and-using-shared-code

mdellweg:
Fixes #326

### DIFF
--- a/plugins/module_utils/foreman_helper.py
+++ b/plugins/module_utils/foreman_helper.py
@@ -5,7 +5,7 @@ import re
 import yaml
 import traceback
 
-from ansible.module_utils.basic import AnsibleModule
+from ansible.module_utils.basic import AnsibleModule, missing_required_lib
 
 try:
     import ansible.module_utils.ansible_nailgun_cement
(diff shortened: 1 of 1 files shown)


plugins/module_utils/foreman_helper.py
```diff
@@ -5,7 +5,7 @@
 import yaml
 import traceback
 
-from ansible.module_utils.basic import AnsibleModule
+from ansible.module_utils.basic import AnsibleModule, missing_required_lib
```
evgeni: This is Ansible 2.8+ only :/

evgeni:
marked this as WIP until 2.8 is "old enough"

mdellweg:
"Wait In Progress"
I was about to create a generic "blocked" label, but i think we have nough of em.

evgeni:
a rebased version is in #839

PR #329, opened 2019-07-08
Use assert in tests

mdellweg:
Clean up some test tasks, and straighten out some involved logic.


test/test_playbooks/tasks/setting_fact.yml
```diff
@@ -7,7 +7,10 @@
...
     name: "{{ setting_name | default(omit) }}"
   register: result
-- fail:
-    msg: "Verification that setting is '{{ setting_value }}' failed!"
-  when: (setting_name is defined) and (setting_value is defined) and (result.settings[0].value != setting_value)
+- assert:
+    fail_msg: "Verification that setting is '{{ setting_value }}' failed!"
+    quiet:
+    that:
+      - result.settings[0].value == setting_value)
```
evgeni:
```suggestion
      - result.settings[0].value == setting_value
```

mdellweg: And i thought, i tested all that stuff.

test/test_playbooks/tasks/content_view_filter.yml
```diff
@@ -1,42 +0,0 @@
----
```
evgeni: for those who wonder about this file, it's been unused since that module was introduced in d71ee5eacd3704458aa324ec02d783bfc6fad84e :)

test/test_playbooks/tasks/setting.yml
```diff
@@ -10,10 +10,16 @@
...
     value: "{{ setting_value | default(omit) }}"
   register: result
-- fail:
-    msg: "Ensuring setting is {{ setting_value | default(false) | ternary(\"'\" + (setting_value | string) | default('') + \"'\", 'undefined') }} failed! (expected_change: {{ expected_change | default('unknown') }})"
-  when: (expected_change is defined) and (result.changed != expected_change)
-- fail:
-    msg: "Setting the setting failed! ('{{ result.foreman_setting.value }}' != '{{ setting_value_override | default(setting_value) }}')"
-  when: (setting_value is defined) and (result.foreman_setting.value != setting_value_override | default(setting_value))
+- assert:
+    fail_msg: "Ensuring setting is {{ setting_value | default(false) | ternary(\"'\" + (setting_value | string) | default('') + \"'\", 'undefined') }} failed! (expected_change: {{ expected_change | default('unknown') }})"
```
evgeni: "The task includes an option with an undefined variable. The error was: 'setting_value' is undefined\n\nThe error appears to be in '/home/travis/build/theforeman/foreman-ansible-modules/test/test_playbooks/tasks/setting.yml': line 13, column 3, but may\nbe elsewhere in the file depending on the exact syntax problem.\n\nThe offending line appears to be:\n\n  register: result\n- assert:\n  ^ here\n"

:woman_shrugging:

mdellweg: The problem seems to be that `setting_value` is not defined in `(setting_value | string) | default('')`. While this is easy to fix by postponing the string conversion to after the `|default('')`, it tells me that those fail_msg are compiled even if the whole task is skipped. This is a sad answer to your 2. comment.

evgeni: yepp, thanks for thinking with me (or against me? :))

PR #335, opened 2019-07-10
Refactor foreman_setting module to apypie

mdellweg:
Also, remove deprecated foreman_setting_facts module.


mdellweg:
@sean797 I needed to add some strange workaround for resetting to a default for lists.
Can you please have a look. I think you have more insight on the type-checking matter.

plugins/modules/foreman_setting.py
```diff
@@ -92,32 +78,31 @@
...
-        entity_dict['value'] = entity.default or ""
-
-    entity_dict = sanitize_entity_dict(entity_dict, name_map)
+        default = entity['default'] or ""
+        if isinstance(default, list):
+            default = ','.join(default)
+        entity_dict['value'] = default
 
-    changed, entity = naildown_entity(Setting, entity_dict, entity, 'present', module, check_type=True)
+    # add state='present' as soon as it is available again
```
mdellweg: This would be when #337 is ready.

plugins/modules/foreman_setting_facts.py
```diff
@@ -1,162 +0,0 @@
-#!/usr/bin/python
```
evgeni: is this removal intentional?

evgeni: disregard this comment, it is according to the initial PR comment

mdellweg:
@evgeni @sean797 
This changes the original interface (which we never stated to be stable...) in that lists should be passed as lists now, and order matters.

tests/test_playbooks/setting.yml
```diff
@@ -33,24 +39,28 @@
...
-      setting_value_override: ['bar', 'foo']
+      setting_value: ['foo', 'bar']
       expected_change: false
 
   - include: tasks/setting.yml
     vars:
-      setting_name: content_action_accept_timeout
+      setting_name: token_duration
       setting_value: 99
+      setting_value_override: "99"
```
mdellweg: This is needed, because the 99 already arrives as a string in entity_dict. Maybe it is the jinja magic in tasks/setting.yaml, that converts it. ;(

plugins/modules/foreman_setting.py
```diff
@@ -69,53 +67,45 @@
...
+    entity_dict['value'] = parameter_value_to_str(new_value, settings_type)
+    entity['value'] = parameter_value_to_str(entity['value'], settings_type)
 
-    entity_dict = sanitize_entity_dict(entity_dict, name_map)
+    changed, entity = module.ensure_entity('settings', entity_dict, entity, state='present', entity_spec=entity_spec)
 
-    changed, entity = naildown_entity(Setting, entity_dict, entity, 'present', module, check_type=True)
+    if entity:
+        # Fake the not serialized version as output
+        entity['value'] = new_value
```
evgeni: Why? When I `PUT` a new integer value (regardless whether the JSON I upload has it as a String or an Integer) the returned reply is an Integer, so it should compare correctly already?

```
# curl -X PUT -u admin:changeme -d '{"setting":{"value":361}}' -H 'Content-Type: application/json' -k https://foreman.example.com/api/v2/settings/132
{"value":361,"description":"Time in minutes installation tokens should be valid for, 0 to disable token generation","category":"Setting::Provisioning","settings_type":"integer","default":360,"created_at":"2019-07-16 06:55:35 UTC","updated_at":"2019-08-09 07:58:05 UTC","id":132,"name":"token_duration","full_name":"Token duration","category_name":"Provisioning"}

# curl -X PUT -u admin:changeme -d '{"setting":{"value":"362"}}' -H 'Content-Type: application/json' -k https://foreman.example.com/api/v2/settings/132
{"value":362,"description":"Time in minutes installation tokens should be valid for, 0 to disable token generation","category":"Setting::Provisioning","settings_type":"integer","default":360,"created_at":"2019-07-16 06:55:35 UTC","updated_at":"2019-08-09 07:59:13 UTC","id":132,"name":"token_duration","full_name":"Token duration","category_name":"Provisioning"}
```

evgeni:
@sean797 do you plan/want to review this PR? Or is it good enough if I ACK it? :)

plugins/modules/foreman_setting.py
```diff
@@ -69,53 +67,45 @@
...
+entity_spec = {
+    'name': {},
+    'value': {},
 }
 
 
 def main():
-    module = ForemanAnsibleModule(
+    module = ForemanApypieAnsibleModule(
         argument_spec=dict(
```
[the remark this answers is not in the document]
mdellweg: This one is special, as it has no `state`. Therefore it is not using the `EntityAnsibleModule`.
I don't know how it behaves with the `entity_spec`. Need to check that.

plugins/modules/foreman_setting.py
mdellweg: The problem is not with integers:
`"msg": "Setting the setting failed! ('[\"foo\", \"bar\"]' != '[u'foo', u'bar']')"`

plugins/modules/foreman_setting.py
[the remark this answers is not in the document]
mdellweg: At least in the current version, it is not supported to pass `entity_spec` to the non-entity-module. It sounds kind of reasonable, given the current naming.

plugins/modules/foreman_setting.py
evgeni: Right. But it's the integers where it bites us in the tests. So my thought to update the `if` to:

```python
if entity and entity['settings_type'] in ['array', 'hash', 'yaml', 'json']:
  entity['value'] = new_value
```

plugins/modules/foreman_setting.py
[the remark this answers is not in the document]
evgeni: Right. (And maybe the name is wrong, but let's discuss that at another stage)

plugins/modules/foreman_setting.py
```diff
@@ -69,53 +67,45 @@
...
-        entity_dict['value'] = entity.default or ""
+        entity_dict['value'] = entity['default'] or ''
+
+    settings_type = entity['settings_type']
+    new_value = entity_dict['value']
+    entity_dict['value'] = parameter_value_to_str(new_value, settings_type)
+    entity['value'] = parameter_value_to_str(entity['value'], settings_type)
 
-    entity_dict = sanitize_entity_dict(entity_dict, name_map)
+    changed, entity = module.ensure_entity('settings', entity_dict, entity, state='present', entity_spec=entity_spec)
```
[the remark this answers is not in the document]
evgeni: not here, not today.

plugins/modules/foreman_setting.py
mdellweg: Hmm. It's more complicated than this, because the api wants `string` and gives us `integer`.
So in case, we did not update the setting, the only reasonable thing to return is the "unchanged" current entity. However we changed it already, to make it fit for the comparison in `_update_entity`. So the solution, i found, is to convert the input to integer, if the `settings_type` is `integer`.

tests/test_playbooks/tasks/setting.yml
```diff
@@ -16,6 +16,8 @@
     that:
       - result.changed == expected_change
   when: expected_change is defined
+- debug:
```
evgeni: I think this can be removed again? :)

plugins/modules/foreman_setting.py
evgeni: Works for me, thanks!

tests/test_playbooks/tasks/setting.yml
mdellweg: Oh yes. :fallen_leaf:

PR #337, opened 2019-07-10
Refactor ensure resource

mdellweg:


plugins/module_utils/foreman_helper.py
```diff
@@ -286,8 +303,10 @@ def __init__(self, argument_spec, **kwargs):
             state=dict(choices=['present', 'absent'], default='present'),
         )
         args.update(argument_spec)
+        name_map = kwargs.pop('name_map', {})
```
evgeni: Do we really want to touch the `nailgun` API?

plugins/modules/katello_sync_plan.py
```diff
@@ -143,7 +143,7 @@ def main():
 
     sync_plan = find_sync_plan(module, entity_dict['name'], entity_dict['organization'], failsafe=True)
 
-    entity_dict = sanitize_entity_dict(entity_dict, name_map)
+    entity_dict = sanitize_entity_dict(entity_dict, name_map=name_map)
```
evgeni: why this change? (besides that I am not sure I want to touch the nailgun modules at all), `sanitize_entity_dict` takes exactly two params :)

evgeni:
Overall, I like :)

plugins/modules/katello_sync_plan.py
mdellweg: `sed -i '...' *.py`
I can take it back. ;)

plugins/module_utils/foreman_helper.py
mdellweg: You did for the documentation. But, you are right. There is no benefit.

I thought about moving the not nailgun Classes to the cement file, because i accidentally added my changes there, and wondered, nothing happend.

evgeni: I did, because it was a no-code change ;)

plugins/modules/katello_sync_plan.py
evgeni: I'd prefer not to touch these files.

PR #340, opened 2019-07-11
Add example for enabling RHEL 8 repo

jwbernin:
RHEL 8 repos won't enable properly if an arch is specified. This change demonstrates how to enable the BaseOS release version 8 repository.

### DIFF
--- a/plugins/modules/katello_repository_set.py
+++ b/plugins/modules/katello_repository_set.py
@@ -109,7 +109,7 @@ EXAMPLES = '''
     repositories:
       - basearch: x86_64
 
-- name: "Enable RHEL 8 BaseOS RPMs repositoryi with label"
+- name: "Enable RHEL 8 BaseOS RPMs repository with label"
   katello_repository_set:
     username: "admin"
     password: "changeme"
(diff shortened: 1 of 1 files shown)


plugins/modules/katello_repository_set.py
```diff
@@ -108,6 +108,16 @@
     state: disabled
     repositories:
       - basearch: x86_64
+
+- name: "Enable RHEL 8 BaseOS RPMs repositoryi with label"
```
mdellweg:
```suggestion
- name: "Enable RHEL 8 BaseOS RPMs repository with label"
```

mdellweg:
I am not an expert here. Maybe 'description' is the right place for something like this.
Thoughts? @sean797 @akofink

sean797:
I think the example is good, perhaps there is something we can do on the Katello side as well. @jwbernin Would you mind creating an issue? https://projects.theforeman.org/projects/katello/issues/new

jwbernin:
@sean797 I'm not sure this is an issue on the katello side. If it is, I don't see what it is well enough to open an issue with any sort of specifics, being mostly unfamiliar with the katello project itself.

Issue #319, opened 2019-07-04
default to supports_check_mode=True


evgeni:
##### SUMMARY
Due to the wrapper classes we have for talking to the API, all our modules support check mode at no cost. Yet they all have to define that on their own. Let's move that definition into the base class!

##### ISSUE TYPE
 - Feature Idea

mdellweg:
The moment they implement actions (like publish, revert, promote) on their own, they need to handle the checkmode think for those explicitly. Do we want them to force supporting check_mode, or do we give them the possibility to opt out?

evgeni:
We surely should give them the option to opt out. I just think the default value for `supports_check_mode` should be True.

PR #343, opened 2019-07-12
default to supports_check_mode=True

evgeni:
Fixes: #319


evgeni:
(now mdellweg correctly will yell at me that I touched nailgun modules -.-)

Issue #305, opened 2019-06-06
Problem serializing network in foreman_subnet


nstrug:
##### SUMMARY
I get the following error when trying to create a subnet:

```
 "msg": "Error while creating Subnet: Object of type IPAddress is not JSON serializable"

```

##### ISSUE TYPE
 - Bug Report

##### ANSIBLE VERSION

```
$ ansible --version
ansible 2.7.10
  config file = /home/nstrug/projects/clone-satellite/ansible.cfg
  configured module search path = ['/home/nstrug/projects/clone-satellite/foreman-ansible-modules/modules']
  ansible python module location = /usr/lib/python3.7/site-packages/ansible
  executable location = /usr/bin/ansible
  python version = 3.7.3 (default, Mar 27 2019, 13:36:35) [GCC 9.0.1 20190227 (Red Hat 9.0.1-0.8)]




```

##### KATELLO/FOREMAN VERSION

```
# rpm -q tfm-rubygem-katello foreman
tfm-rubygem-katello-3.7.0.rc1.4-0.1.rc1.4.el7sat.noarch
foreman-1.18.0.1-1.el7sat.noarch

```

##### NAILGUN VERSION

```
$ pip3 show nailgun | grep Version
Version: 0.32.0

$ pip3 show netaddr | grep Version
Version: 0.7.19


```

##### STEPS TO REPRODUCE

```
- name: Generate Subnets
  foreman_subnet:
    server_url: "{{ target_foreman_server_url }}"
    username: "{{ target_foreman_username }}"
    password: "{{ target_foreman_password }}"
    verify_ssl: "{{ foreman_verify_ssl }}"
    name: "{{ item.Name }}"
    boot_mode: "{{ item['Boot Mode'] }}"
    cidr: "{{ item['Network Prefix'] }}"
    dhcp_proxy: "{{ inventory_hostname }}"
    dns_primary: "{{ item['Primary DNS'] }}"
    dns_secondary: "{{ item['Secondary DNS'] }}"
    dns_proxy: "{{ inventory_hostname }}"
    tftp_proxy: "{{ inventory_hostname }}"
    from_ip: "{{ item['Start of IP Range'] }}"
    to_ip: "{{ item['End of IP Range'] }}"
    gateway: "{{ item['Gateway Addr'] }}"
    ipam: "{{ (item.IPAM == 'None')|ternary(omit,item.IPAM) }}"
    mtu: "{{ item.MTU|int }}"
[...]

PR #346, opened 2019-07-12
foreman_subnet: cast netmask to a string after computing it

evgeni:
Fixes: #305


evgeni:
I did not re-record the tests (yet) because it still needs fixing the nested locations bug.

evgeni:
I've opened this also as https://github.com/theforeman/foreman-ansible-modules/pull/347 against the `nailgun` branch, but skipped updating tests, as re-recording is currently broken and I don't want to fix that too.

evgeni:
now with 100% more working tests

PR #347, opened 2019-07-12
foreman_subnet: cast netmask to a string after computing it

evgeni:
Fixes: #305


evgeni:
@nstrug wanna test this fix?

evgeni:
I'd lean to just merge this, as it's the same fix we did in `master` for the `apypie` implementation

mdellweg:
Now `nailgun` is a branch for a reason. :imp:

PR #349, opened 2019-07-15
Refactor katello product and repository

mdellweg:


plugins/modules/katello_repository.py
```diff
@@ -120,43 +120,28 @@
...
-    )
-except ImportError:
-    pass
 
-from ansible.module_utils.foreman_helper import KatelloEntityAnsibleModule
+from ansible.module_utils.foreman_helper import KatelloEntityApypieAnsibleModule
 
 
 # This is the only true source for names (and conversions thereof)
 name_map = {
```
evgeni: I wonder why there is no `organization(_id)` in the `name_map` here

evgeni: Or is it because a Product already has an Org and thus it gets the id via that route?

mdellweg: Good question. I guess, because the repository is scoped by the product, which in turn is scoped by the organization. The tests passing suggest that it is sufficient.

plugins/modules/katello_content_credential.py
```diff
@@ -89,7 +89,6 @@ def main():
             content_type=dict(required=True, choices=['gpg_key', 'cert']),
             content=dict(required=True),
         ),
-        supports_check_mode=True,
```
evgeni: doesn't belong here, but I don't care too much :)

PR #352, opened 2019-07-16
re-record tapes after apypie route selection changes

evgeni:


evgeni:
compute_attribute is still broken, as I can't get it to re-record without seting up a whole libvirt install here

evgeni:
well... updated

mdellweg:
I know, but we need a compute_resource to do compute_attributes. And there is no fake_resource in foreman. Also, i think the compute_resource playbook is missing taxonomies fixtures. I am going to ack this anyway, to get this green again.

PR #354, opened 2019-07-16
Refactor operating system

mdellweg:


plugins/module_utils/foreman_helper.py
```diff
@@ -364,6 +364,15 @@ def sanitize_entity_dict(entity_dict, name_map):
     return {value: entity_dict[key] for key, value in name_map.items() if key in entity_dict}
 
 
+# Helper for (global, operatingsystem, ...) parameters
```
evgeni: Can I be a real docstring please?

(Why is there no Pinocchio emoji?)

plugins/module_utils/foreman_helper.py
```diff
@@ -364,6 +364,15 @@ def sanitize_entity_dict(entity_dict, name_map):
...
+def parameter_value_to_str(value, parameter_type):
```
evgeni: I think I'd prefer:

```python
def parameter_value_to_str(value, parameter_type):
    if parameter_type in ['real']:
        parameter_string = str(value)
    elif parameter_type in ['array', 'hash', 'yaml', 'json']:
        parameter_string = json.dumps(value)
    else:
        parameter_string = value
    return parameter_string
```

plugins/modules/foreman_operating_system.py
```diff
@@ -192,81 +184,80 @@ def main():
...
+    # Try to find the Operating System to work on
+    # name is however not unique, but description is, as well as "<name> <major>[.<minor>]"
+    entity = None
+    # If we have a description, search for it
+    if 'description' in entity_dict and entity_dict['description'] != '':
+        entity = module.find_resource_by_title('operatingsystems', entity_dict['description'], failsafe=True)
+    # If we did not yet find a unique OS, search by name & version
+    if entity is None:
+        search_parts = ['{}="{}"'.format(key, entity_dict[key]) for key in ('name', 'major', 'minor') if key in entity_dict]
+        entity = module.find_resource('operatingsystems', ','.join(search_parts), failsafe=True)
```
evgeni: would it be more readable to have `search_string = ','.join(search_parts)` as an own line?

plugins/modules/foreman_operating_system.py
```diff
@@ -192,81 +184,80 @@ def main():
...
-    if 'architectures' in operating_system_dict:
-        operating_system_dict['architectures'] = find_entities_by_name(
-            Architecture, operating_system_dict['architectures'], module)
-
-    # Set Installation Media of Operating System
-    if 'media' in operating_system_dict:
-        operating_system_dict['media'] = find_entities_by_name(
-            Media, operating_system_dict['media'], module)
+    if not module.desired_absent:
+        # Set Architectures of Operating System
```
evgeni: I find these comments (here and in the following assignments) quite redundant to the actual code, so I'd drop them (but up to you if you'd do that ;))

plugins/modules/foreman_operating_system.py
```diff
@@ -192,81 +184,80 @@ def main():
...
+        if module.state == 'present' or (module.state == 'present_with_defaults' and entity is None):
+            search_params = {'operatingsystem_id': operating_system['id']}
             if entity:
-                current_parameters = OperatingSystemParameter(operatingsystem=operating_system).search()
-                current_parameters = {p.name: p for p in current_parameters}
+                current_parameters = {p['name']: p for p in module.list_resource('parameters', params=search_params)}
             else:
                 current_parameters = {}
-            desired_parameters = {name: {'name': name, 'value': value, 'operatingsystem': operating_system} for name, value in desired_parameters.items()}
+            desired_parameters = {p['name']: p for p in parameters}
```
evgeni:
```suggestion
            desired_parameters = {parameter['name']: parameter for parameter in parameters}
```

plugins/modules/foreman_operating_system.py
```diff
@@ -192,81 +184,80 @@ def main():
             ptables=dict(type='list'),
             provisioning_templates=dict(type='list'),
             password_hash=dict(choices=['MD5', 'SHA256', 'SHA512']),
-            parameters=dict(type='dict'),
+            parameters=dict(type='list', elements='dict', options=dict(
```
evgeni: this is a backwards incompatible change. I think I'm okay with it as we don't promise a stable API (yet), but I wanted to point it out :)

plugins/modules/foreman_operating_system.py
```diff
@@ -113,6 +113,9 @@
     release_name: stretch
     family: Debian
     major: 9
+    parameters:
+      - name: additional-packages
+      - value: python vim
```
evgeni:
```suggestion
        value: python vim
```

evgeni: I think this also needs an update of the `type: dict` in the docs above?

plugins/modules/foreman_operating_system.py
mdellweg: probably. `',' ` makes for a really bad parameter. ;)

plugins/modules/foreman_operating_system.py
mdellweg: Right. It is kind of tied to the api change of parameters in foreman 1.22.

plugins/module_utils/foreman_helper.py
mdellweg: Those comments in foreman_helper currently serve more as a structuring device for that file.

plugins/modules/foreman_operating_system.py
evgeni: But the module remains compatible with 1.21 and older because our magic will just throw away that type param if its not in the apidoc, right?

mdellweg: That's what we all hope for. ;)

plugins/modules/foreman_operating_system.py
mdellweg: I changed it so that we do not even save the list.

mdellweg:
With this diff:
```
diff --git a/tests/test_playbooks/operating_system.yml b/tests/test_playbooks/operating_system.yml
index 00aeb92..5ba56a7 100644
--- a/tests/test_playbooks/operating_system.yml
+++ b/tests/test_playbooks/operating_system.yml
@@ -28,7 +28,6 @@
             value: value1
           - name: param2
             value: value2
-            parameter_type: json
         expected_change: true
     - include: tasks/operating_system.yml
       vars:
@@ -38,7 +37,6 @@
             value: value1
           - name: param2
             value: value2
-            parameter_type: json
         expected_change: false
     - include: tasks/operating_system.yml
       vars:
```
I get:
```
$ ansible-playbook -i tests/inventory/ tests/test_playbooks/operating_system.yml
<...>
PLAY RECAP ***********************************************************************************************************************************************************************************
fixtures                   : ok=1    changed=1    unreachable=0    failed=0    skipped=1    rescued=0    ignored=0   
tests                      : ok=14   changed=4    unreachable=0    failed=0    skipped=0    rescued=0    ignored=0
```
Running against forklift/centos7-foreman-1.21

mdellweg:
Not quite there yet it seems.

evgeni:
Why does it want a subnet id? It's optional if we can trust the docs

evgeni:
Or is it getting the params from the wrong route?

evgeni:
Yupp, definitely wrong route 😭

evgeni:
My best guess would be that os_id isn't set when we update the parameter, but I don't understand how/why that would be happening.

evgeni:
Should
```
changed |= module.ensure_resource_state('parameters', desired_parameters[name], current_parameter, state="present", name_map=parameter_name_map)
```
be
```
changed |= module.ensure_resource_state('parameters', desired_parameter, current_parameter, state="present", name_map=parameter_name_map)
```

Even though I'd expect it to be the same

evgeni:
And
```
current_parameter['operatingsystem_id'] = operating_system
```
should be
```
current_parameter['operatingsystem'] = operating_system
```

mdellweg:
I suspect the update function to throw it away, because it does not change.

mdellweg:
This seems to work. But honestly, does any one know, what that update_resource function _is_ doing?

evgeni:
> This seems to work. But honestly, does any one know, what that update_resource function _is_ doing?

see the comment in the very first line of that function :(

plugins/modules/foreman_operating_system.py
```diff
@@ -139,47 +166,38 @@
...
+    'family': {},
+    'major': {},
+    'minor': {},
+    'release_name': {},
+    'architectures': {'type': 'entity_list', 'flat_name': 'architecture_ids'},
+    'media': {'type': 'entity_list', 'flat_name': 'medium_ids'},
+    'ptables': {'type': 'entity_list', 'flat_name': 'ptable_ids'},
+    'provisioning_templates': {'type': 'entity_list', 'flat_name': 'provisioning_template_ids'},
+    'password_hash': {},
+}
```
mdellweg: @evgeni After having real trouble updating nested entities, i started to elaborate on your foreman_spec idea. I came up with an entity_spec and a rewrite of the complete ensure mechanism (this time tailored for dict:dict instead of dict:nailgun).
One nice but nonobvious side effect is that i could reduce the doubled translation via name_map and flattening into a single step by applying the entity_spec.
The _new_ update mechanism does not yet know the specialities like `force_update`, `check_missing` ... I think, we can add those on the go, and specify their need in the entity_spec. I hope, we keep the update function clean on the way.

Fodoj: Please consider adding organization_id here or on global for all resources

Fodoj: True, sorry, confused it with other resources - https://github.com/theforeman/foreman-ansible-modules/issues/365

plugins/module_utils/foreman_helper.py
```diff
@@ -364,6 +462,34 @@ def sanitize_entity_dict(entity_dict, name_map):
     return {value: entity_dict[key] for key, value in name_map.items() if key in entity_dict}
 
 
+def _flatten_entity(entity, entity_spec):
+    """Flatten entity according to spec"""
+    result = {}
+    for key, value in entity.items():
+        if key in entity_spec:
```
evgeni: Won't this remove all keys that are already in the flattened name? (When you pass an existing entity)

plugins/module_utils/foreman_helper.py
```diff
@@ -364,6 +462,34 @@ def sanitize_entity_dict(entity_dict, name_map):
...
+            spec = entity_spec[key]
+            _type = spec.get('type', 'str')
+            if _type == 'entity':
```
evgeni: I think having
```python
```
here and use flat_name in all accesses below would be more readable

plugins/modules/foreman_operating_system.py
```diff
@@ -139,47 +166,37 @@
...
-    'major': 'major',
-    'minor': 'minor',
-    'release_name': 'release_name',
-    'architectures': 'architecture',
-    'media': 'medium',
-    'ptables': 'ptable',
-    'provisioning_templates': 'provisioning_template',
-    'password_hash': 'password_hash',
+entity_spec = {
+    'id': {},
```
evgeni: I liked to have the id implicit :p

plugins/module_utils/foreman_helper.py
```diff
@@ -285,14 +287,108 @@ def ensure_resource(self, resource, entity_dict, entity, state=None, name_map=No
...
+                resource (string): Plural name of the api resource to manipulate
+                desired_entity (dict): Desired properties of the entity
+                current_entity (dict): Current properties of the entity
+                params (dict): Lookup parameters (i.e. parent_id for nested entities) (optional)
+                entity_spec (dict): Description of the entity structure
+            Return value:
+                Pair of boolean indicating whether something changed and the new current state if the entity
+        """
+        payload = {}
+        desired_entity = _flatten_entity(desired_entity, entity_spec)
```
evgeni: Should the flattening just happen "once" in ensure_entity instead in each create/update/(delete) function?

evgeni:
Instead of passing in params=search_params, why don't we mark some params as required in the spec?

plugins/module_utils/foreman_helper.py
mdellweg: you mean, that you cannot 'flatten' it twice? That would be correct. But that nothing new compared to the existing sanitize name_map mechanism.

evgeni: But did we ever run sanitize on the data received from the api? We only did flatten_value before, and that did not take into account the map.

mdellweg:
> Instead of passing in params=search_params, why don't we mark some params as required in the spec?

Sounds like a practical idea.

plugins/module_utils/foreman_helper.py
mdellweg: We could make it idempotent by adding all the flat_names as keys with empty dictionary. Do you think it is worth?

plugins/module_utils/foreman_helper.py
mdellweg: I'll try that.

plugins/modules/foreman_operating_system.py
mdellweg: I prefer explicit, but could be convinced.
Also for the cases, where we rename things like `from_ip -> `from`, we need to add both here in order not to delete it from the current_entity. At least, i could not think of an easier way to encode alias names.

plugins/module_utils/foreman_helper.py
mdellweg: We call it exactly _once_, just not in one place. That is due to the fact, that we do not want to call it on the nonexisting current_entity in case of create, and that it is not needed on the desired_entity in case of delete.

plugins/modules/foreman_operating_system.py
evgeni: I think my argument for implicit is "we carry only user visible or better user input in that list"

plugins/module_utils/foreman_helper.py
evgeni: Well, I think it's mandatory. But maybe I'm also missreading the code -- only on my mobile at the moment.

You call flatten_entity on current_entity in update, and current contains the api generated answer. So after flatten it will be missing entries. And then you compare it to desired_entity and stuff is missing. Right?

plugins/modules/foreman_operating_system.py
evgeni: We could stop this renaming (from/from_ip) madness. It's a remainder from nailgun which could not use python reserved words. You can use them in dicts tho. And an alias in the argument spec is cheap.

plugins/module_utils/foreman_helper.py
evgeni: Yeah. I know we call it once. Just confuses me where we call it. Not against the current implementation as there are slight differences, more of a question that popped up in my head.

plugins/module_utils/foreman_helper.py
mdellweg: Not as such. We only call flatten once on either entities. But yes, if the api answer contained e.g. `fullname` and we decide to call it `description` (for whatever reason?!?) we'd end up with something like 
https://github.com/theforeman/foreman-ansible-modules/blob/f169172326368851d4836f0d413b9898fcdb2d25/plugins/modules/foreman_domain.py#L110-L111

For referenced entities, we get the hydrated nested (thin) entities in the api answer, like `domain: {id:1}`, therefore we can employ the same flattening mechanism there.

plugins/modules/foreman_operating_system.py
mdellweg: We also need to specify all other properties, that are different in the api answer.

mdellweg:
> > Instead of passing in params=search_params, why don't we mark some params as required in the spec?
> 
> Sounds like a practical idea.

I am no longer sure, how this should look like. Would you be ok with implementing that in a separate PR? @evgeni

evgeni:
Sure.

plugins/module_utils/foreman_helper.py
```diff
@@ -285,14 +287,108 @@ def ensure_resource(self, resource, entity_dict, entity, state=None, name_map=No
...
+        current_entity = _flatten_entity(current_entity, entity_spec)
+        for key, value in desired_entity.items():
+            if key in current_entity and current_entity[key] == value:
```
```suggestion
            if current_entity.get(key) == value:
```

Is that more readable? Or do we prefer the explicit `in` check? Thinking out aloud.

plugins/module_utils/foreman_helper.py
```diff
@@ -364,6 +462,35 @@ def sanitize_entity_dict(entity_dict, name_map):
...
 
 
+def _flatten_entity(entity, entity_spec):
+    """Flatten entity according to spec"""
+    result = {}
+    for key, value in entity.items():
+        if key in entity_spec:
+            spec = entity_spec[key]
+            flat_name = spec.get('flat_name', key)
+            _type = spec.get('type', 'str')
```
evgeni: Maybe `param_type`? I don't like starting vars with `_`.

plugins/module_utils/foreman_helper.py
evgeni: OK

plugins/module_utils/foreman_helper.py
```diff
@@ -285,14 +287,108 @@ def ensure_resource(self, resource, entity_dict, entity, state=None, name_map=No
...
+            if current_entity is None:
+                changed, current_entity = self._create_entity(resource, desired_entity, params, entity_spec)
+            else:
+                changed, current_entity = self._update_entity(resource, desired_entity, current_entity, params, entity_spec)
+        elif state == 'absent':
+            if current_entity is not None:
+                changed, current_entity = self._delete_entity(resource, current_entity, params)
+        else:
+            self.fail_json(msg='Not a valid state: {}'.format(state))
+        return changed, current_entity
```
evgeni: It kinda feels weird that you return `current_entity` and not `updated_entity`. Don't think it hurts at all, just found it weird when reading that you technically overwrite `current`.

evgeni:
@mdellweg thanks A LOT for the new ensure functions, they read a lot better than the ones I copy-wasted from our old interface.
Could you also migrate another entity (something small like domain or architecture) to the new functions? Two questions I want to answer by that: will we need to re-record tapes? is the migration just changing the spec and the function call? (I guess the answers are no and yes, but I'd like to see them) :)

plugins/module_utils/foreman_helper.py
mdellweg: I thought about calling these things _properties_.

plugins/module_utils/foreman_helper.py
mdellweg: What can possibly go wrong? Value will _never_ be `None` because it comes from `clean_params`.

plugins/module_utils/foreman_helper.py
mdellweg: I can make it copy_on_write. If it helps readabilitiy.

plugins/module_utils/foreman_helper.py
mdellweg: And in cas we do not get the hydrated version from the api, it looks like:
https://github.com/theforeman/foreman-ansible-modules/blob/c88ffc2e10a8e7a486421ce22f62407bb24c0009/plugins/modules/foreman_hostgroup.py#L138-L139

mdellweg:
@evgeni I worked on some more modules to add the nested parameters, where i needed to employ this api, because tho old one did not respond with the entity id for some reason. Have a look at #357 .
I know, it's grown confusing. I can have all of this in one PR, if you prefer.

plugins/module_utils/foreman_helper.py
evgeni: I can live with either.

plugins/module_utils/foreman_helper.py
evgeni: I don't think you have to, but you certainly can :)

evgeni:
test failure should be fixed by https://github.com/theforeman/foreman-ansible-modules/pull/371

mdellweg:
Tests are fixed again. Do you want to renew your review? Or otherwise state, that it is / is not ready for merge?
@evgeni 
I want to get this merged soon, as i think it solves some real world problems around nested entities.

mdellweg:
Architecture migrated without rerecording. ;)

plugins/modules/foreman_architecture.py
```diff
@@ -105,7 +107,7 @@ def main():
 
     entity = module.find_resource_by_name('architectures', name=entity_dict['name'], failsafe=True)
 
-    changed = module.ensure_resource_state('architectures', entity_dict, entity)
+    changed, _ = module.ensure_entity('architectures', entity_dict, entity)
```
evgeni: do we want to have a shortcut function for that, that only returns `changed`? like `ensure_entity_state`?

mdellweg: I guess, i was too lazy. Will add it.

plugins/module_utils/foreman_helper.py
evgeni: I think we can make that nicer, but that would make this PR even bigger, so let's discuss that in #357

plugins/modules/foreman_architecture.py
mdellweg: On the other hand i dream of always returning the changed entity from the module.

evgeni: I have a "implement diff mode for all apypie modules" todo entry, so at some point... but until then the shortcut is handy :)

mdellweg: Ahh, that is why you are interested in the `changed_entity`. ;)

PR #356, opened 2019-07-16
Port foreman_ptable to apypie

mdellweg:


evgeni:
This still has nailgun imports?

mdellweg:
Oh, i forgot that.

PR #357, opened 2019-07-16
Reusable parameters

mdellweg:
I had the thought, that this piece of logic could be reused across entities that can define parameters.
Subnet, Domain, HostGroup come to my mind.

### DIFF
--- a/plugins/module_utils/foreman_helper.py
+++ b/plugins/module_utils/foreman_helper.py
@@ -26,6 +26,13 @@ except ImportError:
     APYPIE_IMP_ERR = traceback.format_exc()
 
 
+parameter_entity_spec = {
+    'name': {},
+    'value': {},
+    'parameter_type': {},
+}
+
+
 class ForemanBaseAnsibleModule(AnsibleModule):
 
     def __init__(self, argument_spec, **kwargs):
@@ -368,9 +375,14 @@ class ForemanApypieAnsibleModule(ForemanBaseAnsibleModule):
                 Pair of boolean indicating whether something changed and the new current state if the entity
         """
         payload = _flatten_entity(desired_entity, entity_spec)
-        if params:
-            payload.update(params)
-        return self._resource_action(resource, 'create', payload)
+        if not self.check_mode:
+            if params:
+                payload.update(params)
+            return self._resource_action(resource, 'create', payload)
+        else:
+            fake_entity = desired_entity.copy()
+            fake_entity['id'] = -1
+            return True, fake_entity
 
     def _update_entity(self, resource, desired_entity, current_entity, params, entity_spec):
         """Update a given entity with given properties if any diverge
@@ -388,15 +400,21 @@ class ForemanApypieAnsibleModule(ForemanBaseAnsibleModule):
         desired_entity = _flatten_entity(desired_entity, entity_spec)
         current_entity = _flatten_entity(current_entity, entity_spec)
         for key, value in desired_entity.items():
-            if current_entity.get(key) == value:
-                continue
-            payload[key] = value
+            if current_entity.get(key) != value:
+                payload[key] = value
         if payload:
             payload['id'] = current_entity['id']
-            if params:
-                payload.update(params)
-            return self._resource_action(resource, 'update', payload)
+            if not self.check_mode:
+                if params:
+                    payload.update(params)
+                return self._resource_action(resource, 'update', payload)
+            else:
+                # In check_mode we emulate the server updating the entity
+                fake_entity = current_entity.copy()
+                fake_entity.update(payload)
+                return True, fake_entity
         else:
+            # Nothing needs changing
             return False, current_entity
 
     def _revert_entity(self, resource, current_entity, params):
@@ -427,7 +445,8 @@ class ForemanApypieAnsibleModule(ForemanBaseAnsibleModule):
         payload = {'id': current_entity['id']}
         if params:
             payload.update(params)
-        return self._resource_action(resource, 'destroy', payload)
+        self._resource_action(resource, 'destroy', payload)
+        return True, None
 
     def _resource_action(self, resource, action, params):
         resource_payload = self.foremanapi.resource(resource).action(action).prepare_params(params)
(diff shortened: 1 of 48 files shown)


mdellweg:
`$ cat foreman.json | jq '.docs.resources.parameters.methods[0].apis|map({api_url})[]'`
```json
{
  "api_url": "/api/hosts/:host_id/parameters"
}
{
  "api_url": "/api/hostgroups/:hostgroup_id/parameters"
}
{
  "api_url": "/api/domains/:domain_id/parameters"
}
{
  "api_url": "/api/operatingsystems/:operatingsystem_id/parameters"
}
{
  "api_url": "/api/locations/:location_id/parameters"
}
{
  "api_url": "/api/organizations/:organization_id/parameters"
}
{
  "api_url": "/api/subnets/:subnet_id/parameters"
}
```
So i can add host, location and organization to the list of entities to benefit.

plugins/module_utils/foreman_helper.py
```diff
@@ -285,14 +295,132 @@ def ensure_resource(self, resource, entity_dict, entity, state=None, name_map=No
...
+        desired_entity = _flatten_entity(desired_entity, entity_spec)
+        current_entity = _flatten_entity(current_entity, entity_spec)
+        for key, value in desired_entity.items():
+            # query entity_spec[key] for special treatment
+            spec = entity_spec.get(key)
+            if spec:
+                # properties with 'force_update' must always be updated
+                if spec.get('force_update', False):
+                    payload[key] = value
+                    continue
```
mdellweg: This is an example for values, that need a special treatment. Though i do not know whether `force_update` is really needed, because if the key is in `desired_entity` and not in `current_entity` (i.e. not reported by the server), it will end up in payload by the logic further down. Also i do not understand the conceptional difference between `force_update` and `check_missing`. So i guess, we don't need either of them.

plugins/module_utils/foreman_helper.py
```diff
@@ -285,14 +295,132 @@ def ensure_resource(self, resource, entity_dict, entity, state=None, name_map=No
...
+                # properties with 'check_type' ???
+                # if spec.get('check_type', False):
+                #     # TODO
+                #     pass
```
mdellweg: I do not fully understand, how `check_type` is supposed to work, but is only used in the settings module. And we might be able to so something even more clever there, because the entity provided by the server contains a hint to the type of the setting.

plugins/module_utils/foreman_helper.py
```diff
@@ -285,14 +295,132 @@ def ensure_resource(self, resource, entity_dict, entity, state=None, name_map=No
...
+            payload['id'] = current_entity['id']
+            if not self.check_mode:
+                if params:
+                    payload.update(params)
+                return self._resource_action(resource, 'update', payload)
+            else:
+                # In check_mode we emulate the server updating the entity
+                fake_entity = current_entity.copy()
+                fake_entity.update(payload)
+                return True, fake_entity
```
mdellweg: Are modules in check_mode supposed to provide useful output apart from the changed status?

plugins/module_utils/foreman_helper.py
sean797: `check_missing` was introduced for Content View Filters, depending on the type of CVF we get a different set of attributes returned.  

https://github.com/theforeman/foreman-ansible-modules/blob/c92cf72879f765e98a2059291c2043b36ae5218d/plugins/module_utils/foreman_helper.py#L243-L246

https://github.com/Katello/katello/blob/master/app/views/katello/api/v2/content_view_filter_rules/show.json.rabl


I dont know the history of `force_update`

mdellweg: I think, by looping over the user provided properties, we should end up including everything, the user wanted to have. But i guess/hope, we will see, whether it works when we get to the tests there.

plugins/modules/foreman_hostgroup.py
```diff
@@ -120,10 +154,16 @@ def main():
             description=dict(),
             architecture=dict(),
             operatingsystem=dict(),
-            media=dict(),
+            medium=dict(),
```
evgeni: can you add an alias for `media` so we don't break existing playbooks?

plugins/module_utils/foreman_helper.py
evgeni: They *can* provide the diff if run with `--check --diff`

plugins/modules/foreman_hostgroup.py
```diff
@@ -102,14 +131,19 @@
...
-    'operatingsystem': 'operatingsystem_id',
-    'media': 'medium_id',
-    'ptable': 'ptable_id',
-    'root_pass': 'root_pass',
+entity_spec = {
+    'id': {},
+    'name': {},
+    'description': {},
+    'architecture': {'type': 'entity', 'flat_name': 'architecture_id'},
+    'architecture_id': {},
```
evgeni: So I dislike this duplication (plus I have the vision that some day we'd use `entity_spec` to generate Ansible's `argument_spec`…).

How about calculating this?

```python
def _flatten_entity(entity, entity_spec):
    """Flatten entity according to spec"""
    extended_entity_spec = {entry['flat_name']:{} for entry in entity_spec.values() if 'flat_name' in entry} # le magic
    extended_entity_spec.update(entity_spec)
    result = {}
    for key, value in entity.items():
        if key in extended_entity_spec:
            spec = extended_entity_spec[key]
            _type = spec.get('type', 'str')
            if _type == 'entity':
                result[flat_name] = value['id']
            elif _type == 'entity_list':
                result[flat_name] = sorted(val['id'] for val in value)
            else:
                result[flat_name] = value
    return result
```

plugins/module_utils/foreman_helper.py
mdellweg: So this might help with check mode. 8)

evgeni: yeah

plugins/module_utils/foreman_helper.py
```diff
@@ -360,15 +380,28 @@ def _update_entity(self, resource, desired_entity, current_entity, params, entit
         desired_entity = _flatten_entity(desired_entity, entity_spec)
         current_entity = _flatten_entity(current_entity, entity_spec)
         for key, value in desired_entity.items():
-            if current_entity.get(key) == value:
-                continue
-            payload[key] = value
+            # query entity_spec[key] for special treatment
+            # spec = entity_spec.get(key)
+            # if spec:
+                # properties with 'check_type' ???
```
```
plugins/module_utils/foreman_helper.py:386:17: E117 over-indented (comment)
plugins/module_utils/foreman_helper.py:387:17: E117 over-indented (comment)
plugins/module_utils/foreman_helper.py:388:17: E117 over-indented (comment)
plugins/module_utils/foreman_helper.py:389:17: E117 over-indented (comment)
```

:stuck_out_tongue:

mdellweg: I guess we should not bother this, until we need it.

plugins/module_utils/foreman_helper.py
```diff
@@ -384,7 +403,8 @@ def _delete_entity(self, resource, current_entity, params):
         payload = {'id': current_entity['id']}
         if params:
             payload.update(params)
-        return self._resource_action(resource, 'destroy', payload)
+        self._resource_action(resource, 'destroy', payload)
+        return True, None
```
evgeni: isn't this identical to the previous code?

mdellweg: No, the previous code returned the deleted entity. This now states: There is no entity (after deletion). We do not use the return value atm iirc.

evgeni:
I like this :)

Probably the only drawback here is the quite high duplication in the `entity_spec` (due to the `_ids` entries) that will be again removed in #376 -- so I wonder if we should get that in first? Or at least a basic part of it?

mdellweg:
I removed that from this PR, because it went out of scope in a crazy way. I don't mind the order in which we want to merge those two.

evgeni:
Can you please update this to what you implemented in #376? Otherwise it's good to go I think.

mdellweg:
I thought i did, but i missed hostgroup. Do you see anything else?

evgeni:
That's the only one missing, yeah.

evgeni:
@mdellweg I'd prefer not to squash these, as the commits make sense on their own. can you squash the last fixup please and "rebase and merge" it afterwards?

Issue #355, opened 2019-07-16
katello_repository_set can not enable repo. set that has no substitutions


ogajduse:
##### SUMMARY

`katello_repository_set` module can not enable repository set that has no substitutions available.

##### ISSUE TYPE
 - Bug Report

##### ANSIBLE VERSION

```
ansible 2.8.1
  config file = /home/ogajduse/foreman/foreman-ansible-modules/ansible.cfg
  configured module search path = ['/home/ogajduse/foreman/foreman-ansible-modules/plugins/modules']
  ansible python module location = /usr/lib/python3.7/site-packages/ansible
  executable location = /usr/bin/ansible
  python version = 3.7.3 (default, May 11 2019, 00:38:04) [GCC 9.1.1 20190503 (Red Hat 9.1.1-1)]

```

##### KATELLO/FOREMAN VERSION

```
katello-3.12.0-0.5.master.el7sat.noarch
foreman-1.22.0.6-1.el7sat.noarch
```

##### NAILGUN VERSION

```
nailgun==0.32.0
```

##### STEPS TO REPRODUCE

```yaml
- hosts: all
  gather_facts: no
  vars:
    ansible_python_interpreter: ./foreman-ansible-modules-venv/bin/python3
    repository_sets:
      - label: satellite-tools-6.5-for-rhel-8-x86_64-rpms
        repositories:
          - releasever: "8"
      - label: satellite-tools-6.5-for-rhel-8-x86_64-rpms
        repositories: []
  tasks:
    - name: Enable Satellite tools RPMs repositories
      katello_repository_set:
        username: "{{ username }}"
        password: "{{ password }}"
        server_url: "{{ server_url }}"
        verify_ssl: "{{ verify_ssl }}"
        label: "{{ item.label }}"
        organization: "{{ pit_org_name }}"
        repositories: "{{ item.repositories }}"
        state: enabled
      loop: "{{ repository_sets }}"

```

##### EXPECTED RESULTS

Enabled repository set without specifying `repositories` variable in the module.
##### ACTUAL RESULTS

No repository enabled by the module.

Output with 
```yaml
repositories:
  - releasever: "8"
```
is
```
ok: [localhost] => (item={'label': 'satellite-tools-6.5-for-rhel-8-x86_64-rpms', 'repositories': [{'releasever': '8'}]}) => changed=false 
  ansible_loop_var: item
  invocation:
    module_args:
[...]

PR #359, opened 2019-07-17
do not require repositories arg

ogajduse:
There should be a case when there is only one repo
available and no desired substitutions.
For example: satellite-tools-6.5-for-rhel-8-x86_64-rpms

Addresses #355 
@akofink @evgeni Could you have a look?


mdellweg:
I don't have enough expertise with repository sets. But can you try to rebuild the special case in the test_playbook?

evgeni:
I like Andrews compromise.

evgeni:
I posted an alternative in https://github.com/theforeman/foreman-ansible-modules/pull/650

evgeni:
closing as #650 is merged

PR #360, opened 2019-07-17
Fix updating of resources with nested associations

Fodoj:
Previously values to be passed to API were flattened on creation, but not on updates. This leads to inability to update certain fields of probably quite a few modules. With this change values are flattened also on the update.

### DIFF
--- a/plugins/module_utils/foreman_helper.py
+++ b/plugins/module_utils/foreman_helper.py
@@ -361,7 +361,7 @@ class KatelloEntityApypieAnsibleModule(ForemanEntityApypieAnsibleModule):
 
 def sanitize_entity_dict(entity_dict, name_map):
     name_map['id'] = 'id'
-    return {value: entity_dict[key] for key, value in name_map.items() if key in entity_dict}
+    return {key: entity_dict[value] for key, value in name_map.items() if value in entity_dict}
 
 
 # Helper for templates
--- a/plugins/module_utils/foreman_helper.py
+++ b/plugins/module_utils/foreman_helper.py
@@ -249,7 +249,7 @@ class ForemanApypieAnsibleModule(ForemanBaseAnsibleModule):
             new_data = {'id': entity_id}
             for key, value in volatile_entity.items():
                 if key in fields:
-                    new_data[key] = value
+                    new_data[key] = self._flatten_value(value)
             return self._resource_action(resource, 'update', params=new_data)
         return False, result
 


mdellweg:
Right now, any touching of this update function scares me to the bones.
OTOH it sounds reasonable.

evgeni:
I think the change makes sense. Could you provide a test or tell us in which scenario the additional flattening is required? I think most fields should be flattened in https://github.com/theforeman/foreman-ansible-modules/blob/master/plugins/module_utils/foreman_helper.py#L228 already, but I can see cases (from the code) when they are not because their key is not in entity_dict, I just can't come up with an actual example.

Or maybe we should just create volatile entity by flattening all of old entity at the beginning of the function?

Fodoj:
Even though they are flattented here https://github.com/theforeman/foreman-ansible-modules/blob/master/plugins/module_utils/foreman_helper.py#L228, they are not flatteted for the fields where check_missing is enabled, for instance. So the first conditional block, the one that flattens values, can be skipped due to key being not there, and then key (and value) appears later inside `if check_missing is not None:` block.

One example would be any field that references external resource in Hostgroup. I did initial PR  for adding check_missing to Hostgroup, but I still need to polish it and fix tests: https://github.com/theforeman/foreman-ansible-modules/pull/361.

Please tell how I should proceed. I can sure update the code to flatten it all in the beginning and then add some specs.

Fodoj:
I found another bug: in sanitize_entity_dict key and value needed to be swapped. Otherwise look up will fail, because `entity_dict` has _api_ fields as keys, and these are _values_ in `name_map`. Let's see if test pass. In my scenario, every entry was always marked as `changed` because of this.

Fodoj:
Ok, last fix appears not to be that straightforward. I will dig deeper into it and will also add some extra tests.

mdellweg:
@Fodoj I am trying to reinvent the whole ensure mechanism in #354 , because the current version emerged from using Nailgun entities.
Hopefully the new way will be better suited to work with dictionaries on both sides.
Maybe you want to take a look.

Fodoj:
@mdellweg then I will close this one as the code will be essentially removed anyways.

mdellweg:
@Fodoj Well, it is not clear, the new code will be accepted. It still needs more review.
If you want to give it a try, there is some more progress within #357 .
It would be interesting, if you can demonstrate whether it fixes the issues you have.

PR #361, opened 2019-07-17
Add parent attribute for the hostgroup module

Fodoj:
Enable nested hostgroups management via parent parameter. Also added check_missing so that parent can be assigned on update and not just on create (note that same needs to be done for architecture, operatingsystem, media and ptable, but I didn't want to bundle too many changes in single PR).


mdellweg:
> (note that same needs to be done for architecture, operatingsystem, media and ptable, but I didn't want to bundle too many changes in single PR).

Those entities are not nestable iirc. So i don't get the point.

Fodoj:
> Those entities are not nestable iirc. So i don't get the point.

Sorry, I was not clear. It's not about those entities being nested, it's about them being unupdateable once you set them (or if you don't set them on creation).

If you create the hostgroup without ptable and then you add ptable to the yaml, then it won't be set, because `check_missing` for ptable is not in place.

mdellweg:
Ahh, so i misread it to be a problem of the architecture, opertingsystem, you_name_it modules.
If this is all about fixing the same problem multiple times in the same module, just go for it.

tests/test_playbooks/fixtures/hostgroup-0.yml
```diff
@@ -6282,7 +6282,7 @@ interactions:
       Accept: [application/json;version=2]
       Cookie: [_session_id=368cd63b7e87604931a457c832b26c60]
     method: GET
-    uri: https://centos7-katello-3-12/api/operatingsystems?per_page=4294967296&search=name%3D%22TestOS%22&thin=True
+    uri: https://centos7-katello-3-12/api/operatingsystems?per_page=4294967296&search=title%3D%22TestOS%22&thin=True
```
mdellweg: Oh, those files are _not_ meant to be edited by hand.
Have a look at:
https://github.com/theforeman/foreman-ansible-modules/blob/master/tests/README.md#guidedeline-to-writing-tests

Fodoj: Do you have a quick guide on setting up test environment just for this? I am testing changes on real env, but I dont want to generate VCR yamls out of it

evgeni: https://github.com/theforeman/forklift can easily deploy vagrant VMs for you

mdellweg: OK, this section of the main README might help:
https://github.com/theforeman/foreman-ansible-modules#how-to-test-modules-in-this-repository

I can really recommend using forklift here: `vagrant up centos7-katello-2.12`. Keep in mind, that the hostname would be `centos7-katello-2-12` with no dot.
In the best case, you would record this test against one of the foreman hosts, but i guess that is impossible until the operatingsystem module uses apypie.

Fodoj: > In the best case, you would record this test against one of the foreman hosts, but i guess that is impossible until the operatingsystem module uses apypie.

Could you please elaborate on that? So I should test against katello, not foreman, otherwise it won't work, right?

Fodoj: I think you meant katello 3.12, not 2.12, is that right?

evgeni: Yes. Both is correct. We have modules that don't work against plain Foreman, despite being officially for Foreman (the underlying library is Katello/Satellite centric). Ans yes 3.12 is the latest Katello release and what should be used.

Fodoj:
@mdellweg @evgeni for time being I removed `check_missing`, as it uncovers the bug with `sanitize_dict_entity` method that will always report `changed` for any resource that has nested resources, like parent id. I might tackle it in a different pull request. 

I also updated searching for OS from using `name` to `title` attribute. `name` is not unique - for "Centos 7.1" and "Centos 7.2" name would be "Centos" in both cases, while title is always different.

Fodoj:
Closing in favor of https://github.com/theforeman/foreman-ansible-modules/pull/366. cc @evgeni @mdellweg

PR #362, opened 2019-07-17
don't make the module files executable

evgeni:


ekohl:
They have a shebang and a main so technically they can be executed.

evgeni:
> They have a shebang and a main so technically they can be executed.

I *think* both are Ansible conventions, not a statement that these files can be used "standalone".

mdellweg:
My editor automatically marks files executable if they have a shebang. As i guess, ansible is meant to run them with a _hand picked_ interpreter anyway, please also remove the shebangs.

evgeni:
So all Ansible "proper" modules have a shebang, and the docs also say (without saying why) to have a shebang.

I'll talk to a few Ansible people and come back to this then.

sean797:
I think it's because you **_can_** create modules in any language.

evgeni:
So after consulting the right docs, the modules (well, the Python ones) *must* have a shebang of `/usr/bin/python` (so that Ansible can easily adjust that to the *real* interpreter it will be using) and *must not* be executable. See https://docs.ansible.com/ansible/latest/dev_guide/testing/sanity/shebang.html

sean797:
Can you add `--test shebang` to 

https://github.com/theforeman/foreman-ansible-modules/blob/c176e97dc9e073602a16e8b0e8e0a125350db478/tests/extras/sanity.yml#L27-L28

evgeni:
@sean797 sure, done

mdellweg:
For those suffering from this change like me:
```vim
" Set executable flag on file if shebang is found
" Exclude ansible modules in a collection
func! MakeScriptExecutable()
  if getline(1) =~ "^#!"
    if expand("%:p:h") =~ "plugins/modules"
      silent !chmod a-x <afile>
    else
      silent !chmod a+x <afile>
    endif
  endif
endfunc
autocmd BufWritePost * :call MakeScriptExecutable()
```

Issue #244, opened 2019-02-27
refactor affects_multiple logic into a helper


evgeni:
##### SUMMARY

as discussed in #240, the affects_multiple logic is currently repeated in every module that supports it, but it should be moved to the helper functions we have.

##### ISSUE TYPE
 - Feature Idea

PR #364, opened 2019-07-18
Port foreman_provisioning_template to apypie

mdellweg:


plugins/modules/foreman_ptable.py
```diff
@@ -320,11 +321,11 @@ def main():
 
     if affects_multiple:
         entities = module.list_resource('ptables')
-        if not module.desired_absent:  # not 'thin'
-            entities = [module.show_resource('ptables', entity['id']) for entity in entities]
         if not entities:
             # Nothing to do; shortcut to exit
             module.exit_json(changed=False)
+        if not module.desired_absent:  # not 'thin'
```
evgeni: This does not really belong in here ;-)

plugins/modules/foreman_provisioning_template.py
```diff
@@ -360,59 +338,48 @@ def main():
...
     module.connect()
 
-    try:
-        if affects_multiple:
-            entities = find_entities(ProvisioningTemplate)
-        else:
-            entities = find_entities(ProvisioningTemplate, name=entity_dict['name'])
-    except Exception as e:
-        module.fail_json(msg='Failed to search for entities: %s ' % e)
+    if affects_multiple:
```
evgeni: Thinking aloud: I've seen this block multiple times now, time for a helper func?

plugins/modules/foreman_provisioning_template.py
```diff
@@ -360,59 +338,48 @@ def main():
...
 
     if not affects_multiple:
-        entity_dict = find_template_kind(entity_dict, module)
-
-    entity_dict = sanitize_entity_dict(entity_dict, name_map)
+        entity_dict = find_template_kind(module, entity_dict)
 
     changed = False
     if not affects_multiple:
-        if len(entities) == 0:
```
evgeni: same comment as the previous "multiple" block :)

evgeni:
I like! Tell me if you want address the comments here or if we should do them later.

plugins/modules/foreman_ptable.py
mdellweg: It came along as a minor optimisation.
And i have two other PR's open that grew a lot more out their original scope. %)

plugins/modules/foreman_provisioning_template.py
mdellweg: Apropos growing out of scope. I think this can be another pr (without rerecording, if done right).

evgeni: Sure, we even have an old issue for that (https://github.com/theforeman/foreman-ansible-modules/issues/244) from when y'all added that multiple logic.

PR #366, opened 2019-07-22
Hostgroup enhancement

bagasse:
Add following hostgroup fields support:
* parent
* organizations
* locations
* subnets (IPv4 and IPv6)
* domain
* realm
* compute_resource
* compute_profile
* pxe_loader
* environment
* config_groups
* puppet_proxy
* puppet_ca_proxy

Fixes ptables and operating_systems tests issues by adding some attributes for ptables and operating_systems tasks.

Not tested via test_playbooks:
* realm (no setup available in my test env)


bagasse:
Just saw that #361 was already submited with parent attribute...

Fodoj:
My PR can be closed then, as this one adds much more. One comment from my side - I would fix operating systems here as well, so that they are searched by title, not by name.

Fodoj:
Also adding opescap_proxy would be great

bagasse:
> Also adding opescap_proxy would be great

@Fodoj , we use foreman_openscap also, but I think handling non core proxies need some reflexion, so I wanted to do this in separated PR.

I  fixed operating system and ptables tests (there was some missing attributes in os and ptables tasks to make things works)

bagasse:
Current test failures related to #369

bagasse:
Rebased to fix sanity checks.

bagasse:
Added test for config_groups param.

plugins/modules/foreman_hostgroup.py
```diff
@@ -118,31 +211,99 @@ def main():
...
+
+        if 'realm' in entity_dict:
+            entity_dict['realm'] = module.find_resource_by_name('realms', name=entity_dict['realm'], failsafe=False, thin=True)
+
         if 'architecture' in entity_dict:
             entity_dict['architecture'] = module.find_resource_by_name('architectures', name=entity_dict['architecture'], failsafe=False, thin=True)
 
         if 'operatingsystem' in entity_dict:
-            entity_dict['operatingsystem'] = module.find_resource_by_name('operatingsystems', name=entity_dict['operatingsystem'], failsafe=False, thin=True)
+            entity_dict['operatingsystem'] = module.find_resource_by_title('operatingsystems', title=entity_dict['operatingsystem'], failsafe=False, thin=True)
```
evgeni: Maybe out of scope for this PR, but in other places we use `title ~ STRING` when searching for OSes:
https://github.com/theforeman/foreman-ansible-modules/blob/61fb6c25cfb894a03290192fbd638490b3c8cdae/plugins/modules/foreman_installation_medium.py#L144-L145

https://github.com/theforeman/foreman-ansible-modules/blob/61fb6c25cfb894a03290192fbd638490b3c8cdae/plugins/modules/foreman_architecture.py#L103-L104

While `find_resource_by_title` uses `=`:
https://github.com/theforeman/foreman-ansible-modules/blob/61fb6c25cfb894a03290192fbd638490b3c8cdae/plugins/module_utils/foreman_helper.py#L186-L188

I think the idea behind this was to make the OS search a bit fuzzier, as OS names/titles are weird.

Given you didn't write this code in the first place in this PR, I don't think we need a decision/solution here now, but I wanted to raise awareness for this.

Maybe also @mdellweg can chime in.

plugins/modules/foreman_hostgroup.py
```diff
@@ -118,31 +211,99 @@ def main():
...
+        if 'config_groups' in entity_dict:
+            entity_dict['config_groups'] = module.find_resources_by_name('config_groups', entity_dict['config_groups'], failsafe=False, thin=True)
+
+        if 'puppet_proxy' in entity_dict:
+            entity_dict['puppet_proxy'] = module.find_resource_by_name('smart_proxies', name=entity_dict['puppet_proxy'], failsafe=False, thin=True)
+
+        if 'puppet_ca_proxy' in entity_dict:
+            entity_dict['puppet_ca_proxy'] = module.find_resource_by_name('smart_proxies', name=entity_dict['puppet_ca_proxy'], failsafe=False, thin=True)
+
+    entity = module.find_resource('hostgroups', search='title="{}"'.format(build_fqn(name, parent)), failsafe=True)
```
evgeni: Any reason not to use `find_resource_by_title` here?

evgeni:
This is great @bagasse! I've added two comments, but nothing major ;)

As for OpenSCAP, yeah this sounds like a good addition, I don't mind it being in this PR or an own one. Just remember that we'll need to use a different `apidoc.json` then ;)

CC @Manisha15 @Fobhep @mdellweg if anyone of you wants to review this too!

plugins/modules/foreman_hostgroup.py
mdellweg: Oh yes, definitely `~` for comparison, please It seems to come up in more and more places. I know we chatted about that, @evgeni , but how about _some_ support for this in the helper?

plugins/modules/foreman_hostgroup.py
bagasse: No reason. Changed.

plugins/modules/foreman_hostgroup.py
bagasse: @evgeni, @mdellweg  I think `~` for comparison to get something can go to not predictable behavior because it can return more than one entity and the code in foreman_helper enforce that we get only one entity when we search an entity (eg searching title ~ CentOS on one of my systems will return 13 entries). I prefer the same behavior as for names, if we don't find any entity with the exact name, it fail (or return None depending on failsafe value).
The title search is mostly used for searching nested resources, operating system is an exception, if we change the helper implementation to use `~` operator, it will break nested resources search.

bagasse:
> This is great @bagasse! I've added two comments, but nothing major ;)
> 
> As for OpenSCAP, yeah this sounds like a good addition, I don't mind it being in this PR or an own one. Just remember that we'll need to use a different `apidoc.json` then ;)

@evgeni It would be nice to add extensible smart_proxy support if we can. We use katello and it would be nice to add content proxy also, but maybe we have to make a subclass katello_hostgroup for that ? I didn't wanted to make this happen in this PR that I think  is big enough :)

plugins/modules/foreman_hostgroup.py
evgeni: Yeah, the *default* behavior should definitely be `=`. It's just with OperatingSystems we've used fuzzy (`~`) search to find minor versions if the users passes only parts of it (`CentOS 7.3` → `CentOS 7.3.1234`).

Also, `find_resource` will only ever allow the user to get one result and abort if there are muplitple:
https://github.com/theforeman/foreman-ansible-modules/blob/d4e5ce331f1a26fd7319b8f1ff422ec6b4903577/plugins/module_utils/foreman_helper.py#L171-L176

How about we add a flag to `find_resource_by_title` to optionally allow fuzzy (`~`) search. And then add an explicit `find_operatingsystem` that will first try an exact search, and if that fails do a fuzzy one?

bagasse: Ok, so if it's ok for you I would prefer to manage this in another PR that will change how we manage OSes in all places to be consistent and go ahead as is  for now :)

mdellweg: Fine for me.

evgeni: Absolutely!

evgeni: I've opened https://github.com/theforeman/foreman-ansible-modules/issues/373 for that.

bagasse:
@mdellweg, @evgeni thanks for your time !

Fodoj:
@bagasse do you have some hints for setting up test environment for this? I started working on puppetclasses support, but it seems that now hostgroups require much bigger test infra - dns proxy, compute resource provider etc. Bare forklift install is not enough anymore.

bagasse:
@Fodoj, Yes, I install and start `libvirtd` on the foreman server after the box setup for compute resource. For the dns proxy part I've a custom box that add some forman installer options like `--foreman-proxy-dns=true`. My setup is mostly discussed in #375 sorry I didn't find the time to go forward on this issue for now...

Fodoj:
Thanks for info, I will proceed with this. How do I install libvirt provider? Re-run foreman-installer?
Am 26. Okt. 2019, 12:18 +0200 schrieb Baptiste Agasse <notifications@github.com>:

PR #368, opened 2019-07-23
Port foreman_os_default_templates to apypie

mdellweg:


plugins/modules/foreman_os_default_template.py
```diff
@@ -126,35 +103,30 @@ def main():
...
-    (entity_dict, state) = module.parse_params()
+    entity_dict = module.clean_params()
 
     module.connect()
 
-    entity_dict['operatingsystem'] = find_operating_system_by_title(module, entity_dict['operatingsystem'])
-    entity_dict['template_kind'] = find_entities_by_name(
-        TemplateKind, [entity_dict['template_kind']], module)[0]
+    search = 'title~"{}"'.format(entity_dict['operatingsystem'])
+    entity_dict['operatingsystem'] = module.find_resource('operatingsystems', search, thin=True)
```
evgeni: Given my comment in https://github.com/theforeman/foreman-ansible-modules/pull/366#discussion_r306633638, I really think we should move that to a helper now ;)

plugins/modules/foreman_os_default_template.py
```diff
@@ -80,42 +80,19 @@
...
-        'provisioning_template': 'provisioning_template',
-    }
-    result = {}
-    for key, value in name_map.items():
-        if key in entity_dict:
-            result[value] = entity_dict[key]
-    return result
+name_map = {
+    'operatingsystem': 'operatingsystem',
+    'template_kind': 'template_kind',
```
evgeni: I "like" that we use `template_kind` here, but `kind` in the provisioning template. Time to consolidate? :)

evgeni:
needs rebase (prov template was merged) and rerecord :)

plugins/modules/foreman_os_default_template.py
mdellweg: How about a fuzzy Flag?

evgeni: Sounds good

bagasse: I given my comment in https://github.com/theforeman/foreman-ansible-modules/pull/366#discussion_r306678268 on this too :)

Issue #369, opened 2019-07-23
Sanity tests fails on ansible devel branch


bagasse:
##### SUMMARY

Sanity checks are failing on ansible devel branch. Fixing branch to stable-2.8 in `tests/extras/sanity.yml` make tests pass again.

##### ISSUE TYPE
 - Bug Report

##### STEPS TO REPRODUCE

```
make sanity
```

##### ACTUAL RESULTS

```
fatal: [fixtures]: FAILED! => {"changed": true, "cmd": ["/home/bagasse/projects/contrib/foreman/foreman-ansible-modules/tests/extras/../../.tmp/ansible/test/runner/ansible-test", "sanity", "--local", "--python", "2.7", "--test", "ansible-
doc", "--test", "compile", "--test", "shebang", "lib/ansible/modules/foreman-test/", "lib/ansible/module_utils/foreman-test/"], "delta": "0:00:00.088052", "end": "2019-07-23 11:16:00.140537", "msg": "non-zero return code", "rc": 1, "start
": "2019-07-23 11:16:00.052485", "stderr": "ERROR: The current working directory must be at or below one of:\n\n - Ansible source: /home/bagasse/projects/contrib/foreman/foreman-ansible-modules/.tmp/ansible/\n - Ansible collection: {...}/
ansible_collections/{namespace}/{collection}/\n\nCurrent working directory: /home/bagasse/projects/contrib/foreman/foreman-ansible-modules/tests/extras", "stderr_lines": ["ERROR: The current working directory must be at or below one of:",
 "", " - Ansible source: /home/bagasse/projects/contrib/foreman/foreman-ansible-modules/.tmp/ansible/", " - Ansible collection: {...}/ansible_collections/{namespace}/{collection}/", "", "Current working directory: /home/bagasse/projects/c
ontrib/foreman/foreman-ansible-modules/tests/extras"], "stdout": "", "stdout_lines": []}
```

evgeni:
I think this should be fixed by https://github.com/theforeman/foreman-ansible-modules/pull/371 too

evgeni:
39c7a1c

bagasse:
@evgeni, thanks !

PR #370, opened 2019-07-23
Add config_groups support

bagasse:
Add config groups support.
Needed by #366 to add config_groups tests on hostgroups.

### DIFF
--- a/plugins/modules/foreman_config_group.py
+++ b/plugins/modules/foreman_config_group.py
@@ -1,0 +1,109 @@
+#!/usr/bin/python
+# -*- coding: utf-8 -*-
+# (c) 2019 Baptiste Agasse
+#
+# This file is part of Ansible
+#
+# Ansible is free software: you can redistribute it and/or modify
+# it under the terms of the GNU General Public License as published by
+# the Free Software Foundation, either version 3 of the License, or
+# (at your option) any later version.
+#
+# Ansible is distributed in the hope that it will be useful,
+# but WITHOUT ANY WARRANTY; without even the implied warranty of
+# MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+# GNU General Public License for more details.
+#
+# You should have received a copy of the GNU General Public License
+# along with Ansible.  If not, see <http://www.gnu.org/licenses/>.
+
+DOCUMENTATION = '''
+---
+module: foreman_config_group
+short_description: Manage (Puppet) config groups using Foreman API
+description:
+  - Create and Delete Foreman (Puppet) config groups using Foreman API
+author:
+  - "Baptiste Agasse (@bagasse)"
+requirements:
+  - "apypie"
+options:
+  name:
+    description: The config group name
+    required: true
+  puppetclasses:
+    description: List of puppet classes to include in this group
+    required: false
+    type: list
+  state:
+    description: config group presence
+    default: present
+    choices: ["present", "absent"]
+extends_documentation_fragment: foreman
+'''
+
+EXAMPLES = '''
+- name: create new config group
+  foreman_config_group:
+    name: "My config group"
+    puppetclasses:
+      - ntp
+      - mymodule::myclass
+    server_url: "https://foreman.example.com"
+    username: "admin"
+    password: "secret"
+    state: present
+'''
+
+RETURN = ''' # '''
+
+from ansible.module_utils.foreman_helper import ForemanEntityApypieAnsibleModule
+
+# This is the only true source for names (and conversions thereof)
+name_map = {
+    'name': 'name',
+    'puppetclasses': 'puppetclass_ids',
+}
+
+
+def main():
+    module = ForemanEntityApypieAnsibleModule(
+        argument_spec=dict(
+            name=dict(required=True),
+            puppetclasses=dict(type='list'),
+        ),
+    )
+
+    entity_dict = module.clean_params()
+
+    module.connect()
+
+    entity = module.find_resource_by_name('config_groups', name=entity_dict['name'], failsafe=True)
+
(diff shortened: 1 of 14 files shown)


bagasse:
Sanity error related to #369

plugins/modules/foreman_config_group.py
```diff
@@ -0,0 +1,111 @@
...
+from ansible.module_utils.foreman_helper import (
+    split_fqn,
+    ForemanEntityApypieAnsibleModule,
+    sanitize_entity_dict,
+)
+
+# This is the only true source for names (and conversions thereof)
+name_map = {
+    'name': 'name',
+    'puppetclasses': 'puppetclass_ids'
```
mdellweg:
```suggestion
    'puppetclasses': 'puppetclass_ids',
```

bagasse:
@mdellweg, thanks for review. Suggestion applied (+ some variables names fixes)

bagasse:
Rebased to fix sanity checks.

tests/test_playbooks/config_group.yml
```diff
@@ -0,0 +1,48 @@
...
+        config_group_state: present
+        expected_change: true
+    - include: tasks/config_group.yml
+      vars:
+        config_group_state: present
+        config_group_puppetclasses: []
+        expected_change: false
+    - include: tasks/config_group.yml
+      vars:
+        config_group_puppetclasses: ['ntp', 'ntp::config']
```
evgeni: does that mean we need to import the `ntp` puppet class before this playbook can run?

evgeni: (or be recorded, to be precise)

bagasse: Yes, as stated in the comment at the top of the file.

plugins/module_utils/foreman_helper.py
```diff
@@ -393,12 +393,12 @@ def parse_template_from_file(file_name, module):
 
 
 # Helper for titles
-def split_fqn(title):
+def split_fqn(title, separator='/'):
     """ Split fully qualified name (title) in name and parent title """
-    fqn = title.split('/')
+    fqn = title.split(separator)
```
```suggestion
    fqn = title.split(separator, 1)
```

then you don't need to `join` anymore :)

plugins/modules/foreman_config_group.py
```diff
@@ -0,0 +1,111 @@
...
+
+    entity = module.find_resource_by_name('config_groups', name=entity_dict['name'], failsafe=True)
+
+    if not module.desired_absent:
+        if 'puppetclasses' in entity_dict:
+            # puppet classes API return puppet classes grouped by puppet module name
+            puppet_classes = []
+            for puppet_class in entity_dict['puppetclasses']:
+                search = 'name="{}"'.format(puppet_class)
+                module_name, class_name = split_fqn(puppet_class, '::')
```
evgeni: where are those two used?

plugins/modules/foreman_config_group.py
```diff
@@ -0,0 +1,111 @@
...
+                results = module.list_resource('puppetclasses', search)
+                if len(results) == 1 and len(results[results.keys()[0]]) == 1:
+                    result = results[results.keys()[0]]
```
evgeni: this seems unsed too

plugins/modules/foreman_config_group.py
```diff
@@ -0,0 +1,111 @@
...
+
+    if not module.desired_absent:
+        if 'puppetclasses' in entity_dict:
+            # puppet classes API return puppet classes grouped by puppet module name
+            puppet_classes = []
+            for puppet_class in entity_dict['puppetclasses']:
+                search = 'name="{}"'.format(puppet_class)
+                module_name, class_name = split_fqn(puppet_class, '::')
+                results = module.list_resource('puppetclasses', search)
+                if len(results) == 1 and len(results[results.keys()[0]]) == 1:
```
evgeni:
```suggestion
                if len(results) == 1 and len(results.values()[0]) == 1:
```
IMHO better readable (you want to ensure the first (and only) entry of the dict has len 1.

plugins/modules/foreman_config_group.py
```diff
@@ -0,0 +1,111 @@
...
+                    result = results[results.keys()[0]]
+                    puppet_classes.append({'id': results[results.keys()[0]][0]['id']})
```
evgeni: this is hard to read.

tests/test_playbooks/config_group.yml
evgeni: Right. Comments. Sorry :)

plugins/module_utils/foreman_helper.py
bagasse: Thanks for the pointer :), but I reverted this change as I don't use split_fqn() anymore.

plugins/modules/foreman_config_group.py
bagasse: Not used anymore.

plugins/modules/foreman_config_group.py
bagasse: Yes, sorry for the noise...

plugins/modules/foreman_config_group.py
bagasse: Applied, thanks !

plugins/modules/foreman_config_group.py
bagasse: Added some comment and simplified with your suggestion above.

bagasse:
@evgeni, thanks a lot for the review ! I think that I fixed all that you pointed, please take another look on the part that is hard to read if it's ok for you now :)

evgeni:
Thanks @bagasse!

PR #376, opened 2019-07-29
Entity spec

mdellweg:


mdellweg:
This is the first attempt. We need to agree on the naming of parameters.
The goal is to completely deduplicate the information in entity_spec and argument_spec and make that transparent to the user (module writer in this case).

mdellweg:
The user can still insert additional entries in the argument_spec by passing argument_spec. I am thinking about {domain, ...}_parameters here.

plugins/module_utils/foreman_helper.py
```diff
@@ -26,6 +26,34 @@
     APYPIE_IMP_ERR = traceback.format_exc()
 
 
+def _entity_spec_helper(spec):
```
evgeni: Given this is pure Python code, with no relation to Ansible whatsoever, I think this should get a set of unittests :)

plugins/module_utils/foreman_helper.py
```diff
@@ -26,6 +26,34 @@
...
+    entity_spec = {'id': {}}
+    argument_spec = {}
+    for key, value in spec.items():
+        if 'alt_names' in value:
+            value = value.copy()
+            alt_names = value.pop('alt_names')
+            for alt_name in alt_names:
+                entity_spec[alt_name] = value
+        entity_spec[key] = value
+        if 'flat_name' in value:
```
evgeni: I know this `flat` comes from the original `_flatten_value` I wrote, but I think this rather should be `foreman_name` or something?

plugins/module_utils/foreman_helper.py
```diff
@@ -26,6 +26,34 @@
...
 
 
+def _entity_spec_helper(spec):
+    """Extend an entity spec by adding entries for all alt_ and flat_names.
+    Extract ansible compatible argument_spec on the way.
+    """
+    entity_spec = {'id': {}}
+    argument_spec = {}
+    for key, value in spec.items():
+        if 'alt_names' in value:
```
evgeni: Can you elaborate on the usecase here? As your example doesn't use it ;)

evgeni:
@sean797 @ekohl this is what we discussed in Nürnberg and matches quite closely my vision of an unified "data spec", so I'd love your :eyes: on this :)

plugins/module_utils/foreman_helper.py
mdellweg: Well i like it to be called 'flat', because it distinguishes between the inflated version of the entity (with dictionaries for related entities) as provided by the user _and_ api responses, and the 'flat' version with only ids as required in the api requests.

plugins/module_utils/foreman_helper.py
```diff
@@ -26,6 +26,34 @@
...
+            value = value.copy()
+            alt_names = value.pop('alt_names')
```
ekohl: What's the benefit of copying and then popping a value?

plugins/module_utils/foreman_helper.py
mdellweg: At some point, you said we get aliases almost for free. This is the implementation for it. The name 'aliases' however had already been taken by ansible.
A usecase would be that we wanted to use a different name for something than foreman, e.g. 'from_ip' instead of 'from'.

plugins/module_utils/foreman_helper.py
mdellweg: The reason for copying before changing is, that those dictionaries in Python are always passed around by value. So if i do not copy it, i might change a dictionary that was already assigned for another purpose in entity_spec, which i should not.
Maybe this first occurance of copy is not _desperately needed_, but i think it is a good custom not to change the `spec` variable that was passed originally.

mdellweg:
Now with unittests.

plugins/module_utils/foreman_helper.py
evgeni: I *do* remember myself saying (or rather writing) that, yeah. But I think I actually meant Ansible aliases.

Let's see what I can remember from that conversation:
* [Foreman uses `from` as the name for the start IP of a Subnet](https://theforeman.org/api/1.22/apidoc/v2/subnets/create.html)
* `nailgun` uses `from_` internally, as `from` is a reserved word in Python (https://github.com/SatelliteQE/nailgun/blob/60a38229de3f2e7edb076f470ca35f18c4466afb/nailgun/entities.py#L7124)
* we use `from_ip` which does sound nicer, but my brain likes to have our and Foremans params to match (if possible)

If we're going to rename our param to `from`, the following spec should do fine:
```python
'from': {'aliases': ['from_ip']}
```
and if we keep `from_ip`, we can use:
```python
'from_ip': {'aliases': ['from'], 'flat_name': 'from'}
```

In both cases, `alt_names` aren't needed?
Please note, I wrote the `dict` directly, as you cannot write `dict(from=something)` without getting a `SyntaxError`.

The only case where I'd need `alt_names` as you implemented them is when I'd want to pass both `from` and `from_ip` to Foreman, but I don't want/need that.

evgeni: FWIW, here is my original statement: https://github.com/theforeman/foreman-ansible-modules/pull/354#discussion_r305461148

And I'm kinda sure I meant Ansible aliases ;-)

mdellweg: I can see your point now, and will remove the unneeded complexity here.

plugins/module_utils/foreman_helper.py
evgeni: naming is hard, I can live with either :)

mdellweg:
Added an example for domain which already involves quite weird aliases.

plugins/modules/foreman_domain.py
```diff
@@ -31,10 +31,14 @@
   name:
     description: The full DNS domain name
     required: true
-  dns_proxy:
+  dns:
```
ekohl: I actually liked the dns_proxy name more because it was more explicit.

mdellweg: >     * we use `from_ip` which does sound nicer, but my brain likes to have our and Foremans params to match (if possible)

We can not have it both. And there is still the alias.

ekohl: This is tricky because you don't want to literally use `dns_id`. Because we already do some conversion on top (by retrieving the actual object and getting the ID) maybe this is fine?

mdellweg: `dns_id` is not visible to ansible (not even mentioned in the `argument_spec`). It is _only_ used internally towards the api.

ekohl: Let me try to be a bit less brief :) We don't want to use the literal API default, I think we agree on that. Then the question becomes what we should use. Just reading `dns` without context makes me think it's a boolean (manage or not) while with `dns_proxy` it's clearer to me that I need to specify a Foreman Proxy with the DNS feature. That's why I liked the `dns_proxy` option here. Can we use `dns_proxy` as default and `dns` as alias perhaps?

mdellweg: I think, we can even use dns_proxy (which i prefer, too) and not even provide dns as an alias as before. But we'd need something like https://github.com/theforeman/foreman-ansible-modules/pull/376#discussion_r308188471 again.
I'd like @evgeni 's thought before flip-flopping that one.

evgeni:
Yeah. I might have been too brief in my above comment too.
In the case of from vs from_ip I prefer from as that is the same as the API exposes *and* makes sense on it's own in the context of a subnet. One could also call that start and everyone would still know what's in there.

Now in the domain context, dns might be a bit unambiguous and dns_proxy the better choice.

I think we should default to the API naming (minus _id/_ids) unless the api name is not explicit enough.

plugins/modules/foreman_domain.py
evgeni: See below (thanks fasthub!) and tl;dr: dns_proxy, no alias is fine here for ne

mdellweg:
OK, what about the `fullname` vs `description` mess in the domain api?

evgeni:
There I'd probably add "fullname" as an alias to match the api, but the name as such is IMHO nonsense and description is much better

evgeni:
(that doesn't strictly belong into this PR tho, as long as we keep the possibility of adding aliases, we're good here)

plugins/modules/foreman_domain.py
mdellweg: In the current implementation (i.e. without `alt_names`), i think, this is not possible, as the `flat_name` _must _ be `dns_id` and we _need_ to have `dns` in the generated `entity_spec` to correctly translate what foreman delivers in the `find_resource`.

mdellweg: There is one more option if we supported both names: We could lie in the docs about which one is the alias.

plugins/module_utils/foreman_helper.py
```diff
@@ -26,6 +26,31 @@
     APYPIE_IMP_ERR = traceback.format_exc()
 
 
+def _entity_spec_helper(spec):
+    """Extend an entity spec by adding entries for all alt_ and flat_names.
```
ekohl:
```suggestion
    """Extend an entity spec by adding entries for all flat_names.
```

plugins/module_utils/foreman_helper.py
```diff
@@ -26,6 +26,31 @@
...
+    entity_spec = {'id': {}}
+    argument_spec = {}
+    for key, value in spec.items():
+        entity_value = {}
+        argument_value = value.copy()
+        if 'flat_name' in argument_value:
+            flat_name = argument_value.pop('flat_name')
+            entity_value['flat_name'] = flat_name
+            entity_spec[flat_name] = {}
+        if argument_value.get('type') == 'entity':
```
ekohl: Maybe a newline for readability?
```suggestion

        if argument_value.get('type') == 'entity':
```

plugins/modules/foreman_domain.py
evgeni: Uh, and how does it work now with dns_proxy?

mdellweg: Atm dns_proxy is the alias.

plugins/module_utils/foreman_helper.py
```diff
@@ -433,13 +461,16 @@ def __init__(self, argument_spec, **kwargs):
 
 class ForemanEntityApypieAnsibleModule(ForemanApypieAnsibleModule):
 
-    def __init__(self, argument_spec, **kwargs):
+    def __init__(self, argument_spec=None, **kwargs):
+        entity_spec, gen_args = _entity_spec_helper(kwargs.pop('entity_spec', {}))
+        if argument_spec is None:
```
ekohl: Perhaps this should be lower and reversed?
```python
if argument_spec is not None:
    args.update(argument_spec)
```

mdellweg: you mean without changing the logic, we can scrap a few lines of code?
Maybe yes. About beeing reversed, i think whatever comes through the `argument_spec` parameter to `__init__` should have the highest priority.

ekohl: Yep, this is what I meant.

plugins/modules/foreman_domain.py
evgeni: No, not in https://github.com/theforeman/foreman-ansible-modules/blob/827eb5a8629c41a7e70311e08ff1ea967b258433/plugins/modules/foreman_domain.py#L92 ;)

mdellweg: Ah, i meant in this pr.
And honestly, i think in the current master it is actually broken: See #360

evgeni: But the API response to a PUT/POST or GET will include both, `dns_id` and `dns` 
https://github.com/theforeman/foreman-ansible-modules/blob/827eb5a8629c41a7e70311e08ff1ea967b258433/tests/test_playbooks/fixtures/domain-1.yml#L43-L48

so technically we don't need to "flatten" `dns` into `dns_id` as that's already there?

mdellweg: I conclude, that means we don't need it here, and therefore we can add it later if needed.

evgeni: So @mdellweg pointed out to me on IRC that while this is true for the domain→dns relationship, it's not true for organization→media and organization→templates. However, we don't diverge from the API name in the Ansible spec there so the issue does not arise.

We could add an `api_name` to the spec in the future to avoid that.

Bit *I* think doing this in the future is sufficient.

plugins/module_utils/foreman_helper.py
```diff
@@ -467,6 +471,32 @@ def sanitize_entity_dict(entity_dict, name_map):
...
+        entity_value = {}
+        argument_value = value.copy()
+        if 'flat_name' in argument_value:
+            flat_name = argument_value.pop('flat_name')
+            entity_value['flat_name'] = flat_name
+            entity_spec[flat_name] = {}
+
+        if argument_value.get('type') == 'entity':
+            argument_value.pop('type')
+            entity_value['type'] = 'entity'
```
evgeni: you could write that as

```python
entity_value['type'] = argument_value.pop('type')
```

plugins/module_utils/foreman_helper.py
```diff
@@ -467,6 +471,32 @@ def sanitize_entity_dict(entity_dict, name_map):
...
+        elif value.get('type') == 'entity_list':
```
```suggestion
        elif argument_value.get('type') == 'entity_list':
```

yes, they are copies, it just feels weird if one if branch uses `value` and the other `argument_value`

plugins/modules/foreman_domain.py
```diff
@@ -74,26 +78,15 @@
...
 def main():
     module = ForemanEntityApypieAnsibleModule(
-        argument_spec=dict(
+        entity_spec=dict(
             name=dict(required=True),
-            description=dict(),
-            dns_proxy=dict(),
-            locations=dict(type='list'),
-            organizations=dict(type='list'),
+            fullname=dict(aliases=['description']),
```
evgeni: IMHO we also should update our test playbooks to use the new primary name

plugins/modules/foreman_domain.py
```diff
@@ -32,9 +32,13 @@
     description: The full DNS domain name
     required: true
   dns_proxy:
+    aliases:
+      - dns
     description: DNS proxy to use within this domain for managing A records
     required: false
-  description:
+  fullname:
```
evgeni: the examples below still use `description`

plugins/module_utils/foreman_helper.py
mdellweg: sure

plugins/modules/foreman_domain.py
mdellweg: I changed it back to make description the default name.

PR #380, opened 2019-07-30
Modify foreman_compute_resource module for apypie

Manisha15:


Manisha15:
@mdellweg @evgeni  This is ready for review.

plugins/modules/foreman_compute_resource.py
```diff
@@ -81,29 +82,30 @@
     provider: vmware
```
mdellweg: this is inconsistent with the description of the task

plugins/modules/foreman_compute_resource.py
```diff
@@ -208,25 +188,20 @@ def main():
...
-    if 'organizations' in entity_dict:
-        entity_dict['organizations'] = find_organizations(module, entity_dict['organizations'])
+    if not module.desired_absent:
+        if 'updated_name' in entity_dict:
+            entity_dict['name'] = entity_dict['updated_name']
 
-    if 'locations' in entity_dict:
-        entity_dict['locations'] = find_locations(module, entity_dict['locations'])
+        if 'organizations' in entity_dict:
+            entity_dict['organizations'] = module.find_resources('organizations', entity_dict['organizations'], thin=True)
```
mdellweg: I think, it is find_resources_by_name

plugins/modules/foreman_compute_resource.py
```diff
@@ -208,25 +188,20 @@ def main():
...
 
-    entity_dict = sanitize_entity_dict(entity_dict, name_map)
+        if 'locations' in entity_dict:
+            entity_dict['locations'] = module.find_resources('locations', entity_dict['locations'], thin=True)
```
mdellweg: And this one should be find_resources_by_title

plugins/modules/foreman_compute_resource.py
```diff
@@ -238,7 +213,7 @@ def main():
             if key in provider_params:
                 entity_dict[key] = provider_params[key]
 
-    changed = naildown_entity_state(provider_infos['class'], entity_dict, entity, state, module)
+    changed = module.ensure_entity_state('compute_resources', entity_dict, entity, entity_spec=entity_spec)
```
mdellweg:
```suggestion
    changed = module.ensure_entity_state('compute_resources', entity_dict, entity)
```

plugins/modules/foreman_compute_resource.py
```diff
@@ -71,39 +72,40 @@
...
-      datacenter: aa92fb54-0736-4066-8fa8-b8b9e3bd75ac
-      ovirt_quota: 24868ab9-c2a1-47c3-87e7-706f17d215ac
-      use_v4: true
+      url: libvirt.example.com
+      display_type: vnc
+    server_url: foreman.example.com
+    username: admin
+    password: secret
+    state: present
+- name: delete livirt compute resource
```
evgeni: missing empty line and this task has `state: present` while the name is "delete"? :)

plugins/modules/foreman_compute_resource.py
```diff
@@ -71,78 +72,67 @@
 '''
 
 EXAMPLES = '''
-- name: vmware compute resource
+- name: Create livirt compute resource
```
evgeni: I think I'd leave the vmware and ovirt create examples, just to show how that works with different CRs

plugins/modules/foreman_compute_resource.py
```diff
@@ -185,19 +171,15 @@ def main():
             locations=dict(type='list'),
             organizations=dict(type='list'),
             state=dict(type='str', default='present', choices=['present', 'absent', 'present_with_defaults']),
-
-            # Deprecated provider's specific params, use nested keys in provider_params param instead
             provider_auth=dict(type='dict'),
```
evgeni: any reason not to remove this too? as it's deprecated?

@bagasse what do you think? IIRC that's your code initially

plugins/modules/foreman_compute_resource.py
```diff
@@ -71,78 +72,67 @@
...
-    'name': 'name',
-    'description': 'description',
-    'provider': 'provider',
-    'organizations': 'organization',
-    'locations': 'location',
-    'datacenter': 'datacenter',
+from ansible.module_utils.foreman_helper import ForemanEntityApypieAnsibleModule
+
+
+entity_spec = {
```
evgeni: with the merge of #376 this can be made simpler I guess?

plugins/modules/foreman_compute_resource.py
bagasse: I moved this to provider_params some time ago because some providers needs some specific params other than auth (datacenter, quota...) when I improved oVirt support. I don't know the deprecation policy for foreman ansible modules, but since the commit that introduced it is more than 5 month old, I think we can remove this param.

evgeni: Yeah, we don't have a policy (yet). We should once we start shipping this in a versioned manner. But until then, I too think just remove it is the right path.

plugins/modules/foreman_compute_resource.py
Manisha15: Modified, according to #376 .

plugins/modules/foreman_compute_resource.py
```diff
@@ -112,133 +154,93 @@
...
 
 def get_provider_infos(provider):
     provider_name = provider.lower()
 
     if provider_name == 'libvirt':
         return {
-            'params': ['url', 'display_type'],
-            'class': LibvirtComputeResource
+            'params': ['url', 'display_type']
         }
```
mdellweg: Please either put the dict in one line, or add a comma even on the last line.

plugins/modules/foreman_compute_resource.py
```diff
@@ -112,133 +154,93 @@
...
-            locations=dict(type='list'),
-            organizations=dict(type='list'),
-            state=dict(type='str', default='present', choices=['present', 'absent', 'present_with_defaults']),
-
-            # Deprecated provider's specific params, use nested keys in provider_params param instead
-            provider_auth=dict(type='dict'),
-            url=dict(type='str'),
-            display_type=dict(type='str'),
-            datacenter=dict(type='str'),
+            display_type=dict(),
```
evgeni: shouldn't all these be nested  under `provider_params` or would that require #385 for the nesting to work?

plugins/modules/foreman_compute_resource.py
```diff
@@ -112,133 +154,93 @@
...
+            datacenter=dict(),
+            url=dict(),
+            user=dict(),
+            password=dict(),
```
mdellweg: I think those rwo should not be here as being part of the `provider_params`.

plugins/modules/foreman_compute_resource.py
```diff
@@ -112,133 +154,93 @@
...
+            use_v4=dict(),
+            ovirt_quota=dict(),
+        ),
+        argument_spec=dict(
+            parameters=dict(type='list', elements='dict', options=dict(
+                state=dict(type='str', default='present', choices=['present', 'absent', 'present_with_defaults']),
```
mdellweg: The state should really be a direct member of the `argument_spec`. Also i do not see parameters anywhere in the Documentation.

plugins/modules/foreman_compute_resource.py
```diff
@@ -112,133 +154,93 @@
...
-    if 'locations' in entity_dict:
-        entity_dict['locations'] = find_locations(module, entity_dict['locations'])
-
-    entity_dict = sanitize_entity_dict(entity_dict, name_map)
+        if 'locations' in entity_dict:
+            entity_dict['locations'] = module.find_resources_by_title('locations', entity_dict['locations'], thin=True)
 
     # Add provider specific params
-    if state in ['present', 'present_with_defaults']:
+    if not module.desired_absent:
```
mdellweg: This duplicates line 224, and i do not see why.

plugins/modules/foreman_compute_resource.py
bagasse: You should also update module documentation documentation accordingly.

plugins/modules/foreman_compute_resource.py
Manisha15: Yeah, but while creating or updating compute_resource API directly use these parameters and not provider_params dictionary. So, these parameters are to match API acceptable arguments.

plugins/modules/foreman_compute_resource.py
mdellweg: No, this sort of nesting works in the old way.

plugins/modules/foreman_compute_resource.py
mdellweg: In that case we might need to wait for the invisible type.

plugins/modules/foreman_compute_resource.py
```diff
@@ -112,133 +150,91 @@
...
-        entity_dict['locations'] = find_locations(module, entity_dict['locations'])
+        if 'locations' in entity_dict:
+            entity_dict['locations'] = module.find_resources_by_title('locations', entity_dict['locations'], thin=True)
 
-    entity_dict = sanitize_entity_dict(entity_dict, name_map)
-
-    # Add provider specific params
-    if state in ['present', 'present_with_defaults']:
+        # Add provider specific params
         if not provider_infos and not entity:
```
evgeni: shouldn't this rather check `if not 'provider' in entity_dict`? (but given this was the same in the old code, not blocking)

plugins/modules/foreman_compute_resource.py
```diff
@@ -112,133 +150,91 @@
...
             module.fail_json(msg='To create a compute resource a valid provider must be supplied')
 
         for key in provider_infos['params']:
```
evgeni: should we add a check that `provider_params` has no entries that are not in `provider_infos['params']`?

smth like:
```python
given_provider_params = set(provider_params.keys())
supported_provider_params = set(provider_infos['params'])
unsupported_provider_params = given_provider_params - supported_provider_params
if unsupported_provider_params:
  module.fail_json(msg="Provider {} does not support the following params: {}".format(entity_dict.get('provider'), unsupported_provider_params))
```

plugins/modules/foreman_compute_resource.py
```diff
@@ -112,133 +150,91 @@
...
 
 def get_provider_infos(provider):
     provider_name = provider.lower()
 
     if provider_name == 'libvirt':
-        return {
-            'params': ['url', 'display_type'],
-            'class': LibvirtComputeResource
-        }
+        return {'params': ['url', 'display_type']}
```
evgeni: given `params` is the only thing we return now, maybe rename the function to `get_provider_params` and return the list directly, instead of wrapping it in a dict?

evgeni:
@bagasse do you have any comments on this? or is it good to go from your POV?

plugins/modules/foreman_compute_resource.py
mdellweg: Sounds like a good idea.

plugins/modules/foreman_compute_resource.py
mdellweg: should we check for both, to make up a 'valid provider'?

plugins/modules/foreman_compute_resource.py
mdellweg: Yes, i like this.

plugins/modules/foreman_compute_resource.py
evgeni: provider already has `choices`, so it can be only valid or not present :)

mdellweg:
Made a new attempt on the parameter validation.

plugins/modules/foreman_compute_resource.py
```diff
@@ -154,20 +154,20 @@
 from ansible.module_utils.foreman_helper import ForemanEntityApypieAnsibleModule
 
 
-def get_provider_infos(provider):
+def get_provider_param_keys(provider):
     provider_name = provider.lower()
```
evgeni: is this really required btw?

mdellweg: I don't know, but it allows for some validation before the server api yells at us.

plugins/modules/foreman_compute_resource.py
```diff
@@ -226,14 +220,23 @@ def main():
         if 'locations' in entity_dict:
             entity_dict['locations'] = module.find_resources_by_title('locations', entity_dict['locations'], thin=True)
 
+        if 'provider' in entity_dict:
+            entity_dict['provider'] = entity_dict['provider'].title()
```
evgeni: I know this was in the old code too. But I don't see the reason for it. The API will `.downcase` it anyways: https://github.com/theforeman/foreman/blob/d84dfbbd486d98d3cc1160dab13dd7ebf6686460/app/models/compute_resource.rb#L89-L96

plugins/modules/foreman_compute_resource.py
```diff
@@ -226,14 +220,23 @@ def main():
...
+
+            provider_param_keys = get_provider_param_keys(provider=entity_dict['provider'])
+            provider_params = {k: v for k, v in entity_dict.pop('provider_params', dict()).items() if v is not None}
```
evgeni: we used to have a helper for that ;)

plugins/modules/foreman_compute_resource.py
evgeni: and the "real" names are different anyways: https://github.com/theforeman/foreman/blob/develop/app/models/compute_resource.rb#L45-L55

bagasse:
This is good for me :+1:

plugins/modules/foreman_compute_resource.py
mdellweg: I knew it looked familiar. %)

plugins/modules/foreman_compute_resource.py
mdellweg: I fear we need something like this, because foreman does not accept the downcased version.

evgeni: interesting. thanks for testing. then let's keep it.

mdellweg: Have a look again: I reintroduced the provider_info thing, because title would not work for CamelCase.

evgeni: Not that we have CamelCase (today), but you're right, let's be explicit.

PR #381, opened 2019-07-31
Add find_operatingsystem

### LINKED ISSUE #373: Make OperatingSystem handling consistent across all modules
Maybe out of scope for this PR, but in other places we use `title ~ STRING` when searching for OSes:
https://github.com/theforeman/foreman-ansible-modules/blob/61fb6c25cfb894a03290192fbd638490b3c8cdae/plugins/modules/foreman_installation_medium.py#L144-L145

https://github.com/theforeman/foreman-ansible-modules/blob/61fb6c25cfb894a03290192fbd638490b3c8cdae/plugins/modules/foreman_architecture.py#L103-L104

While `find_resource_by_title` uses `=`:
https://github.com/theforeman/foreman-ansible-modules/blob/61fb6c25cfb894a03290192fbd638490b3c8cdae/plugins/module_utils/foreman_helper.py#L186-L188

I think the idea behind this was to make the OS search a bit fuzzier, as OS names/titles are weird.

Given you didn't write this code in the first place in this PR, I don't think we need a decision/solution here now, but I wanted to raise awareness for this.

Maybe also @mdellweg can chime in.

_Originally posted by @evgeni in https://github.com/theforeman/foreman-ansible-modules/pull/366#discussion_r306633638_

mdellweg:
Allow the search to return an operatingsystem whose title i the prefix of
another. exdend the test at the architecture module, and demonstrate in the
test, that operatingsystem can create anoperatingsystem with said property.

Fixes #373

### DIFF
--- a/plugins/module_utils/foreman_helper.py
+++ b/plugins/module_utils/foreman_helper.py
@@ -211,6 +211,35 @@ class ForemanApypieAnsibleModule(ForemanBaseAnsibleModule):
     def find_resources_by_title(self, resource, titles, **kwargs):
         return [self.find_resource_by_title(resource, title, **kwargs) for title in titles]
 
+    def find_operatingsystem(self, name, params=None, failsafe=False, thin=None):
+        search = 'title~"{}"'.format(name)
+        list_params = {}
+        if params is not None:
+            list_params.update(params)
+        if thin is None:
+            thin = self._thin_default
+        list_params['thin'] = thin
+        results = self.list_resource('operatingsystems', search, list_params)
+        result = None
+        unique_str = ''
+        if len(results) == 1:
+            result = results[0]
+        elif len(results) >= 1:
+            # return the exact match (if available)
+            result = next((result for result in results if result['title'] == name), None)
+            unique_str = 'unique '
+        if not failsafe and not result:
+            self.fail_json(msg="No {}operatingsystem found for '{}'".format(unique_str, name))
+        if result:
+            if thin:
+                result = {'id': result['id']}
+            else:
+                result = self.show_resource(resource, result['id'], params=params)
+        return result
+
+    def find_operatingsystems(self, names, **kwargs):
+        return [self.find_operatingsystem(name, **kwargs) for name in names]
+
     def create_resource(self, resource, entity_dict):
         new_entity = {}
         # FIXME: (?) entity_dict can contain other resources, when it does, translate these entries to only list their IDs
--- a/plugins/module_utils/foreman_helper.py
+++ b/plugins/module_utils/foreman_helper.py
@@ -212,29 +212,10 @@ class ForemanApypieAnsibleModule(ForemanBaseAnsibleModule):
         return [self.find_resource_by_title(resource, title, **kwargs) for title in titles]
 
     def find_operatingsystem(self, name, params=None, failsafe=False, thin=None):
-        search = 'title~"{}"'.format(name)
-        list_params = {}
-        if params is not None:
-            list_params.update(params)
-        if thin is None:
-            thin = self._thin_default
-        list_params['thin'] = thin
-        results = self.list_resource('operatingsystems', search, list_params)
-        result = None
-        unique_str = ''
-        if len(results) == 1:
-            result = results[0]
-        elif len(results) >= 1:
-            # return the exact match (if available)
-            result = next((result for result in results if result['title'] == name), None)
-            unique_str = 'unique '
-        if not failsafe and not result:
-            self.fail_json(msg="No {}operatingsystem found for '{}'".format(unique_str, name))
-        if result:
-            if thin:
-                result = {'id': result['id']}
-            else:
-                result = self.show_resource(resource, result['id'], params=params)
+        result = self.find_resource_by_title('operatingsystems', name, params=params, failsafe=True, thin=thin)
+        if not result:
+            search = 'title~"{}"'.format(name)
+            result = self.find_resource('operatingsystems', search, params=params, failsafe=failsafe, thin=thin)
         return result
 
     def find_operatingsystems(self, names, **kwargs):
--- a/plugins/modules/foreman_operating_system.py
+++ b/plugins/modules/foreman_operating_system.py
@@ -39,7 +39,7 @@ options:
   name:
     description:
       - Name of the Operating System
-    required: true
+    required: false
   release_name:
     description:
       - Release name of the operating system (recommended for debian)
@@ -50,11 +50,11 @@ options:
   family:
     description:
       - distribution family of the Operating System
-    required: true
+    required: false
   major:
     description:
       - major version of the Operating System
-    required: true
+    required: false
   minor:
     description:
       - minor version of the Operating System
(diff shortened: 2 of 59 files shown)


plugins/module_utils/foreman_helper.py
```diff
@@ -202,6 +202,32 @@ def find_resources_by_name(self, resource, names, **kwargs):
...
+        list_params = {}
+        if params is not None:
+            list_params.update(params)
+        if thin is None:
+            thin = self._thin_default
+        list_params['thin'] = thin
+        results = self.list_resource('operatingsystems', search, list_params)
+        if len(results) >= 1:
+            # return the best match (with the shortest title)
+            result = sorted(results, key=lambda x: len(x['title']))[0]
```
evgeni: What happens if the user enters `CentOS 7`, but they only have `CentOS 7.3.123` and `CentOS 7.4.123` defined? Given both entries have the same length, the result is (potentially) unstable?

mdellweg: yes, sadly. It's the same Problem if you had `CentOS 1.2.345` without a desription and another one with `CentOS unstable` as a description.

mdellweg: With a little more effort, we can error out on such ambiguities.

m-bucher: Just my opinion, if we have multiple results for the fuzzy search and among them is one perfect match, return that, otherwise fail with 'search result not distinct' or something like that.
Because, If you try to match the shortest result and there are multiple results, it's always a coin-flip.

evgeni: @m-bucher do I read this as a "yay" to my "refering operating systems" comment in https://github.com/theforeman/foreman-ansible-modules/issues/373#issuecomment-516737370?

mdellweg: That corresponds to the second part of that answer, which i agree:
https://github.com/theforeman/foreman-ansible-modules/issues/373#issuecomment-516737370
I think, we don't have to call the api twice though.

evgeni: Yeah, how you *implement* that is a detail. Doing two API calls would have the benefit that you can just use `find_resource` twice instead of effectively duplicating its content.

mdellweg:
I think, i have found an implementation reflecting the concensus found here and in #373 .
Can ya all verify that it is serving the right cornercases?
@evgeni @m-bucher @bagasse (i am aiming for two positive reviews)

tests/test_playbooks/architecture.yml
```diff
@@ -61,10 +84,10 @@
...
-      operating_system_name: "TestOS1"
-      operating_system_state: absent
-  - include: tasks/operating_system.yml
-    vars:
-      operating_system_name: "TestOS2"
+      operating_system_name: "{{ item }}"
       operating_system_state: absent
+    loop:
+      - "TestOS1"
+      - "TestOS1 SP1"
```
evgeni: I thought the name cannot contain spaces?

plugins/modules/foreman_operating_system.py
```diff
@@ -216,6 +216,14 @@ def main():
             )),
             state=dict(default='present', choices=['present', 'present_with_defaults', 'absent']),
         ),
+        required_if=[
+            ['state', 'present', ['name', 'major', 'family']],
+            ['state', 'present_with_defaults', ['name', 'major', 'family']],
+        ],
+        required_one_of=[
```
evgeni: just so that I parse it right. you require name, major, family for present/with_defaults as that's what we need for creation, but for absent, we can either go by name+major OR description. right?

evgeni: talking to myself, is there a way to better express that in the docs?

mdellweg: that _is_ the intention.

tests/test_playbooks/architecture.yml
mdellweg: I guess you are right. We can not be sure without attempting to rerecord.

plugins/module_utils/foreman_helper.py
```diff
@@ -211,6 +211,35 @@ def find_resources_by_name(self, resource, names, **kwargs):
     def find_resources_by_title(self, resource, titles, **kwargs):
         return [self.find_resource_by_title(resource, title, **kwargs) for title in titles]
 
+    def find_operatingsystem(self, name, params=None, failsafe=False, thin=None):
```
evgeni: So I still think that this is more readable, while taking two API calls:

```python
def find_operatingsystem(self, name, params=None, failsafe=False, thin=None):
  operatingsystem = self.find_resource_by_title('operatingsystems', name, params=params, failsafe=True, thin=thin)
  if not operatingsystem:
    search = 'title~"{}"'.format(name)
    operatingsystem = self.find_resource('operatingsystems', search, params=params, failsafe=failsafe, thin=thin)
  return operatingsystem
```

But I wouldn't block on that.

mdellweg: I prefer to do less api calls. (They are expensive)

evgeni: I'd disagree that they are expensive for *operatingsystems*. Hosts yes. But OS...
We're just adding a single `index` call (because if that succeeds, the second find is not executed, and if it does not succeed, no further API calls inside `find_resource` are executed).

Anyways. I said I won't block on this, and I stand by that. :)

plugins/module_utils/foreman_helper.py
```diff
@@ -211,6 +211,35 @@ def find_resources_by_name(self, resource, names, **kwargs):
...
+            thin = self._thin_default
+        list_params['thin'] = thin
+        results = self.list_resource('operatingsystems', search, list_params)
+        result = None
+        unique_str = ''
+        if len(results) == 1:
+            result = results[0]
+        elif len(results) >= 1:
+            # return the exact match (if available)
+            result = next((result for result in results if result['title'] == name), None)
```
evgeni: funfact, I think this would not work if `openratingsystems` would support the `thin` parameter (they do not) as `title` is not returned for thin (for hosts: https://github.com/theforeman/foreman/blob/f3e9f1f0308fb883637b0a57cbe014ef9f287ec3/app/controllers/api/v2/hosts_controller.rb#L47-L52)

mdellweg: I guess, that means the other solution is future-thin-proof?
I'll change it due to majority request.

evgeni: It's more future proof by the definition in https://twitter.com/lunasorcery/status/1156650841216049152, yes ;-)

tests/test_playbooks/vars/server.yml.example
```diff
@@ -5,8 +5,10 @@ foreman_password: "changeme"
 foreman_server_url: "https://foreman.example.com"
 foreman_validate_certs: false
 
+foreman_proxy: "https://foreman.example.com"
```
evgeni: is it correct to have it *with* the `https://` schema prefix?

mdellweg: I guess not...

mdellweg:
Let's hope this fixes it once and for all. ;)

PR #382, opened 2019-08-01
katello_manifest tests

evgeni:
this does not include the recorded tapes, as I am unsure which manifest I can legally include here yet ;)


evgeni:
cc @sean797 @akofink

evgeni:
There we go, I've added a (deleted) manifest, so it can't actually talk to Red Hat, but is good enough to be imported into Katello.
I also had to modify the VCR glue as the body of the upload would never match :woman_shrugging:

evgeni:
The failure is actually legit. Nice.

evgeni:
:green_apple:

tests/vcr_python_wrapper.py
```diff
@@ -47,6 +47,19 @@ def domain_query_matcher(r1, r2):
...
+def katello_manifest_body_matcher(r1, r2):
+    if r1.headers.get('content-type').startswith('multipart/form-data') and r2.headers.get('content-type').startswith('multipart/form-data'):
+        if r1.path.endswith('/subscriptions/upload') and r2.path.endswith('/subscriptions/upload'):
+            r1_copy = vcr.request.Request(r1.method, r1.uri, r1.body, r1.headers)
+            r2_copy = vcr.request.Request(r2.method, r2.uri, r2.body, r2.headers)
+            # the body is a huge binary blob, which seems to differ on every run, so we just ignore it
+            body1 = body2 = {}
+            r1_copy.body = json.dumps(body1)
+            r2_copy.body = json.dumps(body2)
+            return body_json_l2_matcher(r1_copy, r2_copy)
```
mdellweg: I guess we don't need that line. We can just fall through.

mdellweg: ... if we reused the names r1 and r2.

evgeni: Yes, but for some reason we've been keeping this as copies. And I didn't want to try to understand why ;)

mdellweg: If you change the original requests, that is, what is persisted in the cassettes. (And maybe even sent to the server.)
But instead of naming it r1_copy we could just reuse the local name r1. We would still only modify the copy.

evgeni:
I'll leave that open for a bit if Sean or Andrew want to comment and merge it on Monday if there are no complaints :)

tests/test_playbooks/katello_manifest.yml
```diff
@@ -1,43 +1,67 @@
...
+#    include_tasks: tasks/katello_manifest.yml
+#    vars:
+#      manifest_state: "refreshed"
+#      expected_change: true
+  - name: "drop manifest"
+    include_tasks: tasks/katello_manifest.yml
+    vars:
+      manifest_state: "absent"
+      expected_change: true
+
```
sean797: Is it worth testing that `manifest_state: "absent"` is idempotent?

evgeni: Is it idempotent?

evgeni: seems it is, added test

evgeni:
this has two acks, and I fixed the test that sean was asking for, so I am going to merge this.
AND NOONE IS GOING TO STOP ME! ;-)

PR #383, opened 2019-08-05
Test realm

mdellweg:
Finally managed to get this running.
And it shows, that realm is _not_ idempotent.

### DIFF
--- a/plugins/modules/foreman_realm.py
+++ b/plugins/modules/foreman_realm.py
@@ -41,6 +41,8 @@ options:
       - Password for user accessing Foreman server
     required: true
   validate_certs:
+    aliases:
+      - verify_ssl
     description:
       - Verify SSL of the Foreman server
     required: false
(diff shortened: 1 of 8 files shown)


plugins/modules/foreman_realm.py
```diff
@@ -121,7 +121,7 @@ def main():
             server_url=dict(required=True),
             username=dict(required=True, no_log=True),
             password=dict(required=True, no_log=True),
-            verify_ssl=dict(type='bool', default=True),
+            validate_certs=dict(type='bool', default=True),
```
evgeni: alias. plz.

plugins/modules/foreman_realm.py
```diff
@@ -139,15 +139,15 @@ def main():
...
+    verify_ssl = realm_dict.pop('validate_certs')
     state = realm_dict.pop('state')
 
     try:
         create_server(server_url, (username, password), verify_ssl)
     except Exception as e:
         module.fail_json(msg="Failed to connect to Foreman server: %s " % e)
 
-    ping_server(module)
+    # ping_server(module)
```
evgeni: cheater (but fair)

plugins/modules/foreman_realm.py
evgeni: (or don't do the change, up to you)

evgeni:
Thank you @mdellweg!

Do you want to make it idempotent in this PR? Or should we ignore it for now?
If not, I'd just merge it as-is, because it is a huge improvement already.

@claudiol do you want to review this PR, which updates your #85?

mdellweg:
I decided to do as little change to the module as necessary. And then i just rewrote it for apypie.
#384

evgeni:
I think that's fair. Thank you!

PR #384, opened 2019-08-05
Port foreman_realm to apypie

mdellweg:


tests/test_playbooks/fixtures/realm-0.yml
```diff
@@ -0,0 +1,150 @@
...
+      x-runtime: ['0.014028']
+      x-xss-protection: [1; mode=block]
+    status: {code: 200, message: OK}
+- request:
+    body: null
+    headers:
+      Accept: [application/json;version=2]
+      Cookie: [_session_id=41776ea8f0ba0c93cdd3da10a3beb1f7]
+    method: GET
+    uri: https://centos7-foreman-1-22/api/smart_proxies?per_page=4294967296&search=name%3D%22centos7-foreman-1-22.anubis.example.com%22&thin=True
```
evgeni: wrong proxy name :)

plugins/modules/foreman_realm.py
```diff
@@ -74,92 +57,36 @@
...
-        argument_spec=dict(
-            server_url=dict(required=True),
-            username=dict(required=True, no_log=True),
-            password=dict(required=True, no_log=True),
-            verify_ssl=dict(type='bool', default=True),
+    module = ForemanEntityApypieAnsibleModule(
+        entity_spec=dict(
             name=dict(required=True),
-            realm_proxy=dict(type='int', required=True),
+            realm_proxy=dict(type='entity', flat_name='realm_proxy_id', required=True),
```
evgeni: This needs docs updates, as you changed from ID to name. Also backward incompatible change, but I think that's fine.

tests/test_playbooks/fixtures/realm-0.yml
mdellweg: Oh, actually it's just not yet ignoring it on replay.

tests/test_playbooks/realm.yml
```diff
@@ -19,11 +19,10 @@
...
       expected_change: True
-  # This is nonfunctional
-  # - include: tasks/realm.yml
-  #   vars:
-  #     realm_state: present
-  #     expected_change: False
+  - include: tasks/realm.yml
+    vars:
+      realm_state: present
+      expected_change: False
```
evgeni: :+1:

plugins/modules/foreman_realm.py
```diff
@@ -68,100 +47,46 @@
...
-            entity = entities[0]
-        else:
-            entity = None
-    except Exception as e:
-        module.fail_json(msg='Failed to find entity: %s ' % e)
+    entity = module.find_resource_by_name('realms', name=entity_dict['name'], failsafe=True)
 
-    realm_dict = sanitize_realm_dict(realm_dict)
+    if not module.desired_absent:
+        if 'realm_proxy' in entity_dict:
```
evgeni: the proxy is required, so no need to check if it's in the dict.

PR #386, opened 2019-08-05
add a test for the module "state"

evgeni:
it ensures that the module has tests and uses our abstraction layer

can also be used to generate a report of untested modules by just calling `python -m tests.test_module_state`


tests/test_crud.py
```diff
@@ -7,48 +7,8 @@
...
-    'role',
-    'search_facts',
-    'setting',
-    'snapshot',
-    'subnet',
-    'sync_plan',
-    'upload',
-    'user',
-]
+from .conftest import TESTED_MODULES
```
ekohl: You could make this a fixture and use it again elsewhere. https://github.com/theforeman/obal/blob/2f21ce6999b337bb1efd1f8c6cc12287242510d4/tests/conftest.py#L5-L7 defines a shared fixture. https://github.com/theforeman/obal/blob/2f21ce6999b337bb1efd1f8c6cc12287242510d4/tests/test_playbooks.py#L17-L19 parametrizes on a list.

tests/test_module_state.py
```diff
@@ -0,0 +1,79 @@
...
+
+from conftest import TESTED_MODULES
+
+SPECIAL_MODULES = {
+    'katello_content_view_version': 'tested as part of katello_content_view',
+    'foreman_host_power': 'tested as part of foreman_host',
+    'katello_sync': 'the test is called repository_sync',
+}
+
+MODULES_PATH = py.path.local(__file__).realpath() / '../..' / 'plugins/modules'
```
ekohl: Interesting combination of using slashes.

tests/test_module_state.py
```diff
@@ -0,0 +1,79 @@
...
+
+
+def find_all_modules():
+    for module in MODULES_PATH.listdir(sort=True):
+        module = module.basename
+        module = module.replace('.py', '')
+        if module == '__init__':
+            continue
```
ekohl: I'm not a fan of continue when it's not needed. Should this be:
```python
if module != '__init__':
    yield module
```

mdellweg:
This kind of nails us down to the _one-playbook-per-module_ concept. Do we want that?

evgeni:
> This kind of nails us down to the _one-playbook-per-module_ concept. Do we want that?

Does it? Or rather: does it more than we do already?

`TESTED_MODULES` could also be `TEST_PLAYBOOKS` (which it really is, if we're honest) and if we really want it, we could define the "modules tested by this playbook" in a separate list (or in a comment in the playbook itself). But this all also applies to todays use of that list.

tests/test_crud.py
evgeni: But I also need is a plain list in https://github.com/theforeman/foreman-ansible-modules/blob/229cf0a1a573d998dddff3c7a24d6704cf7d1635/tests/test_module_state.py#L33-L37

would that still work?

mdellweg:
Well the name of the variable implies it, but it is really just a list of playbooks we run. I am not opposed to a one-to-one relationship there. It would be a mess for contributors if the didn't know which playbook were affected by their change.
otoh, do we need a test to be sure to have tests for all modules? (Now that we are almost there...)
Testing for the abstraction layer would be interesting. Love to see that.
Would it be like "We added a new convention, those modules still use the old one."?

tests/test_crud.py
ekohl: I'd have a plain list and then build fixture from that.

From https://gist.github.com/ekohl/bcb4b815153f491d08d551ed4ed98ad9#file-test_pages-py-L67-L69

```
PAGES = [
  # ...
]

@pytest.fixture(params=PAGES)
def url(request, base_url):
    return base_url + request.param
```

evgeni:
> Well the name of the variable implies it, but it is really just a list of playbooks we run. I am not opposed to a one-to-one relationship there. It would be a mess for contributors if the didn't know which playbook were affected by their change.

We don't have a one-to-one relationship today already:
https://github.com/theforeman/foreman-ansible-modules/blob/229cf0a1a573d998dddff3c7a24d6704cf7d1635/tests/test_module_state.py#L10-L11

I'll rename the var to `TEST_PLAYBOOKS` to remove the implication, but the 1:1 relation sip is another discussion.

> otoh, do we need a test to be sure to have tests for all modules? (Now that we are almost there...)
> Testing for the abstraction layer would be interesting. Love to see that.
> Would it be like "We added a new convention, those modules still use the old one."?

So this started as a script to get a list for #77, and then I thought "why not to add this as a test, so noone will ever contribute a module without tests" but also to give a place to test conventions etc (as you mention). E.g. I have now added a test that will emit warnings (not failures) for modules that don't use our apypie wrapper.

tests/test_crud.py
evgeni: What's the benefit compared to directly using `@pytest.mark.parametrize('module', TEST_PLAYBOOKS)`?

tests/test_module_state.py
```diff
@@ -0,0 +1,104 @@
...
+SPECIAL_MODULES = {
+    'katello_content_view_version': 'tested as part of katello_content_view',
+    'foreman_host_power': 'tested as part of foreman_host',
+    'katello_sync': 'the test is called repository_sync',
+}
+
+MODULES_PATH = py.path.local(__file__).realpath() / '..' / '..' / 'plugins' / 'modules'
+
+
+def find_all_modules():
```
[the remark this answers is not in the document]
evgeni: probably less smart than whats in stdlib, but yeah.

tests/test_crud.py
ekohl: If it's one occurrence then probably little to none.

tests/test_module_state.py
```diff
@@ -7,6 +7,11 @@
 
 from .conftest import TEST_PLAYBOOKS
 
+try:
+    ast_try = ast.TryExcept
+except AttributeError:
+    ast_try = ast.Try
```
mdellweg: I think you can use six for things like this.

evgeni: you mean `if six.PY2` things? yeah. I'm too used to not use six for "reasons" :)

tests/test_module_state.py
```diff
@@ -0,0 +1,133 @@
...
+from .conftest import TEST_PLAYBOOKS
+
+if six.PY2:
+    ast_try = ast.TryExcept
+else:
+    ast_try = ast.Try
+
+SPECIAL_MODULES = {
+    'katello_content_view_version': 'tested as part of katello_content_view',
+}
```
mdellweg: This is about to be obsolete.

tests/test_module_state.py
```diff
@@ -0,0 +1,133 @@
...
+        module_ast = ast.parse(module_file.read())
+    return _module_framework_from_body(module_ast.body) or 'unknown'
+
+
+def _module_framework_from_body(body):
+    framework = None
+    for entry in body:
+        if isinstance(entry, ast.ImportFrom):
+            if entry.module == 'ansible.module_utils.ansible_nailgun_cement':
+                framework = 'nailgun'
```
mdellweg: Should not happen anymore.

tests/test_module_state.py
evgeni: yupp, updated.

tests/test_module_state.py
evgeni: you say the check as such can be removed too? yeah, probably right, updated.

PR #388, opened 2019-08-06
Rename foreman_operatingsystem

mdellweg:


mdellweg:
This is kind of a now or never, as it breaks existing playbooks. But in all other places we refer to oses as `operatingsystem`.

evgeni:
:+1: from me, because the API also always uses `operatingsystem(s)` as one word.
given the implication of this: ping @theforeman/ansible-modules

mdellweg:
"Three ACK's for a git mv" sounds like a movie title from the 80s.

Thank you folks!

PR #389, opened 2019-08-06
Fixes #374 - Extend hostgroup module with non-Foreman options

### LINKED ISSUE #374: extend hostgroup module with non-Foreman options
> This is great @bagasse! I've added two comments, but nothing major ;)
> 
> As for OpenSCAP, yeah this sounds like a good addition, I don't mind it being in this PR or an own one. Just remember that we'll need to use a different `apidoc.json` then ;)

@evgeni It would be nice to add extensible smart_proxy support if we can. We use katello and it would be nice to add content proxy also, but maybe we have to make a subclass katello_hostgroup for that ? I didn't wanted to make this happen in this PR that I think  is big enough :)

_Originally posted by @bagasse in https://github.com/theforeman/foreman-ansible-modules/pull/366#issuecomment-514528037_

bagasse:
* Update `foreman.json` and `katello.json` to with openscap support
* Add following params to hostgroup:
  * openscap_proxy
  * organization
  * content_source
  * lifecycle_environment
  * content_view

* Add helper to limit API calls when fetching smart proxies
* Update docs to add examples


bagasse:
Ok, @mdellweg, @evgeni, @Manisha15 I think I'm close to the end on this, if you can take a look it would be nice :)

plugins/module_utils/foreman_helper.py
```diff
@@ -201,6 +201,16 @@ def find_resource(self, resource, search, name=None, params=None, failsafe=False
...
+    def find_katello_resource_by_name(self, resource, organization, name, **kwargs):
+        search = 'name="{}"'.format(name)
+        kwargs['name'] = name
+        if 'params' in kwargs:
+            kwargs['params'].update({'organization_id': organization['id']})
+        else:
+            kwargs['params'] = {'organization_id': organization['id']}
+
+        return self.find_resource(resource, search, **kwargs)
+
```
mdellweg: Not sure, i like this. It's easy to keep a scope variable around in the module and pass that to the params arg.
Also there are cases, where you do not need / want to pass the organization_id (for example when obtaining subscriptions for an activation_key [@evgeni might disagree on that example] or reading repositories from a product), because they are already scoped more specifically.

plugins/module_utils/foreman_helper.py
```diff
@@ -504,6 +514,13 @@ def _flatten_entity(entity, entity_spec):
     return result
 
 
+# Helper to minimize API calls to retrive smart proxies
+def get_smart_proxy(module, smart_proxies, smart_proxy_name):
+    if smart_proxy_name not in smart_proxies:
+        smart_proxies[smart_proxy_name] = module.find_resource_by_name('smart_proxies', name=smart_proxy_name, failsafe=False, thin=True)
+    return smart_proxies[smart_proxy_name]
+
```
mdellweg: I like the effort to not search for things twice. But i think, we could have this in a generic way by passing a memory / cache / you-name-it dictionary to the `find_resource` which will be looked into first and will receive the result.
It would, almost transparently, solve the find my organizations problem, too.

tests/test_playbooks/katello_hostgroup.yml
```diff
@@ -0,0 +1,116 @@
...
+    vars:
+      hostgroup_name: "New host group"
+      hostgroup_description: "New host group"
+      hostgroup_locations: "{{ hostgroup.locations }}"
+      hostgroup_organizations: "{{ hostgroup.organizations }}"
+      hostgroup_organization: "{{ hostgroup.organizations | first }}"
+      hostgroup_content_source: "{{ hostgroup.content_source }}"
+      hostgroup_lifecycle_environment: "{{ hostgroup.lifecycle_environment }}"
+      hostgroup_content_view: "{{ hostgroup.content_view.name }}"
+      hostgroup_parameters:
```
mdellweg: Not convinced, we need to retest those here.

tests/test_playbooks/vars/hostgroup.yml
```diff
@@ -1,6 +1,8 @@
 ---
+foreman_proxy: foreman.example.com
```
mdellweg: this should be set in server.yml and if needed, a corresponding match exception added to vcr_wrapper

tests/test_playbooks/vars/hostgroup.yml
```diff
@@ -27,6 +29,13 @@ hostgroup:
   subne6: Test subnet6
   puppet_server: "{{ foreman_proxy }}"
   puppet_ca: "{{ foreman_proxy }}"
+  dns_proxy: "{{ foreman_proxy }}"
+  openscap_proxy: "{{ foreman_proxy }}"
+  content_source: "{{ foreman_proxy }}"
```
mdellweg: I am wondering, this does not collide with the plain-foreman version of the playbook.

plugins/module_utils/foreman_helper.py
evgeni: I won't disagree on the "do not need" (I might on "do not want", but that's not for discussion here :stuck_out_tongue:)

Having scoped resources is not a Katello-only thing, it's just that Katello *requires* them to be scoped, while Foreman does not. It is perfectly fine in a plain Foreman to have two Orgs which both contain a hostgroup called "prod" and without scoping the search you will never find the right one if your user can see both.

So if anything, we should have `find_scoped_resource…` helpers, but then again just passing the right scope as we did in the other Katello modules that were migrated to Apypie is (IMHO) equally easy.

What we could do (in the future) is to teach `ForemanEntityApypieAnsibleModule` to gather the scope "automatically" if a `organization`/`location` param was passed to the module and apply that scope in `find_resource` *unless* the developer provided a different scope while calling the function. But I think we have bigger fish to fry until this optimization is worth it.

plugins/module_utils/foreman_helper.py
evgeni: :+1: on Matthias idea, if anywhere this should be directly in the generic find helper.

mdellweg: I think it's ok to use the `search` string as the key in the memory dict.

plugins/modules/foreman_hostgroup.py
```diff
@@ -303,10 +358,42 @@ def main():
...
+            else:
+                entity_dict['organizations'] = [entity_dict['organization']]
+            entity_dict['organization'] = module.find_resource_by_name('organizations', name=entity_dict['organization'], failsafe=False, thin=True)
+            organization = entity_dict['organization']
+
+        if 'organizations' in entity_dict:
+            entity_dict['organizations'] = module.find_resources_by_name('organizations', entity_dict['organizations'], thin=True)
+
+        if 'content_source' in entity_dict:
+            if not organization:
```
evgeni: I think this can be better expressed by Ansible's `required_by` parameter. (but that's new in 2.8)

Alternatively, if we want to keep pre-2.8 compat, we could do something like https://github.com/theforeman/foreman-ansible-modules/blob/0eff1a589432dd4f6e8929735276f2b4f3768e5a/plugins/modules/foreman_hostgroup.py#L307-L309

But in either case, I think this should happen as early as possible, not after several API requests were made.

plugins/modules/foreman_hostgroup.py
```diff
@@ -303,10 +358,42 @@ def main():
...
+                module.fail_json(msg="Please specify the organization.")
+            entity_dict['content_source'] = get_smart_proxy(module, smart_proxies, entity_dict['content_source'])
+
+        if 'lifecycle_environment' in entity_dict:
+            if not organization:
```
evgeni: same as https://github.com/theforeman/foreman-ansible-modules/pull/389/files#r312440972

plugins/modules/foreman_hostgroup.py
```diff
@@ -303,10 +358,42 @@ def main():
...
+                module.fail_json(msg="Please specify the organization.")
+            entity_dict['lifecycle_environment'] = module.find_katello_resource_by_name('lifecycle_environments', organization,
+                                                                                        name=entity_dict['lifecycle_environment'], failsafe=False, thin=True)
+
+        if 'content_view' in entity_dict:
+            if not organization:
```
evgeni: and here also https://github.com/theforeman/foreman-ansible-modules/pull/389/files#r312440972 ;)

evgeni:
I am not sure I like to include OpenSCAP in the `foreman.json`. Doing so means that our tests never run with a pure Foreman.

In other place (esp forklift), we have the concept of "luna" which is Foreman+Katello+Ansible+OpenSCAP+I_can't_remember (or, as you maybe can guess from the name: whatever is enabled/supported in Red Hat Satellite).

How about we keep `foreman.json` plain Foreman, `katello.json` "plain" Foreman+Katello and add a `luna.json` that represents Luna as described above. You can then run the OpenSCAP related tests against that apidoc and keep the others "clean".

This *does* mean we'd end up with *three* test playbooks: `hostgroup`, `katello_hostgroup`, `luna_hostgroup` and more work for you, though.

What do you think?

plugins/modules/foreman_hostgroup.py
evgeni: I've opened https://github.com/theforeman/foreman-ansible-modules/issues/399 for a general discussion around `required_by`

tests/test_playbooks/vars/hostgroup.yml
bagasse: I don't think, this is doted notation, and we use vars name with underscores for playbooks/tasks variables so for me no collisions.

mdellweg: Sorry, i mean the fact, that you provide `content_source` which is a katello thing in the foreman only test. But i guess, in the end, it is silently filtered by apypie.

bagasse: Yes, I already ran hostgroup tests after adding that and all went fine.

bagasse:
@evgeni, @mdellweg thanks for your feedback, I moved out the cache related things out of this PR to focus on hostgroup. I will revert `foreman.json` and `katello.json` changes and create a `luna.json` to make openscap related tests, not too much work. I will ping you when I' done with this. Thanks !

bagasse:
@evgeni, @mdellweg thanks for your feedback, I moved out the cache related things out of this PR to focus on hostgroup. I will revert `foreman.json` and `katello.json` changes and create a `luna.json` to make openscap related tests, not too much work. I will ping you when I' done with this. Thanks !

bagasse:
@evgeni, @mdellweg,
* Reverted `foreman.json` and `katello.json` changes
* added `luna.json`
* Test openscap related stuff with `luna_hostgroup`
* Added some fixes to `hostgroup` tests related to operating system changes

plugins/modules/foreman_hostgroup.py
```diff
@@ -248,6 +303,11 @@ def main():
     name, parent = split_fqn(entity_dict['name'])
     entity_dict['name'] = name
 
+    katello_params = ['content_source', 'lifecycle_environment', 'content_view']
+
+    if 'organization' not in entity_dict and list(set(katello_params) & set(entity_dict.keys())):
```
evgeni: this *could* be `required_by` (#399) if we'd move to 2.8+ only (not urging you to change, just thinking aloud about the future and making GH reference this PR in #399 so that when we implement that we know where :))

plugins/modules/foreman_hostgroup.py
```diff
@@ -302,11 +359,29 @@ def main():
...
+            else:
+                entity_dict['organizations'] = [entity_dict['organization']]
+            entity_dict['organization'] = module.find_resource_by_name('organizations', name=entity_dict['organization'], failsafe=False, thin=True)
+            scope = {'organization_id': entity_dict['organization']['id']}
+
+        if 'organizations' in entity_dict:
+            entity_dict['organizations'] = module.find_resources_by_name('organizations', entity_dict['organizations'], thin=True)
+
+        if 'lifecycle_environment' in entity_dict:
+            entity_dict['lifecycle_environment'] = module.find_resource_by_name('lifecycle_environments', name=entity_dict['lifecycle_environment'],
```
evgeni: I'd put `params` after `name` to match the order in `find_resource_by_name`

plugins/modules/foreman_hostgroup.py
```diff
@@ -302,11 +359,29 @@ def main():
...
+                                                                                failsafe=False, thin=True, params=scope)
 
-        if 'puppet_ca_proxy' in entity_dict:
-            entity_dict['puppet_ca_proxy'] = module.find_resource_by_name('smart_proxies', name=entity_dict['puppet_ca_proxy'], failsafe=False, thin=True)
+        if 'content_view' in entity_dict:
+            entity_dict['content_view'] = module.find_resource_by_name('content_views', name=entity_dict['content_view'],
```
evgeni: I'd put `params` after `name` to match the order in `find_resource_by_name`

plugins/modules/foreman_hostgroup.py
```diff
@@ -130,6 +130,28 @@
     description: Puppet CA proxy name
     required: false
     default: None
+  openscap_proxy:
+    description: OpenSCAP proxy name. Only available when OpenSCAP plugin is installed in the foreman instance.
```
```suggestion
    description: OpenSCAP proxy name. Only available when the OpenSCAP plugin is installed.
```

plugins/modules/foreman_hostgroup.py
```diff
@@ -130,6 +130,28 @@
...
+    required: false
+    default: None
+  organization:
+    description:
+      - Organization for scoped resources attached to the hosgroup. Only used for katello instance.
```
```suggestion
      - Organization for scoped resources attached to the hosgroup. Only used for Katello installations.
```

plugins/modules/foreman_hostgroup.py
```diff
@@ -130,6 +130,28 @@
...
+      - This organization will be added in ornagizations parameter if needed.
```
```suggestion
      - This organization will be added in organizations parameter if needed.
```

plugins/modules/foreman_hostgroup.py
```diff
@@ -130,6 +130,28 @@
...
+    required: false
+    default: None
+  content_source:
+    description: Katello Content source. Only available for katello instance.
```
```suggestion
    description: Katello Content source. Only available for Katello installations.
```

plugins/modules/foreman_hostgroup.py
```diff
@@ -130,6 +130,28 @@
...
+    required: false
+    default: None
+  lifecycle_environment:
+    description: Katello Lifecycle environment. Only available for katello instance.
```
```suggestion
    description: Katello Lifecycle environment. Only available for Katello installations.
```

plugins/modules/foreman_hostgroup.py
```diff
@@ -130,6 +130,28 @@
...
+    required: false
+    default: None
+  content_view:
+    description: Katello Content view. Only available for katello instance.
```
```suggestion
    description: Katello Content view. Only available for Katello installations.
```

evgeni:
@bagasse awesome! I left a few comments of the "picky" style and Travis seems unhappy with your recordings, but otherwise :100: :+1:

evgeni:
This is labeled WIP, but I think this is ready, right? Removing the label.

bagasse:
I can't add/remove label, but yes, if it's ready for you, it's ready for me :)

evgeni:
> I can't add/remove label, but yes, if it's ready for you, it's ready for me :)

Well, let's change that? https://community.theforeman.org/t/proposing-to-add-bagasse-to-the-foreman-ansible-modules-commiters/14998

plugins/modules/foreman_hostgroup.py
```diff
@@ -130,6 +130,28 @@
...
     required: false
     default: None
+  openscap_proxy:
+    description: OpenSCAP proxy name. Only available when the OpenSCAP plugin is installed.
+    required: false
+    default: None
+  organization:
+    description:
+      - Organization for scoped resources attached to the hosgroup. Only used for katello installations.
+      - This organization will be added in organizations parameter if needed.
```
mdellweg:
```suggestion
      - This organization will implicitly be added to the I(organizations) parameter if needed.
```

plugins/modules/foreman_hostgroup.py
```diff
@@ -130,6 +130,28 @@
     description: Puppet CA proxy name
     required: false
     default: None
+  openscap_proxy:
+    description: OpenSCAP proxy name. Only available when the OpenSCAP plugin is installed.
+    required: false
+    default: None
+  organization:
+    description:
+      - Organization for scoped resources attached to the hosgroup. Only used for katello installations.
```
mdellweg:
```suggestion
      - Organization for scoped resources attached to the hostgroup. Only used for katello installations.
```

plugins/modules/foreman_hostgroup.py
```diff
@@ -237,8 +287,13 @@ def main():
             config_groups=dict(type='entity_list', flat_name='config_group_ids'),
             puppet_proxy=dict(type='entity', flat_name='puppet_proxy_id'),
             puppet_ca_proxy=dict(type='entity', flat_name='puppet_ca_proxy_id'),
+            openscap_proxy=dict(type='entity', flat_name='openscap_proxy_id'),
             parameters=dict(type='nested_list', entity_spec=parameter_entity_spec),
+            content_source=dict(type='entity', flat_name='content_source_id'),
+            lifecycle_environment=dict(type='entity', flat_name='lifecycle_environment_id'),
+            content_view=dict(type='entity', flat_name='content_view_id'),
         ),
+        argument_spec=dict(organization=dict()),
```
mdellweg: Just thinking: Does this render it a (fake) `KatelloEntity{Apypie}AnsibleModule`? Or is the `organization` required there?

tests/fixtures/apidoc/luna_hostgroup.json
```diff
@@ -0,0 +1 @@
+luna.json
```
mdellweg: @evgeni is this really the luna setup?
I know we already have a special api.json for snapshot. Maybe we want one for openscap?

evgeni: I didn't look if the JSON is from a Luna setup, if that's what you're asking.

My original suggestion to use luna here was to allow us to add `foreman_ansible` related hostgroup settings easily in the future. *Today* a Foreman+OpenSCAP setup/JSON would suffice. But that's also harder to produce than `vagrant up luna` :)

plugins/modules/foreman_hostgroup.py
evgeni: In a "real" Katello module, the Org would be required, yes.

tests/fixtures/apidoc/luna_hostgroup.json
mdellweg: Yes, the question is (and i guess to @bagasse then) is this produced from `vagrant up luna`?

bagasse: Yes, that's vagrant up luna ;)

mdellweg: This renaming should not be needed in any case. That is why we do all that extra matching dance in the vcr_wrapper.

bagasse:
@evgeni, @mdellweg, as usual, thank a lot for your time/patience ! Merged just before my holidays ;)

PR #392, opened 2019-08-07
migrate katello_manifest to apypie

evgeni:


evgeni:
@mdellweg @akofink @sean797 the tests are green, but the code is a tad ugly because the manifest api is ugly.

plugins/module_utils/foreman_helper.py
```diff
@@ -448,12 +449,15 @@ def _delete_entity(self, resource, current_entity, params):
         self._resource_action(resource, 'destroy', payload)
         return True, None
 
-    def _resource_action(self, resource, action, params):
+    def resource_action(self, *args, **kwargs):
```
mdellweg: Why not renaming the method altogether?
And if we think we need the hidden version for the transition, we should make it the alias and deprecate it.

evgeni: mostly me being lazy and not motivated to do `sed -i …` :)

mdellweg: Then you should consider this as "external motivation (tm)".

evgeni: I motivated my vim to do the change for me.

plugins/module_utils/foreman_helper.py
```diff
@@ -357,20 +358,31 @@ def _delete_entity(self, resource, current_entity, params):
...
         return True, None
 
-    def _resource_action(self, resource, action, params):
+    def resource_action(self, resource, action, params, **kwargs):
         resource_payload = self.foremanapi.resource(resource).action(action).prepare_params(params)
         try:
             result = None
             if not self.check_mode:
-                result = self.foremanapi.resource(resource).call(action, resource_payload)
+                result = self.foremanapi.resource(resource).call(action, resource_payload, **kwargs)
```
mdellweg: I guess, this is to pass `file=`?

Maybe the same reasoning as in
https://github.com/Apipie/apypie/pull/42#issuecomment-518642495
applies.

plugins/module_utils/foreman_helper.py
```diff
@@ -454,6 +466,33 @@ def __init__(self, argument_spec=None, **kwargs):
             args.update(argument_spec)
         super(KatelloEntityApypieAnsibleModule, self).__init__(argument_spec=args, **kwargs)
 
+    def connect(self, ping=True):
+        super(KatelloEntityApypieAnsibleModule, self).connect(ping)
+        self._patch_subscription_upload_api()
+        self._patch_organization_update_api()
+
+    def _patch_subscription_upload_api(self):
```
mdellweg: I am wondering whether we should put those in a separate file, as they start to clunk up the foreman_helper.

plugins/module_utils/foreman_helper.py
```diff
@@ -357,20 +358,31 @@ def _delete_entity(self, resource, current_entity, params):
...
                 action, resource, str(e)))
         return True, result
 
+    def wait_for_task(self, task, duration=60, poll=4):
+        if task['state'] not in ['paused', 'stopped']:
+            if duration > 0:
+                time.sleep(poll)
+                duration -= poll
+                _, task = self.resource_action('foreman_tasks', 'show', {'id': task['id']})
+                task = self.wait_for_task(task, duration, poll)
```
ekohl: Could this use a while loop instead of recursion?

plugins/modules/katello_manifest.py
```diff
@@ -76,116 +61,71 @@
...
-    organization = module.params['organization']
-    manifest_path = module.params['manifest_path']
-    redhat_repository_url = module.params['redhat_repository_url']
-    state = module.params['state']
+    organization = module.find_resource_by_name('organizations', name=entity_dict['organization'], thin=False)
+    scope = {'organization_id': organization['id']}
 
-    create_server(server_url, (username, password), verify_ssl)
-    ping_server(module)
+    existing_manifest = organization.get('owner_details', {}).get('upstreamConsumer')
```
mdellweg: `None` should have a dull `get`. ;)

plugins/modules/katello_manifest.py
```diff
@@ -76,116 +61,71 @@
...
+            org_spec = dict(redhat_repository_url=dict())
+            changed_url, organization = module.ensure_entity('organizations', payload, organization, state='present', entity_spec=org_spec)
+        else:
+            changed_url = False
+
+        try:
+            with open(entity_dict['manifest_path'], 'rb') as manifest_file:
+                files = {'content': (entity_dict['manifest_path'], manifest_file, 'application/zip')}
+                headers = {'content_type': 'multipart/form-data', 'multipart': 'true'}
+                params = scope.copy()
```
mdellweg: I think i'd prefer `params.update(scope)` later, but that's up to you.

plugins/modules/katello_manifest.py
```diff
@@ -76,116 +61,71 @@
...
+            payload = {'redhat_repository_url': entity_dict['repository_url']}
+            org_spec = dict(redhat_repository_url=dict())
+            changed_url, organization = module.ensure_entity('organizations', payload, organization, state='present', entity_spec=org_spec)
+        else:
+            changed_url = False
+
+        try:
+            with open(entity_dict['manifest_path'], 'rb') as manifest_file:
+                files = {'content': (entity_dict['manifest_path'], manifest_file, 'application/zip')}
+                headers = {'content_type': 'multipart/form-data', 'multipart': 'true'}
```
ekohl: Can't requests do this for us?

plugins/modules/katello_manifest.py
```diff
@@ -76,116 +61,71 @@
...
+                params = scope.copy()
+                if 'repository_url' in entity_dict:
+                    params['repository_url'] = entity_dict['repository_url']
+                changed, result = module.resource_action('subscriptions', 'upload', params, options={'skip_validation': True}, files=files)
+                task = module.wait_for_task(result, 300)  # 5 minutes
```
ekohl: Would writing `5 * 60` self-document it's 5 minutes?

plugins/modules/katello_manifest.py
```diff
@@ -76,116 +61,71 @@
...
+                for error in task['humanized']['errors']:
```
mdellweg: Should we manage errors in exceptions?
We can postpone this.

plugins/modules/katello_manifest.py
```diff
@@ -76,116 +61,71 @@
...
+                    if "same as existing data" in error:
+                        changed = False
+                    elif "older than existing data" in error:
+                        module.fail_json(msg="Manifest is older than existing data.")
+                    else:
+                        module.fail_json(msg="Upload of the manifest failed: %s" % error)
+                changed |= changed_url
+        except IOError as e:
+            module.fail_json(msg="Unable to open the manifest file: %s" % e)
+    elif module.state == 'absent' and existing_manifest:
```
mdellweg:
```suggestion
    elif module.desired_absent and existing_manifest:
```

plugins/modules/katello_manifest.py
```diff
@@ -76,116 +61,71 @@
...
+                for error in task['humanized']['errors']:
+                    if "same as existing data" in error:
+                        changed = False
+                    elif "older than existing data" in error:
+                        module.fail_json(msg="Manifest is older than existing data.")
+                    else:
+                        module.fail_json(msg="Upload of the manifest failed: %s" % error)
+                changed |= changed_url
+        except IOError as e:
+            module.fail_json(msg="Unable to open the manifest file: %s" % e)
```
ekohl:
```suggestion
            module.fail_json(msg="Unable to read the manifest file: %s" % e)
```

tests/vcr_python_wrapper.py
```diff
@@ -48,8 +48,8 @@ def query_matcher_ignore_proxy(r1, r2):
 
 
 def katello_manifest_body_matcher(r1, r2):
-    if r1.headers.get('content-type').startswith('multipart/form-data') and r2.headers.get('content-type').startswith('multipart/form-data'):
-        if r1.path.endswith('/subscriptions/upload') and r2.path.endswith('/subscriptions/upload'):
+    if r1.path.endswith('/subscriptions/upload') and r2.path.endswith('/subscriptions/upload'):
+        if r1.headers.get('content-type').startswith('multipart/form-data') and r2.headers.get('content-type').startswith('multipart/form-data'):
```
mdellweg: Is this a performance thing?

plugins/module_utils/foreman_helper.py
```diff
@@ -357,20 +358,31 @@ def _delete_entity(self, resource, current_entity, params):
...
             result = None
             if not self.check_mode:
-                result = self.foremanapi.resource(resource).call(action, resource_payload)
+                result = self.foremanapi.resource(resource).call(action, resource_payload, **kwargs)
         except Exception as e:
             self.fail_json(msg='Error while performing {} on {}: {}'.format(
                 action, resource, str(e)))
         return True, result
 
+    def wait_for_task(self, task, duration=60, poll=4):
```
evgeni: would this be more readable?

```python
    def wait_for_task(self, task, duration=60, poll=4):
        while duration > 0:
            if task['state'] not in ['paused', 'stopped']:
                time.sleep(poll)
                duration -= poll
                _, task = self.resource_action('foreman_tasks', 'show', {'id': task['id']})
            else:
                break
        else:
            self.fail_json(msg="Timout waiting for Task {}".format(task['id']))
        return task
```

or maybe even

```python
    def wait_for_task(self, task, duration=60, poll=4):
        while duration > 0:
            if task['state'] not in ['paused', 'stopped']:
                time.sleep(poll)
                duration -= poll
                _, task = self.resource_action('foreman_tasks', 'show', {'id': task['id']})
            else:
                return task
        self.fail_json(msg="Timout waiting for Task {}".format(task['id']))
```

plugins/module_utils/foreman_helper.py
mdellweg: I heard of a theorem...

evgeni: see https://github.com/theforeman/foreman-ansible-modules/pull/392/files#r311952665? :)

plugins/module_utils/foreman_helper.py
evgeni: yes.

plugins/modules/katello_manifest.py
evgeni: I see no `None.get` here, just a `{}.get` ;)

plugins/module_utils/foreman_helper.py
ekohl: The tail is not very nice. You do a show but you never check the result which is inefficient. So how about:

```python
def wait_for_task(self, task, duration=60, poll=4):
    while task['state'] not in ['paused', 'stopped']:
        duration -= poll
        if duration <= 0:
            self.fail_json(msg="Timout waiting for Task {}".format(task['id']))
        time.sleep(poll)

        _, task = self.resource_action('foreman_tasks', 'show', {'id': task['id']})

    return task
```

plugins/modules/katello_manifest.py
evgeni: probably, this is copy&paste from the old code, I didn't try without. will.

plugins/modules/katello_manifest.py
evgeni: Sometimes I miss rails *whistle*

```
> 5.minutes.to_i
=> 300
```

plugins/modules/katello_manifest.py
evgeni: Not sure I follow what you mean? `wait_for_tasks` does not raise anything at the moment.

plugins/modules/katello_manifest.py
ekohl: Perhaps the longer version is more readable:
```suggestion
    try:
        existing_manifest = organization['owner_details']['upstreamConsumer']
    except KeyError:
        existing_manifest = None
```

plugins/modules/katello_manifest.py
ekohl: Exactly what I was thinking. Usually I prefer Python but every now and then :)

plugins/modules/katello_manifest.py
ekohl: In my suggestion I was thinking about `raise TimeoutError` instead of `self.fail_json`

plugins/modules/katello_manifest.py
mdellweg: Sorry, i meant that as a feature request to python. ;)

plugins/module_utils/foreman_helper.py
evgeni: almost, I also pass `skip_validation`, because apypie can't validate the file :D

plugins/modules/katello_manifest.py
mdellweg: The question was whether it should raise exceptions, which would make handling them a bit more comfortable.

plugins/modules/katello_manifest.py
evgeni: it's not passed anyways :rofl:

tests/vcr_python_wrapper.py
evgeni: no, a GET request has no content and thus no content-type.

evgeni: so I want to match only the POST to the upload endpoint

plugins/modules/katello_manifest.py
evgeni: @mdellweg ah, maybe. But I like @ekohl's suggestion

plugins/modules/katello_manifest.py
mdellweg: I never understood, why ansible does not like exceptions for error-reporting.

plugins/module_utils/foreman_helper.py
evgeni: I like yours, applied

plugins/modules/katello_manifest.py
evgeni: Ah, OK, now I get it. Yeah we can make it raise various Exceptions and handle these here.

Not sure there is an added benefit though (besides being more Pythonic than Ansiblig)

evgeni:
> I love how we're fixing so many issues in the API documentation. Feels like nobody uses it.

Yes. Same feeling here.

PR #393, opened 2019-08-07
Refactor activation key

mdellweg:


plugins/modules/katello_activation_key.py
```diff
@@ -145,88 +114,151 @@ def override_to_boolnone(override):
...
-    (entity_dict, state) = module.parse_params()
+    entity_dict = module.clean_params()
 
     module.connect()
 
-    entity_dict['organization'] = find_organization(module, name=entity_dict['organization'])
-    if 'lifecycle_environment' in entity_dict:
-        entity_dict['lifecycle_environment'] = find_lifecycle_environment(module, entity_dict['lifecycle_environment'], entity_dict['organization'])
+    # TODO move to foreman_helper
+    module.foremanapi.apidoc['docs']['resources']['subscriptions']['methods'][0]['params'][7]['required'] = False
```
mdellweg: If i have not overlooked something, this is the WIP in this PR.

tests/fixtures/apidoc/activation_key.json
```diff
@@ -1 +1 @@
-katello.json
\ No newline at end of file
+katello-3.12.json
```
evgeni: Why not just making 3.12 the default?

plugins/modules/katello_activation_key.py
```diff
@@ -145,88 +114,151 @@ def override_to_boolnone(override):
...
-        if state == 'present' or (state == 'present_with_defaults' and changed):
-            if 'subscriptions' in entity_dict:
-                subscriptions = entity_dict['subscriptions']
-                desired_subscription_ids = set(s.id for s in find_subscriptions(module, subscriptions, entity_dict['organization']))
-                current_subscriptions = [Subscription(**result)
-                                         for result in Subscription().search_normalize(activation_key_entity.subscriptions()['results'])]
-                current_subscription_ids = set(s.id for s in current_subscriptions)
+        if module.state == 'present' or (module.state == 'present_with_defaults' and changed):
+            ak_scope = {'activation_key_id': activation_key['id']}
+            # ak_scope.update(scope)  # See Hack above.
```
evgeni: which hack?

plugins/modules/katello_activation_key.py
```diff
@@ -145,88 +114,151 @@ def override_to_boolnone(override):
...
+                changed_content_overrides = []
+
+                for label, override in desired_content_overrides.items():
+                    if override != current_content_overrides.pop(label, None):
+                        changed_content_overrides.append({'content_label': label, 'value': override})
+                for label in current_content_overrides.keys():
+                    changed_content_overrides.append({'content_label': label, 'reset': True})
+
+                if changed_content_overrides:
+                    if not module.check_mode:
```
evgeni: `_resource_action` already checks for check mode, so no need to repeat that here:

https://github.com/theforeman/foreman-ansible-modules/blob/ec9963682435244742e90f9fe26aaf03a5f4ed3f/plugins/module_utils/foreman_helper.py#L453-L462

plugins/modules/katello_activation_key.py
```diff
@@ -145,88 +114,151 @@ def override_to_boolnone(override):
...
+            # ak_scope.update(scope)  # See Hack above.
+            if subscriptions is not None:
+                subscription_names = [subscription['name'] for subscription in subscriptions]
+                desired_subscriptions = module.find_resources_by_name('subscriptions', subscription_names, params=scope, thin=True)
+                desired_subscription_ids = set(item['id'] for item in desired_subscriptions)
+                current_subscriptions = module.list_resource('subscriptions', params=ak_scope)
+                current_subscription_ids = set(item['id'] for item in current_subscriptions)
 
                 if desired_subscription_ids != current_subscription_ids:
                     if not module.check_mode:
```
evgeni: no need to check for checkmode (see below)

plugins/modules/katello_activation_key.py
```diff
@@ -145,88 +114,151 @@ def override_to_boolnone(override):
...
+                    current_host_collection_ids = set(activation_key['host_collection_ids'])
+                else:
+                    current_host_collection_ids = set(item['id'] for item in activation_key['host_collections'])
+                desired_host_collections = module.find_resources_by_name('host_collections', host_collections, params=scope, thin=True)
+                desired_host_collection_ids = set(item['id'] for item in desired_host_collections)
 
-                if desired_content_overrides != current_content_overrides:
+                module.warn(repr((desired_host_collection_ids, current_host_collection_ids)))
+                if desired_host_collection_ids != current_host_collection_ids:
                     if not module.check_mode:
```
evgeni: here too

plugins/modules/katello_activation_key.py
evgeni: What is this? method show, param org_id?

plugins/modules/katello_activation_key.py
mdellweg: You are right, i missed that.

plugins/modules/katello_activation_key.py
mdellweg: copy'n'paste for the win!

tests/fixtures/apidoc/activation_key.json
mdellweg: I will try.

plugins/modules/katello_activation_key.py
mdellweg: the organization_id was `reqired=True` which renders it matter of luck to use the right api_url. And the wrong one delivers the wrong answers.
So it's the API-Hack i made above that still needs to be made nice.

evgeni: ah, ok, now it makes sense. even tho I don't fully understand why apypie would select the wrong url if we pass the org_id here. apypie bug?

evgeni: aah, after looking at the docs, there are multiple routes:

* GET /katello/api/organizations/:organization_id/subscriptions 
List organization subscriptions
* GET /katello/api/activation_keys/:activation_key_id/subscriptions 
List an activation key's subscriptions
* GET /katello/api/subscriptions 

You obviously want the second one. Makes sense.

plugins/modules/katello_activation_key.py
```diff
@@ -145,88 +114,151 @@ def override_to_boolnone(override):
...
-                        for subscription_id in (current_subscription_ids - desired_subscription_ids):
-                            activation_key_entity.remove_subscriptions(data={'subscription_id': subscription_id})
+                        ids_to_remove = current_subscription_ids - desired_subscription_ids
+                        if ids_to_remove:
+                            payload = {
+                                'id': activation_key['id'],
+                                'subscriptions': [{'id': item} for item in ids_to_remove],
+                            }
+                            payload.update(scope)
+                            module._resource_action('activation_keys', 'remove_subscriptions', payload)
```
evgeni: in #392 I made `resource_action` fully public, as it felt weird to call `_resource_action` on anything but `self`. what do you think? or @ekohl?

mdellweg: I think it's not yet merged, but i like it.

plugins/modules/katello_activation_key.py
mdellweg: Index and org_id

evgeni: index, oups :)

plugins/module_utils/foreman_helper.py
```diff
@@ -454,6 +474,20 @@ def __init__(self, argument_spec=None, **kwargs):
...
 
+    def _patch_subscriptions_api(self):
+        """This is a workaround for the broken subscriptions apidoc in foreman.
+        """
+
+        _subscription_methods = self.foremanapi.apidoc['docs']['resources']['subscriptions']['methods']
+        _subscription_index = next(x for x in _subscription_methods if x['name'] == 'index')
+        _subscription_index_params_organization_id = next(x for x in _subscription_index['params'] if x['name'] == 'organization_id')
+        _subscription_index_params_organization_id['required'] = False
+        return
```
evgeni: what for?

plugins/module_utils/foreman_helper.py
```diff
@@ -454,6 +474,20 @@ def __init__(self, argument_spec=None, **kwargs):
...
+
+    def connect(self):
```
```suggestion
    def connect(self, ping=True):
```

plugins/module_utils/foreman_helper.py
```diff
@@ -454,6 +474,20 @@ def __init__(self, argument_spec=None, **kwargs):
...
+        super(KatelloEntityApypieAnsibleModule, self).connect()
```
```suggestion
        super(KatelloEntityApypieAnsibleModule, self).connect(ping)
```

plugins/module_utils/foreman_helper.py
```diff
@@ -417,6 +437,7 @@ def connect(self, ping=True):
         super(KatelloEntityApypieAnsibleModule, self).connect(ping)
         self._patch_subscription_upload_api()
         self._patch_organization_update_api()
+        self._patch_subscription_index_api()
```
evgeni: Not strictly your fault, but can we sort these patches? :)

plugins/module_utils/foreman_helper.py
```diff
@@ -441,6 +462,16 @@ def _patch_organization_update_api(self):
         _organization_update_params_organization = next(x for x in _organization_update['params'] if x['name'] == 'organization')
         _organization_update_params_organization['required'] = False
 
+    def _patch_subscription_index_api(self):
+        """This is a workaround for the broken subscriptions apidoc in katello.
```
evgeni: redmine link? :)

plugins/modules/katello_activation_key.py
```diff
@@ -144,88 +114,144 @@ def override_to_boolnone(override):
...
+
+            if host_collections is not None:
+                if 'host_collection_ids' in activation_key:
+                    current_host_collection_ids = set(activation_key['host_collection_ids'])
+                else:
+                    current_host_collection_ids = set(item['id'] for item in activation_key['host_collections'])
+                desired_host_collections = module.find_resources_by_name('host_collections', host_collections, params=scope, thin=True)
+                desired_host_collection_ids = set(item['id'] for item in desired_host_collections)
+
+                module.warn(repr((desired_host_collection_ids, current_host_collection_ids)))
```
evgeni:
```suggestion
```

plugins/modules/katello_activation_key.py
```diff
@@ -144,88 +114,144 @@ def override_to_boolnone(override):
...
+                    if ids_to_add:
+                        payload = {
+                            'id': activation_key['id'],
+                            'host_collection_ids': list(ids_to_add),
+                        }
+                        module.resource_action('activation_keys', 'add_host_collections', payload)
                     changed = True
 
         module.exit_json(changed=changed)
     except Exception as e:
```
evgeni: what kind of Exception can be raised here that is not yet caught by our `e2fail` decorator?

mdellweg: i guess none.

plugins/module_utils/foreman_helper.py
mdellweg: Can we piggyback https://projects.theforeman.org/issues/27527 ?

evgeni: I'd prefer not to, so I've created https://projects.theforeman.org/issues/27575

PR #394, opened 2019-08-08
Isolate nailgun

mdellweg:


.travis.yml
```diff
@@ -8,6 +8,9 @@ python:
 addons:
   apt:
     packages:
-    - rpm
+      - rpm
+  pip:
```
mdellweg: Seems like travis ignores this. ;(

evgeni: I don't think there is a `pip` addon?

mdellweg: twas worth a try. I woulda expect it to fail, though

plugins/module_utils/ansible_nailgun_cement.py
```diff
@@ -5,56 +5,110 @@
...
+
+
+class ForemanAnsibleModule(ForemanBaseAnsibleModule):
+
+    def check_requirements(self):
+        if not HAS_NAILGUN:
+            self.fail_json(msg='The nailgun Python module is required',
+                           exception=NAILGUN_IMP_ERR)
+
+    def connect(self, ping=True):
```
evgeni: interestingly, this does not actually call `ping()` (and never did it seems).

tests/test_crud.py
```diff
@@ -112,5 +112,4 @@ def run_playbook(module, extra_vars=None, limit=None):
 @pytest.mark.parametrize('module', MODULES)
 def test_crud(tmpdir, module, record):
     run = run_playbook_vcr(tmpdir, module, record=record)
-    print(run.stdout.read())
```
evgeni: this is not needed because because run playbook already will print stuff?

.gitignore
```diff
@@ -7,4 +7,5 @@ __pycache__
 *.pyo
 *.swp
 .pytest_cache/
+.mypy_cache/
```
evgeni: this seems unrelated?

mdellweg: Yes, unrelated. Mypy was able to pinpoint syntaxerrors codestyle did not find.

mdellweg: I'd like to keep it anyways.

tests/test_crud.py
mdellweg: Yes, i noticed the output (in case of a failure) was doubled.

mdellweg:
In general, we could remove the Apypie from the module names now. Do we want to do it in this PR?

evgeni:
I'd prefer not to right now.

.gitignore
evgeni: fine with me

PR #396, opened 2019-08-09
Port katello_lifecycle_environment to apypie

bagasse:
* Port katello_lifecycle_environment to apypie.
* Some changes global to katello to automaticaly scope entity to organization.
See #274.

### DIFF
--- a/plugins/modules/katello_lifecycle_environment.py
+++ b/plugins/modules/katello_lifecycle_environment.py
@@ -1,6 +1,7 @@
 #!/usr/bin/python
 # -*- coding: utf-8 -*-
 # (c) 2017, Andrew Kofink <ajkofink@gmail.com>
+# (c) 2019, Baptiste Agasse <baptiste.agasse@gmail.com>
 #
 # This file is part of Ansible
 #
@@ -23,9 +24,11 @@ module: katello_lifecycle_environment
 short_description: Create and Manage Katello lifecycle environments
 description:
     - Create and Manage Katello lifecycle environments
-author: "Andrew Kofink (@akofink)"
+author:
+  - "Andrew Kofink (@akofink)"
+  - "Baptiste Agasse (@bagasse)"
 requirements:
-    - "nailgun >= 0.28.0"
+    - "apypie"
 options:
   name:
     description:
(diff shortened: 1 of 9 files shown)


plugins/modules/katello_lifecycle_environment.py
```diff
@@ -70,96 +73,45 @@
...
-            desired_environment.prior = find_prior(module, "Library", organization) if prior is None else prior
-            if not module.check_mode:
-                desired_environment.create()
-            changed = True
-    elif current_environment is not None:
-        if not module.check_mode:
-            current_environment.delete()
-        changed = True
-    return changed
+from ansible.module_utils.foreman_helper import (
```
evgeni: this can go into one line :)

tests/test_playbooks/tasks/lifecycle_environment.yml
```diff
@@ -13,6 +13,7 @@
     name: "{{ lifecycle_environment_name }}"
     label: "{{ lifecycle_environment_label }}"
     organization: "{{ organization_name }}"
+    description: "{{ lifecycle_environment_description }}"
```
evgeni:
```suggestion
    description: "{{ lifecycle_environment_description | default(omit) }}"
```

as the description is optional

tests/test_playbooks/tasks/lifecycle_environment.yml
```diff
@@ -13,6 +13,7 @@
     name: "{{ lifecycle_environment_name }}"
     label: "{{ lifecycle_environment_label }}"
```
evgeni:
```suggestion
    label: "{{ lifecycle_environment_label | default(omit) }}"
```

the label is optional

plugins/modules/katello_lifecycle_environment.py
```diff
@@ -70,96 +73,45 @@
...
 
-    validate_params(module, state, label=label, description=description, prior=prior)
+    entity_dict['organization'] = module.find_resource_by_name('organizations', entity_dict['organization'], thin=True)
+    entity = module.find_katello_resource_by_name('lifecycle_environments', entity_dict['organization'], name=entity_dict['name'], failsafe=True)
+    if not module.desired_absent:
+        if 'prior' in entity_dict:
+            entity_dict['prior'] = module.find_katello_resource_by_name('lifecycle_environments', entity_dict['organization'], entity_dict['prior'], thin=True)
+        else:
+            # Default to 'Library' for new env with no 'prior' provided
+            if not entity:
```
evgeni: IMHO you can fold that into the else above, making it `elif not entity`

evgeni:
Looks pretty good, thanks @bagasse!

I think the only thing that I'd like to properly discuss is the helpers you used here and in #389, but you guessed as much already probably

bagasse:
@evgeni, thanks for feedback/suggestions, I removed all non lifecycle env related stuff and fixed that you pointed out.

plugins/modules/katello_lifecycle_environment.py
```diff
@@ -70,96 +73,43 @@
...
+    if not module.desired_absent:
+        if 'prior' in entity_dict:
+            entity_dict['prior'] = module.find_resource_by_name('lifecycle_environments', entity_dict['prior'], params=scope, thin=True)
+        # Default to 'Library' for new env with no 'prior' provided
+        elif not entity:
+            entity_dict['prior'] = module.find_resource_by_name('lifecycle_environments', 'Library', params=scope, thin=True)
+
+    if entity:
+        if 'label' in entity_dict and entity_dict['label'] and entity['label'] != entity_dict['label']:
+            module.fail_json(msg="Label cannot be updated on a lifecycle environment. %s ******* %s" % (entity_dict, entity))
```
evgeni: why do you print the two entities with `%s ******* %s` here?

plugins/modules/katello_lifecycle_environment.py
```diff
@@ -70,96 +73,43 @@
...
+
+        if 'prior' in entity_dict and entity['prior']['id'] != entity_dict['prior']['id']:
+            module.fail_json(msg="Prior cannot be updated on a lifecycle environment. %s ******* %s" % (entity_dict, entity))
```
evgeni: same

evgeni:
@bagasse I left two comments, but I think this is ready to merge as soon as these two are cleared :)

plugins/modules/katello_lifecycle_environment.py
bagasse: :man_facepalming:

bagasse:
@evgeni, sorry for these two last :man_facepalming: ... Fixed.

PR #400, opened 2019-08-09
[WIP] Add apypie API cache

bagasse:
Add API cache to limit API calls in apypie modules. Discussion started [here](https://github.com/theforeman/foreman-ansible-modules/pull/389#discussion_r312370277) with @mdellweg and @evgeni because hostgroup can make a lot of API calls to find out all related smart proxies.
First I started to implement with `search` as `key` as suggested @mdellweg but python dict cannot have keys like search string, so I have to implement it at another layer.

### DIFF
--- a/plugins/module_utils/foreman_helper.py
+++ b/plugins/module_utils/foreman_helper.py
@@ -157,6 +157,37 @@ def _exception2fail_json(msg='Generic failure: %s'):
     return decor
 
 
+class ForemanApiCache:
+
+    def __init__(self):
+        self._entities_cache = {'show': {}, 'index': {}}
+
+    def get_list(self, resource, search=None, params=None):
+        if resource in self._entities_cache['index']:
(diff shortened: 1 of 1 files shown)


mdellweg:
Atm, i see nothing, that puts anything into the cache.
Also, i had something siglhtly different in mind, where the user provides the function with a dictionary that contains earlier found instances, and the find_* fills fills that cache.
Also we need to incoporate the params into the cache lookup because it can scope the search and change the result (thin,...).

bagasse:
@mdellweg, the cache is working now. the The `thin` param is managed at an upper level. I've added the params because yes you are right it's mandatory.

bagasse:
I tested it on hostgroup tests and I see one API call less than without it (on smart proxies, because we use thin here and in our tests, the puppet_proxy and puppet_ca proxy are the same)

PR #405, opened 2019-08-13
Refactor content view

mdellweg:


plugins/modules/katello_content_view.py
```diff
@@ -95,106 +98,108 @@
...
                         # When changing to latest=False & version is the latest we must send 'content_view_version' to the server
-                        force_update.append('content_view_version')
+                        # Let's fake, it wasn't there...
+                        cvc_matched.pop('content_view_version', None)
+                        cvc_matched.pop('content_view_version_id', None)
                 if cvc_matched:
-                    cvc['composite_content_view'] = content_view_entity
-                    cvc_dict = sanitize_entity_dict(cvc, cvc_map)
-                    cvc_changed = naildown_entity_state(ContentViewComponent, cvc_dict, cvc_matched, 'present', module, force_update=force_update)
+                    changed |= module.ensure_entity_state('content_view_components', cvc, cvc_matched, state='present', entity_spec=cvc_entity_spec, params=ccv_scope)
```
mdellweg:
```suggestion
                    changed |= module.ensure_entity_state(
                        'content_view_components', cvc, cvc_matched, state='present', entity_spec=cvc_entity_spec, params=ccv_scope)
```

plugins/modules/katello_content_view.py
```diff
@@ -95,106 +98,109 @@
...
-                        cvc_matched = _cvc
-                force_update = list()
-                if 'version' in component:
-                    cvc['version'] = find_content_view_version(module, cvc['content_view'], version=component['version'])
+    # Also only act for composite content views
+    if (module.state == 'present' or (module.state == 'present_with_defaults' and changed)) and content_view_entity['composite'] and components is not None:
+        if not changed:
+            content_view_entity['content_view_components'] = entity['content_view_components']
+        current_cvcs = content_view_entity.get('content_view_components', [])
+        if components:
```
evgeni: we already checked for components 5 lines before

plugins/modules/katello_content_view.py
```diff
@@ -95,106 +98,109 @@
...
+            components_to_add = []
+            ccv_scope = {'composite_content_view_id': content_view_entity['id']}
+            for component in components:
+                cvc = {
+                    'content_view': module.find_resource_by_name('content_views', name=component['content_view'], params=scope),
+                    'latest': component['latest'],
```
evgeni: this will have to change if you drop the default `latest=False` above

plugins/modules/katello_content_view.py
```diff
@@ -95,106 +98,109 @@
...
+                }
+                cvc_matched = next((item for item in current_cvcs if item['content_view']['id'] == cvc['content_view']['id']), None)
+                if not cvc['latest']:
+                    version = component.get('content_view_version')
+                    if version is None:
```
evgeni: should this check happen BEFORE we did any changes? I think it should.

plugins/modules/katello_content_view.py
```diff
@@ -95,106 +98,109 @@
...
+                        module.fail_json(msg="Content View Component must either have latest=True or provide a Content View Version.")
+                    search = "content_view_id={},version={}".format(cvc['content_view']['id'], component['content_view_version'])
+                    cvc['content_view_version'] = module.find_resource('content_view_versions', search=search, thin=True)
                     cvc['latest'] = False
-                    if cvc_matched and cvc_matched.latest:
+                    if cvc_matched and cvc_matched['latest']:
                         # When changing to latest=False & version is the latest we must send 'content_view_version' to the server
-                        force_update.append('content_view_version')
+                        # Let's fake, it wasn't there...
```
evgeni: not sure what you're faking by removing things

evgeni: or do you fake it by removing stuff here, so that `ensure_entity` will add it to the fields to send?

plugins/modules/katello_content_view.py
```diff
@@ -95,106 +98,109 @@
...
-}
 
-cvc_map = {
-    'composite_content_view': 'composite_content_view',
-    'content_view': 'content_view',
-    'latest': 'latest',
-    'version': 'content_view_version',
+cvc_entity_spec = {
+    'content_view': {'type': 'entity', 'flat_name': 'content_view_id', 'required': True},
+    'latest': {'type': 'bool', 'default': False},
```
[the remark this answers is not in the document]
evgeni: after reading the code, the way it's currently is easier :)

plugins/modules/katello_content_view.py
```diff
@@ -95,106 +98,109 @@
...
     module.connect()
 
-    entity_dict['organization'] = find_organization(module, name=entity_dict['organization'])
-    if 'repositories' in entity_dict and not entity_dict['composite']:
-        entity_dict['repositories'] = find_repositories(module, entity_dict['repositories'], entity_dict['organization'])
+    entity_dict['organization'] = module.find_resource_by_name('organizations', entity_dict['organization'], thin=True)
+    scope = {'organization_id': entity_dict['organization']['id']}
+    if not module.desired_absent:
+        if 'repositories' in entity_dict:
+            if entity_dict['composite']:
```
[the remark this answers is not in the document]
evgeni: `components` != `composite`, @evgeni just can't read.

evgeni:
as @mdellweg is out, I took the liberty and addressed my comments myself.

@sean797 if you want to have a second look? :)

plugins/module_utils/foreman_helper.py
```diff
@@ -455,6 +455,9 @@ def resource_action(self, resource, action, params, options=None, data=None, fil
             result = None
             if not self.check_mode:
                 result = self.foremanapi.resource(resource).call(action, resource_payload, options=options, data=data, files=files)
+                if isinstance(result, dict) and 'action' in result and 'state' in result and 'pending' in result:
```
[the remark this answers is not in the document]
mdellweg: I think, this means the product should also wait for it's task.

plugins/modules/katello_content_view.py
mdellweg: exactly. This was called `force_update` before, i think.

mdellweg:
I cannot add a review on my own PR, but i think this can be merged.

mdellweg:
Thanks @evgeni

plugins/module_utils/foreman_helper.py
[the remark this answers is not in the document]
evgeni: in general, yes, I just didn't include it here for the sake of "let's decide on an api for that first" :)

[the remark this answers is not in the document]
mdellweg: Well, then lets agree on this one, though i would have called the parameter `non_blocking`.

evgeni:
updated.

@sean797 @akofink could you have a look here (and #423) for a second pair of eyes? @mdellweg and I agree on the code, but we both touched it :)

plugins/module_utils/foreman_helper.py
```diff
@@ -444,28 +447,32 @@ def _delete_entity(self, resource, current_entity, params):
...
+    def resource_action(self, resource, action, params, options=None, data=None, files=None, synchronous=True):
         resource_payload = self.foremanapi.resource(resource).action(action).prepare_params(params)
         if options is None:
             options = {}
         try:
             result = None
             if not self.check_mode:
                 result = self.foremanapi.resource(resource).call(action, resource_payload, options=options, data=data, files=files)
+                if synchronous and isinstance(result, dict) and 'action' in result and 'state' in result and 'pending' in result:
+                    # looks like a task
```
akofink: Nitpick: if you have to put a comment in a complicated condition like this, why not make a boolean variable that explains such?

```rb
is_foreman_task = synchronous and isinstance(result, dict) and 'action' in result and 'state' in result and 'pending' in result
if is_foreman_task:
  ...
```

plugins/modules/katello_content_view.py
```diff
@@ -25,20 +25,23 @@
     - Create and Manage Katello content views
 author: "Eric D Helms (@ehelms)"
 requirements:
-    - "nailgun >= 0.28.0"
+    - apypie
```
akofink: Not just a refactor, I guess?

plugins/modules/katello_content_view.py
```diff
@@ -95,106 +98,113 @@
...
-                else:
-                    for attr in ['latest', 'version']:
-                        if attr not in cvc:
-                            cvc[attr] = None
-                    ContentViewComponent(composite_content_view=content_view_entity, content_view=cvc['content_view'],
-                                         latest=cvc['latest'], content_view_version=cvc['version']).add()
-                    changed = True
-        for cvc in current_cvcs:
+    # Also only act for composite content views
+    if (module.state == 'present' or (module.state == 'present_with_defaults' and changed)) and content_view_entity['composite'] and components is not None:
```
akofink: Same here - consider using a boolean variable:

```suggestion
    composite_cv = (module.state == 'present' or (module.state == 'present_with_defaults' and changed)) and content_view_entity['composite'] and components is not None
    if compsite_cv:
```

plugins/modules/katello_content_view.py
```diff
@@ -95,106 +98,113 @@
...
+        ccv_scope = {'composite_content_view_id': content_view_entity['id']}
+        for component in components:
+            cvc = {
+                'content_view': module.find_resource_by_name('content_views', name=component['content_view'], params=scope),
+                'latest': component['latest'],
+            }
+            cvc_matched = next((item for item in current_cvcs if item['content_view']['id'] == cvc['content_view']['id']), None)
+            if not cvc['latest']:
+                search = "content_view_id={},version={}".format(cvc['content_view']['id'], component['content_view_version'])
+                cvc['content_view_version'] = module.find_resource('content_view_versions', search=search, thin=True)
```
akofink: `find_resource` inside a for loop makes me a bit concerned. Hopefully we've already ruled out using only a single API call?

mdellweg: At least not so easy.

plugins/modules/katello_content_view.py
mdellweg: This is more a question of whether we need to update the list of components.

plugins/modules/katello_content_view.py
mdellweg: Well it is refactoring in that the change of used library should be invisible to the user.
But yes this is primarily a conversion from Nailgun to apypie.

plugins/module_utils/foreman_helper.py
mdellweg: I am hesitant to define a variable to use it exactly once, when a little comment in the right place can provide the same information.

akofink: Never fear, your variable used only once will be optimized away, I'm sure. DRY is not the only reason to use a variable IMO.

plugins/modules/katello_content_view.py
akofink: I guess the comment isn't so clear, then... ;) `# Also only act for composite content views`

akofink:
Btw, I'm not blocking merge (I would have 'Requested Changes' if so). It looks fine to merge.

plugins/modules/katello_content_view.py
akofink: Something to consider: when you ask for my review these days, I'm coming at the code as more of an outsider now that I've been less involved. It's not simple code to follow IMO for new contributors or old ones like me. It's these long methods or long one liner conditionals that really get me confused.

plugins/module_utils/foreman_helper.py
mdellweg: changing it. ;)

plugins/modules/katello_content_view.py
mdellweg: Oh it goes together with `# only update CVC's of newly created or updated CCV's` (unchanged and above the changed block).
As a variable name, `update_components` come to my mind.

mdellweg: > Something to consider: when you ask for my review these days, I'm coming at the code as more of an outsider now that I've been less involved. It's not simple code to follow IMO for new contributors or old ones like me. It's these long methods or long one liner conditionals that really get me confused.

That is what is making your review still very helpful, because being deeply involved like evgeni and me gives you blind spots on different levels.

mdellweg: I'll split the boolean expression into two. Hopefully it's more readable.

mdellweg:
ping @akofink

PR #409, opened 2019-08-13
Correct documentation

m-czernek:
Hi,

the URL parameter is not mandatory when modifying an existing repository. Not sure whether it's mandatory otherwise. 

WDYT?

### DIFF
--- a/plugins/modules/katello_repository.py
+++ b/plugins/modules/katello_repository.py
@@ -56,7 +56,7 @@ options:
   url:
     description:
       - Repository URL to sync from
-    required: true
+    required: false (in case you're modifying an existing repository)
   docker_upstream_name:
     description:
       - name of the upstream docker repository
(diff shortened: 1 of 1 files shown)


plugins/modules/katello_repository.py
```diff
@@ -56,7 +56,7 @@
   url:
     description:
       - Repository URL to sync from
-    required: true
+    required: false (in case you're modifying an existing repository)
```
evgeni:
```suggestion
    required: false
```
`required` is a special param for an Ansible documentation block, so passing a string here might not have the expected result. Also, I don't think it's actually required when you create a new repo, as it's totally fine to have a repository that has no "upstream" to sync from.

evgeni: this actually failed on travis:

```
ansible.errors.AnsibleError: Invalid required value 'false (in case you're modifying an existing repository)' for parameter 'url' in 'katello_repository' (must be truthy)
```

m-czernek:
Ha, apologies. Should be better now.

evgeni:
We should also update tests for this. But that can go in an own PR/issue.

PR #412, opened 2019-08-13
search_facts: allow fetching full details of found resources

evgeni:
yes, this is missing tests atm :)


sean797:
:+1: to this -  thanks @evgeni

evgeni:
Added (simple) tests and ready for review :)

plugins/modules/foreman_search_facts.py
```diff
@@ -84,18 +91,26 @@ def main():
...
 
     module.connect()
 
-    response = module.list_resource(entity, search)
+    response = module.list_resource(resource, search)
+
+    if module_params['full_details']:
+        resources = []
+        for res in response:
+            resources.append(module.show_resource(resource, res['id']))
```
mdellweg: Why not as a comprehension (modern python)?
`resources = [module.show_resource(resource, res['id'])) for res in response]`
Nice how it it unclear whether res is the abbreviation of resource or response. :smiling_imp:

tests/test_playbooks/search_facts.yml
```diff
@@ -13,6 +13,11 @@
     vars:
       resource: subnets
       return_length: 0
+  - include: tasks/search_facts.yml
+    vars:
+      resource: organizations
+      full_details: true
+      return_length: 2
```
mdellweg: Is this `2` a good test result? I think it depends a lot on your test setup. And you did not even include the fixtures for organization, so i guess, 'Test Organization' isn't even one of those.

sean797:
Would you mind also adding something to the EXAMPLE section?

plugins/modules/foreman_search_facts.py
evgeni: > Nice how it it unclear whether res is the abbreviation of resource or response.

Damn, you noticed it :stuck_out_tongue_closed_eyes:

evgeni:
@mdellweg @sean797 adressed your comments, please have another look :)

PR #413, opened 2019-08-14
Shorten names of AnsibleModule derivatives

mdellweg:
The namespacing is already complete on the module level.

@evgeni Is it time for some sed magic?


evgeni:
You mean on the "module_utils" level? But yeah, I think this is fine. Even tho it means I'll have to update the detection code I wrote in #386

PR #414, opened 2019-08-14
drop deprecated katello modules

evgeni:


README.md
```diff
@@ -191,7 +191,6 @@ This is a list of modules currently in the repository (please add to the list if
 
  * katello_sync: sync Katello repositories and products
  * katello_upload: upload files, rpms, etc. to repositories. Note, rpms & files are idempotent.
- * katello_content_view_publish: publish Katello content views
```
mdellweg: Can you add katello_content_view_version to the list above? I think it is missing.

evgeni: I actually considered dropping the list completely in favor of https://theforeman.org/plugins/foreman-ansible-modules

mdellweg: ... which is autogenerated. :1st_place_medal: 
For the riddance!

evgeni: I'd do that in a separate PR tho.

mdellweg: Why? Do you want to collect PR-Points?

PR #415, opened 2019-08-14
Implement deprecation_marker in docs

mdellweg:


evgeni:
NACK. Ansible upstream renames the modules with a leading `_` to mark deprecations:
https://github.com/ansible/ansible/blob/69bc24f607c2e128768c50236df35ed355105fa4/hacking/build_library/build_ansible/command_plugins/plugin_formatter.py#L196-L215

(don't ask why they don't use the metadata)

See e.g. https://github.com/ansible/ansible/tree/devel/lib/ansible/modules/remote_management/foreman

docs/_templates/plugin.rst.j2
```diff
@@ -155,7 +160,7 @@ Parameters
                     {% endif %}
                     {# Show default value, when multiple choice or no choices #}
                     {% if value.default is defined and value.default not in value.choices %}
-                        <b>Default:</b><br/><div style="color: blue">@{ value.default | escape }@</div>
+                        <b>Default:</b><br/><div style="color: blue">@{ value.default | tojson | escape }@</div>
```
mdellweg: This line is an actual reimport.

Issue #419, opened 2019-08-15
katello_activation_key is missing params


evgeni:
##### SUMMARY

Looking at https://theforeman.org/plugins/katello/3.13/api/apidoc/v2/activation_keys/create.html, we're missing the following params when creating AKs:

* max_hosts
* unlimited_hosts

and starting with katello 3.13:

* purpose_usage
* purpose_role
* purpose_addons

##### ISSUE TYPE
 - Bug Report

##### ANSIBLE VERSION

```

```

##### KATELLO/FOREMAN VERSION

```

```

##### NAILGUN VERSION

```

```

##### STEPS TO REPRODUCE

```yaml

```

##### ACTUAL RESULTS

```

```

evgeni:
originally reported by @m-czernek on IRC

evgeni:
We fixed auto attach in #420 and release/service level i #421, so I dropped these from the description

PR #420, opened 2019-08-15
katello_activation_key: idempotent auto_attach

evgeni:
as written in #419, creating an AK with `auto_attach=false` does not work properly, it needs an `update` for the setting to work.


ekohl:
Should we open a Redmine issue to Katello and link to it as a comment?

evgeni:
@ekohl if I read the code correctly, this is intentional as this is basically exposing the candlepin API.

But yeah, the create+update could be done in Katello, so I will open a Redmine.

m-czernek:
+1 from me for the code change; I applied changes in the `katello_activation_key.py` file, and tested with both `True` and `False` values. This now works for both creating a new activation key, as well as modifying an already existing key (i.e. key's auto attach was switched to false, but we describe true in the code).

Thank you!

PR #421, opened 2019-08-15
k_activation_key: add release_version and support_level params

evgeni:
there are no tests for that because they would require working RHEL
subscriptions and synced RHEL content available

### DIFF
--- a/plugins/modules/katello_activation_key.py
+++ b/plugins/modules/katello_activation_key.py
@@ -57,6 +57,16 @@ options:
     description:
       - Set Auto-Attach on or off
     type: bool
+  release_version:
+    description:
+      - Set the content release version
+  service_level:
+    description:
+      - Set the service level
+    choices:
+      - Self-Support
+      - Standard
+      - Premium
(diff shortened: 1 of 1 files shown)


evgeni:
Actually, this does not completely fix #419 because I didn't add max_hosts and unlimited_hosts :woman_shrugging:

m-czernek:
Tested manually, after updating to the newest apypie package. The `release_version` parameter was a semi-blocking for us for configuration of Satellite. I have tested this PR in the following conditions:

- No release_version given: module seems to work as before
- release_version the same as the already-set version: returns OK, no change
- release_version is given, but not set in Satellite: returns changed, UI reflects release version change
-  release_version is given, but is different from an already set version in satellite: changed, UI reflects release version change 

The same works with creating new keys, not only updating already existing keys. I checked mainly the `release_version` parameter in depth, much less so the `service_level` one. 

Thank you for the patch!

evgeni:
dropped fixes, will add the others some time later

PR #424, opened 2019-08-16
port katello_sync to apypie

evgeni:


evgeni:
@Manisha15 @mdellweg @ehelms @sean797 plz review :)

plugins/modules/katello_sync.py
```diff
@@ -95,52 +95,38 @@
...
-    except Exception as e:
-        module.fail_json(msg=e)
+    params['organization'] = module.find_resource_by_name('organizations', params['organization'], thin=True)
+    scope = {'organization_id': params['organization']['id']}
+    params['product'] = module.find_resource_by_name('products', params['product'], params=scope, thin=True)
+    if 'repository' in params:
+        product_scope = {'product_id': params['product']['id']}
+        params['repository'] = module.find_resource_by_name('repositories', params['repository'], params=product_scope, thin=True)
+        changed, task = module.resource_action('repositories', 'sync', {'id': params['repository']['id']})
+        if params['synchronous']:
```
evgeni: this might become easier after we merge #405

mdellweg: ... which is the case now.

evgeni: just pushed an updated version

PR #427, opened 2019-08-16
 make foreman_host tests more modular

evgeni:
this PR achieves two things:
* it allows foreman_host tests to be recorded on a setup without provisioning
* it splits out the foreman_host_power tests into an own playbook, thus allowing to refactor both modules independantly

to achieve that, the following changes were made:
* implemented the managed param in foreman_host, so we can run tests with managed=false
* moved foreman_host_power tests into an own playbook
* enhanced the foreman_host test playbooks with better variable and fixture support


evgeni:
@mdellweg @Fobhep @Manisha15 I'd welcome some review :)

nesanton:
Hi @evgeni,

The test playbooks need some work for #429, but you've done most of it already in this PR. I could just add on top tests for the new ```build``` param that I introduced. Only https://github.com/theforeman/foreman-ansible-modules/pull/427/commits/9d4107ddb56e7d85482c72a09926b3f7346d5518 is off. I added the ```managed``` option too, so the rest of your changes still make sense.

evgeni:
@nesanton yeah, feel free to base your PR on this one, makes total sense imho

nesanton:
LGTM. Let's merge

PR #429, opened 2019-08-17
Port foreman_host to apypie and add build parameter

### LINKED ISSUE #402: ability to build a host
##### SUMMARY
When creating a host via API one can pass ```build: true```. This will setup all the necessary context required to build/provision this host. If this host is powered on within build token lifetime - it will get booted from PXE and installed according to context and settings of the host group it belongs to.

foreman_host does not seem to provide such an option.
Have I missed somthing?

In case there's indeed no such feature at the moment I can suggest a couple of options on how to add it:

1. Add a boolean ```build``` argument to the foreman_host module. In case the build token exists and has not expired yet, the module would not change anything on existing host. Otherwise, if host exists, but build token does not - the module would update the host, resulting in ```changed: true```

2. Add foreman_host_build module that does does the same as above but with its own idempotency logic, not affecting the idempotent behavior of the foreman_host module.

If you like I can try developing either of these two or other solution that you see fit.

##### ISSUE TYPE
 - Feature Idea

### LINKED ISSUE #426: foreman_host: requires hostgroup for state=absent
##### SUMMARY

It's documented that hostgroup is required for present/with_defaults, but on absent the code still tries to read it and fails.

##### ISSUE TYPE
 - Bug Report

##### ANSIBLE VERSION

```

```

##### KATELLO/FOREMAN VERSION

```

```

##### NAILGUN VERSION

```

```

##### STEPS TO REPRODUCE

```yaml

```

##### ACTUAL RESULTS

```
TASK [Ensure host 'test-host' is absent] ***************************************
task path: /home/egolov/Devel/theforeman/foreman-ansible-modules/tests/test_playbooks/tasks/host.yml:2
<tests> ESTABLISH LOCAL CONNECTION FOR USER: egolov
<tests> EXEC /bin/sh -c 'echo ~egolov && sleep 0'
<tests> EXEC /bin/sh -c '( umask 77 && mkdir -p "` echo /home/egolov/.ansible/tmp/ansible-tmp-1565957283.44-72843192982690 `" && echo ansible-tmp-1565957283.44-72843192982690="` echo /home/egolov/.ansible/tmp/ansible-tmp-1565957283.44-72843192982690 `" ) && sleep 0'
Using module file /home/egolov/Devel/theforeman/foreman-ansible-modules/plugins/modules/foreman_host.py
<tests> PUT /home/egolov/.ansible/tmp/ansible-local-7947hWnpLs/tmpT3O3ze TO /home/egolov/.ansible/tmp/ansible-tmp-1565957283.44-72843192982690/AnsiballZ_foreman_host.py
<tests> EXEC /bin/sh -c 'chmod u+x /home/egolov/.ansible/tmp/ansible-tmp-1565957283.44-72843192982690/ /home/egolov/.ansible/tmp/ansible-tmp-1565957283.44-72843192982690/AnsiballZ_foreman_host.py && sleep 0'
<tests> EXEC /bin/sh -c '../vcr_python_wrapper.py /home/egolov/.ansible/tmp/ansible-tmp-1565957283.44-72843192982690/AnsiballZ_foreman_host.py && sleep 0'
<tests> EXEC /bin/sh -c 'rm -f -r /home/egolov/.ansible/tmp/ansible-tmp-1565957283.44-72843192982690/ > /dev/null 2>&1 && sleep 0'
The full traceback is:
  File "../vcr_python_wrapper.py", line 130, in <module>
    exec(code)
  File "/home/egolov/.ansible/tmp/ansible-tmp-1565957283.44-72843192982690/AnsiballZ_foreman_host.py", line 114, in <module>
    _ansiballz_main()
  File "/home/egolov/.ansible/tmp/ansible-tmp-1565957283.44-72843192982690/AnsiballZ_foreman_host.py", line 106, in _ansiballz_main
    invoke_module(zipped_mod, temp_path, ANSIBALLZ_PARAMS)
  File "/home/egolov/.ansible/tmp/ansible-tmp-1565957283.44-72843192982690/AnsiballZ_foreman_host.py", line 49, in invoke_module
    imp.load_module('__main__', mod, module, MOD_DESC)
  File "/tmp/ansible_foreman_host_payload_yLJrhc/__main__.py", line 151, in <module>
  File "/tmp/ansible_foreman_host_payload_yLJrhc/__main__.py", line 130, in main
KeyError: 'hostgroup'

fatal: [tests]: FAILED! => {
    "changed": false, 
    "module_stderr": "Traceback (most recent call last):\n  File \"../vcr_python_wrapper.py\", line 130, in <module>\n    exec(code)\n  File \"/home/egolov/.ansible/tmp/ansible-tmp-1565957283.44-72843192982690/AnsiballZ_foreman_host.py\", line 114, in <module>\n    _ansiballz_main()\n  File \"/home/egolov/.ansible/tmp/ansible-tmp-1565957283.44-72843192982690/AnsiballZ_foreman_host.py\", line 106, in _ansiballz_main\n    invoke_module(zipped_mod, temp_path, ANSIBALLZ_PARAMS)\n  File \"/home/egolov/.ansible/tmp/ansible-tmp-1565957283.44-72843192982690/AnsiballZ_foreman_host.py\", line 49, in invoke_module\n    imp.load_module('__main__', mod, module, MOD_DESC)\n  File \"/tmp/ansible_foreman_host_payload_yLJrhc/__main__.py\", line 151, in <module>\n  File \"/tmp/ansible_foreman_host_payload_yLJrhc/__main__.py\", line 130, in main\nKeyError: 'hostgroup'\n", 
    "module_stdout": "", 
    "msg": "MODULE FAILURE\nSee stdout/stderr for the exact error", 
    "rc": 1
}

```

nesanton:
migrated the module to apypie and also added the ```build``` option

Known issues (or features):
* ```name``` has to be specified without domain part
Not sure if that's how it should be, but I decided not to change the behavior of the original module just yet. If there no domain defined in the ```hostgroup```, module will fail due to RFC in-compliant name (ending with period), which was also the case in the original module.
* ```hostgroup``` is required. It makes sense up until this module gets all the possible individual parameters for a host

Fixes #402 
Fixes #426


evgeni:
Thanks!

Yeah, this "short name to foreman_host and then we generate the full name with the hostgroup" is something I'd personally would like to get rid of (or at least make it optional).

As for requiring the HG at all: in https://github.com/theforeman/foreman-ansible-modules/pull/427 I added a "managed" param, if you set that to false, Foreman will require just the name to create a host, which I found useful.

plugins/modules/foreman_host.py
```diff
@@ -74,74 +74,43 @@
...
-# This is the only true source for names (and conversions thereof)
-name_map = {
-    'name': 'name',
-    'enabled': 'enabled',
-    'hostgroup': 'hostgroup',
-    'location': 'location',
-    'organization': 'organization',
-}
+from ansible.module_utils.foreman_helper import (
+    ForemanEntityAnsibleModule,
```
evgeni: This is unused, right?

plugins/modules/foreman_host.py
```diff
@@ -74,74 +74,43 @@
...
-        argument_spec=dict(
+        entity_spec=dict(
             name=dict(required=True),
-            hostgroup=dict(),
-            location=dict(),
-            organization=dict(),
+            hostgroup=dict(type='entity', required=True, flat_name='hostgroup_id'),
+            location=dict(type='entity', flat_name='location_id'),
+            organization=dict(type='entity', flat_name='organization_id'),
             enabled=dict(default='true', type='bool'),
```
evgeni: Did you drop the "present with defaults" state intentionally? I think the host had no parameters yet where this was useful, just checking.

plugins/modules/foreman_host.py
```diff
@@ -74,74 +74,43 @@
...
-            state=dict(default='present', choices=[
-                       'present_with_defaults', 'present', 'absent']),
-        ),
-        required_if=(
-            ['state', 'present_with_defaults', ['hostgroup']],
-            ['state', 'present', ['hostgroup']],
-        ),
```
evgeni: I think it doesn't need a default

plugins/modules/foreman_host.py
nesanton: correct, removing

plugins/modules/foreman_host.py
nesanton: Yes, I thought the same. Could it maybe be used to drop the hostgroup from a host?

plugins/modules/foreman_host.py
nesanton: build? agreed!

nesanton:
> This will require to re-record the test fixtures. Do you need help with that?

I'm afraid, yes. I can contribute to the test playbook itself

nesanton:
Reworked the code with the following in mind:
* using fqdn instead of shortname in name
This allows to search for entity without relying on ```hostgroup```. E.g. when removing a host without specifying a ```hostgroup```. As @evgeni suggested we could add another option for ```shortname``` that would require ```hostgroup``` and make them ```require_one_of(name, shortname)```
* using ```managed``` bool to decide if ```hostgroup``` is required.
This allows for creating a dummy host without ```hostgroup``` now and when more options will be added there won't be need to change much

plugins/modules/foreman_host.py
evgeni: dunno. that's @Fobhep's code :)

Fobhep:
@nesanton awesome thanks a lot already! I have no time today, but I will check tomorrow in detail!

plugins/modules/foreman_host.py
```diff
@@ -63,6 +86,25 @@
...
     state: present
 
+- name: "Create a host with build context"
+  foreman_host:
+    username: "admin"
+    password: "changeme"
+    server_url: "https://foreman.example.com"
+    name: "new_host"
+    hostgroup: my_hostgroup
+    build: yes
```
evgeni:
```suggestion
    build: true
```
(Ansible accepts "yes", but I prefer to have "real" booleans in the examples)

plugins/modules/foreman_host.py
```diff
@@ -63,6 +86,25 @@
...
+    state: present
+
+- name: "Create an unmanaged host"
+  foreman_host:
+    username: "admin"
+    password: "changeme"
+    server_url: "https://foreman.example.com"
+    name: "new_host"
+    managed: no
```
evgeni:
```suggestion
    managed: false
```

plugins/modules/foreman_host.py
```diff
@@ -74,74 +116,48 @@
...
-        argument_spec=dict(
+        entity_spec=dict(
             name=dict(required=True),
-            hostgroup=dict(),
-            location=dict(),
-            organization=dict(),
+            hostgroup=dict(type='entity', flat_name='hostgroup_id'),
+            location=dict(type='entity', flat_name='location_id'),
+            organization=dict(type='entity', flat_name='organization_id'),
             enabled=dict(default='true', type='bool'),
```
evgeni: while not changed by you, what do you think about dropping the default from `enabled` too?

evgeni:
This looks pretty good to me! If we get #427 in soon -- or just base this PR ontop of it directly, we can update the test fixtures easily too and should have everything neat and green.

plugins/modules/foreman_host.py
nesanton: sure, it must be homogeneous

plugins/modules/foreman_host.py
nesanton: yep

plugins/modules/foreman_host.py
nesanton: agree

evgeni:
Thanks @nesanton!

A few things to think about/consider/whatever:
* this PR changes the behavior of `name` from "short name" to "fqdn"
  * I think this is the correct thing to do, but would like an ACK from @Fobhep 
  * this (indirectly) solves https://github.com/theforeman/foreman-ansible-modules/issues/426, yay!
  * this also allows moving hosts between hostgroups, which I think is a nice feature
  * we might want to add a check that `name` contains at least one `.`?
* this PR drops support for `state=present_with_defaults`
  * again, I think that's fine, but would like to have an ACK from @Fobhep

Fobhep:
so as I see @evgeni did already a review (thanks ;) ) few for me to add :D

>     * this PR changes the behavior of `name` from "short name" to "fqdn"
>       
>       * I think this is the correct thing to do, but would like an ACK from @Fobhep

I think that is a good idea. I was originally considering to do that anyways. For reasons I can't recall anymore, I went for the short name. Let's change that!
>       * we might want to add a check that `name` contains at least one `.`?
I also agree upon that check

>     * this PR drops support for `state=present_with_defaults`
>       * again, I think that's fine, but would like to have an ACK from @Fobhep
I implemented present_with_defaults without really questioning myself wether it makes sense or not. Are there scenarios where you need to create an emtpy host shell in foreman and change it later on?

Besides that thank you very much @nesanton! Great work :)

plugins/modules/foreman_host.py
```diff
@@ -86,76 +116,48 @@
...
+        entity_spec=dict(
             name=dict(required=True),
-            hostgroup=dict(),
-            location=dict(),
-            organization=dict(),
-            enabled=dict(default='true', type='bool'),
+            hostgroup=dict(type='entity', flat_name='hostgroup_id'),
+            location=dict(type='entity', flat_name='location_id'),
+            organization=dict(type='entity', flat_name='organization_id'),
+            enabled=dict(type='bool'),
```
Fobhep: just for my understanding. you dropped the default enabled=true, because enabled: true and managed: false make no sense together?

evgeni: my reasoning was: enabled, build and managed are *optional* args in the API and the API will set them to reasonable defaults, so there is no reason to copy this default from Foreman into the module

Fobhep: ok, that make sense. thx

plugins/modules/foreman_host.py
```diff
@@ -86,76 +119,62 @@
...
+    entity_dict = module.clean_params()
+
+    # additional param validation
+    if not module.desired_absent:
+        if 'hostgroup' not in entity_dict and entity_dict.get('managed', True):
+            module.fail_json(msg='Hostgroup can be omitted only with managed=False')
+
+        if 'build' in entity_dict:
+            # When 'build'=True, 'managed' has to be True.
+            # Assuming that user's priority is to build
```
nesanton: I tried a few different approaches. This one seems most logical to me.
There are two contradicting facts:
* when ```build```=true, managed has to be true, otherwise call fails
* when ```managed=false``` (we want to unmanage host) and existing host has both at true, we have to pass ```build=false```, otherwise call fails
So I decided that if user passes ```build``` explicitly then it takes precedence.

plugins/modules/foreman_host.py
```diff
@@ -86,76 +119,62 @@
...
+            enabled=dict(type='bool'),
             managed=dict(type='bool'),
-            state=dict(default='present', choices=[
-                       'present_with_defaults', 'present', 'absent']),
+            build=dict(type='bool'),
         ),
         required_if=(
-            ['state', 'present_with_defaults', ['hostgroup']],
-            ['state', 'present', ['hostgroup']],
+            ['managed', True, ['hostgroup']],
```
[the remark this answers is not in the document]
evgeni: lets not over-complicate this, I think the current dependency (`(build, managed) → hostgroup)`) is just fine and we can re-visit later if that ever occurs

plugins/modules/foreman_host.py
```diff
@@ -86,76 +119,62 @@
...
+            module.fail_json(msg='Hostgroup can be omitted only with managed=False')
+
+        if 'build' in entity_dict:
+            # When 'build'=True, 'managed' has to be True.
+            # Assuming that user's priority is to build
+            if entity_dict['build']:
+                if 'managed' in entity_dict:
+                    if not entity_dict['managed']:
+                        # Give a warning only when 'managed' is explicitly passed as False
+                        module.warn('when \'build\'=True, \'managed\' is ignored and forced to True')
```
evgeni:
```suggestion
                        module.warn("when 'build'=True, 'managed' is ignored and forced to True")
```

plugins/modules/foreman_host.py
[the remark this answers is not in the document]
nesanton: Thought the same

plugins/modules/foreman_host.py
nesanton: yeah, I come from place where double quotes were banned ;) Single quotes around 'build' are printed unescaped by ansible, hence the choice.

plugins/modules/foreman_host.py
evgeni: I think that totally makes sense. Just wondering if the following code would be more readable:

```python
if 'build' in entity_dict and entity_dict['build'] and 'managed' in entity_dict and not entity_dict['managed']:
     # When 'build'=True, 'managed' has to be True. Assuming that user's priority is to build.
      module.warn("when 'build'=True, 'managed' is ignored and forced to True")
      entity_dict['managed'] = True
elif 'build' not in entity_dict and 'managed' in entity_dict and not entity_dict['managed']:
    # When 'build' is not given and 'managed'=False, have to clear 'build' context that might exist on the server
    entity_dict['build'] = False
```

evgeni:
FTR, I tried re-recording the test fixtures for this PR, and the only change I needed was (the obvious):

```
diff --git a/tests/test_playbooks/tasks/host.yml b/tests/test_playbooks/tasks/host.yml
index 4267712..2a34fbb 100644
--- a/tests/test_playbooks/tasks/host.yml
+++ b/tests/test_playbooks/tasks/host.yml
@@ -1,7 +1,7 @@
 ---
 - name: "Ensure host '{{ host_name }}' is {{ host_state }}"
   vars:
-    host_name: "test-host"
+    host_name: "test-host.{{ host.domain }}"
     host_state: present
   foreman_host:
```

After thinking a bit, I'd do the following tho:

```
diff --git a/tests/test_playbooks/host.yml b/tests/test_playbooks/host.yml
index 8b06d42..f1f66ab 100644
--- a/tests/test_playbooks/host.yml
+++ b/tests/test_playbooks/host.yml
@@ -34,7 +34,7 @@
         hostgroup_state: present
     - include_tasks: tasks/host.yml
       vars:
-        host_hostgroup: "{{ host.hostgroup }}"
+        host_domain: "{{ host.domain }}"
         host_state: absent
 
 - hosts: tests
@@ -46,6 +46,7 @@
     - name: create host
       include_tasks: tasks/host.yml
       vars:
+        host_domain: "{{ host.domain }}"
         host_organization: "{{ host.organization }}"
         host_location: "{{ host.location }}"
         host_hostgroup: "{{ host.hostgroup }}"
@@ -55,6 +56,7 @@
     - name: create host again, no change
       include_tasks: tasks/host.yml
       vars:
+        host_domain: "{{ host.domain }}"
         host_organization: "{{ host.organization }}"
         host_location: "{{ host.location }}"
         host_hostgroup: "{{ host.hostgroup }}"
@@ -65,14 +67,14 @@
       include_tasks: tasks/host.yml
       vars:
         host_state: absent
-        host_hostgroup: "{{ host.hostgroup }}"
+        host_domain: "{{ host.domain }}"
         expected_change: true
 
     - name: delete host again, no change
       include_tasks: tasks/host.yml
       vars:
         host_state: absent
-        host_hostgroup: "{{ host.hostgroup }}"
+        host_domain: "{{ host.domain }}"
         expected_change: false
 
 - hosts: fixtures
diff --git a/tests/test_playbooks/tasks/host.yml b/tests/test_playbooks/tasks/host.yml
index 4267712..631544d 100644
--- a/tests/test_playbooks/tasks/host.yml
+++ b/tests/test_playbooks/tasks/host.yml
@@ -1,7 +1,7 @@
 ---
 - name: "Ensure host '{{ host_name }}' is {{ host_state }}"
   vars:
-    host_name: "test-host"
+    host_name: "test-host.{{ host_domain }}"
     host_state: present
   foreman_host:
```

(this drops the `host_hostgroup` from the absent calls, because it's not needed, and passes `host_domain` from the main playbook, removing the dependency on the `host` dict in `tasks/host.yml`)

plugins/modules/foreman_host.py
nesanton: Sure, will collapse the conditions.
This one above will allow for managed=false build=true if managed omitted and host exists with managed=false. Will fail with "can't build unmanaged host"

evgeni:
@nesanton the tests currently fail because you recorded them against `localhost:8080` and we have the `port` matcher enabled in VCR. I've opened https://github.com/theforeman/foreman-ansible-modules/pull/434 to disable it. You can also do a `sed -i 's/localhost:8080/foreman.example.com/g' tests/test_playbooks/fixtures/host-*` ;)

plugins/modules/foreman_host.py
```diff
@@ -50,18 +51,29 @@
     description:
       - Name of related organization
     required: false
+  build:
+    description:
+      - Whether or not to setup build context for the host
+    type: bool
+    required: false
+    default: None
```
evgeni:
```suggestion
```

plugins/modules/foreman_host.py
```diff
@@ -50,18 +51,29 @@
...
   enabled:
     description:
       - Include this host within Foreman reporting
     type: bool
     required: false
-    default: true
+    default: None
```
evgeni:
```suggestion
```

plugins/modules/foreman_host.py
```diff
@@ -50,18 +51,29 @@
...
   managed:
     description:
-      - whether a host is managed or unmanaged
+      - Whether a host is managed or unmanaged.
+      - Forced to true when I(build=true)
     type: bool
     required: false
     default: None
```
evgeni: this conflicts after I've merged https://github.com/theforeman/foreman-ansible-modules/pull/433 -- sorry.

nesanton: merged
and added docstring for state

nesanton:
> @nesanton the tests currently fail because you recorded them against `localhost:8080` and we have the `port` matcher enabled in VCR. I've opened #434 to disable it. You can also do a `sed -i 's/localhost:8080/foreman.example.com/g' tests/test_playbooks/fixtures/host-*` ;)

haha, yeah I noticed. I was using proxy to have insight on what was being sent to api

plugins/modules/foreman_host.py
```diff
@@ -50,17 +51,27 @@
...
     required: false
-    default: true
   managed:
     description:
-      - whether a host is managed or unmanaged
+      - Whether a host is managed or unmanaged.
+      - Forced to true when I(build=true)
     type: bool
     required: false
+    default: None
```
evgeni:
```suggestion
```

plugins/modules/foreman_host.py
```diff
@@ -85,76 +115,57 @@
...
+            ['managed', True, ['hostgroup']],
+            ['build', True, ['hostgroup']],
         ),
     )
 
-    (host_dict, state) = module.parse_params()
+    entity_dict = module.clean_params()
 
-    module.connect()
+    # additional param validation
```
evgeni: this should add the "is the hostname a FQDN" check we discussed
```suggestion
    # additional param validation
    if '.' not in entity_dict['name']:
        module.fail_json(msg="The host name must include the domain")
```

tests/test_playbooks/tasks/host.yml
```diff
@@ -1,7 +1,7 @@
 ---
 - name: "Ensure host '{{ host_name }}' is {{ host_state }}"
   vars:
-    host_name: "test-host"
+    host_name: "test-host.{{ host.domain }}"
```
evgeni: this makes `tasks/host.yml` dependant on `vars/host.yml` and makes it hard for the user to override the domain of one host. how about we omit `host_name` here completely and explicitly set it in every call in `test_playbooks/host.yml`?

plugins/modules/foreman_host.py
nesanton: silly check to my taste, but we can't really do any more yet. Done

tests/test_playbooks/tasks/host.yml
nesanton: true, moving it there

plugins/modules/foreman_host.py
evgeni: Silly because it's just checking for a dot? Yeah. But better than nothing :)
Thanks!

tests/test_playbooks/tasks/host.yml
nesanton: will also remove host_state hardcoded to present. State is present by default anyway

evgeni:
all my comments and concerns were addressed, so absolute :+1: from me

@Fobhep any last words from you before we merge this?

Fobhep:
> all my comments and concerns were addressed, so absolute +1 from me
> 
> @Fobhep any last words from you before we merge this?

only lgtm :+1: and thanks again  ;)

Issue #428, opened 2019-08-16
foreman_subnet fails on subsequent re-runs


m-czernek:
##### SUMMARY

It seems like foreman_subnet is not idempotent. On the first pass, I can create a network. On the second pass, I get an error.

(latest Ansible, latest foreman-ansible-modules, latest Satellite, running on RHEL)

##### ISSUE TYPE
 - Bug Report

##### ACTUAL RESULTS

```
WARNING: The below traceback may *not* be related to the actual failure.
  File "/tmp/ansible_foreman_subnet_payload_1mnM8N/ansible_foreman_subnet_payload.zip/ansible/module_utils/foreman_helper.py", line 457, in resource_action
    result = self.foremanapi.resource(resource).call(action, resource_payload, options=options, data=data, files=files)
  File "/home/ansible/foreman-ansible-modules-venv/lib/python2.7/site-packages/apypie/resource.py", line 30, in call
    return self.api.call(self.name, action, params, headers, options, data, files)
  File "/home/ansible/foreman-ansible-modules-venv/lib/python2.7/site-packages/apypie/api.py", line 157, in call
    action.validate(params, data, files)
  File "/home/ansible/foreman-ansible-modules-venv/lib/python2.7/site-packages/apypie/action.py", line 53, in validate
    self._validate(self.params, values, data, files)
  File "/home/ansible/foreman-ansible-modules-venv/lib/python2.7/site-packages/apypie/action.py", line 74, in _validate
    raise MissingArgumentsError(message)

failed: [dh2.test.local -> localhost] (item={u'dns_primary': u'192.168.20.1', u'to_ip': u'192.168.20.30', u'name': u'test-network', u'from_ip': u'192.168.20.10', u'domains': [u'test.local'], u'cidr': 24, u'gateway': u'192.168.20.1', u'network': u'192.168.20.0'}) => {"ansible_loop_var": "item", "changed": false, "item": {"cidr": 24, "dns_primary": "192.168.20.1", "domains": ["test.local"], "from_ip": "192.168.20.10", "gateway": "192.168.20.1", "name": "test-network", "network": "192.168.20.0", "to_ip": "192.168.20.30"}, "msg": "Error while performing update on subnets: The following required parameters are missing: subnet"}
```

evgeni:
Weird. Can you post the corresponding play book?

evgeni:
[...]

PR #430, opened 2019-08-19
patch apidoc to include rex proxy ids for subnets

evgeni:
Fixes: #428

### DIFF
--- a/plugins/module_utils/foreman_helper.py
+++ b/plugins/module_utils/foreman_helper.py
@@ -204,6 +204,10 @@ class ForemanAnsibleModule(ForemanBaseAnsibleModule):
             see https://projects.theforeman.org/issues/19086
         """
 
+        if 'remote_execution_features' not in self.foremanapi.apidoc['docs']['resources']:
+            # the system has no foreman_remote_execution installed, no need to patch
+            return
+
         _subnet_rex_proxies_parameter = {
             u'validations': [],
             u'name': u'remote_execution_proxy_ids',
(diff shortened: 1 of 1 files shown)


m-czernek:
I applied the patch and tested in my environment. It fixes the fact that:

- update calls are not idempotent
- remote execution proxy settings is ignored in the create calls

evgeni:
@bagasse @mdellweg thoughts? What I don't like is that this is applied *always*, not only to setups that actually have REX enabled. But this shouldn't be a problem?

evgeni:
@m-czernek can you try the latest version of this? should work just fine on your setup, but avoid patching when REX is not installed.

m-czernek:
I have this in my mind. I'm now working on something different, but will definitely test this within the next week.

mdellweg:
Is this part of api only missing in the saved apidoc from a plain foreman without rex?

evgeni:
No, this is missing from instances with REX installed too, see https://projects.theforeman.org/issues/19086

evgeni:
@m-czernek I'll merge this now. Feel free to yell at me if it turns out it doesn't work for you.

m-czernek:
@evgeni Just wanted to get back to you, this still works (tested it today against master).

It's slightly unfortunate that the exit value is always `changed`, but that's a very minor thing.

Thank you once again!

evgeni:
The exit is always changed? Mhh. I'll have to look at that. But later ;)

evgeni:
Oh right, of course it is, the api doesn't answer with the id eitjer

PR #431, opened 2019-08-20
make katello_repository_set tests reproducible

evgeni:
this adds a setup/teardown step and creates an org and manifest (no
manifest, no subs, no reposets)


evgeni:
@akofink @sean797 :eyes:

PR #432, opened 2019-08-21
migrate katello_repository_set to apypie

evgeni:


evgeni:
@akofink @sean797 :eyes:

evgeni:
this requires latest apypie master with https://github.com/Apipie/apypie/pull/51 applied, thanks to https://projects.theforeman.org/issues/27673 ;)

plugins/modules/katello_repository_set.py
```diff
@@ -140,68 +132,68 @@ def get_desired_repos(desired_substitutions, available_repos):
...
-    if product:
-        product = find_product(module, product, organization)
-    repo_set = find_repository_set(module, name=name, product=product, organization=organization, label=label)
+def main():
+    module = KatelloEntityAnsibleModule(
+        argument_spec=dict(
+            name=dict(default=None),
+            product=dict(default=None),
+            label=dict(default=None),
+            repositories=dict(required=True, type='list'),
```
sean797: Should this use `type='nested_list'` ?

evgeni: `nested_list` is only really useful when we actually use the `entity_spec`, but this module doesn't because the `repository_set` API is very katello-like ;)

evgeni: Did you mean to use Ansible's own `elements='dict'`, and `options=…` (as done by `nested_list`)

plugins/modules/katello_repository_set.py
```diff
@@ -140,68 +132,68 @@ def get_desired_repos(desired_substitutions, available_repos):
...
+            state=dict(default='enabled', choices=['disabled', 'enabled']),
+        ),
+        required_one_of=[['label', 'name']],
+    )
```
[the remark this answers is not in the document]
evgeni: But is it required? Looking at the old code and the api docs, passing a product(_id) is optional and our tests include a "name only" task:

https://github.com/theforeman/foreman-ansible-modules/blob/853f984ef98b5a105a94b14e62a4287e23342193/tests/test_playbooks/repository_set.yml#L20-L23

plugins/modules/katello_repository_set.py
evgeni: I've set [`elements='dict'`](https://docs.ansible.com/ansible/latest/dev_guide/developing_program_flow_modules.html#elements) now,

plugins/modules/katello_repository_set.py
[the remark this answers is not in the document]
sean797: Ah yes, you can send one of name + product (+org), name + org or label + org.

akofink:
Nice work! Sorry I'm not as available to review these days :(

PR #435, opened 2019-08-23
use the new ansible doc building framework

evgeni:


.travis.yml
```diff
@@ -12,4 +12,6 @@ addons:
 install:
   - pip install pytest-travis-fold
   - make test-setup doc-setup
```
ekohl: Do you want to save time and skip doc-setup on Python 2.7 here?

evgeni: good idea, done

PR #437, opened 2019-08-23
 error out when trying to delete a location that has children

evgeni:


sean797:
Can we solve this by propagating error messages returned from the server instead?

evgeni:
Yes and no. In this particular case the http code is 200 :(

evgeni:
See https://projects.theforeman.org/issues/26937

mdellweg:
Can we do a scoped search to find children? Also do we want to give an option to delete the children, too?

evgeni:
You can't search for parent id, no :(
Recursive delete might be possible, but would require more code and I'm lazy (and the UI doesn't support that either).

What we could do instead of the search is to issue the delete and inspect the API answer, as it does contain the error, just not with the right http code.

evgeni:
I've pushed two iterations moving the error detection to parsing the result instead of predicting the problem and surfacing the error message received from the API. Still a workaround, but feels nicer?

plugins/module_utils/foreman_helper.py
```diff
@@ -444,8 +444,14 @@ def _delete_entity(self, resource, current_entity, params):
         payload = {'id': current_entity['id']}
         if params:
             payload.update(params)
-        self.resource_action(resource, 'destroy', payload)
-        return True, None
+
+        changed, entity = self.resource_action(resource, 'destroy', payload)
+
+        # this is a workaround for https://projects.theforeman.org/issues/26937
+        if entity and 'error' in entity and 'message' in entity['error']:
```
mdellweg: Is this generic enough to have it in resource_action?

evgeni: this is at least true for location and organization, as both are a taxonomy, and the bug linked is affecting all taxonomies

mdellweg: Sorry, i meant whether it can happen on other actions than delete, and whether we should employ it for create, update, you name it.

evgeni: ah, I wasn't sure about that. and the json *this* API endpoint returns on errors is slightly different then all the other endpoints, so let's keep it in delete for now.

PR #438, opened 2019-08-26
make "make dist" generate a working, clean collection

evgeni:
this commit ensures that the built collection contains
1. only files we need (no .github etc)
2. no nailgun-based modules
3. imports that work inside a collection


mdellweg:
Nice one.
Is there a way to test-install the built collection?

evgeni:
@mdellweg `mazer installer` or `ansible-galaxy collection install` (from Ansible 2.9/git) should be able to install that, yes, so we could try installing it on travis. Let's see if I can cook something up.

PR #440, opened 2019-08-26
Port foreman_job_template to new foreman_helper

mdellweg:


plugins/modules/foreman_job_template.py
```diff
@@ -249,95 +249,88 @@
...
-        JobTemplate,
-        TemplateInput,
-        find_entities,
-        find_entities_by_name,
-        find_template_input,
-        ForemanEntityAnsibleModule,
-        naildown_entity,
-        naildown_entity_state,
-        sanitize_entity_dict,
-    )
```
evgeni: Where is `os` used? I can't find it in neither the old nor the new code

plugins/modules/foreman_job_template.py
```diff
@@ -249,95 +249,88 @@
...
 }
 
 
 def main():
     module = ForemanEntityAnsibleModule(
-        argument_spec=dict(
-            # Entity parameter
+        entity_spec=dict(
             audit_comment=dict(),
             description_format=dict(),
```
evgeni: Can we remove this effective user line?

plugins/modules/foreman_job_template.py
mdellweg: there is some os.path manipulation around line 374.

evgeni: Ooh, I've been fooled by the diff view, didn't see that there were collapsed lines. Or rather not used to see that in a nailgun to apypie PR ;)

plugins/modules/foreman_job_template.py
mdellweg: sure. A quick look at the api shows, that its implementation is at least not that easy.

plugins/modules/foreman_job_template.py
mdellweg: Maybe now it not the best time anymore to _get used_ to seeing nailgun-to-apypie-prs.

plugins/modules/foreman_job_template.py
```diff
@@ -388,72 +380,66 @@ def main():
     affects_multiple = name == '*'
     # sanitize user input, filter unuseful configuration combinations with 'name: *'
     if affects_multiple:
-        if state == 'present_with_defaults':
+        if module.state == 'present_with_defaults':
             module.fail_json(msg="'state: present_with_defaults' and 'name: *' cannot be used together")
-        if state == 'absent':
+        if module.state == 'absent':
```
evgeni:
```suggestion
        if module.desired_absent:
```

Even tho I am not sure the check is *really* useful ;)

PR #441, opened 2019-08-26
Remove alias 'verify_ssl' from cleaned parameters

mdellweg:
When using the alias, is is left in the module parameters, thus failing the
validation with affects_multiple.

### DIFF
--- a/plugins/module_utils/foreman_helper.py
+++ b/plugins/module_utils/foreman_helper.py
@@ -35,7 +35,7 @@ class ForemanBaseAnsibleModule(AnsibleModule):
         )
         args.update(argument_spec)
         supports_check_mode = kwargs.pop('supports_check_mode', True)
-        self._aliases = set(alias for arg in args.values() for alias in arg.get('aliases', []))
+        self._aliases = {alias for arg in args.values() for alias in arg.get('aliases', [])}
         super(ForemanBaseAnsibleModule, self).__init__(argument_spec=args, supports_check_mode=supports_check_mode, **kwargs)
 
         self._params = self.params.copy()
--- a/plugins/module_utils/foreman_helper.py
+++ b/plugins/module_utils/foreman_helper.py
@@ -35,6 +35,7 @@ class ForemanBaseAnsibleModule(AnsibleModule):
         )
         args.update(argument_spec)
         supports_check_mode = kwargs.pop('supports_check_mode', True)
+        self._aliases = set(alias for arg in args.values() for alias in arg.get('aliases', []))
         super(ForemanBaseAnsibleModule, self).__init__(argument_spec=args, supports_check_mode=supports_check_mode, **kwargs)
 
         self._params = self.params.copy()
@@ -45,6 +46,8 @@ class ForemanBaseAnsibleModule(AnsibleModule):
         self._foremanapi_username = self._params.pop('username')
         self._foremanapi_password = self._params.pop('password')
         self._foremanapi_validate_certs = self._params.pop('validate_certs')
+        if 'verify_ssl' in self._params:
+            self.warn("Please use 'validate_certs' instead of deprecated 'verify_ssl'.")
 
         self.task_timeout = 60
         self.task_poll = 4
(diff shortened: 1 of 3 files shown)


plugins/module_utils/foreman_helper.py
```diff
@@ -54,7 +54,7 @@ def parse_params(self):
         return {k: v for (k, v) in self._params.items() if v is not None}
 
     def clean_params(self):
-        return {k: v for (k, v) in self._params.items() if v is not None}
+        return {k: v for (k, v) in self._params.items() if v is not None and k is not 'verify_ssl'}
```
ekohl: Can we clean any alias and only keep the "original" name?

mdellweg: I don't know how easy that would be, and i'd expect ansible to do this already.

mdellweg:
@ekohl i gave it a shot. Also i added a warning if this specific alias was used.

plugins/module_utils/foreman_helper.py
```diff
@@ -35,6 +35,7 @@ def __init__(self, argument_spec, **kwargs):
         )
         args.update(argument_spec)
         supports_check_mode = kwargs.pop('supports_check_mode', True)
+        self._aliases = set(alias for arg in args.values() for alias in arg.get('aliases', []))
```
ekohl: You can use set comprehension

```suggestion
        self._aliases = {alias for arg in args.values() for alias in arg.get('aliases', [])}
```

plugins/modules/foreman_installation_medium.py
```diff
@@ -116,7 +116,8 @@ def main():
             module.fail_json(msg="'state: present_with_defaults' and 'name: *' cannot be used together")
         if module.desired_absent:
             if list(entity_dict.keys()) != ['name']:
-                module.fail_json(msg='When deleting all installation media, there is no need to specify further parameters %s ' % entity_dict.keys())
+                entity_dict.pop('name', None)
+                module.fail_json(msg='When deleting all installation media, there is no need to specify further parameters: %s ' % entity_dict.keys())
```
ekohl: Removing an entity doesn't feel very clean. Would the following work and provide useful feedback?
```suggestion
                module.fail_json(msg='When deleting all installation media, only name may be passed as a parameter', passed=entity_dict.keys())
```

mdellweg: Do you have documentation for this. I cannot find any about how `passed` is used here.

ekohl: It is entirely untested but a passing module essentially returns a dict where `msg` is a bit of a special key. When using the YAML output, you can see all keys.

ekohl: It's not that common in the Ansible codebase, but it does show up:
https://github.com/ansible/ansible/blob/c2469648e4235deb10e3c314bb668e412f2d6ffb/lib/ansible/modules/cloud/vmware/vsphere_file.py#L203-L206

mdellweg: I liked the concept of only showing the offending keys. And since we exit in the next line anyway, it seemed appropriate to do it the easy way. This list is already lacking a lot of the actually passed parameters (username, password, server_url, ...).

ekohl: I don't feel very strongly about this. Exactly for the reasons you mention I debated it. If this becomes a pattern then we can add a helper:
```python
def only_allow_params(params, allowed):
    params = set(params)
    allowed = set(allowed)
    if params > allowed:
        message = 'Only %s can be passed; %s passed' % (', '.join(allowed), ', '.join(params))
        raise ValueError(message)
```

However, let's merge this for now.

mdellweg: I have seen two other places, that just did not report the superfluous parameters.
At some point, we should look how much of this whole affects_multiple can be dryed up.

mdellweg:
Thanks for the review, @ekohl

PR #443, opened 2019-08-27
Fix foreman_os_default_template

mdellweg:
Is seems, os_default_template does not have a scoped search on
template_kind_id. So we search for the right template by hand.

### DIFF
--- a/plugins/modules/foreman_os_default_template.py
+++ b/plugins/modules/foreman_os_default_template.py
@@ -107,6 +107,8 @@ def main():
     entity_dict['template_kind'] = module.find_resource_by_name('template_kinds', entity_dict['template_kind'], thin=True)
 
     scope = {'operatingsystem_id': entity_dict['operatingsystem']['id']}
+    # Default templates do not support a scoped search
+    # see: https://projects.theforeman.org/issues/27722
     entities = module.list_resource('os_default_templates', params=scope)
     entity = next((item for item in entities if item['template_kind_id'] == entity_dict['template_kind']['id']), None)
 
--- a/plugins/modules/foreman_os_default_template.py
+++ b/plugins/modules/foreman_os_default_template.py
@@ -107,8 +107,8 @@ def main():
     entity_dict['template_kind'] = module.find_resource_by_name('template_kinds', entity_dict['template_kind'], thin=True)
 
     scope = {'operatingsystem_id': entity_dict['operatingsystem']['id']}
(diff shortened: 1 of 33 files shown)


mdellweg:
I needed to change some `operating_system` to `operatingsystem` in order to rerecord. So i thought, i should go all the way there.

evgeni:
Why wasn't that caught by the tests?

mdellweg:
I guess, there was only one os_default_template, and therefore the find_resource succeeded.

evgeni:
If we pass a search scoped_search doesn't understand, it will (should?) raise a 500 unprocessible error, so that sounds unlikely?

evgeni:
That said, there is no scoped search defined in https://github.com/theforeman/foreman/blob/develop/app/models/os_default_template.rb, so the bug is valid, I'd just would love to be able to verify it ;)

mdellweg:
Well rails is fast at silently ignoring information it cannot digest...
Trying to improve our test playbook.

tests/test_playbooks/os_default_template.yml
```diff
@@ -10,12 +10,17 @@
...
       location_state: present
-  - include: tasks/operating_system.yml
+  - include: tasks/operatingsystem.yml
     vars:
-      operating_system_state: present
+      operatingsystem_state: present
   - include: tasks/provisioning_template.yml
     vars:
       provisioning_template_state: present
+  - include: tasks/provisioning_template.yml
```
evgeni: you use two separate tasks here, vs loops later:

```
    loop:
      - provision
      - finish
```

and

```
    loop:
      - Timetravel provision
      - Timetravel finish
```

IMHO we should have one way to do it ;)

plugins/modules/foreman_os_default_template.py
```diff
@@ -107,8 +107,8 @@ def main():
     entity_dict['template_kind'] = module.find_resource_by_name('template_kinds', entity_dict['template_kind'], thin=True)
 
     scope = {'operatingsystem_id': entity_dict['operatingsystem']['id']}
-    search = 'template_kind_id={}'.format(entity_dict['template_kind']['id'])
-    entity = module.find_resource('os_default_templates', search, params=scope, failsafe=True)
+    entities = module.list_resource('os_default_templates', params=scope)
```
evgeni: can you add a comment, linking to https://projects.theforeman.org/issues/27722 for the explanation

mdellweg:
Adressed both comments. I am going to merge, once travis' head turns green.

PR #445, opened 2019-08-28
Port katello_content_view_version to foreman_helper

mdellweg:


plugins/modules/katello_content_view_version.py
```diff
@@ -22,71 +22,73 @@
...
-            - Force metadata regeneration when performing Publish and Promote tasks
-    synchronous:
-        description:
-            - Wait for the Publish or Promote task to complete if True. Immediately return if False.
-        default: true
-    current_lifecycle_environment:
-        description:
-            - The lifecycle environment that is already associated with the content view version
-            - Helpful for promoting a content view version
+  server_url:
```
evgeni: these all can go if you use the doc fragment (this module was not using the nailgun cement, so not the fragment already)

plugins/modules/katello_content_view_version.py
```diff
@@ -143,58 +145,34 @@
...
+def promote_content_view_version(module, content_view_version, environments, synchronous, force, force_yum_metadata_regeneration):
     changed = False
 
-    current_environment_ids = map(lambda environment: environment.id, content_view_version.environment)
-    desired_environment_ids = map(lambda environment: environment.id, environments)
-    promote_to_environment_ids = list(set(desired_environment_ids) - set(current_environment_ids))
-
-    request_data = {'environment_ids': promote_to_environment_ids}
-    request_data.update({k: v for k, v in kwargs.items() if v is not None})
+    current_environment_ids = {environment['id'] for environment in content_view_version['environments']}
```
evgeni: TIL: `{a, b, c}` is a `set`, totally not confusing with dicts…

plugins/modules/katello_content_view_version.py
```diff
@@ -203,66 +181,58 @@ def main():
...
-    elif 'version' in params_dict:
-        content_view_version = find_content_view_version(module, content_view, version=params_dict['version'], failsafe=True)
+    if 'current_lifecycle_environment' in entity_dict:
+        entity_dict['current_lifecycle_environment'] = module.find_resource_by_name(
+            'lifecycle_environments', name=entity_dict['current_lifecycle_environment'], params=scope)
+        search = "content_view_id={},environment_ids=[{}]".format(content_view['id'], entity_dict['current_lifecycle_environment']['id'])
+        content_view_version = module.find_resource('content_view_versions', search=search, thin=True)
+    elif 'version' in entity_dict:
+        search = "content_view_id={},version={}".format(content_view['id'], entity_dict['version'])
+        content_view_version = module.find_resource('content_view_versions', search=search, failsafe=True)
```
evgeni: why failsafe here vs thin in line 201?

plugins/modules/katello_content_view_version.py
mdellweg: Right. That's what i was not used to see. 8)

plugins/modules/katello_content_view_version.py
mdellweg: I guess it could be thin here, too, but the `failsafe` are ok. If we specify lce, and there is no voersion, we want to error out. If we specify a version, and there is none, we ought to create one.
Also we never call update on an existing version. So always thin should be ok.

evgeni: Sounds good. I was basically thinking: "two similar lines, could they be collapsed to one" but your explanation is sound that they can't :)

mdellweg: Changed it to thin.

plugins/modules/katello_content_view_version.py
```diff
@@ -22,71 +22,55 @@
...
+    type: bool
+    aliases:
+      - force
+  force_yum_metadata_regeneration:
+    description:
+      - Force metadata regeneration when performing Publish and Promote tasks
+  synchronous:
+    description:
+      - Wait for the Publish or Promote task to complete if True. Immediately return if False.
+    default: true
```
evgeni: synchronous should be type:bool

plugins/modules/katello_content_view_version.py
```diff
@@ -143,58 +127,34 @@
...
-    )
-    has_import_error = False
-except ImportError as e:
-    has_import_error = True
-    import_error_msg = str(e)
-from ansible.module_utils.basic import AnsibleModule
+from ansible.module_utils.foreman_helper import KatelloEntityAnsibleModule
 
 
-def promote_content_view_version(module, content_view_version, organization, environments, synchronous, **kwargs):
```
evgeni: Should force be force_promote?

plugins/modules/katello_content_view_version.py
```diff
@@ -143,58 +127,34 @@
...
-            password=dict(required=True, no_log=True),
-            verify_ssl=dict(type='bool', default=True),
-            content_view=dict(required=True),
-            organization=dict(required=True),
-            state=dict(default='present', choices=['present', 'absent']),
+    module = KatelloEntityAnsibleModule(
+        entity_spec=dict(
+            content_view=dict(type='entity', flat_name='content_view_id', required=True),
             version=dict(),
             lifecycle_environments=dict(type='list', default=['Library']),
```
evgeni: This contradicts the doc, which lists force_promote as the primary name

plugins/modules/katello_content_view_version.py
```diff
@@ -143,58 +127,34 @@
...
     if promote_to_environment_ids:
-        if not module.check_mode:
-            content_view_version.promote(synchronous, data=request_data)
-        changed = True
+        payload = {
+            'environment_ids': promote_to_environment_ids,
+            'force': force,
+            'force_yum_metadata_regeneration': force_yum_metadata_regeneration,
+        }
+
```
evgeni: Seems promote_content_view_version doesn't pass synchronous to the resource action?

sean797: Bonus points if you can add a test like 

https://github.com/theforeman/foreman-ansible-modules/blob/3a85d43dffb4f93daf791d05069862d1c8b743e0/tests/test_playbooks/tasks/katello_sync.yml#L21-L28

plugins/modules/katello_content_view_version.py
mdellweg: The api wants force, but since we control that part, we can let ansible behave the way we like it.

mdellweg: I think in the context of this function 'force' is clear.

plugins/modules/katello_content_view_version.py
mdellweg: I will change it to adhere the docs, because in the context of the whole module force is not that clear.

plugins/modules/katello_content_view_version.py
mdellweg: There is a problem here. I think we can only trigger the promote task, once the publish task is done. So we cannot have them both synchronous. What should we do about it?
Can we fold the promotion into the publish request? Can we add a wait for _this_ task to the promote request?

evgeni: I think we can have them both sync, just not both unsync?

We issue the publish here:
https://github.com/theforeman/foreman-ansible-modules/blob/8d57b2f211201d9b0cc46d7245378a3dfab3554c/plugins/modules/katello_content_view_version.py#L209

If sync is `True`, this step will wait until the publish is done and we can do a sync promote too.

If sync=False for the publish, then already this line won't work (I think, didn't verify):
https://github.com/theforeman/foreman-ansible-modules/blob/8d57b2f211201d9b0cc46d7245378a3dfab3554c/plugins/modules/katello_content_view_version.py#L210

And we can't promote until the publish is done.

IMHO we should do the following:
* no current_lifecycle_environment, no version and sync=False → error out during param validation
* all other cases look happy?

plugins/modules/katello_content_view_version.py
```diff
@@ -151,126 +136,98 @@
...
+                payload['force_yum_metadata_regeneration'] = entity_dict['force_yum_metadata_regeneration']
+            if 'version' in entity_dict:
+                split_version = list(map(int, str(entity_dict['version']).split('.')))
+                payload['major'] = split_version[0]
+                payload['minor'] = split_version[1]
+
+            changed, response = module.resource_action('content_views', 'publish', params=payload, synchronous=entity_dict['synchronous'])
+            content_view_version = module.show_resource('content_view_versions', response['output']['content_view_version_id'])
+
+        if 'lifecycle_environments' in entity_dict:
```
evgeni:
```suggestion
        if entity_dict['lifecycle_environments'] != ['Library']:
```

plugins/modules/katello_content_view_version.py
```diff
@@ -151,126 +136,98 @@
...
-
-        if 'lifecycle_environments' in params_dict:
-            lifecycle_environments = find_lifecycle_environments(module, names=params_dict['lifecycle_environments'], organization=organization)
-            le_changed = promote_content_view_version(module, content_view_version, organization, lifecycle_environments, params_dict['synchronous'],
-                                                      force=params_dict['force'],
-                                                      force_yum_metadata_regeneration=params_dict['force_yum_metadata_regeneration'])
-    elif state == 'absent':
-        changed = naildown_entity_state(ContentViewVersion, dict(), content_view_version, state, module)
+            # Do a sanity check, whether we can perform this non-synchronous
+            if 'lifecycle_environments' in entity_dict and not entity_dict['synchronous']:
```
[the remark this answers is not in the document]
mdellweg: I know, there is the possibility to publish an older version to library, but I think a new one would end up in library anyway. So how about not defaulting to library?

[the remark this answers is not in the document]
evgeni: So a *publish* always goes to Library. You can then *promote* to any existing LCE (incl Library if the version in Library is different than the one you're promoting). And with that said, yes, dropping the default and then using your original code sounds sane to me.

@akofink @sean797 can you back us up here?

plugins/modules/katello_content_view_version.py
```diff
@@ -151,126 +136,98 @@
...
+                module.fail_json("Cannot perform non-blocking publishing and promoting in the same module call.")
```
evgeni:
```suggestion
                module.fail_json(msg="Cannot perform non-blocking publishing and promoting in the same module call.")
```

plugins/modules/katello_content_view_version.py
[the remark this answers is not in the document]
sean797: Agreed :+1:

plugins/modules/katello_content_view_version.py
```diff
@@ -151,126 +136,98 @@
...
-
-        if 'lifecycle_environments' in params_dict:
-            lifecycle_environments = find_lifecycle_environments(module, names=params_dict['lifecycle_environments'], organization=organization)
-            le_changed = promote_content_view_version(module, content_view_version, organization, lifecycle_environments, params_dict['synchronous'],
-                                                      force=params_dict['force'],
-                                                      force_yum_metadata_regeneration=params_dict['force_yum_metadata_regeneration'])
-    elif state == 'absent':
-        changed = naildown_entity_state(ContentViewVersion, dict(), content_view_version, state, module)
+            # Do a sanity check, whether we can perform this non-synchronous
+            if 'lifecycle_environments' in entity_dict and not entity_dict['synchronous']:
```
evgeni: we end up here when `content_view_version is None` which can be in the following cases:
1. the user did not pass neither `version` nor `current_lifecycle_environment`
2. the user *did* pass a `version`, but it was a not-yet-existing one

obviously, for 2 we need to talk to the API to find out that the version didn't exist yet and we need to publish it first.
but for 1 we can know that we need to do a publish before doing any API interaction.

I wonder if we should add a similar test directly after `module.connect()` and thus fail as early as possible?

```python
if 'lifecycle_environments' in entity_dict and 'version' not in entity_dict and 'current_lifecycle_environment' not in entity_dict and not entity_dict['synchronous']:
    module.fail_json(msg="Cannot perform non-blocking publishing and promoting in the same module call.")
```

mdellweg: I think, i'll put this even before `module.connect()`.

mdellweg:
If i do not find the time to add the tests today, feel free to merge this one.

evgeni:
opened #464 for the tests and merging

PR #446, opened 2019-08-28
Add a filter to produce candle_pin labels

mdellweg:


sean797:
Whats the use case for this?

mdellweg:
When you create products and repositories with certain names, you can use this to construct the corresponding urls.

sean797:
Are we better getting it from the API since the user can customise it? 

Screenshot from 2019-08-29 11-09-22

mdellweg:
I think both ways are valid.

mdellweg:
I must admit, at least partially i wanted to write _a_ filter. But also, we use this one.

PR #449, opened 2019-09-02
don't patch subscription upload api

evgeni:
apypie doesn't fail on this (anymore)


sean797:
Don't we have to wait for https://github.com/Apipie/apipie-rails/pull/669 ?

evgeni:
No, because we just don't verify file uploads

PR #455, opened 2019-09-06
enable validate-modules sanity

evgeni:


evgeni:
Note to self, that test has a --collection param that should avoid the version added error

evgeni:
I started with `Found 510 validate-modules issue(s) which need to be resolved`. This is now down to 246 \o/

plugins/modules/katello_activation_key.py
```diff
@@ -159,7 +179,7 @@ def main():
             )),
             content_overrides=dict(type='list', elements='dict', options=dict(
                 label=dict(required=True),
-                override=dict(choises=['enabled', 'disabled']),
```
mdellweg: How can this ever have worked...
(I know it just skipped the choices validation)

evgeni: The joys of `**kwargs`

PR #458, opened 2019-09-09
redhat_manifest: modern exceptions

evgeni:


evgeni:
@ekohl this got changes after you ACK'ed, so if you want to re-review :)

Issue #452, opened 2019-09-06
Use installed ansible-test


evgeni:
It got promoted to a real tool in https://github.com/ansible/ansible/pull/60718

So we don't need to do so many hacks anymore, probably

PR #461, opened 2019-09-10
use installed ansible-test

evgeni:
fixes: #452


evgeni:
this has the side effect that the nailgun based modules aren't tested anymore for sanity, but I think that's OK at the current stage of migration

PR #463, opened 2019-09-11
enable pylint sanity test

evgeni:


plugins/module_utils/foreman_helper.py
```diff
@@ -349,7 +349,7 @@ def ensure_entity(self, resource, desired_entity, current_entity, params=None, s
         if entity_spec is None:
             entity_spec = self.entity_spec
         else:
-            entity_spec, _ = _entity_spec_helper(entity_spec)
+            entity_spec = _entity_spec_helper(entity_spec)[0]
```
evgeni: I am not strictly happy about this one, but using `_argument_spec_ignored` seemed also weird?

plugins/module_utils/foreman_helper.py
```diff
@@ -588,7 +588,7 @@ def _entity_spec_helper(spec):
         elif argument_value.get('type') == 'nested_list':
             argument_value['type'] = 'list'
             argument_value['elements'] = 'dict'
-            _, argument_value['options'] = _entity_spec_helper(argument_value.pop('entity_spec'))
+            argument_value['options'] = _entity_spec_helper(argument_value.pop('entity_spec'))[1]
```
evgeni: I am not strictly happy about this one, but using `_entyty_spec_ignored` seemed also weird?

mdellweg: You could name it dummy instead of _

evgeni: also meh, but did so now as it feels better

PR #467, opened 2019-09-13
enable import sanity test

evgeni:


plugins/module_utils/foreman_helper.py
```diff
@@ -233,6 +238,9 @@ def check_requirements(self):
         if not HAS_APYPIE:
             self.fail_json(msg='The apypie Python module is required', exception=APYPIE_IMP_ERR)
 
+        if not HAS_PYYAML:
+            self.fail_json(msg='The PyYAML Python module is required', exception=PYYAML_IMP_ERR)
```
sean797: Is it worth DRY'ing this? Something like? Perhaps even in module_utils

```
if FAILED_IMPORT:
    self.fail_json(msg='The {} Python module is required'.format(FAILED_IMPORT), exception=IMP_ERR)

evgeni: Dunno. There is #328 for when we decide to depend on Ansible 2.8, which is kinda DRYing it?

evgeni: One thing to think about: We *really* require `PyYAML` only in modules that call `parse_template(_from_file)`, so `foreman_job_template`, `foreman_provisioning_template` and `foreman_ptable`. Maybe this `fail` should be moved there?

sean797: Maybe, we just need to be sure to do the same if/when we create modules for more template types

evgeni: All (current) modules call `parse_template` before doing any *real* API interaction, so we could just do the `fail` inside `parse_template`?

sean797: I think thats a nice idea:1st_place_medal:

evgeni: I can't read. pushed.

PR #470, opened 2019-09-13
add PyYAML to requirements

evgeni:
this is only needed when the module runs on a different node than
ansible, which should be seldom, but possible


ekohl:
There's a merge conflict

evgeni:
rebased

PR #469, opened 2019-09-13
move requirements to doc_fragment

evgeni:

PR #471, opened 2019-09-13
content_view_filter: migrate to apypie

evgeni:
fixes: #274 
:fireworks:


evgeni:
this fails tests on Python 3.5 (and 3.5 only) so probably some sorting weirdness, yay

plugins/modules/katello_content_view_filter.py
```diff
@@ -252,51 +234,63 @@ def main():
...
-            entity_dict['uuid'] = find_package_group(module, name=entity_dict['rule_name']).uuid
+            package_group = module.find_resource_by_name('package_groups', entity_dict['rule_name'], params=scope)
+            entity_dict['uuid'] = package_group['uuid']
+
+        # drop 'name' from the dict, as otherwise it might override 'rule_name'
+        rule_dict = entity_dict.copy()
+        rule_dict.pop('name', None)
 
-        content_view_filter_rule_dict = sanitize_entity_dict(entity_dict, rule_map)
         check_missing = ['min_version', 'max_version', 'version', 'start_date', 'end_date', 'architecture', 'date_type']
```
evgeni: this is not used anymore and I don't completely understand @sean797's old comment in the cement code:

```
# check_missing is a special case, Foreman sometimes returns different values
# depending on what 'type' of same object you are requesting. Content View
# Filters are a prime example. We list these attributes in `check_missing`
# so we can ensure the entity is as the user specified.
```

mdellweg: I also am not sure, but i know, that we switched the searching for properties to compare from the (server responded) object to the (user provided) dictionary of desired values. This made the update method look so much simpler. Maybe it fixed this one also. Can we rely on the existing tests here?

evgeni: the test do fail on `master` when I remove the `check_missing` code:

```diff
diff --git a/plugins/module_utils/ansible_nailgun_cement.py b/plugins/module_utils/ansible_nailgun_cement.py
index 162ff3b..cb8ed46 100644
--- a/plugins/module_utils/ansible_nailgun_cement.py
+++ b/plugins/module_utils/ansible_nailgun_cement.py
@@ -292,21 +292,9 @@ def update_entity(old_entity, entity_dict, module, check_missing, check_type, fo
                 if not fields_equal(value, entity_dict[key]):
                     volatile_entity.__setattr__(key, entity_dict[key])
                     fields.append(key)
-            # check_missing is a special case, Foreman sometimes returns different values
-            # depending on what 'type' of same object you are requesting. Content View
-            # Filters are a prime example. We list these attributes in `check_missing`
-            # so we can ensure the entity is as the user specified.
-            if check_missing is not None and key not in entity_dict and key in check_missing:
-                volatile_entity.__setattr__(key, None)
-                fields.append(key)
         if force_update:
             for key in force_update:
                 fields.append(key)
-        if check_missing is not None:
-            for key in check_missing:
-                if key in entity_dict and key not in volatile_entity.get_values():
-                    volatile_entity.__setattr__(key, entity_dict[key])
-                    fields.append(key)
         if len(fields) > 0:
             if not module.check_mode:
                 result = volatile_entity.update(fields)
```

So "maybe"? :)

mdellweg: OK for me, but i'd say it _is_ @sean797 s call.

sean797: >  we switched the searching for properties to compare from the (server responded) object to the (user provided) dictionary of desired values. This made the update method look so much simpler. Maybe it fixed this one also.

I think that probably removed any need for `check_missing` :+1:

plugins/modules/katello_content_view_filter.py
```diff
@@ -33,8 +33,17 @@
     - Create and Manage Katello content View filters
 author: "Sean O'Keeffe (@sean797)"
 requirements:
-    - "nailgun >= 0.28.0"
+    - apypie
```
sean797: I think you can remove this since https://github.com/theforeman/foreman-ansible-modules/pull/469

requirements.txt
```diff
@@ -1,4 +1,3 @@
-nailgun
```
sean797: :heart_eyes:

plugins/modules/katello_content_view_filter.py
sean797: Yes - we can rely on tests for this, previously without it most filter types would always report back as `changed`.
So this can be removed.

plugins/modules/katello_content_view_filter.py
evgeni: yes!

PR #472, opened 2019-09-16
sync nailgun branch readme with master

evgeni:


README.md
```diff
@@ -1,15 +1,43 @@
...
+A list of all modules and their documentation can be found at [theforeman.org/plugins/foreman-ansible-modules](https://theforeman.org/plugins/foreman-ansible-modules/).
+
+## Branches
+
+* `master` - current development branch, using the `apypie` library.
+* `nailgun` - the state of the repository before the switch to the `apypie` library started, `nailgun` is the only dependency
+
+## Supported Foreman and plugins versions
+
+### `nailgun` based modules
```
mdellweg: There is nothing else than "nailgun based", right? I guess we don't need that header.

evgeni: yes

evgeni:
this fails tests, but I am not going to fix them here.

README.md
```diff
@@ -1,15 +1,41 @@
...
 
 ## Goals
 
 The intent of this repository is to be a place that community members can develop or contribute modules. The goals of this repository are:
 
   * centralized location for community modules
   * a single repository to clone for interacting with Foreman & plugins
-  * an intermediate landing place for modules before pushing them to Ansible community
-  * repository maintainers will be working to push the modules into Ansible proper [https://github.com/ansible/ansible/tree/devel/lib/ansible/modules/remote_management/foreman](https://github.com/ansible/ansible/tree/devel/lib/ansible/modules/remote_management/foreman)
+  * source for the official Ansible collection (TBD)
```
mdellweg: Since almost nothing on this branch is going to be part of that Ansible collection, we could reference to the master branch here.

mdellweg:
> this fails tests, but I am not going to fix them here.

This sounds fair, given that it is a docs only change.

README.md
evgeni: `s/almost//` ;)

Issue #474, opened 2019-09-16
katello_content_view_version failed with two lifecycle environments


sthirugn:
##### SUMMARY
katello_content_view_version failed with two lifecycle environments

##### ISSUE TYPE
 - Bug Report

##### ANSIBLE VERSION
```
# ansible --version
ansible 2.8.2
  config file = /etc/ansible/ansible.cfg
  configured module search path = ['/root/.ansible/plugins/modules', '/usr/share/ansible/plugins/modules']
  ansible python module location = /usr/lib/python3.7/site-packages/ansible
  executable location = /usr/bin/ansible
  python version = 3.7.2 (default, Mar 21 2019, 10:09:12) [GCC 8.3.1 20190223 (Red Hat 8.3.1-2)]

# cat /etc/redhat-release 
Fedora release 29 (Twenty Nine)

# rpm -qa | grep theforeman
ansible-collection-theforeman-foreman-0.0.0-0.1.pre.master.20190913083800gitc924973.fc29.noarch
```

##### KATELLO/FOREMAN VERSION
```
# rpm -q tfm-rubygem-katello foreman
tfm-rubygem-katello-3.10.0.54-1.el7sat.noarch
foreman-1.20.1.35-1.el7sat.noarch

```

##### NAILGUN VERSION
```
N/A
```

##### STEPS TO REPRODUCE
1. Create org called test_org
2. Create a content view called testcv
3. Run the publish/promote task as shown below

```yaml
- hosts: localhost
  tasks:
   - name: "Publish a content view to Library, and promote that version to prod, not idempotent"
     theforeman.foreman.katello_content_view_version:
       username: "admin"
       password: "changeme"
       server_url: "https://satellite.example.com"
       content_view: "testcv"
       organization: "test_org"
       lifecycle_environments:
         - Library  
         - prod
```

##### EXPECTED RESULTS
- Publish/Promote to Library/Prod works

##### ACTUAL RESULTS
- Promotion to prod environment fails
- publish to Library worked

```
# ansible-playbook foreman_playbook.yml 

PLAY [localhost] ****************************************************

TASK [Gathering Facts] **********************************************************************
ok: [satellite.example.com]

TASK [Publish a content view to Library, and promote that version to prod, not idempotent] ***
fatal: [satellite.example.com]: FAILED! => {"changed": false, "msg": "Error while performing promote on content_view_versions: The following required parameters are missing: id"}

PLAY RECAP **********************************************************************************
[...]

PR #476, opened 2019-09-17
content_view_version tests and fixes

evgeni:
the tests in the content_view playbook were not sufficient, so I've split them out, added more testcases and fixed bugs along the way:
* pass the content_view_version id when promoting to a new env
* correct the resource when deleting a CV version
* don't fetch version as thin, we need the environments data

Fixes: #474

### DIFF
--- a/plugins/modules/katello_content_view_version.py
+++ b/plugins/modules/katello_content_view_version.py
@@ -202,11 +202,11 @@ def main():
     if 'current_lifecycle_environment' in entity_dict:
         entity_dict['current_lifecycle_environment'] = module.find_resource_by_name(
             'lifecycle_environments', name=entity_dict['current_lifecycle_environment'], params=scope)
-        search = "content_view_id={0},environment_ids=[{1}]".format(content_view['id'], entity_dict['current_lifecycle_environment']['id'])
-        content_view_version = module.find_resource('content_view_versions', search=search, thin=True)
+        search_scope = {'content_view_id': content_view['id'], 'environment_id': entity_dict['current_lifecycle_environment']['id']}
+        content_view_version = module.find_resource('content_view_versions', search=None, params=search_scope)
     elif 'version' in entity_dict:
         search = "content_view_id={0},version={1}".format(content_view['id'], entity_dict['version'])
-        content_view_version = module.find_resource('content_view_versions', search=search, thin=True, failsafe=True)
+        content_view_version = module.find_resource('content_view_versions', search=search, failsafe=True)
     else:
         content_view_version = None
 
--- a/plugins/modules/katello_content_view_version.py
+++ b/plugins/modules/katello_content_view_version.py
@@ -156,6 +156,7 @@ def promote_content_view_version(module, content_view_version, environments, syn
 
     if promote_to_environment_ids:
         payload = {
+            'id': content_view_version['id'],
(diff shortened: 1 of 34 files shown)


evgeni:
@mdellweg please have a look ;)

@sthirugn can you please test if that fixes your issue? (it should :))

plugins/modules/katello_content_view_version.py
```diff
@@ -201,18 +202,18 @@ def main():
...
-        content_view_version = module.find_resource('content_view_versions', search=search, thin=True, failsafe=True)
+        content_view_version = module.find_resource('content_view_versions', search=search, failsafe=True)
     else:
         content_view_version = None
 
     changed = False
     le_changed = False
     if module.desired_absent:
-        changed = module.ensure_entity_state('content_view_version', None, content_view_version, params=scope)
+        changed = module.ensure_entity_state('content_view_versions', None, content_view_version, params=scope)
```
mdellweg: :+1:

evgeni:
I went ahead and removed cv_version recording during content_view tests, this is imho cleaner.

plugins/modules/katello_content_view_version.py
```diff
@@ -201,18 +202,18 @@ def main():
     if 'current_lifecycle_environment' in entity_dict:
         entity_dict['current_lifecycle_environment'] = module.find_resource_by_name(
             'lifecycle_environments', name=entity_dict['current_lifecycle_environment'], params=scope)
-        search = "content_view_id={0},environment_ids=[{1}]".format(content_view['id'], entity_dict['current_lifecycle_environment']['id'])
-        content_view_version = module.find_resource('content_view_versions', search=search, thin=True)
+        search_scope = {'content_view_id': content_view['id'], 'environment_id': entity_dict['current_lifecycle_environment']['id']}
+        content_view_version = module.find_resource('content_view_versions', search=None, params=search_scope)
```
mdellweg: Is there a reason for this change other than the `thin`thing?

plugins/modules/katello_content_view_version.py
```diff
@@ -156,6 +156,7 @@ def promote_content_view_version(module, content_view_version, environments, syn
 
     if promote_to_environment_ids:
         payload = {
+            'id': content_view_version['id'],
```
mdellweg: :+1:This i guess is 'the fix' for the referred to problem?

plugins/modules/katello_content_view_version.py
evgeni: Yes,  you can't search for `environment_ids=[{1}]`, you need to pass `environment_id` as a param.

plugins/modules/katello_content_view_version.py
evgeni: Yes.

evgeni:
> Just curious: Do you think, the old version had the same problem?

Old = nailgun? No, I don't think it had that problem. Actually, I am sure it hadn't, as I used it exactly how it was failing now.

evgeni:
ok, not recording it doesn't work as expected, force-pushed the old version.

sthirugn:
> @sthirugn can you please test if that fixes your issue? (it should :))

@evgeni, I can confirm that issue #474 is now resolved and working great in latest master.  Thank you for the quick fix.

PR #480, opened 2019-09-18
use flake8 instead of pycodestyle for linting

evgeni:
flake uses pycodestyle internally, so we getting a superset of checks
and don't ignore E722 anymore, we fixed that

### DIFF
--- a/plugins/modules/foreman_host_power.py
+++ b/plugins/modules/foreman_host_power.py
@@ -114,8 +114,6 @@ def main():
 
     module.connect()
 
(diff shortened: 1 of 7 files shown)


plugins/modules/foreman_host_power.py
```diff
@@ -114,8 +114,6 @@ def main():
 
     module.connect()
 
-    entity = module.find_resource_by_name('hosts', name=entity_dict['name'], failsafe=False, thin=True)
```
mdellweg: lol

PR #485, opened 2019-09-19
add description and homepage to the galaxy metadata

evgeni:
also make license a list as documented in [1]

[1] https://docs.ansible.com/ansible/devel/dev_guide/collections_galaxy_meta.html#collections-galaxy-meta

### DIFF
--- a/galaxy.yml
+++ b/galaxy.yml
@@ -1,5 +1,6 @@
 namespace: "theforeman"
 name: "foreman"
+description: Ansible Modules to manage Foreman and Katello installations
 authors:
   - "Andrew Kofink <ajkofink@gmail.com>"
   - "Anthony Green <green@moxielogic.com>"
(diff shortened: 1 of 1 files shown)


galaxy.yml
```diff
@@ -28,13 +29,15 @@ authors:
...
-license: "GPL-3.0-or-later"
+license:
+  - "GPL-3.0-or-later"
 tags:
   - foreman
   - katello
   - satellite
   - orcharhino
 readme: "README.md"
+homepage: "https://theforeman.org/"
```
mdellweg: This looks a bit generic, but there might not be a better (more specific) landing page atm.

evgeni: Yeah, agreed. The docs link already goes to the "plugin" subpage, so that felt redundant too.

Issue #484, opened 2019-09-18
Get rid of the base class


evgeni:
We don't have multiple backends anymore, so the Basemodule can be merged with foreman module

PR #486, opened 2019-09-19
drop ForemanBaseAnsibleModule

evgeni:
fixes: #484

### DIFF
--- a/plugins/module_utils/foreman_helper.py
+++ b/plugins/module_utils/foreman_helper.py
@@ -33,45 +33,6 @@ parameter_entity_spec = dict(
 )
 
 
-class ForemanBaseAnsibleModule(AnsibleModule):
-
-    def __init__(self, argument_spec, **kwargs):
-        args = dict(
-            server_url=dict(required=True),
-            username=dict(required=True),
-            password=dict(required=True, no_log=True),
-            validate_certs=dict(type='bool', default=True, aliases=['verify_ssl']),
-        )
-        args.update(argument_spec)
-        supports_check_mode = kwargs.pop('supports_check_mode', True)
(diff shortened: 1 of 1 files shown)


plugins/module_utils/foreman_helper.py
```diff
@@ -166,14 +127,48 @@ def inner(self, *args, **kwargs):
...
 
-    def __init__(self, *args, **kwargs):
-        super(ForemanAnsibleModule, self).__init__(*args, **kwargs)
         self._thin_default = False
         self.state = 'undefined'
         self.entity_spec = {}
 
+    def parse_params(self):
+        self.warn("Use of deprecated method parse_params")
+        return {k: v for (k, v) in self._params.items() if v is not None}
```
mdellweg: Anyone still using this?

plugins/module_utils/foreman_helper.py
```diff
@@ -166,14 +127,48 @@ def inner(self, *args, **kwargs):
...
+
+    def clean_params(self):
+        return {k: v for (k, v) in self._params.items() if v is not None and k not in self._aliases}
+
+    def get_server_params(self):
+        return (self._foremanapi_server_url, self._foremanapi_username, self._foremanapi_password, self._foremanapi_validate_certs)
```
mdellweg: Connect is using those variables directly, so this might be dead code also.

plugins/module_utils/foreman_helper.py
evgeni: git says no, removed.

plugins/module_utils/foreman_helper.py
evgeni: same

PR #487, opened 2019-09-19
diff support

evgeni:
this adds diff output when ansible is called with --diff

create:

    --- before
    +++ after
    @@ -1 +1,7 @@
    -{}
    +{
    +    "fullname": "three.example",
    +    "id": 13,
    +    "location_ids": [],
    +    "name": "three.example.net",
    +    "organization_ids": []
    +}

    fixtures | CHANGED => {
        "changed": true, 
        "entity": {
            "created_at": "2019-09-20 07:34:22 UTC", 
            "dns": null, 
            "dns_id": null, 
            "fullname": "three.example", 
            "id": 13, 
            "interfaces": [], 
            "locations": [], 
            "name": "three.example.net", 
            "organizations": [], 
            "parameters": [], 
            "subnets": [], 
            "updated_at": "2019-09-20 07:34:22 UTC"
        }
    }

update:

    --- before
    +++ after
    @@ -1,5 +1,5 @@
     {
    -    "fullname": "three",
    +    "fullname": "three.example",
         "id": 12,
         "location_ids": [],
         "name": "three.example.net",

    fixtures | CHANGED => {
        "changed": true, 
        "entity": {
            "created_at": "2019-09-20 07:31:40 UTC", 
            "dns": null, 
            "dns_id": null, 
            "fullname": "three.example", 
            "id": 12, 
            "interfaces": [], 
            "locations": [], 
            "name": "three.example.net", 
            "organizations": [], 
            "parameters": [], 
            "subnets": [], 
            "updated_at": "2019-09-20 07:33:07 UTC"
        }
    }

delete:

    --- before
    +++ after
    @@ -1,3 +1 @@
    -{
    -    "id": 12
    -}
    +{}

    fixtures | CHANGED => {
        "changed": true, 
        "entity": null
    }

### DIFF
--- a/plugins/module_utils/foreman_helper.py
+++ b/plugins/module_utils/foreman_helper.py
@@ -162,6 +162,9 @@ class ForemanAnsibleModule(AnsibleModule):
         self.state = 'undefined'
         self.entity_spec = {}
 
+        self._before = {}
+        self._after = {}
+
     def clean_params(self):
         return {k: v for (k, v) in self._params.items() if v is not None and k not in self._aliases}
 
@@ -292,6 +295,8 @@ class ForemanAnsibleModule(AnsibleModule):
                 result = {'id': result['id']}
             else:
                 result = self.show_resource(resource, result['id'], params=params)
+        if self._before == {}:
+            self._before = result
         return result
 
     def find_resource_by_name(self, resource, name, **kwargs):
@@ -369,6 +374,13 @@ class ForemanAnsibleModule(AnsibleModule):
                 changed, updated_entity = self._delete_entity(resource, current_entity, params, synchronous)
         else:
             self.fail_json(msg='Not a valid state: {0}'.format(state))
+
+        if self._after == {}:
+            self._after = updated_entity
+        else:
+            # ensure_entity was called multiple times, we can't be sure which entity is the real one
+            self._after = {'invalid': True}
+
         return changed, updated_entity
 
     def _create_entity(self, resource, desired_entity, params, entity_spec, synchronous):
@@ -549,6 +561,15 @@ class ForemanEntityAnsibleModule(ForemanAnsibleModule):
                     changed |= self.ensure_entity_state(
                         'parameters', None, current_parameter, state="absent", entity_spec=parameter_entity_spec, params=scope)
         return changed
+
+    def exit_json(self, **kwargs):
+        if self._after is None or 'invalid' not in self._after:
+            if 'diff' not in kwargs:
+                kwargs['diff'] = {'before': _flatten_entity(self._before or {}, self.entity_spec),
+                                  'after': _flatten_entity(self._after or {}, self.entity_spec)}
+            if 'entity' not in kwargs:
+                kwargs['entity'] = self._after
+        super(ForemanEntityAnsibleModule, self).exit_json(**kwargs)
 
 
 class KatelloAnsibleModule(KatelloMixin, ForemanAnsibleModule):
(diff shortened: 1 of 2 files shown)


evgeni:
to make delete more meaningful, we could make `thin=True` only *pass* `thin` to the API, but not actually *filter* the results.

evgeni:
and this obviously doesn't catch changes in the params

sean797:
Looks good, I wonder if we should filter out params we cannot change or the module does not support yet. For example, with this domain module: `id, subnet, interfaces, dns, dns_id, updated_at, created_at` ?

evgeni:
I said the same yesterday to @ekohl while showing him the code on irc, and he mentioned that this information could still be useful. After giving this a bit more thought, I think the "result" should be split. We should return the **diff** for what we can change, and the **whole** new entity, so the user can inspect on their leisure.

I'll try to extend this PR to do all that.

evgeni:
okay, and now I've implemented it for all entity modules that don't call `ensure_entity` more than once.

sean797:
LGTM :+1:

plugins/module_utils/foreman_helper.py
```diff
@@ -369,6 +374,13 @@ def ensure_entity(self, resource, desired_entity, current_entity, params=None, s
                 changed, updated_entity = self._delete_entity(resource, current_entity, params, synchronous)
         else:
             self.fail_json(msg='Not a valid state: {0}'.format(state))
+
+        if self._after == {}:
+            self._after = updated_entity
+        else:
+            # ensure_entity was called multiple times, we can't be sure which entity is the real one
```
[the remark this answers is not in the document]
evgeni: we could add explicit `save` params to `find_resource` and `ensure_entity`, thus making this less magic in such cases

plugins/module_utils/foreman_helper.py
```diff
@@ -550,6 +562,15 @@ def ensure_scoped_parameters(self, scope, entity, parameters):
                         'parameters', None, current_parameter, state="absent", entity_spec=parameter_entity_spec, params=scope)
         return changed
 
+    def exit_json(self, **kwargs):
+        if self._after is None or 'invalid' not in self._after:
+            if 'diff' not in kwargs:
+                kwargs['diff'] = {'before': _flatten_entity(self._before or {}, self.entity_spec),
+                                  'after': _flatten_entity(self._after or {}, self.entity_spec)}
```
mdellweg: I do not think, this `or` can ever be hit (unless something in the `__init__` calls `exit_json`).

plugins/module_utils/foreman_helper.py
mdellweg: This can only happen if there are sub-entities, right? Do you think, we can add a `subentity_path`, where the diff can be placed? It would of course complicate the before/after tracking.

plugins/module_utils/foreman_helper.py
evgeni: sure, `ensure_entity` with `state=absent` will make `self._after` `None` and `_flatten_entity` can't cope with that.

evgeni: (you could argue we can fix that in flatten)

plugins/module_utils/foreman_helper.py
evgeni: params and stuff, and modules that support `affect_multiple`

plugins/module_utils/foreman_helper.py
evgeni: (fixed flatten_entity)

evgeni:
there were no objections, so I'm going to merge this and then we'll see how it continues

PR #494, opened 2019-09-24
restructure documentation

evgeni:
see individual commit messages for their changes, TL;DR: move non-user docs to own files, remove "goals" section, move support sections further up


README.md
```diff
@@ -2,14 +2,6 @@
 
 Ansible modules for interacting with the Foreman API and various plugin APIs such as Katello.
 
-## Goals
```
ekohl: I still like the concept of goals. For example, it can clarify whether modules for plugin's resources are accepted or not.

evgeni: My original goal with dropping the goals section was to make the excerpt on Galaxy (https://galaxy.ansible.com/theforeman/foreman) more meaningful.

As to your explicit comment: doesn't the main description "Ansible modules for interacting with the Foreman API and various plugin APIs such as Katello" include all modules?

README.md
```diff
@@ -2,25 +2,31 @@
...
-  * centralized location for community modules
-  * a single repository to clone for interacting with Foreman & plugins
-  * source for the official Ansible collection
-
 ## Documentation
 
 A list of all modules and their documentation can be found at [theforeman.org/plugins/foreman-ansible-modules](https://theforeman.org/plugins/foreman-ansible-modules/).
 
-## Supported Foreman and plugins versions
+Documentation how to [write](docs/developing.md), [test](docs/testing.md) and [debug](docs/debugging.md) modules is available in the [`docs`](docs/) folder.
```
mdellweg: Are those links working when viewed from Galaxy?
And i guess, i know the answer... Probably not.

evgeni: probably not, no.

evgeni: definitely not, just tested with the (old) `tests/README.md` link we had, and it points to https://galaxy.ansible.com/tests/README.md :see_no_evil:

evgeni:
@ekohl @mdellweg any objections against merging after your/my comments?

README.md
ekohl: I guess the summary on GH is also clear enough:
> Repository for Ansible modules that communicate with Foreman and plugin APIs

mdellweg:
Apart from the link problem, I didn't look closely, but what I saw was reasonable.

evgeni:
Given the link problem was there before, I am going to ignore it for now and see that we fix it in another way later.

evgeni:
(I've opened #496 to track that)

PR #495, opened 2019-09-24
include stable ansible versions for tests

evgeni:
This adds tests against Ansible 2.7 and 2.8 (didn't bother to add older releases).

One thing to consider: Travis only gives us 5 parallel jobs, but this increases our need to 6, so one will always wait :( Might want to ask Travis people if they can up our (theforeman) limits (see https://twitter.com/travisci/status/651856570784083968).

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,18 +1,41 @@
 dist: xenial
 language: python
-python:
-  - "2.7"
-  - "3.5"
-  - "3.6"
-  - "3.7"
+matrix:
+  include:
+    - name: Ansible devel with Python 2.7
+      python: "2.7"
+      env:
+        - ANSIBLE=devel
+    - name: Ansible devel with Python 3.5
+      python: "3.5"
+      env:
+        - ANSIBLE=devel
+    - name: Ansible devel with Python 3.6
+      python: "3.6"
+      env:
+        - ANSIBLE=devel
+    - name: Ansible devel with Python 3.7
+      python: "3.7"
+      env:
+        - ANSIBLE=devel
+    - name: Ansible 2.7 with Python 3.7
+      python: "3.7"
+      env:
+        - ANSIBLE=stable-2.7
+    - name: Ansible 2.8 with Python 3.7
+      python: "3.7"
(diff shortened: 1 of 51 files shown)


.travis.yml
```diff
@@ -1,18 +1,41 @@
...
+    - name: Ansible devel with Python 3.6
+      python: "3.6"
+      env:
+        - ANSIBLE=devel
+    - name: Ansible devel with Python 3.7
+      python: "3.7"
+      env:
+        - ANSIBLE=devel
+    - name: Ansible 2.7 with Python 3.7
+      python: "3.7"
```
ehelms: This should be 2.7

ehelms: Nvm, I miss read this

.travis.yml
```diff
@@ -1,18 +1,41 @@
 dist: xenial
 language: python
-python:
-  - "2.7"
-  - "3.5"
-  - "3.6"
-  - "3.7"
+matrix:
+  include:
+    - name: Ansible devel with Python 2.7
```
ehelms: This may sound a little silly, but you could either swap the order of the variables or the wording to they match? Makes trying to see the matrix easier for me.

.travis.yml
```diff
@@ -1,18 +1,41 @@
 dist: xenial
 language: python
-python:
-  - "2.7"
-  - "3.5"
-  - "3.6"
-  - "3.7"
+matrix:
```
ehelms: Do we need to test the same version of Ansible on different pythons? Should we be able to trust that Ansible itself has tested its code on different pythons and for us we only care that we work for certain pythons and certain Ansible versions?

evgeni: Ansible, not, but we occasionally run into differences between Python 2.7, 3.5 and 3.7 (3.6 so far always behaved like 3.7) around sorting of dicts, which at least once found an actual bug in my code.

.travis.yml
evgeni: matrix

evgeni:
closing in favor of #501

PR #499, opened 2019-09-30
Add updated name support

bagasse:
Add ability to change the name of:
* hostgroup
* subnet
* domain
* partition_template
* provisioning_template
* installation_media
* config_group
* global_parameter
* operatingsystem
* architecture
* host_collection


mdellweg:
I know, i have seen something like this before...
Should we add a note, that this can never be idempotent?

bagasse:
@mdellweg, yes, you're right. Added statement on idempotency.

mdellweg:
I believe, you must code a conflict of renaming with `affects_multiple`.

plugins/modules/foreman_installation_medium.py
```diff
@@ -137,6 +137,8 @@ def main():
     if affects_multiple:
         if module.state == 'present_with_defaults':
             module.fail_json(msg="'state: present_with_defaults' and 'name: *' cannot be used together")
+        if module.params['updated_name']:
+            module.fail_json(msg="updated_name not allowed if 'name: *'!")
```
mdellweg: :+1:

plugins/modules/foreman_ptable.py
```diff
@@ -341,8 +341,8 @@ def main():
         changed = module.ensure_entity_state('ptables', entity_dict, entity)
     else:
         entity_dict.pop('name')
-        for entity in entities:
-            changed |= module.ensure_entity_state('ptables', entity_dict, entity)
+        for ent in entities:
+            changed |= module.ensure_entity_state('ptables', entity_dict, ent)
```
mdellweg: why should this be needed?

bagasse: This not needed, reverted.

plugins/modules/foreman_installation_medium.py
```diff
@@ -124,12 +130,15 @@ def main():
 
     module.connect()
     name = entity_dict['name']
+    entity = None
```
mdellweg: why is this? I think we want to be extra sure, that `entity` as a variable is not even defined, in case of `affects_multiple`.

bagasse: If this variable is not defined, I've an `UnboundLocalError: local variable 'entity' referenced before assignment` with python 3 because I check that entity is not None before applying the updated name.

mdellweg: I don't understand why it should only be problematic in py3? Also it should not be needed if you move that transformation in one of the `not affects_multiple` blocks.

bagasse: Rewritten condition so that's not needed anymore.

plugins/modules/foreman_installation_medium.py
```diff
@@ -138,14 +146,16 @@ def main():
     if affects_multiple:
         entities = module.list_resource('media')
         if not module.desired_absent:  # not 'thin'
-            entities = [module.show_resource('media', entity['id']) for entity in entities]
+            entities = [module.show_resource('media', ent['id']) for ent in entities]
```
mdellweg:
```suggestion
            entities = [module.show_resource('media', entity['id']) for entity in entities]
```

plugins/modules/foreman_provisioning_template.py
```diff
@@ -345,11 +349,13 @@ def main():
             # Nothing to do; shortcut to exit
             module.exit_json(changed=False)
         if not module.desired_absent:  # not 'thin'
-            entities = [module.show_resource('provisioning_templates', entity['id']) for entity in entities]
+            entities = [module.show_resource('provisioning_templates', ent['id']) for ent in entities]
```
mdellweg:
```suggestion
            entities = [module.show_resource('provisioning_templates', entity['id']) for entity in entities]
```

plugins/modules/foreman_ptable.py
```diff
@@ -317,11 +322,13 @@ def main():
             # Nothing to do; shortcut to exit
             module.exit_json(changed=False)
         if not module.desired_absent:  # not 'thin'
-            entities = [module.show_resource('ptables', entity['id']) for entity in entities]
+            entities = [module.show_resource('ptables', ent['id']) for ent in entities]
```
mdellweg:
```suggestion
            entities = [module.show_resource('ptables', entity['id']) for entity in entities]
```

PR #501, opened 2019-09-30
migrate to GH actions

evgeni:


evgeni:
I would've expected this would trigger an acrion run, like you can see in https://github.com/evgeni/foreman-ansible-modules/pull/2/ -- weird

evgeni:
this is btw an alternative to #495 :)

evgeni:
(closed the other PR, now it triggered, weird)

.github/workflows/main.yml
```diff
@@ -0,0 +1,56 @@
+name: CI
+
+on: [pull_request]
```
evgeni: this should also probably include "push to master"

evgeni:
@ehelms I've pushed a few minor changes, mind having a second look before I merge?

mdellweg:
I am kind of pleased to see that the tests are working independently from the chosen CI.

tests/extras/sanity.yml
```diff
@@ -3,7 +3,12 @@
 - hosts: fixtures
   tasks:
     - name: Run ansible-test
-      command: "ansible-test sanity --venv --python {{ lookup('env', 'TRAVIS_PYTHON_VERSION') | default(ansible_python_version.split('.')[:2] | join('.'), true) }} plugins/"
+      command: "ansible-test sanity --venv --python {{ ansible_python['version']['major'] }}.{{ ansible_python['version']['minor'] }} plugins/"
```
mdellweg: Yay!

requirements.txt
```diff
@@ -1,3 +1,2 @@
 apypie>=0.0.5
-ansible>=2.3
```
mdellweg: Isn't that still a requirement?

evgeni: yeah. I needed to remove it from the dev file so I can install different ansible versions from the matrix, but the end user requirement is still valid.

mdellweg: Can you install the other ansible version beforehand? Than this requirement will already be satisfied and not touched again.

requirements.txt
```diff
@@ -1,2 +1,3 @@
 apypie>=0.0.5
```
mdellweg: Btw. Don't we need 0.1.0 here?

evgeni: not strictly, I don't think we *depend* on any 0.1 features (but then again, I didn't test it either)

Issue #502, opened 2019-10-01
error when waiting for a long running task


evgeni:
##### SUMMARY
when waiting for a long running task (like `katello_sync` of all RHEL), the job fails with

```
fatal: [localhost]: FAILED! => {"changed": false, "msg": "Error while performing show on foreman_tasks: maximum recursion depth exceeded in cmp"}
```

##### ISSUE TYPE
 - Bug Report

##### ANSIBLE VERSION

```

```

##### KATELLO/FOREMAN VERSION

```

```

##### APYPIE VERSION

```

```

##### STEPS TO REPRODUCE

```yaml

```

##### ACTUAL RESULTS

```

```

PR #503, opened 2019-10-01
make foreman_tasks show asynchronous

evgeni:
otherwise the code will go into a deep recursion because the API returns
a foreman_task and then we wait for that to finish, query it, get a
task, and so on.

Fixes: #502

### DIFF
--- a/plugins/module_utils/foreman_helper.py
+++ b/plugins/module_utils/foreman_helper.py
@@ -515,7 +515,7 @@ class ForemanAnsibleModule(AnsibleModule):
                 self.fail_json(msg="Timout waiting for Task {0}".format(task['id']))
             time.sleep(self.task_poll)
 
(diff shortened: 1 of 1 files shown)


evgeni:
@mdellweg <insert some mumbling about Evgeni didn't want `sync=True` by default ;)>

mdellweg:
Should we not try to convert the recursion into a simple loop?

evgeni:
@mdellweg it *is* a simple loop. besides the fact that `resource_action` will autodetect that we're looking at a foreman task and try to wait for it, while waiting, it will notice that the result is a foreman task, and will try to wait for it, …

evgeni:
We *could* extend the check in to be

```python
is_foreman_task = isinstance(result, dict) and 'action' in result and 'state' in result and 'pending' in result and resource != 'foreman_tasks'
```

mdellweg:
Sorry, i misread the code change. I think, the change you proposed, breaks the recursion. ACK!

PR #505, opened 2019-10-02
pass on errors from foreman

evgeni:
before:
```
fixtures | FAILED! => {
    "changed": false, 
    "msg": "Error while performing create on hosts: 422 Client Error: Unprocessable Entity for url: https://foreman.example.com/api/hosts"
}
```

after:
```
fixtures | FAILED! => {
    "changed": false, 
    "error": {
        "errors": {
            "interfaces.name": [
                "hostname can contain only lowercase letters, numbers, dashes and dots according to RFC921, RFC952 and RFC1123"
            ], 
            "name": [
                "hostname can contain only lowercase letters, numbers, dashes and dots according to RFC921, RFC952 and RFC1123"
            ]
        }, 
        "full_messages": [
            "Name hostname can contain only lowercase letters, numbers, dashes and dots according to RFC921, RFC952 and RFC1123", 
            "Name hostname can contain only lowercase letters, numbers, dashes and dots according to RFC921, RFC952 and RFC1123"
        ], 
        "id": null
    }, 
    "msg": "Error while performing create on hosts: 422 Client Error: Unprocessable Entity for url: https://foreman.example.com/api/hosts"
}
```

### DIFF
--- a/plugins/module_utils/foreman_helper.py
+++ b/plugins/module_utils/foreman_helper.py
@@ -16,6 +16,7 @@ from ansible.module_utils.basic import AnsibleModule
 
 try:
     import apypie
+    import requests.exceptions
     HAS_APYPIE = True
 except ImportError:
     HAS_APYPIE = False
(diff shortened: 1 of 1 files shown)


evgeni:
pondering if we should just override `fail_json` and do it directly there.

mdellweg:
Why not make `fail_from_exc` a _thing_ and call `fail_json` from there without returning first?

Issue #407, opened 2019-08-13
katello_upload should use `rpm`s Python bindings, not `check_output`


evgeni:
I think this should use `rpm`s Python bindings, not `check_output`.

_Originally posted by @evgeni in https://github.com/theforeman/foreman-ansible-modules/pull/403_

PR #508, opened 2019-10-04
use rpm's Python bindings to get details of an RPM

evgeni:
Fixes: #407

### DIFF
--- a/plugins/modules/katello_upload.py
+++ b/plugins/modules/katello_upload.py
@@ -33,6 +33,8 @@ short_description: Upload content to Katello
 description:
     - Allows the upload of content to a Katello repository
 author: "Eric D Helms (@ehelms)"
+requirements:
+  - rpm
 options:
   src:
     description:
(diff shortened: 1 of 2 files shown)


plugins/modules/katello_upload.py
```diff
@@ -94,6 +121,9 @@ def main():
         ),
     )
 
+    if not HAS_RPM:
+        module.fail_json(msg='The rpm Python module is required', exception=RPM_IMP_ERR)
```
mdellweg: As we do this lazy import failing anyway, i think we can be even lazyer and move this in the "if yum" section, or inside the `get_rpm_info`. No need to fail file uploads here.

evgeni: `get_rpm_info` has no access to `module`, but I can move that to the `if yum` section, yes.

and need to fix tests. I hate RPM sometimes

evgeni:
@mdellweg mind having a second look at the latest changes? it passes tests now and I need a shower.

evgeni:
Then you didn't look what exactly `rpm-py-installer` does ;-)

Merging before you do ;)

mdellweg:
Ugh! Let's sure hope, that we can still satisfy the dependencies when distributing the collection as a package.

evgeni:
That's why I added it only to the dev requirements. Normal installs shall just use the system python's bindings.

PR #510, opened 2019-10-04
Add debian package upload

mdellweg:


mdellweg:
"Cannot have a has_many :through association 'Katello::Deb#repositories' which goes through 'Katello::Deb#repository_debs' before the through association is defined."
Hmmm, i think i might be up to something.

plugins/modules/katello_upload.py
```diff
@@ -59,7 +59,7 @@
     required: true
     type: str
 notes:
-    - Currently only idempotent when uploading to an RPM & file repository
+    - Currently only uploading to an deb, RPM & file repository is supported
```
evgeni:
```suggestion
    - Currently only uploading to an deb, RPM & file repositories is supported
```

plugins/modules/katello_upload.py
```diff
@@ -146,7 +158,14 @@ def main():
     checksum = checksum.hexdigest()
 
     content_unit = None
-    if entity_dict['repository']['content_type'] == 'yum':
+    if entity_dict['repository']['content_type'] == 'deb':
+        if not HAS_DEBFILE:
+            module.fail_json(msg='The python-debian module is required', exception=DEBFILE_IMP_ERR)
+
+        name, version, architecture = get_deb_info(entity_dict['src'])
+        query = 'name = "{0}" and version = "{1}" and architecture = "{2}"'.format(name, version, architecture)
```
evgeni: no epoch support?

requirements-dev.txt
```diff
@@ -2,6 +2,7 @@ flake8
 pytest
 vcrpy
 ansible_runner
+python-debian
```
evgeni: should that also be present for normal installs? if so, `requirements-common` is better suited.

(we don't have to make the built rpm depend on that)

plugins/modules/katello_upload.py
mdellweg: Is part of the Version.

requirements-dev.txt
mdellweg: Well, it's optional. otherwise we would not need the guarding of its import.

plugins/modules/katello_upload.py
mdellweg:
```suggestion
    - Currently only uploading to deb, RPM & file repositories is supported
```

plugins/modules/katello_upload.py
evgeni: Oh, okay

requirements-dev.txt
evgeni: The guarding is an ansible convention, we do the same for apypie. But I see your point. shall we start an optional.txt? And move ipaddress there too?

tests/test_playbooks/upload.yml
```diff
@@ -16,6 +16,16 @@
   - include: tasks/repository.yml
     vars:
       repository_state: present
+  - include: tasks/repository.yml
+    vars:
+      repository_state: absent
+      repository_name: "Test Deb Repository"
+      repository_content_type: "deb"
```
evgeni: technically not needed here

mdellweg: right, but otherwise the content_type would be defaulted to yum by that task, and that feels wrong.

evgeni: yet you don't specify it on  line 88+ where the main destroy happens :D

mdellweg: Now we do.

PR #513, opened 2019-10-07
increase the sync timeout

evgeni:
30 minutes is too short for a full repo sync sometimes, let's wait for 2
hours.

in the old (nailgun) code this was set to "180000", which we translated
to 30 minutes, but nailgun in reality takes seconds too, so the old code
really waited up to 50 hours.

### DIFF
--- a/plugins/modules/katello_sync.py
+++ b/plugins/modules/katello_sync.py
@@ -117,7 +117,7 @@ def main():
         ),
     )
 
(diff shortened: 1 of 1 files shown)


mdellweg:
I suspect 2h is yet not enough for really big repositories (e.g. debian stable main). Can we extend it to 12h?

evgeni:
you need bigger pipes ;)

PR #514, opened 2019-10-07
always ping and early ping

evgeni:
it was optional because of nailgun, which used a katello-specific path
for pinging -- we don't do that anymore.

also call it earlier, so that we can fetch the latest apidoc *before*
patching it


evgeni:
#515 is better

PR #524, opened 2019-10-17
Add option 'description' to katello_content_view_version

### LINKED ISSUE #521: katello_content_view_version: add description option
##### SUMMARY
Description field seems missing. It is used often to store information about filters that the content view had at promotion time of the version in question. 
There's no other simple way to store this info within satellite, so this filed is a must.

##### ISSUE TYPE
 - Feature Idea

nesanton:
Fixes #521.

### DIFF
--- a/tests/test_playbooks/fixtures/content_view_version-0.yml
+++ b/tests/test_playbooks/fixtures/content_view_version-0.yml
@@ -4,22 +4,16 @@ interactions:
     headers:
       Accept:
       - application/json;version=2
-      Accept-Encoding:
-      - gzip, deflate
-      Connection:
-      - keep-alive
-      User-Agent:
-      - apypie (https://github.com/Apipie/apypie)
     method: GET
     uri: https://foreman.example.com/api/status
   response:
     body:
-      string: !!python/unicode '{"result":"ok","status":200,"version":"1.22.0","api_version":2}'
+      string: !!python/unicode '{"result":"ok","status":200,"version":"1.21.3","api_version":2}'
     headers:
+      apipie-checksum:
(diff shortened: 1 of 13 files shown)


evgeni:
Thanks @nesanton! This looks correct from the diff, not sure why CI failed, will investigate.

tests/test_playbooks/fixtures/content_view_version-0.yml
```diff
@@ -4,22 +4,16 @@ interactions:
...
-      - apypie (https://github.com/Apipie/apypie)
     method: GET
     uri: https://foreman.example.com/api/status
   response:
     body:
-      string: !!python/unicode '{"result":"ok","status":200,"version":"1.22.0","api_version":2}'
+      string: !!python/unicode '{"result":"ok","status":200,"version":"1.21.3","api_version":2}'
     headers:
+      apipie-checksum:
+      - d841c30f0b33dc0af1a12bd583b7add23cc87b95
```
evgeni: so it seems `filter_header` didn't work in https://github.com/theforeman/foreman-ansible-modules/blob/master/tests/vcr_python_wrapper.py#L125 :(

@mdellweg help appreciated :)

evgeni: right, because that only applies to the request, not the response m)

evgeni:
@nesanton  should be fixed by re-recording after https://github.com/theforeman/foreman-ansible-modules/pull/525 is merged.

evgeni:
@nesanton merged #525, please rebase and re-record

evgeni:
Спасибо Антон!

PR #527, opened 2019-10-18
Add filters support for roles.

Fodoj:
As proposed here https://github.com/theforeman/foreman-ansible-modules/pull/293, this PR adds management of Role Filters inside Role directly. It doesn't support overriding locations or organizations, which can be added at later point.

Note that this implementation basically re-adds all filters everytime. My assumption is that it is simpler (from code perspetive) and more performant than comparing all existing and yaml filters one by one, but I am open for suggestions if this implementation is not good enough.

### DIFF
--- a/plugins/modules/foreman_role.py
+++ b/plugins/modules/foreman_role.py
@@ -126,7 +126,7 @@ def main():
         if filters is not None:
             for role_filter in filters:
                 role_filter['role_id'] = entity['id']
-                role_filter['permissions'] = module.find_resources_by_name('permissions', role_filter['permissions'])
+                role_filter['permissions'] = module.find_resources_by_name('permissions', role_filter['permissions'], thin=True)
                 module.ensure_entity_state('filters', role_filter, None, None, 'present', filters_entity_spec)
             for old_filter in entity['filter_ids']:
                 module.ensure_entity('filters', None, {'id': old_filter}, {}, 'absent')
--- a/plugins/modules/foreman_role.py
+++ b/plugins/modules/foreman_role.py
@@ -94,7 +94,7 @@ def main():
             description=dict(),
             locations=dict(type='entity_list', flat_name='location_ids'),
             organizations=dict(type='entity_list', flat_name='organization_ids'),
-            filters=dict(type='entity_list', flat_name='filters'),
+            filters=dict(type='entity_list', flat_name='filter_ids'),
         ),
     )
 
@@ -122,17 +122,16 @@ def main():
 
     changed, entity = module.ensure_entity('roles', entity_dict, entity)
 
-    if not module.desired_absent:
+    if not module.desired_absent and not module.check_mode:
         if filters is not None:
-            existing_filters = entity['filters']
             for f in filters:
                 f['role_id'] = entity['id']
                 f['permissions'] = module.find_resources_by_name('permissions', f['permissions'])
                 module.ensure_entity_state('filters', f, None, None, 'present', filters_entity_spec)
-            for old_filter in existing_filters:
-                if not type(old_filter) is dict:
-                    old_filter = {'id': old_filter}
-                module.ensure_entity('filters', None, old_filter, {}, 'absent')
+            for old_filter in entity['filter_ids']:
+                module.ensure_entity('filters', None, {'id': old_filter}, {}, 'absent')
+            if len(entity['filter_ids']) != len(filters):
+                changed = True
 
     module.exit_json(changed=changed)
 
--- a/plugins/modules/foreman_role.py
+++ b/plugins/modules/foreman_role.py
@@ -125,10 +125,10 @@ def main():
     if not module.desired_absent:
         if filters is not None:
             existing_filters = entity['filters']
-            for filter in filters:
-                filter['role_id'] = entity['id']
-                filter['permissions'] = module.find_resources_by_name('permissions', filter['permissions'])
-                module.ensure_entity_state('filters', filter, None, None, 'present', filters_entity_spec)
+            for f in filters:
+                f['role_id'] = entity['id']
+                f['permissions'] = module.find_resources_by_name('permissions', f['permissions'])
+                module.ensure_entity_state('filters', f, None, None, 'present', filters_entity_spec)
             for old_filter in existing_filters:
                 if not type(old_filter) is dict:
                     old_filter = {'id': old_filter}
--- a/plugins/modules/foreman_role.py
+++ b/plugins/modules/foreman_role.py
@@ -50,6 +50,10 @@ options:
     description: List of organizations the role should be assigned to
     required: false
     type: list
+  filters:
+    description: Filters with permissions for this role
+    required: false
+    type: list
   state:
     description: role presence
     default: present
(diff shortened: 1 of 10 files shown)


evgeni:
@Fodoj thanks! I will have a deeper look later. The CI fails because we had a bug in the recording tooling which we fixed earlier today, could you please rebase your branch on top of our master and re-record the fixtures? THis should make CI green.

plugins/modules/foreman_role.py
```diff
@@ -101,7 +118,21 @@ def main():
...
 
-    changed = module.ensure_entity_state('roles', entity_dict, entity)
+    filters = entity_dict.pop("filters", None)
+
+    changed, entity = module.ensure_entity('roles', entity_dict, entity)
+
+    if not module.desired_absent:
+        if filters is not None:
+            existing_filters = entity['filters']
+            for filter in filters:
```
evgeni: I think `filter` is a built in function in Python, so we might want to use some other name here

plugins/modules/foreman_role.py
```diff
@@ -85,9 +94,17 @@ def main():
             description=dict(),
             locations=dict(type='entity_list', flat_name='location_ids'),
             organizations=dict(type='entity_list', flat_name='organization_ids'),
+            filters=dict(type='entity_list', flat_name='filters'),
```
evgeni: you could (but don't have to) add sub-options to this definition, to make Ansible validate them, see

https://github.com/theforeman/foreman-ansible-modules/blob/b02a7bcff1d6648844717636c46db2d74dc5c0b7/plugins/modules/katello_content_view.py#L160-L163

and

https://github.com/theforeman/foreman-ansible-modules/blob/b02a7bcff1d6648844717636c46db2d74dc5c0b7/plugins/modules/katello_content_view.py#L50-L65

plugins/modules/foreman_role.py
Fodoj: replace with `f` :)

plugins/modules/foreman_role.py
Fodoj: I added filters spec below, which should also validate it on creation of filters, if I am not mistaken?

Fodoj:
@evgeni re Ci - saw it, updated cassetes

tests/test_playbooks/tasks/role.yml
```diff
@@ -17,6 +17,11 @@
     description: "{{ role_description }}"
     locations: "{{ role_locations }}"
     organizations: "{{ role_organizations }}"
+    filters:
```
evgeni: this makes the tests create the role *always* with those filters and we can't test idempotentcy well.

could you make it `filters: "{{ role_filters | default(omit) }}" here and pass `role_filters` as a var in `test_playbooks/role.yml` in some (but not all) cases? something like the following workflow should be tested: create a role without a filter, ensure that creating it again without a filter does not yield any changes, add a filter (expect changes), add the same filter (no changes), remove filter (changes).

plugins/modules/foreman_role.py
```diff
@@ -101,7 +118,21 @@ def main():
...
+
+    if not module.desired_absent:
+        if filters is not None:
+            existing_filters = entity['filters']
+            for f in filters:
+                f['role_id'] = entity['id']
+                f['permissions'] = module.find_resources_by_name('permissions', f['permissions'])
+                module.ensure_entity_state('filters', f, None, None, 'present', filters_entity_spec)
+            for old_filter in existing_filters:
+                if not type(old_filter) is dict:
```
evgeni:
```
ERROR: plugins/modules/foreman_role.py:133:0: unidiomatic-typecheck: Type comparison using type() found. Use isinstance() instead
```

plugins/modules/foreman_role.py
evgeni: I would've went with `role_filter` ;)

plugins/modules/foreman_role.py
evgeni: Nope, not for nested params (yet) sadly.

plugins/modules/foreman_role.py
```diff
@@ -101,7 +118,21 @@ def main():
...
             entity_dict['organizations'] = module.find_resources_by_name('organizations', entity_dict['organizations'], thin=True)
 
-    changed = module.ensure_entity_state('roles', entity_dict, entity)
+    filters = entity_dict.pop("filters", None)
+
+    changed, entity = module.ensure_entity('roles', entity_dict, entity)
+
+    if not module.desired_absent:
+        if filters is not None:
+            existing_filters = entity['filters']
```
evgeni: why are you assigning this? just for nicer `for old_filter in existing_filters` later? or do you meant to use `.copy()`?

Fodoj: Good question. Now I realize that entity won't be reloaded with new filters, so this variable is not really needed.

Fodoj:
I've improved handling of "changed" status (it's still very simplistic though) and fixed other things as in previous review. Please verify if it's okay now.

plugins/modules/foreman_role.py
Fodoj: Right, that's much better. Golang infected my mind. Updated.

plugins/modules/foreman_role.py
```diff
@@ -101,7 +118,20 @@ def main():
...
+
+    if not module.desired_absent and not module.check_mode:
+        if filters is not None:
+            for role_filter in filters:
+                role_filter['role_id'] = entity['id']
+                role_filter['permissions'] = module.find_resources_by_name('permissions', role_filter['permissions'])
+                module.ensure_entity_state('filters', role_filter, None, None, 'present', filters_entity_spec)
+            for old_filter in entity['filter_ids']:
+                module.ensure_entity('filters', None, {'id': old_filter}, {}, 'absent')
+            if len(entity['filter_ids']) != len(filters):
```
evgeni: this will not work if you remove and add exactly the same number of (different) filters.

what is the exact reason you don't update existing filters?

Fodoj: That would require pulling each existing filter and compare them field by field with what is in YAML. That would be a much more complex implementation that would indeed result in better update tracking, and I simply can’t invest much time into doing it right now. My suggestion is to tackle it as a next iteration in a follow up pull request.

evgeni: that's fine for me, yeah. just wanted to be sure the downside is known.

plugins/modules/foreman_role.py
```diff
@@ -101,7 +118,20 @@ def main():
...
-    changed = module.ensure_entity_state('roles', entity_dict, entity)
+    filters = entity_dict.pop("filters", None)
+
+    changed, entity = module.ensure_entity('roles', entity_dict, entity)
+
+    if not module.desired_absent and not module.check_mode:
+        if filters is not None:
+            for role_filter in filters:
+                role_filter['role_id'] = entity['id']
+                role_filter['permissions'] = module.find_resources_by_name('permissions', role_filter['permissions'])
```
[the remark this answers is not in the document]
evgeni: correct, but given the `permissions` endpoint doesn't really support this, the performance will be identical.

@Fodoj I'd prefer if you could fix that for consistency reasons, but I wouldn't block the PR on that. Please tell me if you have time, otherwise I'd merge as-is and follow up with a cleanup later.

evgeni:
there are two open comments, but neither are blocking, I am going to merge this this evening with whatever cleanup comes until then or not :)

plugins/modules/foreman_role.py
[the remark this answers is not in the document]
Fodoj: Done!

evgeni:
Thanks @Fodoj! I've opened #534 to track the add/remove enhancement

Fodoj:
Thanks for prompt review!

PR #529, opened 2019-10-18
add subs to activation key by pool_id

### LINKED ISSUE #165: k_activation_key can't add subscriptions that exist multiple times
##### SUMMARY

when adding subscriptions to an activation key, ansible errors out when there are multiple subscriptions with the same name (but different contract dates, e.g.). this is because `handle_find_response` explicitly expects one result, not multiple.

##### ISSUE TYPE
 - Bug Report

##### ANSIBLE VERSION
```
ansible 2.6.2
  configured module search path = [u'/home/remote/egolov/.ansible/plugins/modules', u'/usr/share/ansible/plugins/modules']
  python version = 2.7.15 (default, May 16 2018, 17:50:09) [GCC 8.1.1 20180502 (Red Hat 8.1.1-1)]

```

##### KATELLO/FOREMAN VERSION
```
# rpm -q tfm-rubygem-katello foreman
tfm-rubygem-katello-3.4.5.58-2HOTFIXRHBZ1522912.el7sat.noarch
foreman-1.15.6.34-1.el7sat.noarch

```

##### NAILGUN VERSION
```
git master, 6.3.z branches
```

##### EXPECTED RESULTS
Sub is added

##### ACTUAL RESULTS
```
"msg": "No subscription found for Red Hat Satellite Employee Subscription"
```

wbclark:
resolves #165 - k_activation_key can't add subscriptions that exist multiple times

### DIFF
--- a/plugins/modules/katello_activation_key.py
+++ b/plugins/modules/katello_activation_key.py
@@ -53,14 +53,19 @@ options:
     type: str
   subscriptions:
     description:
-      - List of subscriptions that include name
+      - List of subscriptions that include Name or Pool ID
     type: list
     suboptions:
       name:
         description:
           - Name of the Subscription to be added
         type: str
-        required: true
+        required: false
+      pool_id:
+        description:
+          - Pool ID of the Subscription to be added
+        type: str
+        required: false
   host_collections:
     description:
       - List of host collections to add to activation key
@@ -175,8 +180,11 @@ def main():
         ),
         argument_spec=dict(
             subscriptions=dict(type='list', elements='dict', options=dict(
-                name=dict(required=True),
-            )),
+                name=dict(),
+                pool_id=dict(),
+            ),
+                required_one_of=[['name', 'pool_id']]
+            ),
             content_overrides=dict(type='list', elements='dict', options=dict(
                 label=dict(required=True),
                 override=dict(required=True, choices=['enabled', 'disabled']),
@@ -226,8 +234,12 @@ def main():
 
         ak_scope = {'activation_key_id': activation_key['id']}
         if subscriptions is not None:
-            subscription_names = [subscription['name'] for subscription in subscriptions]
-            desired_subscriptions = module.find_resources_by_name('subscriptions', subscription_names, params=scope, thin=True)
+            desired_subscriptions = []
+            for subscription in subscriptions:
+                if subscription['name'] is not None and subscription['pool_id'] is None:
+                    desired_subscriptions.append(module.find_resource_by_name('subscriptions', subscription['name'], params=scope, thin=True))
+                if subscription['pool_id'] is not None:
+                    desired_subscriptions.append(module.find_resource_by_id('subscriptions', subscription['pool_id'], params=scope, thin=True))
             desired_subscription_ids = set(item['id'] for item in desired_subscriptions)
             current_subscriptions = module.list_resource('subscriptions', params=ak_scope)
             current_subscription_ids = set(item['id'] for item in current_subscriptions)
--- a/plugins/modules/katello_activation_key.py
+++ b/plugins/modules/katello_activation_key.py
@@ -53,7 +53,7 @@ options:
     type: str
   subscriptions:
     description:
-      - List of subscriptions that include Name or Pool ID
+      - List of subscriptions that include either Name or Pool ID. Pool IDs are preferred since Names will fail when one to many. Mutually exclusive.
     type: list
     suboptions:
       name:
@@ -137,7 +137,8 @@ EXAMPLES = '''
         - rhel7-servers
         - rhel7-production
     subscriptions:
-        - name: "Red Hat Enterprise Linux"
+      - pool_id: "8a88e9826db22df5016dd018abdd029b"
+      - pool_id: "8a88e9826db22df5016dd01a23270344"
     content_overrides:
         - label: rhel-7-server-optional-rpms
           override: enabled
@@ -183,7 +184,8 @@ def main():
                 name=dict(),
                 pool_id=dict(),
             ),
-                required_one_of=[['name', 'pool_id']]
+                required_one_of=[['name', 'pool_id']],
+                mutually_exclusive=[['name', 'pool_id']],
             ),
             content_overrides=dict(type='list', elements='dict', options=dict(
                 label=dict(required=True),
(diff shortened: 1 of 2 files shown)


evgeni:
Thanks William! Couple of things that popped up in my mind:

In #165 I linked more searchable data of subs, do you think adding any of them would be beneficial?

Do you know what the difference between id and cp_id (or product_id) is?

Can we easily add tests for this?

Can you please add an example?

plugins/modules/katello_activation_key.py
```diff
@@ -175,8 +180,11 @@ def main():
         ),
         argument_spec=dict(
             subscriptions=dict(type='list', elements='dict', options=dict(
-                name=dict(required=True),
-            )),
+                name=dict(),
+                pool_id=dict(),
+            ),
+                required_one_of=[['name', 'pool_id']]
```
evgeni: What happens if the user enters both? If I read the code in line 237+ correctly, the `pool_id` silently wins. I wonder if we should also set `mutually_exclusive=[['name', 'pool_id']]` so that such playbooks error out. Otherwise the user might end up with a playbook having both, updates one of the values and gets weird results.

Oh, and please add a trailing comma ;)
```suggestion
                required_one_of=[['name', 'pool_id']],
```

m-czernek:
This looks promissing. The PR works when I get input subscription_id. When I only input `name`, it still fails. I suppose that's to be expected. Suggestion for improvement:

* Make pool_id a list. After all, if we have only one id, we can always make it into a list in the yaml, but most of the time, we'll have multiple IDs (and if we have only one ID, I suppose we'd use Name anyways)

Architecturally, I wonder whether it wouldn't be the best if the search by name function was redesigned. For example:

* When ID is specified, add only the specified subscription
* When name is specified, find all IDs that belong to the name and add all of them

I am not 100% sure that's what we want, just floating ideas here.

wbclark:
@evgeni @m-czernek thanks for the feedback! I declared the `name` and `pool_id` mutually exclusive and made it clear in the documentation that `pool_id` should always be used in the case where `name` is one to many. I've also added an example.

About tests, once we can pass an organization or organization-id to `foreman_search_facts` I can get a `pool_id` (created at time of manifest import) corresponding to a subscription on the test manifest and use this in a test playbook.

evgeni:
@wbclark I think you now deleted the name example :(

evgeni:
Marek, not sure what you mean by making the pool id a list. Subscriptions is a list already, so you can pass something like

```yaml
subscriptions:
  - pool_id: 1234
  - pool_id: 5678
```

wbclark:
@evgeni I think it's OK for the example to follow the best practice since I specified in the documentation that "Pool IDs are preferred since Names will fail when one to many" but I can add the line with a name declaration back to the subscription list if you think it's valuable to have.

wbclark:
On the other hand, pool IDs are not useful when you are writing playbooks before a Satellite has even been deployed. The user would have to wait until the manifest if imported before they can declare what pool IDs should be included on activation keys. So I agree with @m-czernek that perhaps additional change to allow better searching by name is needed, for example the option to get all subscriptions matching a given name, with perhaps some additional filters applied such as service level, expiration date, etc.

evgeni:
I prefer to have examples for everything :)

sean797:
> pool IDs are not useful when you are writing playbooks before Katello has even been deployed. The user would have to wait until the manifest is imported and any custom products are created before they can declare what pool IDs should be included on activation keys.

Not true, you can access these on the portal as well..

Screenshot from 2019-10-21 17-36-27

wbclark:
@sean797 That Pool ID is the 'upstream' Pool ID on the Candlepin instance running on the portal. It is unfortunately different from the Pool ID on Katello's Candlepin instance, which gets created at the time when a manifest is imported or a custom product is created.

In the foreman DB:

```
foreman=# \d katello_pools;
                                         Table "public.katello_pools"
      Column       |            Type             |                         Modifiers                          
-------------------+-----------------------------+------------------------------------------------------------
 id                | integer                     | not null default nextval('katello_pools_id_seq'::regclass)
 cp_id             | character varying(255)      | not null
 created_at        | timestamp without time zone | not null
 updated_at        | timestamp without time zone | not null
 subscription_id   | integer                     | 
 account_number    | integer                     | 
 contract_number   | integer                     | 
 virtual           | boolean                     | 
 quantity          | integer                     | 
 start_date        | timestamp without time zone | 
 pool_type         | character varying(255)      | 
 end_date          | timestamp without time zone | 
 ram               | integer                     | 
 multi_entitlement | boolean                     | 
 consumed          | integer                     | 
 virt_only         | boolean                     | not null default false
 unmapped_guest    | boolean                     | not null default false
 virt_who          | boolean                     | not null default false
 hypervisor_id     | integer                     | 
 stacking_id       | character varying           | 
 organization_id   | integer                     | 
 upstream_pool_id  | character varying           | 
```
You can see there is a `cp_id` which is the local pool ID, and a distinct `upstream_pool_id` which is what @sean797 has referred to.

IF the `upstream_pool_id` were defined as a searchable field in using the Katello API at https://github.com/Katello/katello/blob/master/app/models/katello/pool.rb#L24-L42 it would actually be very useful in this module.

sean797:
That’s a shame :/ thanks @wbclark

evgeni:
I agree, long term, we should add more ways to find a specific subscription, but we'll also have to ensure that we still get just exactly one sub -- as otherwise the module won't be idempotent if executed multiple times and the search returns other/differently sorted results.

Given pool_id is the only *unique* value we can reference, I think the future logic should be: either the user passes a pool_id or a set of (name, …) and the search happens accordingly.

So today, adding pool_id will help, but won't be the final solution.

wbclark:
@evgeni requested doc changes are made. thank you for your input

evgeni:
@wbclark thanks! I am happy with the code side of things (for now). Tests are blocked on #537/#540 and we have a follow-up issue for the subscription search improvement in #541, so we could merge this.

@m-czernek does this PR make your life easier? Even if far from perfect/easy yet?

m-czernek:
`Marek, not sure what you mean by making the pool id a list`

Ha, correct! Sorry :))

`@m-czernek does this PR make your life easier?`

Well, it makes the module technically usable. `pool_id` is an internal resource, so we first have to search all the `pool_id`s such that the label matches what we specified, and then map it to a list of pool IDs. I agree that it's not the real solution, but it's a workaround that does work.

Thank you!

PR #531, opened 2019-10-20
Remove ansible_fail interpreter

mdellweg:
This was introduced to expose legacy test playbooks which we do not longer have.


tests/extras/sanity.yml
```diff
@@ -1,6 +1,7 @@
 ---
 # this playbook runs `ansible-test sanity` against our collection
-- hosts: fixtures
+- hosts: localhost
+  gather_subset: min
```
evgeni: TIL

Issue #338, opened 2019-07-10
foreman_search_facts: does not support scoped search


jwbernin:
##### SUMMARY
When I try to retrieve a list of Products using foreman_search_facts, it either complains that a required parameter is missing (organization_id), or complains that an unsupported parameter was provided (organization_id).

##### ISSUE TYPE
 - Bug Report

##### ANSIBLE VERSION

```
$ ansible --version
ansible 2.8.0
  config file = /nfshome/jberning/ansible/ansible.cfg
  configured module search path = ['/nfshome/jberning/Projects/foreman-ansible-modules/plugins/modules']
  ansible python module location = /usr/lib/python3.6/site-packages/ansible
  executable location = /usr/bin/ansible
  python version = 3.6.8 (default, Apr  3 2019, 17:26:03) [GCC 8.2.1 20180905 (Red Hat 8.2.1-3)]

```

##### KATELLO/FOREMAN VERSION

```
tfm-rubygem-katello-3.10.0.46-1.el7sat.noarch
foreman-1.20.1.34-1.el7sat.noarch
```

##### NAILGUN VERSION

```
$ pip3 show nailgun | grep Version
Version: 0.32.0
```

##### STEPS TO REPRODUCE

```yaml
---
- name: enable repos on Satellite
  hosts: localhost
  vars:
    satellite_host: newsat.ncphotography.lan
    satellite_user: admin
    satellite_password: admin
    org_id: 1
  tasks:
    - name: Get all Products
      foreman_search_facts:
        username: "{{ satellite_user }}"
        password: "{{ satellite_password }}"
        server_url: "https://{{ satellite_host }}"
        verify_ssl: false
        resource: products
  register: products
```
OR
```yaml
---
- name: enable repos on Satellite
  hosts: localhost
  vars:
    satellite_host: newsat.ncphotography.lan
    satellite_user: admin
    satellite_password: admin
    org_id: 1
  tasks:
    - name: Get all Products
      foreman_search_facts:
        username: "{{ satellite_user }}"
        password: "{{ satellite_password }}"
        server_url: "https://{{ satellite_host }}"
        verify_ssl: false
        resource: products
        organization_id: "{{ org_id }}"
    register: products
```

##### EXPECTED RESULTS

One of the two plays would work as expected
##### ACTUAL RESULTS

First play:
fatal: [localhost]: FAILED! => {"changed": false, "msg": "Failed to list resource: The following required parameters are missing: organization_id"}

Second play:
[...]

PR #537, opened 2019-10-21
Provide a way to pass parameters to the search_facts module

### LINKED ISSUE #523: foreman_search_facts: add params option
##### SUMMARY
It's not possible to specify params that GET API calls for different resources may have. Sometimes these params are required which makes the module fail. E.g. with ```resource: content_view_filters``` it fails with  "content_view_id is required". Since ```search``` can't be performed for ```content_view_filters``` using ```search: content_view_id=189```, it's impossible to use this module to list content_view_filters.

Suggestion: add ```params``` option that would acceps a dictionary of arbitrary parameters.
```yaml
- name: Ensure CV exists
  katello_content_view:
    other_stuff: blah
    name: cv1
  register: cv
    
- name: Search for CV filters of cv1
  foreman_search_facts:
    other_stuff: blah
    resource: content_view_filters
    params:
       content_view_id: "{{ cv.entity.id }}"
 register: all_cv1_filters
```

This module uses ```list_resource``` function https://github.com/theforeman/foreman-ansible-modules/blob/362161115f693cd8e7bda10c98ff9f82a64cbfa4/plugins/modules/foreman_search_facts.py#L119

that does accept ```params```, which are supposedly exactly what we need. Should be easy to add.
##### ISSUE TYPE
 - Feature Idea

m-czernek:
This PR adds the `params` option to the `foreman_search_facts` module. It resolves #523 as well as #338 .

WDYT?

This seems to work with the `subscriptions` resources, which requires org_id as an extra parameter.

### DIFF
--- a/plugins/modules/foreman_search_facts.py
+++ b/plugins/modules/foreman_search_facts.py
@@ -46,6 +46,11 @@ options:
       - Search query to use
       - If None, all resources are returned
     type: str
+  params:
+    description:
+      - Add parameters to the API call if necessary
+      - If not specified, no additional parameters are passed
+    type: dict
   full_details:
     description:
       - If C(True) all details about the found resources are returned
@@ -88,6 +93,17 @@ EXAMPLES = '''
   register: result
 - debug:
     var: result.resources
+
+- name: Get all existing subscriptions
+  foreman_search_facts:
+    username: "admin"
+    password: "changeme"
+    server_url: "https://foreman.example.com"
+    resource: subscriptions
+    params: "organization_id=1"
+      register: result
+- debug:
+    var: result
 '''
 
 RETURN = '''
@@ -107,16 +123,18 @@ def main():
             resource=dict(type='str', required=True),
             search=dict(default=""),
             full_details=dict(type='bool', aliases=['info'], default='false'),
+            params=dict(type='dict', default=dict()),
         ),
     )
 
     module_params = module.clean_params()
     resource = module_params['resource']
     search = module_params['search']
+    params = module_params['params']
 
     module.connect()
 
-    response = module.list_resource(resource, search)
+    response = module.list_resource(resource, search, params)
 
     if module_params['full_details']:
         resources = []
--- a/plugins/modules/foreman_search_facts.py
+++ b/plugins/modules/foreman_search_facts.py
@@ -100,8 +100,9 @@ EXAMPLES = '''
     password: "changeme"
     server_url: "https://foreman.example.com"
     resource: subscriptions
-    params: "organization_id=1"
-      register: result
+    params:
+      organization_id: 1
+  register: result
 - debug:
     var: result
 '''
@@ -123,14 +124,14 @@ def main():
             resource=dict(type='str', required=True),
             search=dict(default=""),
             full_details=dict(type='bool', aliases=['info'], default='false'),
-            params=dict(type='dict', default=dict()),
+            params=dict(type='dict'),
         ),
     )
 
     module_params = module.clean_params()
     resource = module_params['resource']
     search = module_params['search']
-    params = module_params['params']
+    params = module_params.get('params')
 
     module.connect()
 


plugins/modules/foreman_search_facts.py
```diff
@@ -88,6 +93,17 @@
...
     var: result.resources
+
+- name: Get all existing subscriptions
+  foreman_search_facts:
+    username: "admin"
+    password: "changeme"
+    server_url: "https://foreman.example.com"
+    resource: subscriptions
+    params: "organization_id=1"
+      register: result
```
ekohl: Indenting is off:
```suggestion
  register: result
```

plugins/modules/foreman_search_facts.py
```diff
@@ -88,6 +93,17 @@
...
 - debug:
     var: result.resources
+
+- name: Get all existing subscriptions
+  foreman_search_facts:
+    username: "admin"
+    password: "changeme"
+    server_url: "https://foreman.example.com"
+    resource: subscriptions
+    params: "organization_id=1"
```
evgeni:
```suggestion
    params:
      organization_id: 1
```

plugins/modules/foreman_search_facts.py
```diff
@@ -107,16 +123,18 @@ def main():
             resource=dict(type='str', required=True),
             search=dict(default=""),
             full_details=dict(type='bool', aliases=['info'], default='false'),
+            params=dict(type='dict', default=dict()),
```
evgeni:
```suggestion
            params=dict(type='dict'),
```

no need for a `default`, it will be `None` if not defined and `list_resource` will handle that fine

plugins/modules/foreman_search_facts.py
```diff
@@ -107,16 +123,18 @@ def main():
...
         ),
     )
 
     module_params = module.clean_params()
     resource = module_params['resource']
     search = module_params['search']
+    params = module_params['params']
```
evgeni:
```suggestion
    params = module_params.get('params')
```

that way it will really be None if the user didn't pass it.

plugins/modules/foreman_search_facts.py
m-czernek: IMHO, it's true that users typically don't know these low-level details, such as org_id or scope of a particular call (nor should they). However, I feel this might be useful in addition to adding other parameters. I.e. it would be nice to add organization, but I suspect (and correct me if I'm wrong here) that there will be other, call-specific params. 

For example, I see `scope = {'template_id': job_template['id']}` in `foreman_job_template`. I can't test the module at this particular moment, but I suspect that if we wanted to list the job templates, we'd have to pass a template ID..?

So, I'm all for specific parameters (like `organization`), but isn't it beneficial to have a direct way to pass params to modules? (for other modules)

m-czernek:
Incorporated requested changes. Please allow a bit of time for testing on my side.

plugins/modules/foreman_search_facts.py
evgeni: Oh yeah, that's what I meant by "Or probably both, as organization is something that will happen often, but maybe the user will need to search for something else that requires more input than just the org?"!

Let's add `organization` (and probably `location`) explicitly, but still allow *anything* via `params`.

@sean797 @mdellweg @bagasse what do you think?

sean797: @evgeni suggestion sounds sane to me! :+1:

bagasse: AS @evgeni stated, I think that we should not ask to users to know IDs of resources to use foreman ansible modules if there is a way to use name/title. But for this particular maybe managing mappings `X[s] -> X_id[s] -> find_resource_by_[name|title](Xs, ...)` for every possible relations will be a pain.

But as Katello resources are all scoped to an `organization`, :+1: to add it directly. For `location` OTOMH, I think there is no resources in foreman/katello that rely only on one location (only `locations` are used, but I can be wrong) so I don't think we should add a shortcut for `location`.

evgeni: While nothing *requires* location (as opposed to organization in Katello context), most entities can be scoped/assigned to a location, so if we look at `foreman_search_facts` as a generic "give me a list of entities, having the default scopes (org, loc) available sounds beneficial.

But as it's not *required*, I think we can add that in a different PR if anyone wants to.

m-czernek:
I would like to try to add the `organization` param. Currently, I have the following implementation:

```
if module_params.get('organization'):
    response = module.find_resource_by_name('organizations', module_params['organization'], thin=True)
    if response:
        params['organization_id'] = response['id']
```
We can talk about the code itself later, but my main worry is that:

- This works when the resource itself does require the organization_id
- This fails when the resource does not require organization_id

When the user provides organization and tries to list a resource which does not require a resource, I get:

```
The full traceback is:
WARNING: The below traceback may *not* be related to the actual failure.
  File "/tmp/ansible_theforeman.foreman.foreman_search_facts_payload_toe30h/ansible_theforeman.foreman.foreman_search_facts_payload.zip/ansible_collections/theforeman/foreman/plugins/module_utils/foreman_helper.py", line 44, in inner
    return f(self, *args, **kwargs)
  File "/tmp/ansible_theforeman.foreman.foreman_search_facts_payload_toe30h/ansible_theforeman.foreman.foreman_search_facts_payload.zip/ansible_collections/theforeman/foreman/plugins/module_utils/foreman_helper.py", line 270, in list_resource
  File "/usr/lib/python2.7/site-packages/apypie/resource.py", line 30, in call
    return self.api.call(self.name, action, params, headers, options, data, files)
  File "/usr/lib/python2.7/site-packages/apypie/api.py", line 158, in call
    return self._call_action(action, params, headers, data, files)
  File "/usr/lib/python2.7/site-packages/apypie/api.py", line 167, in _call_action
    headers, data, files)
  File "/usr/lib/python2.7/site-packages/apypie/api.py", line 193, in http_call
    request.raise_for_status()
  File "/usr/lib/python2.7/site-packages/requests/models.py", line 834, in raise_for_status
    raise HTTPError(http_error_msg, response=self)

fatal: [dh2.bpc.local -> localhost]: FAILED! => {
    "changed": false, 
    "error": {
        "message": "Association not found for organization"
    }, 
    "invocation": {
        "module_args": {
            "full_details": false, 
            "organization": "org_BPC", 
            "params": {
                "organization_id": 1, 
                "per_page": 4294967296, 
                "search": ""
            }, 
            "resource": "settings", 
            "search": "", 
            "server_url": "https://dh2.rh.local", 
            "username": "admin", 
            "validate_certs": false
        }
    }, 
    "msg": "Failed to list resource: 500 Server Error: Internal Server Error"
}
```

My question is, can we work around this? I.e. ideally, when the user inputs the explicit `organization` and tries to list a resource that does not require the `organization_id` parameter, the call would pass. Is this possible?

m-czernek:
I would like to try to add the `organization` param. Currently, I have the following implementation:

```
if module_params.get('organization'):
    response = module.find_resource_by_name('organizations', module_params['organization'], thin=True)
    if response:
        params['organization_id'] = response['id']
```
We can talk about the code itself later, but my main worry is that:

- This works when the resource itself does require the organization_id
- This fails when the resource does not require organization_id

When the user provides organization and tries to list a resource which does not require a resource, I get:

```
The full traceback is:
WARNING: The below traceback may *not* be related to the actual failure.
  File "/tmp/ansible_theforeman.foreman.foreman_search_facts_payload_toe30h/ansible_theforeman.foreman.foreman_search_facts_payload.zip/ansible_collections/theforeman/foreman/plugins/module_utils/foreman_helper.py", line 44, in inner
    return f(self, *args, **kwargs)
  File "/tmp/ansible_theforeman.foreman.foreman_search_facts_payload_toe30h/ansible_theforeman.foreman.foreman_search_facts_payload.zip/ansible_collections/theforeman/foreman/plugins/module_utils/foreman_helper.py", line 270, in list_resource
  File "/usr/lib/python2.7/site-packages/apypie/resource.py", line 30, in call
    return self.api.call(self.name, action, params, headers, options, data, files)
  File "/usr/lib/python2.7/site-packages/apypie/api.py", line 158, in call
    return self._call_action(action, params, headers, data, files)
  File "/usr/lib/python2.7/site-packages/apypie/api.py", line 167, in _call_action
    headers, data, files)
  File "/usr/lib/python2.7/site-packages/apypie/api.py", line 193, in http_call
    request.raise_for_status()
  File "/usr/lib/python2.7/site-packages/requests/models.py", line 834, in raise_for_status
    raise HTTPError(http_error_msg, response=self)

fatal: [dh2.rh.local -> localhost]: FAILED! => {
    "changed": false, 
    "error": {
        "message": "Association not found for organization"
    }, 
    "invocation": {
        "module_args": {
            "full_details": false, 
            "organization": "org_rh", 
            "params": {
                "organization_id": 1, 
                "per_page": 4294967296, 
                "search": ""
            }, 
            "resource": "settings", 
            "search": "", 
            "server_url": "https://dh2.rh.local", 
            "username": "admin", 
            "validate_certs": false
        }
    }, 
    "msg": "Failed to list resource: 500 Server Error: Internal Server Error"
}
```

My question is, can we work around this? I.e. ideally, when the user inputs the explicit `organization` and tries to list a resource that does not require the `organization_id` parameter, the call would pass. Is this possible?

mdellweg:
(Unrelated to your last question:) I believe, you did not insert the `"per_page": 4294967296` in the params in the playbook?
Can you please make a copy of the params and work on that. If you mess around with ansibles internal representations of the `module_params`, ansible reports strange things.

plugins/modules/foreman_search_facts.py
```diff
@@ -107,16 +124,18 @@ def main():
...
             search=dict(default=""),
             full_details=dict(type='bool', aliases=['info'], default='false'),
+            params=dict(type='dict'),
         ),
     )
 
     module_params = module.clean_params()
     resource = module_params['resource']
     search = module_params['search']
+    params = module_params.get('params')
```
mdellweg:
```suggestion
    params = module_params.get('params').copy()
```
The `module_params` are a filtered copy of ansibles internal data, but it is only a shallow copy. So this command retrieves another reference to internal data.

m-czernek: I have tried similar. The above-described behavior happens even when I hard-code the call as `response = module.list_resource(resource, search, {"organization_id": "1"} )` and when I'm not touching any params

mdellweg: I think it is the call in:
https://github.com/theforeman/foreman-ansible-modules/pull/537/files#diff-0424e051e61216ca71bbdfedba26401bR138

Could you try again with a copy of params?

m-czernek: For completeness sake:

```
    "error": {
        "message": "Association not found for organization"
    }, 
    "invocation": {
        "module_args": {
            "full_details": false, 
            "organization": "org_RH", 
            "params": {}, 
            "resource": "settings", 
            "search": "", 
            "server_url": "https://dh2.rh.local", 
            "username": "admin", 
            "validate_certs": false
        }
    }, 

```

mdellweg: See, the strangeness in module_args['params'] is gone, where it has been in your previous example:
```
"params": {
                "organization_id": 1, 
                "per_page": 4294967296, 
                "search": ""
            }
```

evgeni: can you try

```python
params = module.foremanapi.resource(resource).action('index').prepare_params(params)
```

after the `if organization` block? that should filter out params that the resource can't understand.

m-czernek:
Folks, I tried the suggestions, both the `.copy()` as well as `module.foremanapi.resource(resource).action('index').prepare_params(params)`, and couldn't figure it out. At this point, it's up to you. We can either keep this open and I can debug this as and when I have the time (time constrained by work for customers), or you can merge as is and finish the rest later.

It's up to you, but I felt there was some pressure to get this merged, so I don't want to hold you up. I am fairly certain there's a small mistake somewhere on my part, and that you'd be able to finish the `organization` param quite quickly if that's on your radar. If you don't mind the long delays, feel free to keep this open and we can continue slowly but surely our debugging :)

mdellweg:
Hi, sorry, i might have not been concise enough. The `copy` thing was not directly related to the problem you started the discussion on. It was something i saw in you debug output, and the `copy` suggestion solved that other problem (namely ansible reporting strange module_args), please keep that anyway.

PR #538, opened 2019-10-21
External Usergroups

Fodoj:
Adds support for external user groups.

The code can cause a lot of questions, so let me explain some of the things there:

1. There is no way to search by name, so I had to iterate through all ext. groups of give user group
2. I had to actually roll out local FreeIPA to make tests run, that's why I disable tls (self-signed cert) and updated domain for ldap auth source (I used vagrant.vm locally)
3. I didn't include support for locations and organizations because its not in normal usergroups neither.

Hope this clarifies things a bit.


plugins/modules/foreman_external_usergroup.py
```diff
@@ -0,0 +1,112 @@
...
+            usergroup=dict(required=True, type='str'),
+            auth_source=dict(required=True, type='entity', flat_name='auth_source_id'),
+        ),
+    )
+
+    entity_dict = module.clean_params()
+
+    module.connect()
+
+    params = {"usergroup_id": entity_dict.pop('usergroup')}
```
evgeni: This is interesting, as the param name is `usergroup_id`, but the API actually also allows the name attribute here.

plugins/modules/foreman_external_usergroup.py
```diff
@@ -0,0 +1,112 @@
...
+
+    # There is no way to find by name via API search, so we need
+    # to iterate over all external user groups of a given usergroup
+    for external_usergroup in module.list_resource("external_usergroups", params=params):
+        if external_usergroup['name'] == entity_dict['name']:
+            entity = external_usergroup
+
+    entity_dict['auth_source'] = module.find_resource_by_name('auth_sources', entity_dict['auth_source'], thin=True)
+
+    if not module.desired_absent and entity and entity['auth_source_ldap']['id'] == entity_dict['auth_source']['id']:
```
evgeni: Ewww. Would this line go away if we'd call the param `auth_source_ldap`? I *think* it would as we'd flatten both old and new entities to only have `auth_source_id` and then the comparison would work. We can still add `aliases=['auth_source']` to the params to make it easier for the users

tests/test_playbooks/external_usergroup.yml
```diff
@@ -0,0 +1,75 @@
...
+      auth_source_ldap_attr_login: dc=vagrant,dc=vm
+      auth_source_ldap_groups_base: cn=groups,cn=accounts,dc=vagrant,dc=vm
+      auth_source_ldap_tls: False
+
+- hosts: tests
+  gather_facts: false
+  vars_files:
+    - vars/server.yml
+  tasks:
+    - name: 'Test external usergroup with minimal params (name)'
```
evgeni: I think you can remove this `block` and have the includes directly, as we only have one test block anyways

plugins/modules/foreman_external_usergroup.py
evgeni: What I wanted to say here: I'm debating with myself whether I like this, or want to see a `find_resource_by_name` just to be consistent.

@bagasse @mdellweg opinions?

Fodoj: Well, saves one API call if we just use the name :-)

plugins/modules/foreman_external_usergroup.py
Fodoj: Let me try that, didn't know about aliases option.

plugins/modules/foreman_external_usergroup.py
evgeni: I know...

The thing is, I think it's a miss-feature of the API to accept the name here too (but it does this in a few places) and a part of me says "the field is called `_id`, so put in a damn id there".

And on the other hand, it saves a request. :woman_shrugging:

Fodoj: My guess is foreman uses friendly_id ruby on rails gem to treat some fields same as IDs. IMO if actual backend API  thinks it's okay to accept name as ID and it's described in API spec then API clients can do so as well. :)

evgeni: Yeah, Foreman does:
https://github.com/theforeman/foreman/blob/cd8d94440318e8ed4f8a943df5142b776bba5d62/app/models/external_usergroup.rb#L2-L3

I'm fine with this as it is I think.

plugins/modules/foreman_external_usergroup.py
Fodoj: I don't see how aliases are used in the code: https://github.com/theforeman/foreman-ansible-modules/blob/master/plugins/module_utils/foreman_helper.py#L626. Searching in this helper for aliases in general didnt return promising results. How do I use it? This didn't work:

```
auth_source=dict(required=True, aliases=['auth_source_ldap'],type='entity', flat_name='auth_source_id'),
```

=> same result, doesn't see changes

```
(Pdb) desired_entity
{'name': 'admins', 'auth_source_id': 10}
(Pdb) current_entity
{'id': 9, 'name': 'admins'}
```

evgeni: can you try

```python
auth_source_ldap=dict(required=True, aliases=['auth_source'],type='entity', flat_name='auth_source_id'),
```

Fodoj: Just realised what you proposed to do. Fixed now, I just renamed the field to auth_source_ldap. I didn't use an alias, because it's unclear how it will behave if there is some other type of external auth source (neither I am aware which ones would that be).

Fodoj:
Fixed comments as in review and squashed commits into one.

plugins/modules/foreman_external_usergroup.py
```diff
@@ -0,0 +1,108 @@
...
+
+
+ANSIBLE_METADATA = {'metadata_version': '1.1',
+                    'status': ['preview'],
+                    'supported_by': 'community'}
+
+DOCUMENTATION = '''
+---
+module: foreman_external_usergroup
+short_description: Manage Foreman User groups
```
evgeni: Maybe
```suggestion
short_description: Manage external user groups in Foreman
```

I kind of miss the `external` here otherwise :)

plugins/modules/foreman_external_usergroup.py
```diff
@@ -0,0 +1,108 @@
...
+    required: true
+    type: str
+  auth_source_ldap:
+    description:
+      - Name of the authentication source to be used for this group
+    required: true
+    type: str
+  state:
+    description:
+      - State of the user
```
evgeni:
```suggestion
      - State of the external user group
```

plugins/modules/foreman_external_usergroup.py
```diff
@@ -0,0 +1,108 @@
...
+RETURN = ''' # '''
+
+from ansible.module_utils.foreman_helper import ForemanEntityAnsibleModule
+
+
+def main():
+    module = ForemanEntityAnsibleModule(
+        entity_spec=dict(
+            name=dict(required=True),
+            usergroup=dict(required=True, type='str'),
```
evgeni:
```suggestion
            usergroup=dict(required=True),
```

`str` is default :)

Fodoj:
That "commit suggestion" feature is pretty neat..

evgeni:
> That "commit suggestion" feature is pretty neat..

Yes it is!

plugins/modules/foreman_external_usergroup.py
mdellweg: Fine for me too.

mdellweg:
> > That "commit suggestion" feature is pretty neat..
> 
> Yes it is!

_Some_ people started using that on their own prs...

plugins/modules/foreman_external_usergroup.py
bagasse: @evgeni for me it's ok, as it is documented in the API doc that it accept `ID or name of user group`, and :cherries:  on the cake it save 1 API call, so :+1:

PR #539, opened 2019-10-21
Update foreman_host.py

TTherouanne:
Hi,
Small fix to allow using of host groups with parents, this is working in my case

### DIFF
--- a/plugins/modules/foreman_host.py
+++ b/plugins/modules/foreman_host.py
@@ -165,7 +165,7 @@ def main():
 
     if not module.desired_absent:
         if 'hostgroup' in entity_dict:
-            entity_dict['hostgroup'] = module.find_resource_by_name('hostgroups', entity_dict['hostgroup'], thin=True)
+            entity_dict['hostgroup'] = module.find_resource_by_title('hostgroups', entity_dict['hostgroup'], thin=True)
 
         if 'location' in entity_dict:
             entity_dict['location'] = module.find_resource_by_title('locations', entity_dict['location'], thin=True)
--- a/tests/test_playbooks/fixtures/host-0.yml
+++ b/tests/test_playbooks/fixtures/host-0.yml
@@ -7,7 +7,7 @@ interactions:
       Connection: [keep-alive]
       User-Agent: ['apypie (https://github.com/Apipie/apypie)']
     method: GET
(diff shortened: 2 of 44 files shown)


Fodoj:
You need to re-record* tests for this:

```
make record_host
```

And for this you need to have a foreman instance running (check out https://github.com/theforeman/forklift). 

See also: https://github.com/theforeman/forklift/blob/master/docs/testing.md

* (Re-recording is basically saving responses from server to fixtures that are the used for running tests without making real connections to Foreman server, in case you wonder).

evgeni:
Thanks @TTherouanne! The change makes absolutely sense.

As @Fodoj pointed out, you'll have to re-record the test fixtures, as they contain recorded API requests and you changed how the hostgroup module uses the API.
I'd also would like to see a test that verifies that the module now works correctly with nested parent HGs. For that you'll have to modify `tests/test_playbooks/hostgroup.yml` and add a block like this

https://github.com/theforeman/foreman-ansible-modules/blob/f4e3cd817f33e4f705babe58dd5044978e973964/tests/test_playbooks/hostgroup.yml#L139-L175

but with `hostgroup_parent: "New host group/Nested New host group"` or similar.

Please reach out if you need further guidance, especially around re-recording, as it might be tricky at first.

mdellweg:
Thanks for this one. Big +1

TTherouanne:
Hello, I'm using the "centos7-foreman-nightly" vagrant box, when I run "make record_hostgroup", the playbook foreman-ansible-modules/tests/test_playbooks/tasks/domain.yml is included and fail because the smartproxy DNS feature is missing. ("does not have the DNS feature")
How do you enable the DNS feature with forklift ?

evgeni:
you can ssh to the vm (`vagrant ssh centos7-foreman-nightly`) and then enable it via `foreman-installer --foreman-proxy-dns --disable-system-checks`

TTherouanne:
Hi @evgeni 
ahah, I did not really understand because I did not find the link with the hostgroup module!
No worries, I will write the test for the hosts, this is not what takes the most time ;)

evgeni:
Hah, I was about to ask, what's with that `pause` in there :)

evgeni:
@TTherouanne thanks a ton! Merged. It's nice how a one-line code change triggered a 10k lines diff with all the tests :D

PR #543, opened 2019-10-22
Add present_with_defaults to sync plans

bagasse:
Add present_with_defaults to sync plan to be able to avoid 'change loops' if we want to manage `sync plan <-> products` association in products and not sync plans.

### DIFF
--- a/plugins/modules/katello_sync_plan.py
+++ b/plugins/modules/katello_sync_plan.py
@@ -82,11 +82,13 @@ options:
     type: list
   state:
     description:
-      - State of the sync plan
+      - State of the Sync plan
+      - C(present_with_defaults) will ensure the entity exists, but won't update existing ones
+    default: present
     choices:
       - present
       - absent
-    default: present
+      - present_with_defaults
     type: str
 extends_documentation_fragment: foreman
 '''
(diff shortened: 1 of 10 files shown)


plugins/modules/katello_sync_plan.py
```diff
@@ -122,6 +124,7 @@ def main():
             enabled=dict(type='bool', required=True),
             sync_date=dict(required=True),
             cron_expression=dict(),
+            state=dict(default='present', choices=['present_with_defaults', 'present', 'absent']),
```
evgeni: I like that the only thing needed to support that is the addition to the `state` dict.

tests/test_playbooks/sync_plan.yml
```diff
@@ -46,6 +46,15 @@
...
       sync_plan_cron_expression: "* * * * */5"
       expected_change: true
+  - include: tasks/product.yml
+    vars:
+      product_sync_plan: Test Sync Plan
+      product_state: present
+      expected_change: true
+  - include: tasks/sync_plan.yml
+    vars:
+      sync_plan_state: present_with_defaults
```
evgeni: shouldn't this test pass some parameter that won't be updated? like the product list or something?

plugins/modules/katello_sync_plan.py
bagasse: Yes ! I like this too !

tests/test_playbooks/sync_plan.yml
bagasse: Yes, you are right, and it permit me to find a bug. Changed.

PR #544, opened 2019-10-22
add basic tests for diff

evgeni:


evgeni:
seems ansible pre-2.9 doesn't like my test expressions, dammit.

evgeni:
I would add them to more as we go

Issue #546, opened 2019-10-22
fix fail_from_exception with katello


evgeni:
##### SUMMARY

it seems that errors from Katello look differently than the ones from Foreman:

```
{"displayMessage":"Missing values for content_view.","errors":["param is missing or the value is empty: content_view"]}",
```

there is no `error` in that response, and that breaks our logic in

https://github.com/theforeman/foreman-ansible-modules/blob/4a42495373d66867f003f708f68434a35c442c33/plugins/module_utils/foreman_helper.py#L528

##### ISSUE TYPE
 - Bug Report

##### ANSIBLE VERSION

```

```

##### KATELLO/FOREMAN VERSION

```

```

##### APYPIE VERSION

```

```

##### STEPS TO REPRODUCE

```yaml

```

##### ACTUAL RESULTS

```

```

PR #547, opened 2019-10-22
handle Katello-style errors in fail_from_exception

evgeni:
Fixes: #546

### DIFF
--- a/plugins/module_utils/foreman_helper.py
+++ b/plugins/module_utils/foreman_helper.py
@@ -525,7 +525,11 @@ class ForemanAnsibleModule(AnsibleModule):
         fail = {'msg': msg}
         if isinstance(exc, requests.exceptions.HTTPError):
             try:
-                fail['error'] = exc.response.json()['error']
+                response = exc.response.json()
+                if 'error' in response:
+                    fail['error'] = response['error']
+                else:
+                    fail['error'] = response
             except Exception:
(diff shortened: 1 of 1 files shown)


plugins/module_utils/foreman_helper.py
```diff
@@ -525,7 +525,11 @@ def fail_from_exception(self, exc, msg):
         fail = {'msg': msg}
         if isinstance(exc, requests.exceptions.HTTPError):
             try:
-                fail['error'] = exc.response.json()['error']
+                response = exc.response.json()
```
evgeni: this still can raise an exception and thus fuck up the traceback, improvements welcome

mdellweg: You mean when the response is not json deserializable?

evgeni: Yes, response.json will throw a valueerror

mdellweg: For the occasional 503?
But there is a `try` to catch that. I think we are fine.

evgeni: Oh, i wasn't verbose enough. Sorry.

Fail json will output a traceback if there is one (without us doing anything). Now if a ValueError happens here, the traceback will be for that and not from the exception we're handling.

evgeni:
merging as this is still better then before.

PR #548, opened 2019-10-23
Add satellite_config_dump role

jwbernin:
The satellite_config_dump role is intended to be used with a
future satellite_config_load role as a way to easily
replicate the configuration of a Satellite to a new Satellite.

Completing the role will require the merging of #537 - the current PR is being 
created to garner early feedback on the role in question.


evgeni:
cc @sthirugn

PR #549, opened 2019-10-23
Correct documentation for the operatingsystem module

m-czernek:
This PR changes:

* The incorrect documentation (`family: Red Hat` caused the module to fail with `422 Client Error: Unprocessable Entity`)
* Restricts the choices of OS system

Question for the experts here:

1. Is there an easy way to get the full list of OS systems? Seems like in the list I provided, some are missing
1. This module has parameter called `family`. The `foreman_ptable`  module has a parameter called `os_family`. Can't we standardize the param name?


m-czernek:
Re question 1., I found the answer in hammer:

```
hammer os create --family a --name delete --major 7 --minor 1
Could not create the operating system:
  Type must be one of [ AIX, Altlinux, Archlinux, Coreos, Debian, Freebsd, Gentoo, Junos, NXOS, Rancheros, Redhat, Solaris, Suse, Windows, Xenserver ]
```

plugins/modules/foreman_operatingsystem.py
```diff
@@ -190,7 +206,24 @@ def main():
...
+                'Gentoo',
+                'Junos',
+                'NXOS',
+                'Rancheros',
+                'Redhat',
+                'Solaris',
+                'Suse',
+                'Windows',
+                'Xenserver',
+                'Windows',
```
evgeni:
```suggestion
```

evgeni:
as for your second question, the difference comes from the API, which names the params differently (see https://www.theforeman.org/api/1.23/apidoc/v2/operatingsystems/create.html and https://www.theforeman.org/api/1.23/apidoc/v2/ptables/create.html)

Adding an `os_family` alias to the operatingsystem module should be fine. Adding `family` to the ptable one probably not, as the name somehow then would imply it's the family of the ptable, not the referenced OS.

plugins/modules/foreman_operatingsystem.py
```diff
@@ -190,7 +206,24 @@ def main():
             name=dict(),
             release_name=dict(),
             description=dict(),
-            family=dict(),
+            family=dict(choices=[
```
evgeni: we have the same list in `plugins/modules/foreman_installation_medium.py` already, so I wonder if we should define it in `foreman_helper` once and reference from here?

m-czernek: From my perspective, absolutely! Will do so, thank you for the suggestion :)

m-czernek:
foreman_utils now has a list with the OS values. I assume there's no way to resolve the OS list within the documentation string? I.e. something like `${OS_LIST}`?

evgeni:
Not easily, no. One could try fiddling around with a documentation fragment, but I'd leave that for later

m-czernek:
In that case I'm ok with the MR. I have also removed `VRP` from the installation media docs as it seems not to be supported by Satellite (a test was failing on that).

PR #552, opened 2019-10-24
Add netgroups option to auth_source_ldap

chsnell:
This PR adds all the code to support the Netgroups checkbox that's available in the web GUI, but not here in ansible.

### DIFF
--- a/docs/testing.md
+++ b/docs/testing.md
@@ -23,7 +23,7 @@ You can (re-)record the cassettes for a specific test with
 make record_global_parameter
 ```
 
-# Guidedeline to writing tests
+# Guideline to writing tests
 
 The tests in this repository run playbooks that can be found in `tests/test_playbooks`.
 To be run, the name of the corresponding playbook must be listed in `tests/test_crud.py`.
--- a/plugins/modules/foreman_auth_source_ldap.py
+++ b/plugins/modules/foreman_auth_source_ldap.py
@@ -96,9 +96,8 @@ options:
     required: false
     type: str
   use_netgroups:
-    description: Whether to use NIS netgroups instead of posix groups
+    description: Whether to use NIS netgroups instead of posix groups, not valid for I(server_type=active_directory)
     required: false
-    default: False
     type: bool
   server_type:
     description: Type of the LDAP server
@@ -185,11 +184,15 @@ def main():
             ldap_filter=dict(),
             locations=dict(type='entity_list', flat_name='location_ids'),
             organizations=dict(type='entity_list', flat_name='organization_ids'),
-            use_netgroups=dict(type='bool', default=False),
+            use_netgroups=dict(type='bool'),
         ),
     )
 
     entity_dict = module.clean_params()
+
+    # additional parameter checks
+    if 'use_netgroups' in entity_dict and entity_dict['server_type'] == 'active_directory':
+        module.fail_json(msg='use_netgroups cannot be used when server_type=active_directory')
 
     module.connect()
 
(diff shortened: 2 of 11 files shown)


chsnell:
I could use some help on the test portion, clearly.

plugins/modules/foreman_auth_source_ldap.py
```diff
@@ -180,6 +185,7 @@ def main():
             ldap_filter=dict(),
             locations=dict(type='entity_list', flat_name='location_ids'),
             organizations=dict(type='entity_list', flat_name='organization_ids'),
+            use_netgroups=dict(type='bool', default=False),
```
evgeni:
```suggestion
            use_netgroups=dict(type='bool'),
```

If you omit the `default`, the param will only be sent to the server when the user has explicitly set it, thus allowing the server to decide on a default.

Also, the APIdoc says "applicable only when server_type is posix or free_ipa", so we might want to add a guard somehow not to set that on AD connections.

chsnell: I'm not seeing an easy way to add the guard you suggested.  I need, I think, something close to the opposite of the required_if call.  Any suggestions here?

chsnell: Nevermind, I think I figured it out.

evgeni:
Thanks @chsnell! The code and the tests look correct to me!

I think the only missing piece now is to actually add the newly recorded test fixtures (those generated by `make record_auth_source_ldap`). Did you forget to `git add` the new files? Or did the recording not work for you? If it didn't, we can debug together, or I just record them on my install if you like to avoid more work ;)

chsnell:
@evgeni Thanks for looking again.  I can't manage to get a test harness setup properly at the moment, so if you could record the fixtures on your install that would be great.

evgeni:
Will do tomorrow!

Out of curiosity and to improve our setup/docs, what did you try and where did you fail for recording?

chsnell:
I can't even get a valid forklift setup at all.  My primary available machines are Windows and Mac, neither of which seems to have what's required for forklift, and the documentation isn't detailed enough for this seasoned sysadmin/neophyte developer to get it running on those platforms.

evgeni:
Oh. Mhh. I would have thought it just works on macOS, but never tried myself.

evgeni:
@chsnell I've force-pushed my changes to your fork :)

tests/test_playbooks/auth_source_ldap.yml
```diff
@@ -42,11 +42,13 @@
   - include: tasks/auth_source_ldap.yml
     vars:
       auth_source_ldap_state: present
+      auth_source_ldap_account_password: "{{ omit }}"
       auth_source_ldap_netgroups: true
       expected_change: true
   - include: tasks/auth_source_ldap.yml
     vars:
       auth_source_ldap_state: present
+      auth_source_ldap_account_password: "{{ omit }}"
```
mdellweg: Why not ` | default(omit)` in the tasks file?

evgeni: mostly because I was too lazy to touch the tasks file, but you're right.

docs/testing.md
```diff
@@ -23,7 +23,7 @@ You can (re-)record the cassettes for a specific test with
 make record_global_parameter
 ```
 
-# Guidedeline to writing tests
+# Guideline to writing tests
```
mdellweg: :+1:

chsnell:
Thanks for all the help.

PR #553, opened 2019-10-25
Foreman_subnet: add IPAM choice None

Boulet-:
Module: foreman_subnet
API Url: https://theforeman.org/api/1.23/apidoc/v2/subnets/create.html

This PR adds a new choice "None" for the IPAM option.

Let me know if I forgot something.

### DIFF
--- a/plugins/modules/foreman_subnet.py
+++ b/plugins/modules/foreman_subnet.py
@@ -95,7 +95,7 @@ options:
     description: IPAM mode for this subnet
     required: false
     default: DHCP
-    choices: ["DHCP","Internal DB"]
+    choices: ["DHCP","Internal DB", "None"]
     type: str
   dhcp_proxy:
     description: DHCP Smart proxy for this subnet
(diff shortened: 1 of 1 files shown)


mdellweg:
It would be nice, if you could add a test for this. I am wondering, whether `None` is really the literal string `"None"` or an unset parameter. (Looking at the apidoc it could even be both.) In the latter case, i think we need to pass a python `None` to the api in order to unset it.

evgeni:
I think it's a literal "None", but yeah, a test would be nice :)

evgeni:
Hi @Boulet-, thanks for your PR! I wanted to reach out and ask if you're planning to add tests for this feature or if you don't have enough time to dive into our tests. The later is absolutely fine, the change looks fine and given it just adds another type of IPAM, I'm OK with merging it without tests.

evgeni:
we discussed this during triage on Friday, and decided this does not require tests. merging.

Boulet-:
Sorry for the late answer, thanks a lot for the merge :)
I'll try to add tests on next MRs.

PR #555, opened 2019-10-25
allow certain resource_action calls in check_mode

evgeni:
some calls are read only, but we don't use the find/list wrappers
because the api is weird. let's allow these.

otherwise e.g. katello_activation_key is broken in check mode when
trying to assign content overrides

### DIFF
--- a/plugins/module_utils/foreman_helper.py
+++ b/plugins/module_utils/foreman_helper.py
@@ -499,13 +499,13 @@ class ForemanAnsibleModule(AnsibleModule):
 
         return changed, None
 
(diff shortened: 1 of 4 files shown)


evgeni:
cheers to @bbuckingham who made me use `k_ak` in check mode on a Friday :D

mdellweg:
If those calls are read only, they are probably not 'actions'. Should we supply a generic `resource_query` for those cases, which is a copy of `resource_action` without `check_mode` logic and a big disclamer in the docstring?

evgeni:
I used `always` and meant `always_run` as in "run this always, regardless of check_mode", but I can live with `ignore_cm` too if you insist ;)

PR #563, opened 2019-10-30
release 0.2.0

evgeni:


evgeni:
I know we have some outstanding fixes, but I wanted to get the current set to the users.

PR #567, opened 2019-10-31
modules that call resource_action can also record before/after state

evgeni:
this only changes the AK module, if the design is ok, I'll update the following modules too:

* katello_content_view
* katello_repository_set
* katello_sync_plan

### DIFF
--- a/plugins/modules/katello_activation_key.py
+++ b/plugins/modules/katello_activation_key.py
@@ -251,6 +251,10 @@ def main():
             current_subscription_ids = set(item['id'] for item in current_subscriptions)
 
             if desired_subscription_ids != current_subscription_ids:
+                module.record_before('activation_keys/subscriptions', {'id': activation_key['id'], 'subscriptions': current_subscription_ids})
+                module.record_after('activation_keys/subscriptions', {'id': activation_key['id'], 'subscriptions': desired_subscription_ids})
+                module.record_after_full('activation_keys/subscriptions', {'id': activation_key['id'], 'subscriptions': desired_subscription_ids})
+
                 ids_to_remove = current_subscription_ids - desired_subscription_ids
                 if ids_to_remove:
                     payload = {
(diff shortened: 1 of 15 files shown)


evgeni:
test failures unrelated, fix in https://github.com/theforeman/foreman-ansible-modules/pull/568

plugins/modules/katello_activation_key.py
```diff
@@ -251,6 +251,10 @@ def main():
             current_subscription_ids = set(item['id'] for item in current_subscriptions)
 
             if desired_subscription_ids != current_subscription_ids:
+                module.record_before('activation_keys/subscriptions', {'id': activation_key['id'], 'subscriptions': current_subscription_ids})
+                module.record_after('activation_keys/subscriptions', {'id': activation_key['id'], 'subscriptions': desired_subscription_ids})
+                module.record_after_full('activation_keys/subscriptions', {'id': activation_key['id'], 'subscriptions': desired_subscription_ids})
+
```
mdellweg: I would have expected to see the 'after' parts after the actions, but i guess we can just assume, that it worked properly if we don't get an error.

evgeni: yeah, that was the assumption I've been working under. can move if you prefer.

ehelms:
Seems like folks like it, are you gonna do the other resources in this PR?

evgeni:
@ehelms yeah, that was my plan yesterday, but then I got distracted

evgeni:
@ehelms @mdellweg this is now ready for review

mdellweg:
Do you want to add this functionality to the nested parameters in (domain, subnet, hostgroup, ...).

evgeni:
You mean those which use `ensure_scoped_parameters`? That should already work as that helper uses `ensure_entity`. I can add a test to verify.

mdellweg:
Perfect. Have i not acked this already?

PR #571, opened 2019-10-31
Test check_mode on foreman_subnet

mdellweg:


evgeni:
so I must admit I dislike that this blows up the fixtures by 2×

can we somehow re-use the stuff we recorded for the main tests? like instead of running it in `record_*` and `test_*` have a second `check_*` target that does the same as test but with check on?

evgeni:
the tests fail due to a bad matcher?

mdellweg:
The check_mode must be interleaved in the test_playbook, to test, whether it reports the right changed result.

evgeni:
given the issue we wanted to test turned out to be not an issue, shall we still persuade this PR?

mdellweg:
It was the matcher....

evgeni:
needs rebase

mdellweg:
I revoke my previous statement about interleaving.
When running the tests in replay mode, it is actually like independent tests for every task.
And while we can not run the playbook in `check_mode` against a real server, we _can_ actually do that in replay mode.
About to close this one in favour of #595 ...

Issue #570, opened 2019-10-31
add organization parameter to search_facts


evgeni:
##### SUMMARY
another thing that came up while talking to @wbclark is whether we want it in the generic way you implemented it which requires the user to know the Org ID (they usually don't) or add organization as a direct parameter as we do with the other (not search) modules.

Or probably both, as organization is something that will happen often, but maybe the user will need to search for something else that requires more input than just the org? What do you think?

from https://github.com/theforeman/foreman-ansible-modules/pull/537
##### ISSUE TYPE
 - Feature Idea

PR #573, opened 2019-10-31
search_facts: allow scoping by org directly

evgeni:
Fixes: #570

### DIFF
--- a/plugins/modules/foreman_search_facts.py
+++ b/plugins/modules/foreman_search_facts.py
@@ -51,12 +51,18 @@ options:
       - Add parameters to the API call if necessary
       - If not specified, no additional parameters are passed
     type: dict
+  organization:
+    description:
+      - Scope the searched resource by organization
+    type: str
   full_details:
     description:
       - If C(True) all details about the found resources are returned
     type: bool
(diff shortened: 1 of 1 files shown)


plugins/modules/foreman_search_facts.py
```diff
@@ -125,16 +142,24 @@ def main():
...
-    params = module_params.get('params')
+    params = module_params.get('params', {})
 
     module.connect()
 
+    if 'organization' in module_params:
+        params['organization_id'] = module.find_resource_by_name('organizations', module_params['organization'], thin=True)['id']
+
+    if params:
+        params = module.foremanapi.resource(resource).action('index').prepare_params(params)
```
bagasse: Can't we move this directly in `list_resource` ?

evgeni: so this will still fail, as it throws away `thin=*` for all resources that don't support it (most), but I think you're right and this doesn't belong into this pr.

evgeni: updated for not doing it in this PR, will open a second one.

@bagasse mind having another look?

PR #575, opened 2019-11-01
clean unknown params in list_resource and show_resource

evgeni:


plugins/module_utils/foreman_helper.py
```diff
@@ -270,6 +274,9 @@ def list_resource(self, resource, search=None, params=None):
             params = {}
         else:
             params = params.copy()
+
+        params = self.foremanapi.resource(resource).action('index').prepare_params(params)
```
evgeni: WIP because this cleaning should happen *after* we add search and per_page, but this breaks external_group and snapshot test which I didn't investigate yet

evgeni:
@bagasse that's how you meant it? :)

evgeni:
@mdellweg @Manisha15 do I read https://github.com/ATIX-AG/foreman_snapshot_management/blob/fbcb709492557d52fa915e79249706053444efef/app/controllers/api/v2/snapshots_controller.rb#L13-L17 right, the snapshot api does not support pagination and just returns ALL snapshots? (we already know that it doesn't support searching).

Manisha15:
> @mdellweg @Manisha15 do I read https://github.com/ATIX-AG/foreman_snapshot_management/blob/fbcb709492557d52fa915e79249706053444efef/app/controllers/api/v2/snapshots_controller.rb#L13-L17 right, the snapshot api does not support pagination and just returns ALL snapshots? (we already know that it doesn't support searching).

Yes, it doesn't support search (for particular snapshot) and pagination right now.

PR #580, opened 2019-11-07
Fixes #579 - Added support to manage foreman SCAP content.

### LINKED ISSUE #579: Add support for foreman SCAP content.
##### SUMMARY
There is no module for managing Foreman SCAP content.

jameerpathan111:
Fixes #579


jameerpathan111:
I have not added `data/ssg-firefox-ds.xml` yet(for tests), I wanted to know if it's ok to add any scap file in `tests/test_playbooks/data`.

mdellweg:
Thank you very much for this addition.
Things that immediately spring to my mind:
* Is it necessary to implement new find functions?
* Yay tests!
* Do you feel brave enough to add those tests to our ci and record test fixtures?

mdellweg:
> I have not added `data/ssg-firefox-ds.xml` yet(for tests), I wanted to know if it's ok to add any scap file in `tests/test_playbooks/data`.

If there's no legal issue with that file, that's the place for those files to go. For tests however i'd prefer a file that has the right format, but no real life applicability.

jameerpathan111:
@mdellweg I have tried to understand how other modules were written and followed same approach with `foreman_scap_contents` module. 
>     * Is it necessary to implement new find functions?
I had tried using `find_resource_by_title` and `find_resource_by_name` but they were not working for `scap_contents`, that's why I added `find_scapcontent`.
>     * Yay tests!
:)
>     * Do you feel brave enough to add those tests to our ci and record test fixtures?

I didn't get it. can you elaborate it?

mdellweg:
Have you found https://github.com/theforeman/foreman-ansible-modules/blob/master/docs/testing.md ?
Please read that first, if anything is unclear, don't hesitate to ask.

@evgeni we need a nop-slide to our write tests guide.

mdellweg:
@jameerpathan111 any update on this? Can we help you with the tests in any way? It's ok if you say you don't want to record them yourself. One of us will have a shot at it then.

jameerpathan111:
> @jameerpathan111 any update on this? Can we help you with the tests in any way? It's ok if you say you don't want to record them yourself. One of us will have a shot at it then.

@mdellweg yes, that would be great. :)
I will update PR. So I have to add scap content file in `tests/test_playbooks/data`, fix conflicts, fix mistakes present in tests, and list `tests/test_playbooks/scap_contents.yml` in `tests/test_crud.py`.

mdellweg:
And you need to record the vcr tapes by configuring your running foreman server in vars/server.yml and run make record_scap_content.

ogajduse:
Hey @jameerpathan111 I really appreciate your effort you put into creating this module. Can you please address the comments? Having this module in the next release would be awesome.

evgeni:
FWIW, we had a bit of refactoring since this PR was opened, so it will take a bit more then just addressing the old comments :(
If one of you is interested, we could do a remote hacking session and get this PR into a good state together?

jameerpathan111:
> Hey @jameerpathan111 I really appreciate your effort you put into creating this module. Can you please address the comments? Having this module in the next release would be awesome.

Was having problem installing foreman server :(

jameerpathan111:
> If one of you is interested, we could do a remote hacking session and get this PR into a good state together?

Great :)

evgeni:
Drop me a mail to evgeni@redhat.com and I'll see that we get an online session together to hack on this :)

evgeni:
updated the PR with the results of the hacking session
things I want to do before merging: more tests, rename to foreman_scap_content (without s)

plugins/modules/foreman_scap_contents.py
```diff
@@ -0,0 +1,140 @@
...
+  organizations:
+    description: List of organizations the SCAP content should be assigned to
+    required: false
+    type: list
+  state:
+    description: SCAP content presence
+    default: present
+    choices: ["present", "absent"]
+    type: str
+extends_documentation_fragment: foreman
```
mdellweg:
```suggestion
extends_documentation_fragment:
  - foreman
```

evgeni:
Okay, I think I have it in a state that I like now.

The only caveat: when you pass `scap_file`, the module becomes not idempotent, as `scap_file` is not part of a `show` api request. There *is* a `digest` field, so we could check if our local digest is different and only update in that case, but I'd leave that as an exercise to the reader.

evgeni:
Oh, and I opened https://projects.theforeman.org/issues/29403 for a weird DELETE behavior

evgeni:
Oh, this could benefit from `required_plugins`

evgeni:
Oh and the digest we see during show is a sha256, so we can compute that too in update and delete the scap content if it's identical

evgeni:
implemented both thoughts.

@jameerpathan111 @ogajduse UX question: while implementing the digest check, I stumbled over the fact that the `scap_file` param is not really a file, but actually a string with the XML content of the file (otherwise I could have used `module.sha256()` instead of using `hashlib` directly). This matches the API of the OpenSCAP plugin, but I wondered if we really need that, or if we could just let the user pass a normal path we'd open ourself instead of doing `{{ lookup('file', …) }}`?

plugins/modules/foreman_scap_content.py
```diff
@@ -130,6 +131,11 @@ def main():
             if not entity and 'scap_file' not in module_params:
                 module.fail_json(msg="Content of scap_file not provided. XML containing SCAP content is required.")
 
+            if entity and 'scap_file' in module_params:
+                digest = hashlib.sha256(module_params['scap_file'].encode("utf-8")).hexdigest()
```
mdellweg: Does `module.sha256` not work here?

evgeni: It expects a path, this is not a path, see my comment :)

mdellweg: Oh right. The notification took me directly to the diff...
Given that we'd probably don't want large amounts of xml embedded in yaml, I suspect passing it as a file seems to be the better UX.

jameerpathan111: yeah, passing it as a file seems to be the better UX.

evgeni: For some reason I fail to calculate the same checksum as Foreman is when I do it as a file, damn -.-

mdellweg: Is he stripping more Whitespace?

evgeni: Winner Winner, Chicken Dinner!

```diff
--- tests/test_playbooks/data/ssg-firefox-ds.xml	2020-03-25 13:55:54.952567896 +0100
+++ /home/evgeni/Downloads/ssg-firefox-ds.xml	2020-03-25 15:05:39.476260242 +0100
@@ -4012,4 +4012,4 @@
       </ns12:cpe-item>
     </ns12:cpe-list>
   </ns0:component>
-</ns0:data-stream-collection>
+</ns0:data-stream-collection>
\ No newline at end of file
```

ogajduse: Sure It is better UX.
Looking at the foreman_openscap plugin itself, I see that it is using SHA2 digest.
https://github.com/theforeman/foreman_openscap/blob/master/lib/foreman_openscap/bulk_upload.rb#L24

ogajduse: I must somehow miss your comment Evgeni. I guess you got it. :+1:

ogajduse:
This module seems to be very similar to the future `foreman_tailoring_file` module. As soon as this one would be finished/merged, I would use it as a template for the new module.

plugins/modules/foreman_scap_content.py
evgeni: Yeah, I think I got it working now.

evgeni:
@ogajduse I think we could even go further and make the tailoring file essentially inherit all the code by moving it into the helper :)

plugins/modules/foreman_scap_content.py
```diff
@@ -129,11 +133,20 @@ def main():
             if not entity and 'scap_file' not in module_params:
                 module.fail_json(msg="Content of scap_file not provided. XML containing SCAP content is required.")
 
+            if 'scap_file' in module_params:
+                with open(module_params['scap_file']) as input_file:
```
mdellweg: Shouldn't this be reading with "rb"?

evgeni: given it's xml, I can see r, but not b :)

evgeni:
Interesting that name vs title change...

I'll see how i can add it once this pr is reviewed

plugins/modules/foreman_scap_content.py
```diff
@@ -0,0 +1,154 @@
...
+  foreman_scap_content:
+    title: "Red Hat firefox default content"
+    updated_title: "Updated scap content title"
+    scap_file: "/home/user/Downloads/updated-ssg-firefox-ds.xml"
+    original_filename: "updated-ssg-firefox-ds.xml"
+    organizations:
+      - "Org One"
+      - "Org Two"
+    locations:
+        - "Loc Tne"
```
ogajduse:
```suggestion
        - "Loc One"
```

plugins/module_utils/foreman_helper.py
```diff
@@ -640,7 +640,7 @@ def _delete_entity(self, resource, current_entity, params):
         entity = self.resource_action(resource, 'destroy', payload)
 
         # this is a workaround for https://projects.theforeman.org/issues/26937
-        if entity and 'error' in entity and 'message' in entity['error']:
+        if entity and isinstance(entity, dict) and 'error' in entity and 'message' in entity['error']:
             self.fail_json(msg=entity['error']['message'])
```
mdellweg: The variable name `entity` feels kind of wrong here.

I know, that's not primarily the scope of this PR.

plugins/modules/foreman_scap_content.py
```diff
@@ -0,0 +1,154 @@
...
+    required: true
+    type: str
+  updated_title:
+    description:
+      - New SCAP content title.
+      - When this parameter is set, the module will not be idempotent.
+    type: str
+  scap_file:
+    description:
+      - XML containing SCAP content.
```
mdellweg:
```suggestion
      - File containing XML SCAP content.
```

plugins/modules/foreman_scap_content.py
mdellweg: you may not need the `.encode("utf-8")` below.

plugins/modules/foreman_scap_content.py
```diff
@@ -0,0 +1,154 @@
...
+
+            if entity and 'scap_file' in module_params:
+                digest = hashlib.sha256(module_params['scap_file'].encode("utf-8")).hexdigest()
+                # workaround for https://projects.theforeman.org/issues/29409
+                digest_stripped = hashlib.sha256(module_params['scap_file'].strip().encode("utf-8")).hexdigest()
+                if entity['digest'] in [digest, digest_stripped]:
+                    module_params.pop('scap_file')
+
+            if 'scap_file' in module_params and 'original_filename' not in module_params:
+                module_params['original_filename'] = os.path.basename(module_params['scap_file'])
```
mdellweg: At this point, `module_params['scap_file']` does not contain the filename anymore.

plugins/module_utils/foreman_helper.py
evgeni: Yeah, scope creep. The patch here is just for https://projects.theforeman.org/issues/29403

plugins/modules/foreman_scap_content.py
mdellweg: Also (and i know i get confused every time), this file needs to live on the machine where the module is executed, right?

plugins/module_utils/foreman_helper.py
mdellweg: :see_no_evil:

plugins/modules/foreman_scap_content.py
evgeni: Yes.

evgeni: `hashlib.sha256` needs a bytestream, not a string, but the API needs a string, not a bytestream.

So I pretend to read text, and push bytestream to hashlib. :boom:

evgeni:
cloud

for the test failures because ubuntu could not install packages

evgeni:
@mdellweg updated/commented

plugins/modules/foreman_scap_content.py
mdellweg: Convinced.

PR #582, opened 2019-11-08
Add puppetclasses for hostgroups

Fodoj:
A bit clumsy implementation of puppetclasses for hostgroups. According to API, to assign and remove classes from hostgroups there is a whole separate endpoint, thus extra complexity in managing them.


Fodoj:
Any feedback here? /cc @evgeni

Fodoj:
@mdellweg ?

evgeni:
thanks @Fodoj, this will require deeper review, and I don't have time for that at the moment.

if someone from @theforeman/ansible-modules wants to review, please do, otherwise I'll do so in about a week

plugins/modules/foreman_hostgroup.py
```diff
@@ -312,6 +317,12 @@ def main():
             updated_name=dict(),
         ),
     )
+
+    puppetclass_spec = dict(
+        hostgroup=dict(type='entity', flat_name='hostgroup_id'),
```
mdellweg: I think, this should not be in the entity_spec, but be provided as `scope` dictionary to the the `params` variable of `find*` and `ensure*`

plugins/modules/foreman_hostgroup.py
```diff
@@ -408,12 +419,36 @@ def main():
...
+    puppetclasses = entity_dict.pop('puppetclasses', None)
+    current_puppetclasses = entity.pop('puppetclasses', None) if entity else []
+
     changed, hostgroup = module.ensure_entity('hostgroups', entity_dict, entity)
 
     if hostgroup:
         scope = {'hostgroup_id': hostgroup['id']}
         changed |= module.ensure_scoped_parameters(scope, entity, parameters)
 
+    if not module.desired_absent and 'environment' in entity_dict:
```
mdellweg: Why is this not `'environment_id' in hostgroup`?

plugins/modules/foreman_hostgroup.py
```diff
@@ -408,12 +419,36 @@ def main():
...
+                scope = {'environment_id': hostgroup['environment_id']}
+                results = module.list_resource("puppetclasses", params=scope, search='name="{0}"'.format(puppet_class_name))
+                if len(results) == 1 and len(list(results.values())[0]) == 1:
+                    puppet_class_id = list(results.values())[0][0]['id']
+                    if puppet_class_id in current_puppetclasses:
+                        current_puppetclasses.remove(puppet_class_id)
+                    else:
+                        payload = {'hostgroup_id': hostgroup['id'], 'puppetclass_id': puppet_class_id}
+                        module.ensure_entity_state('hostgroup_classes', payload, None, None, 'present', puppetclass_spec)
+                        changed = True
```
mdellweg: changed handling has moved into foreman_helper. You might need to rebase.

Fodoj:
Yes, extra endpoint is required. API for updating housegroups doesnt accept puppetclasses (plot twist: web ui does support it).

plugins/modules/foreman_hostgroup.py
Fodoj: Because param name in this module is `environment` and there is `environment_id` in entity_dict, it appears only after parsing of params is done in foreman_helper

mdellweg: But you are using it as `hostgroup['environment_id']` in line 435. So i thought, this would be the natural check.
Or can it appear in `hostgroup` even if it was not specified in `entity_dict`? And in that case we do _not_ want to act?

Fodoj: We dont want to act is there is no environment, as then assigning classes is pointless.. But you are right, checking for hostgroup[env_id] feels safer in this case

Fodoj:
Pushed all fixes as per review. Also updated with master branch changes, code is much cleaner now, thanks @mdellweg

plugins/modules/foreman_hostgroup.py
```diff
@@ -408,12 +414,34 @@ def main():
...
+            for puppet_class_name in puppetclasses:
+                scope = {'environment_id': hostgroup['environment_id']}
+                results = module.list_resource("puppetclasses", params=scope, search='name="{0}"'.format(puppet_class_name))
+                if len(results) == 1 and len(list(results.values())[0]) == 1:
+                    puppet_class_id = list(results.values())[0][0]['id']
+                    if puppet_class_id in current_puppetclasses:
+                        current_puppetclasses.remove(puppet_class_id)
+                    else:
+                        payload = {'hostgroup_id': hostgroup['id'], 'puppetclass_id': puppet_class_id}
+                        module.ensure_entity('hostgroup_classes', None, None, params=payload, state='present')
```
mdellweg: Now, you just put all the information into the `payload`, which is probably ok, but you should still pass it an (now empty as in `{}`) entity_spec, because otherwise it will use the hostgroup_spec with unforeseeable results.
 The same holds true for line 443.
Btw, if you have seen this pattern elsewhere, we should review that also.

plugins/modules/foreman_hostgroup.py
```diff
@@ -408,12 +414,34 @@ def main():
...
         scope = {'hostgroup_id': hostgroup['id']}
         module.ensure_scoped_parameters(scope, entity, parameters)
 
+    if not module.desired_absent and 'environment_id' in hostgroup:
+        if puppetclasses is not None:
+            current_puppetclasses = [p['id'] for p in current_puppetclasses]
+            for puppet_class_name in puppetclasses:
+                scope = {'environment_id': hostgroup['environment_id']}
+                results = module.list_resource("puppetclasses", params=scope, search='name="{0}"'.format(puppet_class_name))
+                if len(results) == 1 and len(list(results.values())[0]) == 1:
```
mdellweg: This is a bit hazy. What do we see here on what level of the result? Maybe we should be explicit, by tearing it apart into separate apropriately named variables.

Fodoj:
@mdellweg I took hazy puppet classes part from config_group module. I did add some better var names here. As for entity_spec, I used the same approach here - https://github.com/theforeman/foreman-ansible-modules/pull/527. But it is much better without whole extra spec.

plugins/modules/foreman_hostgroup.py
```diff
@@ -428,14 +428,21 @@ def main():
...
-                    puppet_class_id = list(results.values())[0][0]['id']
+
+                # puppet classes API return puppet classes grouped by puppet module name
+                puppet_modules = module.list_resource("puppetclasses", params=scope, search='name="{0}"'.format(puppet_class_name))
+                puppet_classes = list(puppet_modules.values())[0]
+
+                # verify that only one puppet module is returned with only one puppet class inside
+                # as provided search results have to be like "results": { "ntp": [{"id": 1, "name": "ntp" ...}]}
+                # and get the puppet class id
+                if len(puppet_modules) == 1 and len(puppet_classes) == 1:
```
mdellweg: That's a lot more readable! But the `len(puppet_modules) == 1` test must be before `puppet_classes = list(puppet_modules.values())[0]` imho.
I think, before it worked due to lazy evaluation of boolean expressions.
I'd say it would be ok to first do the combined check, and then break out the `puppet_classes` in a variable.

plugins/modules/foreman_hostgroup.py
```diff
@@ -408,12 +414,41 @@ def main():
...
+                    if puppet_class_id in current_puppetclasses:
+                        current_puppetclasses.remove(puppet_class_id)
+                    else:
+                        payload = {'hostgroup_id': hostgroup['id'], 'puppetclass_id': puppet_class_id}
+                        module.ensure_entity('hostgroup_classes', None, None, params=payload, state='present', entity_spec={})
+                else:
+                    module.fail_json(msg='No data found for name="%s"' % puppet_class_name)
+            if len(current_puppetclasses) > 0:
+                for leftover_puppetclass in current_puppetclasses:
+                    module.ensure_entity('hostgroup_classes', None, {'id': leftover_puppetclass}, {'hostgroup_id': hostgroup['id']}, 'absent')
```
mdellweg: This is still missing the `entity_spec`.

mdellweg:
Have you taken a look at how the roles module uses the sub_entity_spec nowadays?
https://github.com/theforeman/foreman-ansible-modules/blob/master/plugins/modules/foreman_role.py#L114

plugins/modules/foreman_hostgroup.py
Fodoj: This line can fail if puppet class is not found - so it will fail a bit earlier with a less pretty error message. I moved the var assignment as you proposed.

Fodoj:
@mdellweg I haven't seen it, but looks cleaner indeed. I don't have capacity to do a bigger refactoring of this puppet classes at the moment though. :(

mdellweg:
Sure, i think we are already there.

mdellweg:
Code wise i'm satisfied. Let's see what CI says.

Fodoj:
I had to push one more fix for a runs in check mode. Now CI is happy.

mdellweg:
Lucky me, we got those check_mode tests in early enough. ;)

nesanton:
Hi @Fodoj,
I can't record_hostgroup because of the puppet classes missing. Seems like your tests are not really self-content. Or am I experiencing a PEBCAK?

Fodoj:
@nesanton they are not. But not just mine, tests in general are not self-content here - for example, you need to setup working LDAP server for some tests, as well computer resource for some other. Check this out, for example: https://github.com/theforeman/foreman-ansible-modules/pull/366#issuecomment-546374988

To test puppet classes you need to install Prometheus puppet module, then it should work.

evgeni:
while it is true that not all our tests can setup themself properly, I think we should *try* to make them so, or document it.

@Fodoj do I see it correctly, the user needs to do roughly the following:

```
# puppet module install puppet-prometheus
# hammer proxy import-classes --puppet-environment production --name $(hostname -f)
```

Fodoj:
Should be more or less like this

nesanton:
Thanks @Fodoj and @evgeni, works.
Good that we now have it written down

evgeni:
Awesome. @nesanton I guess you're currently working on something around this module/playbook? Could you add the above steps as a comment at the top of the playbook? So they don't get lost :)

nesanton:
Done in #618

Issue #534, opened 2019-10-21
foreman_role: filters are *always* removed/added


evgeni:
this will not work if you remove and add exactly the same number of (different) filters.

what is the exact reason you don't update existing filters?

_Originally posted by @evgeni in https://github.com/theforeman/foreman-ansible-modules/pull/527_

mdellweg:
Also it seems (at least in 1.22 and 1.24) that the `resource` field on the filters is unused.

PR #587, opened 2019-11-13
Record changed state on module

mdellweg:
I thought, when we are able to handle the complex diff states for the module writer in a common place, we should surely be able to do the change reporting for him, too.
But somehow this breaks the `foreman_role` module. I think the changed logic there is incomplete.

So i believe, this one is blocked by #534 

cc @evgeni 
cc @Fodoj


plugins/module_utils/foreman_helper.py
```diff
@@ -356,10 +365,6 @@ def record_after(self, resource, entity):
     def record_after_full(self, resource, entity):
         self._after_full[resource].append(entity)
 
-    def ensure_entity_state(self, *args, **kwargs):
-        changed, _entity = self.ensure_entity(*args, **kwargs)
-        return changed
-
```
mdellweg: I'm rather pleased to see this go...

plugins/modules/katello_manifest.py
```diff
@@ -118,26 +115,29 @@ def main():
...
-                changed, result = module.resource_action('subscriptions', 'upload', params, files=files)
+                # TODO: add a no_change field to resource_action
+                saved_changed = module.changed
+                result = module.resource_action('subscriptions', 'upload', params, files=files)
                 for error in result['humanized']['errors']:
                     if "same as existing data" in error:
-                        changed = False
+                        # We should not do this, but i see no better option
+                        # "Katello api is a sausage"
+                        module._changed = saved_changed
```
mdellweg: This is the only ugly hack left. I think, it is either this, or the aforementioned `no_change` flag in `resource_action`.

mdellweg: Added the `record_changed` flag.

evgeni:
@mdellweg please rebase

plugins/module_utils/foreman_helper.py
```diff
@@ -534,7 +540,11 @@ def resource_action(self, resource, action, params, options=None, data=None, fil
             msg = 'Error while performing {0} on {1}: {2}'.format(
                 action, resource, str(e))
             self.fail_from_exception(e, msg)
-        return True, result
+        if record_change:
+            if not ignore_check_mode:
```
evgeni: any reason not to fold these two `if`s into one?

plugins/module_utils/foreman_helper.py
```diff
@@ -165,10 +165,19 @@ def __init__(self, argument_spec, **kwargs):
         self.state = 'undefined'
         self.entity_spec = {}
 
+        # State recording for changed and diff reporting
+        self._changed = False
```
evgeni: is there a benefit of having a private variable, a property and a setter, instead of just exposing the var as `self.changed`?

plugins/module_utils/foreman_helper.py
mdellweg: Well, it took me too long to figure, that the task waiting was calling `resource_action` and thereby recording a change (which was only a problem for the katello_manifest module). So i started to loose trust in my boolean logic. Then after that problem was fixed a few lines down, i thought: Hmm, this might be more readable.
But yes, i should probably merge it again.

plugins/module_utils/foreman_helper.py
mdellweg: Yes. Glad that you ask. The setter prevents anyone from resetting it. I think, that goes well with the logic, that if something in the module resulted in a change, nothing else should be able to state the opposite.
And the setter, as well as the property is exposed to the actual modules without a risk.

evgeni: fair enough

plugins/module_utils/foreman_helper.py
evgeni: I do prefer one-line `if`s if possible ;)

PR #588, opened 2019-11-13
Add Scc_account module

Manisha15:
Add scc_account module for foreman_scc_manager plugin


plugins/modules/foreman_scc_account.py
```diff
@@ -0,0 +1,167 @@
...
+    entity_dict['organization'] = module.find_resource_by_name('organizations', entity_dict['organization'], thin=True)
+
+    scope = {'organization_id': entity_dict['organization']['id']}
+
+    entity = module.find_resource_by_name('scc_accounts', name=entity_dict['name'], params=scope, failsafe=True)
+
+    if 'updated_name' in entity_dict:
+        entity_dict['name'] = entity_dict['updated_name']
+
+    if module.state == 'sync':
```
mdellweg: Beeing ansible-desrciptive, should this be called `synced`?

plugins/modules/foreman_scc_account.py
```diff
@@ -0,0 +1,167 @@
...
+        result = module.foremanapi.resource('scc_accounts').call('sync', {'id': entity['id']})
+        if result:
+            changed = True
+    elif module.state == 'test_connection':
+        scc_account_credentials = {'login': entity_dict['login'], 'password': entity_dict['scc_account_password']}
+        result = module.foremanapi.resource('scc_accounts').call('test_connection', scc_account_credentials)
+        if result:
+            changed = True
```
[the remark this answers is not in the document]
Manisha15: But we use dummy scc_account for which test_connection will always fail. How are we going to test other methods then?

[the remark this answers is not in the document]
mdellweg: I think, we should not decrease user experience by limited testing opportunities.
I would be ok, if we ran the tests with `test_connection: false`.
Did i mention, that this parameter were a (boolean) flag?

plugins/modules/foreman_scc_account.py
```diff
@@ -0,0 +1,170 @@
...
+            updated_name=dict(),
+            login=dict(),
+            scc_account_password=dict(no_log=True, flat_name='password'),
+            base_url=dict(),
+            sync_date=dict(),
+            interval=dict(choices=['never', 'daily', 'weekly', 'monthly']),
+            state=dict(default='present', choices=['present', 'absent', 'synced']),
+        ),
+        argument_spec=dict(
+            test_connection=dict(type='bool', default=False))
```
mdellweg:
```suggestion
            test_connection=dict(type='bool', default=False),
        )
```

plugins/modules/foreman_scc_account.py
```diff
@@ -0,0 +1,170 @@
...
+    choices: ["never", "daily", "weekly", "monthly"]
+  sync_date:
+    description: Last Sync time of scc_account
+    required: false
+    type: str
+  organization:
+    description: Name of related organization
+    type: str
+  test_connection:
+    description: Test connection of scc_account
```
mdellweg: Can you be a bit more specific here?
We should say, that this will add a safeguard test to not accidentially change the accounts credentials to something that is not able to connect to the server.

plugins/modules/foreman_scc_account.py
```diff
@@ -0,0 +1,170 @@
...
+    entity = module.find_resource_by_name('scc_accounts', name=entity_dict['name'], params=scope, failsafe=True)
+
+    if entity_dict['test_connection']:
+        if not entity and not module.desired_absent:
+            scc_account_credentials = {'login': entity_dict['login'], 'password': entity_dict['scc_account_password'], 'base_url': entity_dict['base_url']}
+            connected = module.foremanapi.resource('scc_accounts').call('test_connection', scc_account_credentials)
+        elif module.desired_absent:
+            connected = True
+        else:
+            connected = module.foremanapi.resource('scc_accounts').call('test_connection', {'id': entity['id']})
```
mdellweg: This should test the existing account with overwritten credentials (as much as they are provided) to make sure, the state we want to put the account into can do the connection.

plugins/modules/foreman_scc_account.py
```diff
@@ -0,0 +1,171 @@
...
+    entity_dict['organization'] = module.find_resource_by_name('organizations', entity_dict['organization'], thin=True)
+
+    scope = {'organization_id': entity_dict['organization']['id']}
+
+    entity = module.find_resource_by_name('scc_accounts', name=entity_dict['name'], params=scope, failsafe=True)
+
+    if entity_dict['test_connection']:
+        scc_account_credentials = {'login': entity_dict['login'], 'password': entity_dict['scc_account_password'], 'base_url': entity_dict['base_url']}
+        if not entity and not module.desired_absent:
+            connected = module.foremanapi.resource('scc_accounts').call('test_connection', scc_account_credentials)
```
mdellweg: Can we not use a `module.resource_action` here? Probably with `ignore_check_mode`...

plugins/modules/foreman_scc_account.py
```diff
@@ -0,0 +1,171 @@
...
+            scc_account_credentials['id'] = entity['id']
+            connected = module.foremanapi.resource('scc_accounts').call('test_connection', scc_account_credentials)
+    else:
+        connected = True
+
+    if 'updated_name' in entity_dict:
+        entity_dict['name'] = entity_dict['updated_name']
+
+    if module.state == 'synced':
+        result = module.foremanapi.resource('scc_accounts').call('sync', {'id': entity['id']})
```
mdellweg: And also here, the `resource_action` should take care of the changed for us.

plugins/modules/foreman_scc_account.py
```diff
@@ -0,0 +1,171 @@
...
+        scc_account_credentials = {'login': entity_dict['login'], 'password': entity_dict['scc_account_password'], 'base_url': entity_dict['base_url']}
+        if not entity and not module.desired_absent:
+            connected = module.foremanapi.resource('scc_accounts').call('test_connection', scc_account_credentials)
+        elif module.desired_absent:
+            connected = True
+        else:
+            scc_account_credentials['id'] = entity['id']
+            connected = module.foremanapi.resource('scc_accounts').call('test_connection', scc_account_credentials)
+    else:
+        connected = True
```
mdellweg: Why do we carry the `connected` around here? Does the call (via resource_action) not fail for us in case the connection goes wrong?

tests/test_playbooks/scc_account.yml
```diff
@@ -0,0 +1,72 @@
+---
+- hosts: fixtures
+  gather_facts: false
+  vars_files:
+    - vars/server.yml
+  tasks:
+    - include: tasks/scc_account.yml
```
mdellweg: Can you add setup and cleanup tasks for the 'Defaut Organisation' (and maybe rename it to 'Test Organization')?

plugins/modules/foreman_scc_account.py
Manisha15: This is when user passes fake credentials but want to pass the tests. The user can set test_connection to false in playbook in that case.

mdellweg: What i wanted to say is, in case of a failed test_connection, it should exit with `fail_json` and the rest of the module should not run anymore.

tests/test_playbooks/scc_account.yml
```diff
@@ -0,0 +1,83 @@
...
+      scc_account_interval: 'monthly'
+      scc_account_state: present
+      expected_change: false
+
+  - include: tasks/scc_account.yml
+    vars:
+      scc_account_name: 'Test2'
+      test_connection: false
+      scc_account_state: synced
+      expected_change: true
```
mdellweg: This should be expected to fail, because we do not provide proper credentials.
The real problem can however be in the way we report the task in foreman_scc_manager.
I think, either we fix that one now, or we skip that test, until then.

plugins/modules/foreman_scc_account.py
```diff
@@ -0,0 +1,173 @@
...
+            if 'scc_account_password' in entity_dict:
+                scc_account_credentials['password'] = entity_dict['scc_account_password']
+            module.resource_action('scc_accounts', 'test_connection', scc_account_credentials, ignore_check_mode=True)
+
+    if 'updated_name' in entity_dict:
+        entity_dict['name'] = entity_dict['updated_name']
+
+    if module.state == 'synced':
+        task = module.resource_action('scc_accounts', 'sync', {'id': entity['id']})
+        if task['result'] == 'error':
```
evgeni: I *think* this should check for `task['result'] != 'success'` instead, to also catch `result=warning` etc.

see also https://github.com/theforeman/foreman-tasks/blob/904c07ca7f2be446e5453eeb271f9c4bc98628de/app/models/foreman_tasks/task/status_explicator.rb#L4-L8 for an alternative "definition" of "errored task"

evgeni: oh, and back in nailgun, we did https://github.com/SatelliteQE/nailgun/blob/master/nailgun/entity_mixins.py#L111-L116

plugins/module_utils/foreman_helper.py
```diff
@@ -622,7 +623,8 @@ def wait_for_task(self, task):
 
             resource_payload = self._resource_prepare_params('foreman_tasks', 'show', {'id': task['id']})
             task = self._resource_call('foreman_tasks', 'show', resource_payload)
-
+        if not ignore_errors and task['result'] != 'success':
+            self.fail_json(msg='Task {0} did not succeed. Task information: {1}'.format(task['action'], task['humanized']['errors']))
```
evgeni: When the task times out, we use the task ID in the error message, but I also see a value of printing the action. Would you mind updating both places to read something like "Task ACTION (ID)"?

plugins/modules/foreman_scc_account.py
```diff
@@ -0,0 +1,177 @@
...
+                    'status': ['preview'],
+                    'supported_by': 'community'}
+
+
+DOCUMENTATION = '''
+---
+module: foreman_scc_account
+short_description: Manage Foreman SccAccount
+description:
+  - "Manage Foreman SccAccount Entities"
```
evgeni: while the plugin *is* called scc_manager, do we want to at least once write out that it's the "suse customer center"?

plugins/modules/foreman_scc_account.py
```diff
@@ -0,0 +1,177 @@
...
+  - "This module requires the foreman_scc_manager plugin set up in the server"
+  - "See: U(https://github.com/ATIX-AG/foreman_scc_manager)"
+author:
+  - "Manisha Singhal (@manisha15) ATIX AG"
+options:
+  name:
+    description: Name of the scc_account
```
evgeni: actually, it might be better readable writing "suse customer center in these descriptions too"

plugins/modules/foreman_scc_account.py
```diff
@@ -0,0 +1,177 @@
...
+
+
+def main():
+    module = KatelloEntityAnsibleModule(
+        entity_spec=dict(
+            name=dict(required=True),
+            updated_name=dict(),
+            login=dict(),
+            scc_account_password=dict(no_log=True, flat_name='password'),
+            base_url=dict(),
```
evgeni: should this default to `https://scc.suse.com`?

plugins/modules/foreman_scc_account.py
```diff
@@ -0,0 +1,177 @@
...
+            sync_date=dict(),
+            interval=dict(choices=['never', 'daily', 'weekly', 'monthly']),
+            state=dict(default='present', choices=['present', 'absent', 'synced']),
+        ),
+        argument_spec=dict(
+            test_connection=dict(type='bool', default=False)),
```
evgeni:
```suggestion
            test_connection=dict(type='bool', default=False),
        ),
```

plugins/modules/foreman_scc_account.py
```diff
@@ -0,0 +1,177 @@
...
+                scc_account_credentials['password'] = entity_dict['scc_account_password']
+            if 'base_url' in entity_dict:
+                scc_account_credentials['base_url'] = entity_dict['base_url']
+            module.resource_action('scc_accounts', 'test_connection', scc_account_credentials, ignore_check_mode=True)
+
+    if 'updated_name' in entity_dict:
+        entity_dict['name'] = entity_dict['updated_name']
+
+    if module.state == 'synced':
+        module.resource_action('scc_accounts', 'sync', {'id': entity['id']})
```
evgeni: this will fail if `entity` is `None`, which it can be as you searched with `failsafe=True` above.
```suggestion
        if entity:
            module.resource_action('scc_accounts', 'sync', {'id': entity['id']})
        else:
            module.fail_json(msg="Could not find SCC account {0} to sync".format(entity_dict['name']))
```

tests/test_playbooks/scc_account.yml
```diff
@@ -0,0 +1,104 @@
...
+    vars:
+      scc_account_name: 'Test'
+      organization_name: 'Test Organization'
+      scc_account_state: absent
+  - include: tasks/scc_account.yml
+    vars:
+      scc_account_name: 'Test2'
+      organization_name: 'Test Organization'
+      scc_account_state: absent
+- hosts: tests
```
evgeni:
```suggestion

- hosts: tests
```

tests/test_playbooks/tasks/scc_account.yml
```diff
@@ -0,0 +1,34 @@
...
+- assert:
+    fail_msg: "Ensuring scc_account is {{ scc_account_state }} failed! (expected_change: {{ expected_change }})"
+    that:
+      - result.changed == expected_change
+  when: expected_change is defined
+- assert:
+    fail_msg: "Ensuring scc_account failed as expected!"
+    that:
+      - result is failed
+  when: expected_fail is defined and expected_fail == true
```
evgeni:
```suggestion
  when:
    - expected_fail is defined
    - expected_fail == true
```

evgeni: should this be `== true` or `== True` or not have a `==` at all and rely on it being a boolean?

tests/test_playbooks/scc_account.yml
```diff
@@ -0,0 +1,104 @@
...
+      scc_account_state: absent
+      expected_change: true
+
+  - include: tasks/scc_account.yml
+    vars:
+      scc_account_name: 'Test2'
+      organization_name: 'Test Organization'
+      scc_account_state: absent
+      expected_change: false
+- hosts: localhost
```
evgeni:
```suggestion

- hosts: localhost
```

plugins/modules/foreman_scc_account.py
```diff
@@ -0,0 +1,177 @@
...
+            if 'login' in entity_dict:
+                scc_account_credentials['login'] = entity_dict['login']
+            if 'scc_account_password' in entity_dict:
+                scc_account_credentials['password'] = entity_dict['scc_account_password']
+            if 'base_url' in entity_dict:
+                scc_account_credentials['base_url'] = entity_dict['base_url']
+            module.resource_action('scc_accounts', 'test_connection', scc_account_credentials, ignore_check_mode=True)
+
+    if 'updated_name' in entity_dict:
+        entity_dict['name'] = entity_dict['updated_name']
```
evgeni:
```suggestion
        entity_dict['name'] = entity_dict.pop('updated_name')
```

we don't need the updated_name entry afterwards

plugins/modules/foreman_scc_account.py
mdellweg: That's the default on the server side, and providing a default here would reset that value if unspecified.

evgeni: ah, good!

plugins/modules/foreman_scc_account.py
mdellweg: Sorry for being late to the party. But i just realized, we can use `failsafe= (module.state != 'synched')`.
I'm not going to block on this one.

evgeni:
do you want to have all the commits squashed, or is there anything in the history you'd like to preserve?

plugins/modules/foreman_scc_account.py
evgeni: Agreed, feel free to add that. ACK either way holds.

Manisha15:
> do you want to have all the commits squashed, or is there anything in the history you'd like to preserve?

I think squashing all commits is fine.

PR #589, opened 2019-11-13
Fix role filter analysis

mdellweg:
Fixes: #534

### DIFF
--- a/plugins/modules/foreman_role.py
+++ b/plugins/modules/foreman_role.py
@@ -54,6 +54,16 @@ options:
     description: Filters with permissions for this role
     required: false
     type: list
+    elements: dict
+    suboptions:
+      permissions:
+        description: List of permissions
+        required: true
+        type: list
+      search:
+        description: Filter condition for the resources
+        required: false
+        type: str
   state:
     description: role presence
     default: present
@@ -72,8 +82,7 @@ EXAMPLES = '''
     organizations:
       - "Basalt"
     filters:
-      - resource: 'Host'
-        permissions:
+      - permissions:
           - view_hosts
         search: "owner_type = Usergroup and owner_id = 4"
     server_url: "https://foreman.example.com"
@@ -84,7 +93,15 @@ EXAMPLES = '''
 
 RETURN = ''' # '''
 
+import copy
+
 from ansible.module_utils.foreman_helper import ForemanEntityAnsibleModule
+
+
+filter_entity_spec = dict(
+    permissions=dict(type='entity_list', flat_name='permission_ids', required=True),
+    search=dict(),
+)
 
 
 def main():
@@ -94,15 +111,8 @@ def main():
             description=dict(),
             locations=dict(type='entity_list', flat_name='location_ids'),
             organizations=dict(type='entity_list', flat_name='organization_ids'),
-            filters=dict(type='entity_list', flat_name='filter_ids'),
+            filters=dict(type='nested_list', entity_spec=filter_entity_spec),
         ),
-    )
-
-    filters_entity_spec = dict(
-        permissions=dict(type='entity_list', flat_name='permission_ids'),
-        resource=dict(),
-        search=dict(),
-        role_id=dict(required=True)
     )
 
     entity_dict = module.clean_params()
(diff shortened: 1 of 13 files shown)


plugins/modules/foreman_role.py
```diff
@@ -120,19 +128,29 @@ def main():
...
-            for role_filter in filters:
-                role_filter['role_id'] = entity['id']
-                role_filter['permissions'] = module.find_resources_by_name('permissions', role_filter['permissions'], thin=True)
-                module.ensure_entity_state('filters', role_filter, None, None, 'present', filters_entity_spec)
-            old_filters = entity.get('filter_ids', [])
-            for old_filter in old_filters:
-                module.ensure_entity('filters', None, {'id': old_filter}, {}, 'absent')
-            if len(old_filters) != len(filters):
-                changed = True
+    changed, new_entity = module.ensure_entity('roles', entity_dict, entity)
```
evgeni: I find `entity` vs `new_entity` confusing and had to read multiple times until I understood why you use `entity` below again.

plugins/modules/foreman_role.py
```diff
@@ -120,19 +128,29 @@ def main():
...
+
+    if not module.desired_absent and filters is not None:
+        scope = {'role_id': new_entity['id']}
+
+        if entity:
+            current_filters = [module.show_resource('filters', filter['id']) for filter in entity['filters']]
+        else:
+            current_filters = []
+        desired_filters = [filter.copy() for filter in filters]
```
evgeni: `copy.deepcopy(filters)`?

plugins/modules/foreman_role.py
```diff
@@ -72,8 +82,7 @@
     organizations:
       - "Basalt"
     filters:
-      - resource: 'Host'
```
evgeni: This is "interesting". The entry wasn't used before, but it could. When listing permissions, you can pass a `resource_type`. But then again, permission names are unique and there is not really a point in doing so (unless for a nicer grouping/listing in an UI, which we don't have).

plugins/modules/foreman_role.py
```diff
@@ -120,19 +128,29 @@ def main():
...
+        if entity:
+            current_filters = [module.show_resource('filters', filter['id']) for filter in entity['filters']]
+        else:
+            current_filters = []
+        desired_filters = [filter.copy() for filter in filters]
+
+        for desired_filter in desired_filters:
+            # search for an existing filter
+            for current_filter in current_filters:
+                if desired_filter['search'] == current_filter['search']:
```
evgeni: `search` is optional, so `desired_filter` might not have a `search` entry at all? or will it be `None` because we don't filter sub-opts on `clean_params`?

maybe add a test that does not use `search`? this makes the filter "unlimited" in UI speak.

plugins/modules/foreman_role.py
```diff
@@ -54,6 +54,16 @@
     description: Filters with permissions for this role
     required: false
     type: list
+    elements: dict
+    suboptions:
+      permissions:
+        description: List of permissions
+        required: false
```
evgeni: I kind of think that adding a filter w/o permissions is pointless, but I am not sure if setting `required=True` would make it required *always* or just when filters are defined.

plugins/modules/foreman_role.py
```diff
@@ -54,6 +54,16 @@
     description: Filters with permissions for this role
     required: false
     type: list
+    elements: dict
+    suboptions:
```
evgeni: the API also supports an [`override`](https://theforeman.org/api/1.23/apidoc/v2/filters/create.html) param, should this be added too (or in an own PR?)

plugins/modules/foreman_role.py
mdellweg: That's about the level of copying i need, but then again, there will probably be no more levels.

plugins/modules/foreman_role.py
mdellweg: It's listed as optional in the apidoc.
But setting it required here still allows the whole list of subitems to be empty.

plugins/modules/foreman_role.py
mdellweg: As this one is more like a bugfix PR, i'd put that in a separate feature PR.

plugins/modules/foreman_role.py
evgeni: Nobody said the apidoc is always right ;-)

plugins/modules/foreman_role.py
mdellweg: Right. It's confusing. But new_entity is the one with an id, and the old entity can tell us whether we started out blank.

plugins/modules/foreman_role.py
mdellweg: I'm assuming this means, ACK to this removal. Resolving that conversation.

plugins/modules/foreman_role.py
mdellweg: Are you saying it should be required?

evgeni: I would prefer that, but not block.

You can create filters with no permissions in the UI too. They are just pointless.

plugins/modules/foreman_role.py
evgeni: Blank? Don't we just need the list of filters (if any) at this point?

plugins/modules/foreman_role.py
mdellweg: Yes, you can still provide an empty list now. Which is as pointless, but at least much harder...

plugins/modules/foreman_role.py
mdellweg: It's in line 138, where we check, whether the entity was newly created before.

evgeni: And `new_entity` doesn't contain `filters` when it's actually an updated entity, so we can't just do `current_filters = [module.show_resource('filters', filter['id']) for filter in new_entity['filters']]`?

mdellweg: I think, it was something like that. It's similar with nested parameters.

PR #591, opened 2019-11-13
Add inventory plugin from ansible/ansible to collection

willtome:
Relocating foreman/satellite inventory plugin from ansible/ansible to collection

### DIFF
--- a/plugins/inventory/foreman.py
+++ b/plugins/inventory/foreman.py
@@ -63,7 +63,7 @@ DOCUMENTATION = '''
 
 EXAMPLES = '''
 # my.foreman.yml
-plugin: foreman
+plugin: theforeman.foreman.foreman
 url: http://localhost:2222
 user: ansible-tester
 password: secure
@@ -91,7 +91,7 @@ from requests.auth import HTTPBasicAuth
 class InventoryModule(BaseInventoryPlugin, Cacheable, Constructable):
     ''' Host inventory parser for ansible using foreman as source. '''
 
-    NAME = 'foreman'
+    NAME = 'theforeman.foreman.foreman'
 
     def __init__(self):
 
--- a/plugins/inventory/foreman.py
+++ b/plugins/inventory/foreman.py
@@ -59,6 +59,11 @@ DOCUMENTATION = '''
         description: Toggle, if true the inventory will retrieve 'all_parameters' information as host vars
         type: boolean
         default: False
+      want_hostcollections:
+        description: Toggle, if true the plugin will create Ansible groups for host collections
+        type: boolean
+        default: False
+        version_added: '2.10'
       legacy_hostvars:
         description:
             - Toggle, if true the plugin will build legacy hostvars present in the foreman script
@@ -188,6 +193,10 @@ class InventoryModule(BaseInventoryPlugin, Cacheable, Constructable):
         url = "%s/api/v2/hosts/%s/facts" % (self.foreman_url, hid)
         return self._get_json(url)
 
+    def _get_host_data_by_id(self, hid):
+        url = "%s/api/v2/hosts/%s" % (self.foreman_url, hid)
+        return self._get_json(url)
+
     def _get_facts(self, host):
         """Fetch all host facts of the host"""
 
@@ -254,7 +263,22 @@ class InventoryModule(BaseInventoryPlugin, Cacheable, Constructable):
 
                 # set host vars from facts
                 if self.get_option('want_facts'):
-                    self.inventory.set_variable(host_name, 'ansible_facts', self._get_facts(host))
+                    self.inventory.set_variable(host_name, 'foreman_facts', self._get_facts(host))
+
+                # create group for host collections
+                if self.get_option('want_hostcollections'):
+                    host_data = self._get_host_data_by_id(host['id'])
+                    hostcollections = host_data.get('host_collections')
+                    if hostcollections:
+                        # Create Ansible groups for host collections
+                        for hostcollection in hostcollections:
+                            try:
+                                hostcollection_group = to_safe_group_name('%shostcollection_%s' % (self.get_option('group_prefix'),
+                                                                          hostcollection['name'].lower().replace(" ", "")))
+                                hostcollection_group = self.inventory.add_group(hostcollection_group)
+                                self.inventory.add_child(hostcollection_group, host_name)
+                            except ValueError as e:
+                                self.display.warning("Could not create groups for host collections for %s, skipping: %s" % (host_name, to_text(e)))
 
                 strict = self.get_option('strict')
 
--- a/plugins/inventory/foreman.py
+++ b/plugins/inventory/foreman.py
@@ -59,6 +59,13 @@ DOCUMENTATION = '''
         description: Toggle, if true the inventory will retrieve 'all_parameters' information as host vars
         type: boolean
         default: False
+      legacy_hostvars:
+        description:
+            - Toggle, if true the plugin will build legacy hostvars present in the foreman script
+            - Places hostvars in a dictionary with keys `foreman`, `foreman_facts`, and `foreman_params`
+        type: boolean
+        default: False
+        version_added: '2.10'
 '''
 
 EXAMPLES = '''
@@ -193,6 +200,13 @@ class InventoryModule(BaseInventoryPlugin, Cacheable, Constructable):
             raise ValueError("More than one set of facts returned for '%s'" % host)
         return facts
 
+    def _get_hostvars(self, host, vars_prefix='', omitted_vars=()):
+        hostvars = {}
+        for k, v in host.items():
+            if k not in omitted_vars:
+                hostvars[vars_prefix + k] = v
+        return hostvars
+
     def _populate(self):
 
         for host in self._get_hosts():
@@ -207,25 +221,36 @@ class InventoryModule(BaseInventoryPlugin, Cacheable, Constructable):
                     group_name = self.inventory.add_group(group_name)
                     self.inventory.add_child(group_name, host_name)
 
-                # set host vars from host info
-                try:
-                    for k, v in host.items():
-                        if k not in ('name', 'hostgroup_title', 'hostgroup_name'):
-                            try:
-                                self.inventory.set_variable(host_name, self.get_option('vars_prefix') + k, v)
-                            except ValueError as e:
-                                self.display.warning("Could not set host info hostvar for %s, skipping %s: %s" % (host, k, to_text(e)))
-                except ValueError as e:
-                    self.display.warning("Could not get host info for %s, skipping: %s" % (host_name, to_text(e)))
+                if self.get_option('legacy_hostvars'):
+                    hostvars = self._get_hostvars(host)
+                    self.inventory.set_variable(host_name, 'foreman', hostvars)
+                else:
+                    omitted_vars = ('name', 'hostgroup_title', 'hostgroup_name')
+                    hostvars = self._get_hostvars(host, self.get_option('vars_prefix'), omitted_vars)
+
+                    for k, v in hostvars.items():
+                        try:
+                            self.inventory.set_variable(host_name, k, v)
+                        except ValueError as e:
+                            self.display.warning("Could not set host info hostvar for %s, skipping %s: %s" % (host, k, to_text(e)))
 
                 # set host vars from params
                 if self.get_option('want_params'):
-                    for p in self._get_all_params_by_id(host['id']):
-                        try:
-                            self.inventory.set_variable(host_name, p['name'], p['value'])
-                        except ValueError as e:
-                            self.display.warning("Could not set hostvar %s to '%s' for the '%s' host, skipping:  %s" %
-                                                 (p['name'], to_native(p['value']), host, to_native(e)))
+                    params = self._get_all_params_by_id(host['id'])
+                    filtered_params = {}
+                    for p in params:
+                        if 'name' in p and 'value' in p:
+                            filtered_params[p['name']] = p['value']
+
+                    if self.get_option('legacy_hostvars'):
+                        self.inventory.set_variable(host_name, 'foreman_params', filtered_params)
+                    else:
+                        for k, v in filtered_params:
+                            try:
+                                self.inventory.set_variable(host_name, p['name'], p['value'])
+                            except ValueError as e:
+                                self.display.warning("Could not set hostvar %s to '%s' for the '%s' host, skipping:  %s" %
+                                                     (p['name'], to_native(p['value']), host, to_native(e)))
 
                 # set host vars from facts
                 if self.get_option('want_facts'):
--- a/plugins/inventory/foreman.py
+++ b/plugins/inventory/foreman.py
@@ -1,0 +1,254 @@
+# -*- coding: utf-8 -*-
+# Copyright (C) 2016 Guido Günther <agx@sigxcpu.org>, Daniel Lobato Garcia <dlobatog@redhat.com>
+# Copyright (c) 2018 Ansible Project
+# GNU General Public License v3.0+ (see COPYING or https://www.gnu.org/licenses/gpl-3.0.txt)
+from __future__ import (absolute_import, division, print_function)
+__metaclass__ = type
+
+DOCUMENTATION = '''
+    name: foreman
+    plugin_type: inventory
+    short_description: foreman inventory source
+    version_added: "2.6"
+    requirements:
+        - requests >= 1.1
+    description:
+        - Get inventory hosts from the foreman service.
+        - "Uses a configuration file as an inventory source, it must end in ``.foreman.yml`` or ``.foreman.yaml`` and has a ``plugin: foreman`` entry."
+    extends_documentation_fragment:
+        - inventory_cache
+        - constructed
+    options:
+      plugin:
+        description: the name of this plugin, it should always be set to 'foreman' for this plugin to recognize it as it's own.
+        required: True
+        choices: ['foreman']
+      url:
+        description: url to foreman
+        default: 'http://localhost:3000'
+        env:
+            - name: FOREMAN_SERVER
+              version_added: "2.8"
+      user:
+        description: foreman authentication user
+        required: True
+        env:
+            - name: FOREMAN_USER
+              version_added: "2.8"
+      password:
+        description: foreman authentication password
+        required: True
+        env:
+            - name: FOREMAN_PASSWORD
+              version_added: "2.8"
+      validate_certs:
+        description: verify SSL certificate if using https
+        type: boolean
+        default: False
+      group_prefix:
+        description: prefix to apply to foreman groups
+        default: foreman_
+      vars_prefix:
+        description: prefix to apply to host variables, does not include facts nor params
+        default: foreman_
+      want_facts:
+        description: Toggle, if True the plugin will retrieve host facts from the server
+        type: boolean
+        default: False
+      want_params:
+        description: Toggle, if true the inventory will retrieve 'all_parameters' information as host vars
+        type: boolean
+        default: False
+'''
+
+EXAMPLES = '''
+# my.foreman.yml
+plugin: foreman
+url: http://localhost:2222
+user: ansible-tester
+password: secure
+validate_certs: False
+'''
+
+from distutils.version import LooseVersion
+
+from ansible.errors import AnsibleError
+from ansible.module_utils._text import to_bytes, to_native, to_text
+from ansible.module_utils.common._collections_compat import MutableMapping
+from ansible.plugins.inventory import BaseInventoryPlugin, Cacheable, to_safe_group_name, Constructable
+
+# 3rd party imports
+try:
+    import requests
+    if LooseVersion(requests.__version__) < LooseVersion('1.1.0'):
+        raise ImportError
+except ImportError:
+    raise AnsibleError('This script requires python-requests 1.1 as a minimum version')
+
+from requests.auth import HTTPBasicAuth
+
+
+class InventoryModule(BaseInventoryPlugin, Cacheable, Constructable):
+    ''' Host inventory parser for ansible using foreman as source. '''
+
+    NAME = 'foreman'
+
+    def __init__(self):
+
+        super(InventoryModule, self).__init__()
+
+        # from config
+        self.foreman_url = None
+
+        self.session = None
+        self.cache_key = None
+        self.use_cache = None
+
+    def verify_file(self, path):
+
+        valid = False
+        if super(InventoryModule, self).verify_file(path):
+            if path.endswith(('foreman.yaml', 'foreman.yml')):
+                valid = True
+            else:
+                self.display.vvv('Skipping due to inventory source not ending in "foreman.yaml" nor "foreman.yml"')
+        return valid
+
+    def _get_session(self):
+        if not self.session:
+            self.session = requests.session()
+            self.session.auth = HTTPBasicAuth(self.get_option('user'), to_bytes(self.get_option('password')))
+            self.session.verify = self.get_option('validate_certs')
+        return self.session
+
+    def _get_json(self, url, ignore_errors=None):
+
+        if not self.use_cache or url not in self._cache.get(self.cache_key, {}):
+
+            if self.cache_key not in self._cache:
+                self._cache[self.cache_key] = {url: ''}
+
+            results = []
+            s = self._get_session()
+            params = {'page': 1, 'per_page': 250}
+            while True:
+                ret = s.get(url, params=params)
+                if ignore_errors and ret.status_code in ignore_errors:
+                    break
+                ret.raise_for_status()
+                json = ret.json()
+
+                # process results
+                # FIXME: This assumes 'return type' matches a specific query,
+                #        it will break if we expand the queries and they dont have different types
+                if 'results' not in json:
+                    # /hosts/:id dos not have a 'results' key
+                    results = json
+                    break
+                elif isinstance(json['results'], MutableMapping):
+                    # /facts are returned as dict in 'results'
+                    results = json['results']
+                    break
+                else:
+                    # /hosts 's 'results' is a list of all hosts, returned is paginated
+                    results = results + json['results']
+
+                    # check for end of paging
+                    if len(results) >= json['subtotal']:
+                        break
+                    if len(json['results']) == 0:
+                        self.display.warning("Did not make any progress during loop. expected %d got %d" % (json['subtotal'], len(results)))
+                        break
+
+                    # get next page
+                    params['page'] += 1
+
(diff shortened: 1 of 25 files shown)


ares:
I'd prefer not to add yet another Foreman ansible inventory script. We have the official Foreman inventory script now in main Ansible repo https://github.com/ansible/ansible/blob/devel/lib/ansible/plugins/inventory/foreman.py

That has a benefit of getting contribution by ansible core members as well as being automatically synced to AWX. AWX users then have very good experience with fetching the inventory from Foreman but just providing credentials.

The situation right now is not ideal, because there are other and older versions of this inventory script that we're cleaning up. E.g. one in contrib/ directory in ansible repo itself.

evgeni:
@ares this is exactly the script you linked, but @willtome wants to move it from ansible/ansible to our collection.

willtome:
@evgeni you are correct. This is the exact plugin from Ansible core not a "new" one. With the move to collections, this plugin will eventually be migrated out of Ansible core. Especially since the foreman modules are not in core and reside in this collection, it makes since to keep all content together.

@ares the contrib/ scripts are being deprecated in favor of plugins. AWX and Tower are both using the legacy contrib script (https://github.com/ansible/awx/blob/devel/awx/plugins/inventory/foreman.py, https://docs.ansible.com/ansible-tower/latest/html/userguide/inventories.html#inventory-plugins). Users will still be able to use this plugin from the collection by creating a dynamic inventory "sourced from a project". This is already how users leverage inventory that is not hardcoded into AWX/Tower today.

gundalow:
In the next release of Ansible most (read just about all) modules and plugins will be moved into Collections. Since it looks like you've already got an Ansible Collection here in this repo you can get ahead of the curve.

mdellweg:
My question with moving it here is, who is going to maintain it?
Is there someone we can invite who's first use of the green button is merging this plugin?

evgeni:
I think I (personally) would prefer if this plugin can remain in ansible/ansible.

This collection is for users managing Foreman with Ansible, while the inventory plugin is to manage Foreman clients with Ansible and I think that the user bases are not identical.

Having the  inventory in Ansible core allows more users to benefit from it directly, without installing any further collections or dependencies.

gundalow:
It's very unlikely that any Foreman related modules or plugins will still exist in github.com/ansible/ansible in the next release, as it will be stripped down to something along the lines of https://github.com/ansible-community/collection_migration/blob/master/scenarios/minimal/ansible.yml

Everything else will be put in a Collection.

I'm happy for this to end up in the "catch all" Community collection. Though if you already have a Foreman Collection it may make sense to be in there.

evgeni:
@gundalow is that a strict decision? I can totally see why you'd want to move special module groups, but inventories?

evgeni:
Btw, why is apt_repo on that list? I can't see any other altlinux module...

evgeni:
Oh, and I couldn't find any deprecation warnings for any modules not on your list?

willtome:
The foreman expertise resides in this community. It makes sense that the modules, plugins, roles, playbooks, supporting code, etc. reside here as well. As @gundalow mentioned, Foreman is not unique in that the plugin is being migrated out of ansible core, this will happen to all of them, but _is_ unique in that there is already a rather comprehensive collection and development process for creating, testing, and maintaining this content.

I guess it would be up to this community if the inventory plugin should or should not reside in the same collection as the modules. I would argue that it should because if I'm in need of the inventory I probably have use cases that could benefit from the modules as well. And if I will have to install collections I would rather only install one with everything foreman related rather than multiple (ie. 1 for foreman, 1 for katello, 1 for inventory, etc). However, with the collections model, the foreman community now has the authority to make this decision, but I don't think that the plugin should be left behind to a "catch all" collection with the vibrancy that this project has.

evgeni:
I do agree that it should be in this collection if it moves out of core. I just think it should remain in core :)

And we do have sufficient permissions via https://github.com/ansible/ansible/blob/devel/.github/BOTMETA.yml#L1612 to execute changes there and apply our expertise ;)

evgeni:
test failures unrelated, fix in https://github.com/theforeman/foreman-ansible-modules/pull/633

gundalow:
> @gundalow is that a strict decision? I can totally see why you'd want to move special module groups, but inventories?

@evgeni Correct. github.com/ansible/ansible will be stripped down to just the Ansible Engine and enough modules to bootstrap the rest of the content. See the [use-cases](https://github.com/ansible-community/collection_migration/tree/master/scenarios/minimal) for anisble-minimal



> Btw, why is apt_repo on that list? I can't see any other altlinux module...

@evgeni As per the use-cases we included modules to help with installing content:

For other examples see https://github.com/ansible-community/collection_migration/blob/66d0fc20aa7386c48835df30cda5eb452f4eb261/scenarios/minimal/ansible.yml#L146-L159



> Oh, and I couldn't find any deprecation warnings for any modules not on your list?

@evgeni Unsure if that question is directed to me, if so, could you possibly expand so I can help?



> The foreman expertise resides in this community. It makes sense that the modules, plugins, roles, playbooks, supporting code, etc. reside here as well. As @gundalow mentioned, Foreman is not unique in that the plugin is being migrated out of ansible core, this will happen to all of them, but _is_ unique in that there is already a rather comprehensive collection and development process for creating, testing, and maintaining this content.

Yup, you have a solid base here in this repo to build on.


> I guess it would be up to this community if the inventory plugin should or should not reside in the same collection as the modules. I would argue that it should because if I'm in need of the inventory I probably have use cases that could benefit from the modules as well. And if I will have to install collections I would rather only install one with everything foreman related rather than multiple (ie. 1 for foreman, 1 for katello, 1 for inventory, etc). However, with the collections model, the foreman community now has the authority to make this decision, but I don't think that the plugin should be left behind to a "catch all" collection with the vibrancy that this project has.

I agree that "people that use Foreman inventory also use Foreman modules" so they should be in the same Collection.


Naming wise I was wondering if something like

```
# namespace.collection
satellite.katello
satellite.foreman
satellite.candlepin
```

evgeni:
> > @gundalow is that a strict decision? I can totally see why you'd want to move special module groups, but inventories?
> 
> @evgeni Correct. github.com/ansible/ansible will be stripped down to just the Ansible Engine and enough modules to bootstrap the rest of the content. See the [use-cases](https://github.com/ansible-community/collection_migration/tree/master/scenarios/minimal) for anisble-minimal
> 
> > Btw, why is apt_repo on that list? I can't see any other altlinux module...
> 
> @evgeni As per the use-cases we included modules to help with installing content:
> 
> For other examples see https://github.com/ansible-community/collection_migration/blob/66d0fc20aa7386c48835df30cda5eb452f4eb261/scenarios/minimal/ansible.yml#L146-L159

Right, this is a subset of what we have in https://github.com/ansible/ansible/tree/devel/lib/ansible/modules/packaging/os. My question was: why does the list include `apt_repo`, a module aimed at AltLinux (not Debian) repositories, something that we don't further support.

> > Oh, and I couldn't find any deprecation warnings for any modules not on your list?
> 
> @evgeni Unsure if that question is directed to me, if so, could you possibly expand so I can help?

You say that modules will be removed from Ansible 2.10 (or rather "next release"), but I've not seen any official warnings about that in documentation. Our own modules are documented to be removed in 2.12.

gundalow:
### deprecation

@evgeni ah, I see where the confusion I may have created has come from, I missed out a bit of information.

Starting with Ansible 2.10 there is a difference between:
* contained in ansible/ansible
* shipped in the Ansible Community distribution

In Ansible 2.9 you do `$package_manager install ansible` which gives you the Ansible Engine and thousands of modules & plugins. All of this comes from github.com/ansible/ansible


In 2.10 when you install the Community version of Ansible (ie not [Red Hat Ansible Automation Platform](https://www.ansible.com/products/automation-platform) you will still get all of the modules. It just happens that when the package (pip, deb, rpm) is built we will pull in the current stable versions of each of the selected community Collections [1].

Since the vast majority of people will go for the "install the kitchen sink" route they will still get the modules, therefore no deprecation cycle is needed


Once the Collection is up and running, you need to add a `migrated_to: namespace.collection` entry into `ansible/ansible`'s `.github/BOTMETA.yml`. That will be used (future work) to direct existing docs.ansible.com/ module links to Galaxy. The *plan* (i.e. not fully signed off yet) is at that point you can `git rm` the migrated content from ansible/ansible:devel. It's up to you what you want to do with PRs (close or migrate them for people) to https://github.com/theforeman/foreman-ansible-modules

Also once the Foreman Collection is up and running I'm OK with zero further changes being made to ansible/ansible:devel, ie all feature work can go straight into foreman-ansible-modules. Important bug fixes for Ansible 2.9 can still be made, this is done by raising a PR directly against `stable-2.9` (no longer a backport)


[1] In [.github/BOTMETA.yml](https://github.com/ansible/ansible/blob/devel/.github/BOTMETA.yml#L30-L31) we define the new Collection that hosts the content that was in ansible/ansible

### apt_repo

Ah, I totally miss understood this (I didn't realise `ALT Linux` was a distro), thank you for clarifying that. 

It looks like the authors and reviews of ansible-minimal made the same incorrect assumption that `apt_repo` is for Debian based distros. I'll raise a PR to remove it and update the README.md to make it clearer that we are only including package managers for primary distros (RPM & DEB based) as well as pip.



I hope this helps. I'm happy to jump on a call if that helps.

evgeni:
@gundalow thanks! This makes much more sense now, but also opens new questions :D

1. when you "pull in" content from collections into the "ansible community distribution", will you do the necessary steps to rename things back to be not collection-namespaced (ie. reverting https://github.com/theforeman/foreman-ansible-modules/pull/591/commits/17265dea5cb087ebb8c1a3bb61e492cb752cf734)? Or will the users still have to use the new, namespaced name to refer to the plugins?
2. you will only pull what was in ansible/ansible before, right? so the inventory would still be visible, but our modules will remain collection only?
3. is all of the above documented somewhere official instead of a random PR discussion? like ansible.com or so? would love to have something I can share with other module authors.

gundalow:
@evgeni That cool, please keep the questions coming. It really helps us improve the messaging around Collections. A lot of this is in the prototype stage, so we don't have all the answers yet. We have a few Collections which we are starting with [Supported Netapp](https://github.com/ansible-collections/ansible_collections_netapp) or [community Grafana](https://github.com/gundalow-collections/grafana) where we are testing out how this works in practice.

1) `theforeman.foreman.foreman` is an example of a FQCN (Fully Qualified Collection Name).
We have some ideas about how 2.9 Playbooks could be used without modification in 2.10 (at least for content that's packaged up in the Community Edition of Ansible.

2) Good question. What I'm expecting (subject to change is) something like this for the 2.10 Community Edition of Ansible 
   a) Find all the Collections which contain what the content that was in `stable-2.9` - by finding the `migrated_to: $namespace.$collection` lines from ansible/ansible's .github/BOTMETA.yml 
  b) At build time for Community Edition of Ansible we download the stable version of each of those collections
  c) Put all of that into the release
  d) This would mean if any new modules or plugins are placed in a collection that's referenced in BOTMETA.yml then it WOULD be included in the community release.

The exact details of how the Community Edition will be packaged are still being worked out, and may require further changes to go into Ansible 2.10. Please don't assume that this is exactly how it will work. We hope to have some public proposals & design docs live in the next few weeks.


3) We haven't been as good as communicating this as we should. That's partly as a lot is still up in the air. There will be a blog which is currently in review, which should address a lot of the questions that Contributors and Partners have.

evgeni:
@gundalow thanks! I must admit I don't think that including net-new plugins into ansible community edition, just because they are in the same collection as a moved-out plugin is a good idea. This will confuse users and I bet increase (instead of decrease) the number of issues opened against ansible/ansible.

gundalow:
@evgeni Once a collection has been created any issues or PRs would be created in the new repo, ie https://github.com/theforeman/foreman-ansible-modules/

We will look at adding support to move issues and PRs from ansible/ansible to new Collections repo.

Once Ansible 2.10 has been released only Issues and PRs about the Ansible Core Engine will be filed against ansible/ansible

The new Foreman collection (sourced from https://github.com/theforeman/foreman-ansible-modules) will be part of the Ansible Community Distribution. So any new modules added into https://github.com/theforeman/foreman-ansible-modules will automatically end up in the next Ansile Community Distribution.

evgeni:
@gundalow slightly related, do you know where `packaging/os/redhat_subscription.py`, `packaging/os/rhsm_repository.py` and `packaging/os/rhsm_release.py` is being moved to?

gundalow:
> @gundalow slightly related, do you know where `packaging/os/redhat_subscription.py`, `packaging/os/rhsm_repository.py` and `packaging/os/rhsm_release.py` is being moved to?

Those three modules are `support:community` and not listed in [ansible-minimal](https://github.com/ansible-community/collection_migration/blob/master/scenarios/minimal/ansible.yml), therefore they will most likely end up in `community.general` (the home for otherwise homeless content).

Is there something particularly of interest to you in those three modules?

evgeni:
All three can be used to configure Katello clients (as Katello implements the same API as subscription.rhsm.redhat.com), so you could say they are related to us in a way :)

evgeni:
Oh, another q, what will the contribution/maintenance workflow for `community.general` look like? ansibot driven as it is today, so people will retain their current permissions? What will the release candence look like? Once every ansible (major? minor?) release?

gundalow:
> All three can be used to configure Katello clients (as Katello implements the same API as subscription.rhsm.redhat.com), so you could say they are related to us in a way :)

Thanks for the explanation. Although they are similar to existing package modules which are listed in ansible-minimal, the modules you listed are currently `support:community`, so that would need to change to support:core


> Oh, another q, what will the contribution/maintenance workflow for `community.general` look like? ansibot driven as it is today, so people will retain their current permissions? What will the release cadence look like? Once every ansible (major? minor?) release?

For `community.general`  we will have ansibulbot.

For Single repos with single-focused Collections, such as [Grafana](https://github.com/ansible-collections/grafana) we are experimenting with using standard GitHub workflows and non-ansible specific GitHub Bots. The Grafana repos is the test case for these new workflows.



Please keep the questions coming. This is helping me understand what documentation I need to read (spoiler alert, a lot)

evgeni:
> > All three can be used to configure Katello clients (as Katello implements the same API as subscription.rhsm.redhat.com), so you could say they are related to us in a way :)
> 
> Thanks for the explanation. Although they are similar to existing package modules which are listed in ansible-minimal, the modules you listed are currently `support:community`, so that would need to change to support:core

Not sure I follow. Why would you want to move them to `support:core`? I was just curious where they would be moved to so I know where I need to look for fixes (or think about moving them here) ;-)

gundalow:
Ah, OK. They will most likely end up in `community.general`, which once created will live under github.com/ansible-collections.

gundalow:
Hi, I'm back from Annual leave, is there anything I can help with?

MarkusTeufelberger:
I think the following question/concern hasn't been answered yet:

If the Foreman inventory plugin (which is currently in ansible/ansible) gets moved into this collection (instead of community.general), will this mean that once 2.10 drops, also all modules from this collection as well as the Foreman inventory plugin will be rolled up into the Ansible 2.10 "full" release?

There could be an argument made either way, the easier one would probably be to ship the modules too as new content in 2.10.

gundalow:
Any modules (or plugins) that were part of 2.9 will still be installed when Ansible Community Distro (ACD) 2.10 is installed.

To expand on that, BOTMETA.yml is being updated to include `migrated_to: namespace.collection` entries, all those collections will be included in ACD. 

This means that any new modules added to those collections will also be included in ACD.

Community.general will be one large (1,500+ module) GitHub repo. 

Having your own collection (and therefore your own repo) means you are more in control, ie you can have direct commit powers, rather than having to go via Ansibullbot. https://github.com/ansible-collections/grafana is an example of this.

Having your own collection means you can release as and when you want. 

I'd personally make Foreman it's own collection.

gundalow:
Are we agreed that foreman should be it's own collection?

:+1: or :-1: votes on this issue please

evgeni:
But `foreman` already is a collection: https://galaxy.ansible.com/theforeman/foreman
Or are you suggesting a second, `community.foreman` collection? If so :-1:

gundalow:
Oh, I appear to have missed that the collection existed.

Can botmeta be updated?

gundalow:
So what's left to do?

mdellweg:
At the very least, a maintainer for this inventory plugin needs to be determined and added to this repository.

ares:
There were changes to the inventory script meanwhile, we'll need to rebase the script in this PR if we have an agreement

willtome:
I believe what was implemented in the ansible/ansible#67401 should be done with composed vars rather than an option in the inventory plugin. See example below which sets ansible_host to the indicated hostvar.
```
plugin: foreman
user: admin
password: ***
url: https://insights-sat66.usersys.redhat.com
compose:
  ansible_host: foreman_ipv4
```

AlanCoding:
@gundalow can I bump a question asked here:

> Or are you suggesting a second, community.foreman collection? If so 👎

None of the foreman content in ansible/ansible is marked with the `migrated_to` to indicate that it moved here. The NWO scenario here:

https://github.com/ansible-community/collection_migration/blob/42fa01d62dc4f1bcd5b8690534742050bd1cafd8/scenarios/nwo/community.yml#L425

has content marked for `community.general`. This isn't just a question about the inventory plugin. The [already deprecated foreman module](https://github.com/ansible-community/collection_migration/blob/42fa01d62dc4f1bcd5b8690534742050bd1cafd8/scenarios/nwo/community.yml#L2131) is marked for `community.general` while the [module itself](https://github.com/ansible/ansible/blob/308723c3ca1122363e419070f1fa1d76ff5611a9/lib/ansible/modules/remote_management/foreman/_foreman.py#L20) points here, `theforeman.foreman`.

I'm not arguing for one or the other, it doesn't matter to me. What does matter is having an authoritative answer on where where the code will be maintained. @jladdjr has changes we sorely need to get into the collection. All we need is a decisive answer on where the official collection to adopt the content is.

gundalow:
So I think this repo is the collection, and the galaxy.yml defines this collection as `theforeman.foreman`

1) ansible/ansible:devel needs BOTMETA.yml updating to list every plugin with migrated_to
2) scenarios/nwo/theforeman.yml needs creating and entries adding in (and removing from community.yml)

If the above is correct, then @jladdjr can raise PRs against this repo to make the code changes.

evgeni:
This repo is what's published on Galaxy as theforeman.foreman, yes.

And while I think that adding the inventory (and callback) plugin here is correct if they can't remain in ansible/ansible, I see no point in adding the old, deprecated foreman/katello modules here. They are deprecated in favor of this collection, that's it.

AlanCoding:
From everything said, looks like it may be converging on this action:

1. https://github.com/ansible/ansible/compare/devel...AlanCoding:foreman_migrated_to?expand=1
2. https://github.com/ansible-community/collection_migration/compare/master...AlanCoding:foreman_migrated_to?expand=1

With this structure, it follows that [deprecated modules are put in community collection](https://github.com/ansible-community/collection_migration/blob/42fa01d62dc4f1bcd5b8690534742050bd1cafd8/scenarios/nwo/community.yml#L2131-L2132). This would make complete sense to me, except for [the bug blocking deprecated modules in collections](https://github.com/ansible/ansible/issues/60424), but that's a general question and not a consideration for this discussion specifically.

gundalow:
Thanks for creating those branches.

The deprecated modules MUST go in theforeman.foreman collection as well.

evgeni:
Why? We don't really have resources to maintain them. Plus they are said to be removed in 2.12, but this collection does not follow the same lifecycle as ansible.

Wouldn't it be easier to create community.deprecated and put everything from a/a there that is deprecated?

gundalow:
As you said, they need to exist till Ansible 2.12.

If you look at other collections, they correctly contain deprecated plugins alongside the actively developed ones. 

I'm not requiring them to be maintained. This is about easing the pain for end users. 


If the deprecated modules end up anywhere else then they will be totally ignored (full bit rot), also we will add another Collection Dependency on theforeman.foreman so the shared module_utils are available.

evgeni:
If they are not maintained, they do bit rot, per definition ;) And I don't think the users care where their modules are get pulled from.

Also there are no shared module_utils for foreman in ansible/ansible, so there will be no additional dependency.

ehelms:
As the original author of those modules, I would really prefer we not bring those old modules into this collection. We really do not want users using those modules anymore, nor do we want to promote their usage in anyway if we can avoid it.

As for inventory plugin:

After some discussions we are good to include the inventory plugin in this collection. Further, the callback (https://github.com/ansible/ansible/blob/devel/lib/ansible/plugins/callback/foreman.py) should come along for the ride and live in the collection as well. This would mean we have a single collection for:

 * modules
 * inventory plugin
 * callback plugin

We will plan for and perform a release just prior to merging the inventory / callback plugins to this collection as well.

@gundalow If I missed this sorry, what is the timeline to pull the inventory / callback plugins into the codebase? What happens to any open PRs to ansible/ansible?

gundalow:
Thanks for the details. Given the fact that there isn't any shared code (module_utils) I'm OK with the deprecated modules staying in community.general.

Ansible/ansible:devel has just been frozen.

https://github.com/ansible-collections/overview/blob/master/README.rst#time-line contains the time line.

Yes, this collection should contain the various modules and plugins you mentioned. 

Can the scenarios/nwo file please be updated?

We don't yet have a date for when we will bulk close issues and PRs. PRs will be first, and we will have a process to help move PRs.

AlanCoding:
> Can the scenarios/nwo file please be updated?

Specifically, does this mean that changing the BOTMETA file in the ansible/ansible repo will not be necessary to get this done?

Put up https://github.com/ansible-community/collection_migration/pull/472, and linked dependent PRs from there. Contingent on CI passing and being approved, this offers a chain of merges that would carry out what has been discussed.

AlanCoding:
More changes landed in the `devel` branch of ansible/ansible. Put up https://github.com/willtome/foreman-ansible-modules/pull/1 to pull those into this PR.

sjha4:
@willtome @AlanCoding : AWX is using the  https://github.com/ansible/awx/blob/devel/awx/plugins/inventory/foreman.py script as inventory source. Is that going to change following this? We have some changes in the AWX script that we would need to port over to whatever the new source for AWX becomes. ..https://github.com/ansible/awx/commit/b10a71786b2f7fdcab0ef901e1af871534f1dd80

AlanCoding:
Yes, that's right, the changes to the script will need to be ported over to the inventory plugin. That can start once the matter of its new home is settled. AWX will have an additional complication of how to source the collection for the purpose of inventory imports, but this gives the benefit of no longer being tied to Ansible's version, just the collection version.

ehelms:
As I understand, we are mostly good to go here. Here is my plan:

 * On Friday I will merge the inventory plugin and the callback to our collection
 * Given these are the state of the art in Ansible, we will do a collection release after merge
 * All changes to them will need to flow through this collection after Friday when they are merged

@evgeni @ares @sjha4 @willtome @AlanCoding  If anyone has concerns with this plan please raise them

evgeni:
My only concerns are the missing tests and knowledge, both we can (and will have to) acquire soon - preferably in the next few weeks. But that shouldn't block merging.

gundalow:
> My only concerns are the missing tests and knowledge, both we can (and will have to) acquire soon - preferably in the next few weeks. But that shouldn't block merging.

@evgeni is this referring to the integration tests that are currently in ansible/ansible? You can copy those into this collection, and execute them via a GitHub action. 
https://github.com/ansible-collections/grafana/blob/master/.github/workflows/ansible-test.yml#L76

evgeni:
Yeah. Need to have a look what tests are there in a/a and copy them. At least for a start, I remember a/a having a rather crude foreman emulator at some point, which I disliked at that time ;)

gundalow:
Haha, yes.
I'm happy to give suggestions on testing. 

Running `ansible-test sanity` will help protect you from common module mistakes and broken docs.

evgeni:
We already run sanity, yeah.

plugins/inventory/foreman.py
```diff
@@ -0,0 +1,279 @@
...
+# Copyright (c) 2018 Ansible Project
+# GNU General Public License v3.0+ (see COPYING or https://www.gnu.org/licenses/gpl-3.0.txt)
+from __future__ import (absolute_import, division, print_function)
+__metaclass__ = type
+
+DOCUMENTATION = '''
+    name: foreman
+    plugin_type: inventory
+    short_description: foreman inventory source
+    version_added: "2.6"
```
mdellweg: As it is about to be ejected from ansible, this information has no value anymore.
```suggestion
```

plugins/inventory/foreman.py
```diff
@@ -0,0 +1,279 @@
...
+      plugin:
+        description: the name of this plugin, it should always be set to 'foreman' for this plugin to recognize it as it's own.
+        required: True
+        choices: ['foreman']
+      url:
+        description: url to foreman
+        default: 'http://localhost:3000'
+        env:
+            - name: FOREMAN_SERVER
+              version_added: "2.8"
```
mdellweg:
```suggestion
```

plugins/inventory/foreman.py
```diff
@@ -0,0 +1,279 @@
...
+      user:
+        description: foreman authentication user
+        required: True
+        env:
+            - name: FOREMAN_USER
+              version_added: "2.8"
```
mdellweg:
```suggestion
```

plugins/inventory/foreman.py
```diff
@@ -0,0 +1,279 @@
...
+      password:
+        description: foreman authentication password
+        required: True
+        env:
+            - name: FOREMAN_PASSWORD
+              version_added: "2.8"
```
mdellweg:
```suggestion
```

plugins/inventory/foreman.py
```diff
@@ -0,0 +1,279 @@
...
+        description: Toggle, if true the inventory will retrieve 'all_parameters' information as host vars
+        type: boolean
+        default: False
+      legacy_hostvars:
+        description:
+            - Toggle, if true the plugin will build legacy hostvars present in the foreman script
+            - Places hostvars in a dictionary with keys `foreman`, `foreman_facts`, and `foreman_params`
+        type: boolean
+        default: False
+        version_added: '2.10'
```
mdellweg:
```suggestion
```

plugins/inventory/foreman.py
AlanCoding: I don't know what the current policy is with collections. It could become the collection version, which would still have value to the user.

If it is the collection version, then I can say for sure that 0.5.0 does not have any of this.

evgeni: AFAIK the sanity tests explicitly don't test/expect version_added inside collections.

I don't strictly care if this remains here for now or not, regardless of the value.

PR #594, opened 2019-11-15
don't treat pycache as a module in test_module_state

evgeni:


tests/test_module_state.py
```diff
@@ -19,7 +19,7 @@
 def find_all_modules():
     for module in MODULES_PATH.listdir(sort=True):
         module = module.basename.replace('.py', '')
-        if module != '__init__':
```
mdellweg: `if module.basename.endswith(".py"):` ?

mdellweg: Oh, that does not work with `__init__.py`. Darn.

evgeni: that would also work, yes

mdellweg: What about a regexp match:
`r"\([^-].*\)\.py"` and the first matching group is what you want anyway.

evgeni:
not sure why we didn't see that before, maybe something changed how py.path works?

Issue #172, opened 2018-08-20
katello_manifest isn't idempotent


akofink:
##### SUMMARY
Manifest upload should only happen once, regardless of the number of times a playbook is run.

##### ISSUE TYPE
 - Bug Report

##### ANSIBLE VERSION

```
ansible 2.6.2
  config file = /home/akofink/dev/nfs/centos7-hammer-devel/foreman-ansible-modules/ansible.cfg
  configured module search path = [u'/home/akofink/dev/nfs/centos7-hammer-devel/foreman-ansible-modules/modules']
  ansible python module location = /usr/lib/python2.7/site-packages/ansible
  executable location = /usr/bin/ansible
  python version = 2.7.15 (default, May 16 2018, 17:50:09) [GCC 8.1.1 20180502 (Red Hat 8.1.1-1)]
```

##### KATELLO/FOREMAN VERSION

```
tfm-rubygem-katello-3.7.0-1.el7.noarch
foreman-1.18.1-1.el7.noarch
```

##### NAILGUN VERSION

```
Version: 0.30.2
```

##### STEPS TO REPRODUCE

```yaml
  - name: Upload a manifest
    katello_manifest:
      username: "{{ foreman_username }}"
      password: "{{ foreman_password }}"
      server_url: "{{ foreman_server_url }}"
      verify_ssl: "{{ foreman_verify_ssl }}"
      organization: Default Organization
      manifest_path: "{{ katello_manifest_path }}"
      state: present
  - name: Upload the same manifest again
    katello_manifest:
      username: "{{ foreman_username }}"
      password: "{{ foreman_password }}"
      server_url: "{{ foreman_server_url }}"
      verify_ssl: "{{ foreman_verify_ssl }}"
      organization: Default Organization
      manifest_path: "{{ katello_manifest_path }}"
      state: present
```

##### EXPECTED RESULTS

```
TASK [Upload a manifest] ******************************
changed: [localhost]

TASK [Upload the same manifest again] *****************************
ok: [localhost]
```
##### ACTUAL RESULTS

```
TASK [Upload a manifest] ******************************
changed: [localhost]

TASK [Upload the same manifest again] *****************************
changed: [localhost]
```

sean797:
I had a look at doing this a while ago, couldn't find a decent way to verify if we need to upload another one. IIRC there isn't a version number or anything similar that we could compare.

akofink:
[...]

PR #595, opened 2019-11-15
Run tests again in check_mode

mdellweg:


mdellweg:
And it turns out, we are rather bad at `check_mode`. :imp:

plugins/modules/katello_content_view_version.py
```diff
@@ -239,8 +239,11 @@ def main():
 
             changed, response = module.resource_action('content_views', 'publish', params=payload, synchronous=entity_dict['synchronous'])
             # workaround for https://projects.theforeman.org/issues/28138
-            content_view_version_id = response['output'].get('content_view_version_id') or response['input'].get('content_view_version_id')
-            content_view_version = module.show_resource('content_view_versions', content_view_version_id)
+            if not module.check_mode:
+                content_view_version_id = response['output'].get('content_view_version_id') or response['input'].get('content_view_version_id')
+                content_view_version = module.show_resource('content_view_versions', content_view_version_id)
+            else:
+                content_view_version = {'id': -1, 'environments': []}
```
mdellweg: If you can't make it, fake it.

plugins/modules/katello_manifest.py
```diff
@@ -119,13 +119,15 @@ def main():
...
                 params.update(scope)
                 changed, result = module.resource_action('subscriptions', 'upload', params, files=files)
-                for error in result['humanized']['errors']:
-                    if "same as existing data" in error:
-                        changed = False
-                    elif "older than existing data" in error:
-                        module.fail_json(msg="Manifest is older than existing data.")
-                    else:
-                        module.fail_json(msg="Upload of the manifest failed: %s" % error)
+                # This is not completely accurate, because we cannot be sure something changed until we attempt the actual upload
```
mdellweg: Pain point number one.

tests/test_playbooks/katello_manifest.yml
```diff
@@ -27,7 +27,8 @@
     vars:
       manifest_path: "{{ katello_manifest_path }}"
       manifest_state: "present"
-      expected_change: false
+      # TODO: This does not work in check_mode
+      expected_change: "{{ ansible_check_mode }}"
```
mdellweg: Related to pain point number one.

tests/test_playbooks/location.yml
```diff
@@ -81,8 +81,9 @@
     vars:
       location_state: absent
       location_name: Test Location
-      expected_change: false
-      expected_error: true
+      # TODO checkmode does not realize the impossibility of removing the parent
+      expected_change: "{{ ansible_check_mode }}"
+      expected_error: "{{ not ansible_check_mode }}"
```
mdellweg: Pain point number two.

tests/test_playbooks/redhat_manifest.yml
```diff
@@ -31,7 +31,8 @@
     # Test removing subs & export
   - include: tasks/redhat_manifest.yml
     vars:
-      expected_change: true
+      # TODO This looks like an actual bug in redhat_manifest module
+      expected_change: "{{ not ansible_check_mode }}"
```
mdellweg: Pain point number three.

tests/test_playbooks/tasks/location.yml
```diff
@@ -22,5 +22,5 @@
     fail_msg: "Ensuring location is {{ location_state }} didn't fail! (expected_error: {{ expected_error | default('unknown') }})"
     that:
       - result is failed
-  when: expected_error is defined
+  when: expected_error is defined and expected_error
```
mdellweg: Related to pain point number two.

tests/test_playbooks/tasks/katello_sync.yml
```diff
@@ -22,6 +22,7 @@
     that:
       - not synchronous or ((( result.task.ended_at | replace(' UTC', '') | to_datetime ) - ( result.task.started_at | replace(' UTC', '') | to_datetime )).total_seconds() >= 2)
       - synchronous or not result.task.ended_at  # result.task.ended_at is not set if it matches result.task.started_at
+  when: result.task
```
mdellweg: In `check_mode` we simply don't trigger a task, and cannot expect to analyse that one. I think this is OK.

tests/test_playbooks/redhat_manifest.yml
mdellweg: @sean797 maybe you can see, why.

mdellweg: Found it, fixed it.

plugins/modules/katello_manifest.py
evgeni: I think this block lies about the state of the manifest in Katello and we just should say that this module does not support check mode?

we have https://github.com/theforeman/foreman-ansible-modules/issues/172 open and I think we can't fix it until Katello fixes their API.

cc @akofink @sean797 for their knowledge

tests/test_playbooks/tasks/katello_sync.yml
evgeni: I think that's OK

plugins/modules/katello_manifest.py
mdellweg: Fine with just _not_ stating to support checkmode.

tests/test_playbooks/tasks/location.yml
evgeni:
```suggestion
  when:
   - expected_error is defined
   - expected_error
```

but ack otherwise

evgeni:
Thanks @mdellweg!

I think the only two comments I have are:
* katello_manifest should just not support check mode in the current state of the API
* this almost doubles the test execution time, can we somehow run that in a separate "branch", probably only against one python/ansible release is sufficient too?

plugins/modules/katello_manifest.py
akofink: Yep, Katello needs to fix their API first. +1 to stating no support of checkmode or not stating support, either one.

mdellweg: No check_mode it is...
Funny thing: `AnsibleModule`, when called with `check_mode` despite it not beeing stated as supported calls `self.exit_json` from its initializer to state `changed: false`.
This broke our derived modules, because we overloaded `exit_json` in a way that it expects the initializer to have run.
Does anyone think, this is an ansible bug?

.github/workflows/main.yml
```diff
@@ -43,8 +43,12 @@ jobs:
         run: pip install git+https://github.com/ansible/ansible.git@${{ matrix.ansible }}#egg=ansible
       - name: Install dependencies
         run: make test-setup
+      - name: Run basic tests
+        run: make test-basic
+        if: matrix.python != "3.6"
```
evgeni: apparently this is invalid syntax?!

Makefile
```diff
@@ -30,8 +30,11 @@ lint:
 test:
 	pytest -v $(TEST)
 
+test-basic:
```
evgeni: how about we re-structure the way we call our tests a bit:

```Makefile
test: test-crud test-checkmode test-other

test-crud:
	pytest -v 'tests/test_crud.py::test_crud'

test-checkmode:
	pytest -v 'tests/test_crud.py::test_check_mode'

test-other:
	pytest -v -k 'not test_crud.py'
```

see [pytest docs how to specify test selection](https://docs.pytest.org/en/latest/usage.html#specifying-tests-selecting-tests)

and then the user still can do `make test` while CI does `make test-crud test-other` in most cases and once more with `test-checkmode`?

doesn't have to be in this PR tho if you get your version to work.

evgeni: oh, and this is missing the `$(TEST)` var we have above, so would need some more thought ;)

(or maybe just leave the old `make test` definition`)

mdellweg: Can we drop the `$(TEST)` variable? Does anyone use it?
Also how about moving the static analysis tests into `lint` or `sanity`?
For now, i think, i'll "just" try add another step to do the check_mode tests isolated.

mdellweg: Actually i'll leave the old `make test`, because that one runs all tests in one `pytest` instance, giving a better overview at the end.

PR #598, opened 2019-11-19
Fix wrong value in the examples for "server_url"

jerrejkw:
This PR fixes the wrong value of "server_type" in the examples of the following modules:

- foreman_compute_profile
- foreman_compute_resource

This will prevent the following error from occurring when users enter an FQDN in server_url without prefixing it with "http://" or "https://"

`    "msg": "Failed to connect to Foreman server: Could not load data from foreman.example.com: Invalid URL '/apidoc/v2.json': No schema supplied. Perhaps you meant http:///apidoc/v2.json?\n                  - is your server down?\n                  - was rake apipie:cache run when using apipie cache? (typical production settings) "
`

### DIFF
--- a/plugins/modules/foreman_compute_profile.py
+++ b/plugins/modules/foreman_compute_profile.py
@@ -71,7 +71,7 @@ EXAMPLES = '''
 - name: compute profile
   foreman_compute_profile:
     name: example_compute_profile
-    server_url: foreman.example.com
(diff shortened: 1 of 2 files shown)


bagasse:
Thanks @jerrejkw !

jerrejkw:
Thanks @bagasse for approving the PR so quickly.

PR #604, opened 2019-11-21
Add scc_product module

Manisha15:
Add scc_product module for scc_manager plugin api (v2)


mdellweg:
Given, that this module really only able to subcribe products, we might want to call it an action plugin, not an entity plugin. @evgeni what do you think?

plugins/modules/foreman_scc_product.py
```diff
@@ -0,0 +1,118 @@
...
+        ),
+    )
+
+    entity_dict = module.clean_params()
+
+    module.connect()
+
+    entity_dict['organization'] = module.find_resource_by_name('organizations', entity_dict['organization'], thin=True)
+
+    scope = {'organization_id': entity_dict['organization']['id']}
```
evgeni: these two lines can be replaced with `handle_organization_param`

plugins/modules/foreman_scc_product.py
```diff
@@ -0,0 +1,118 @@
...
+from ansible.module_utils.foreman_helper import KatelloEntityAnsibleModule
+
+
+def main():
+    module = KatelloEntityAnsibleModule(
+        entity_spec=dict(
+            name=dict(required=True),
+            friendly_name=dict(),
+            scc_account=dict(required=True),
+            state=dict(default='subscribe', choices=['subscribe']),
```
evgeni: what is the point of `state` here if we only can have `subscribe` and that is the default?

plugins/modules/foreman_scc_product.py
```diff
@@ -0,0 +1,118 @@
...
+
+    entity_dict['scc_account'] = module.find_resource_by_name('scc_accounts', name=entity_dict['scc_account'], params=scope, thin=True)
+
+    scope = {'scc_account_id': entity_dict['scc_account']['id']}
+
+    # Try to find the SccProduct to work on
+    # name is however not unique, but friendly_name is
+    entity = None
+    # If we have a friendly_name, search for it
+    if 'friendly_name' in entity_dict and entity_dict['friendly_name'] != '':
```
evgeni: and we don't have `friendly_name` we don't search at all? then `friendly_name` should be required.

plugins/modules/foreman_scc_product.py
```diff
@@ -0,0 +1,118 @@
...
+        search_string = 'friendly_name="{0}"'.format(entity_dict['friendly_name'])
+        entity = module.find_resource('scc_products', search_string, params=scope, failsafe=True)
+
+    if module.state == 'subscribe':
+        result = module.foremanapi.resource('scc_products').call('subscribe', {'id': entity['id'], 'scc_account_id': scope['scc_account_id']})
```
evgeni:
```suggestion
        params = {'id': entity['id']}
        params.update(scope)
        result = module.foremanapi.resource('scc_products').call('subscribe', params)
```

also, any reason not to use `resource_action` here?

plugins/modules/foreman_scc_product.py
```diff
@@ -0,0 +1,118 @@
...
+                    'status': ['preview'],
+                    'supported_by': 'community'}
+
+
+DOCUMENTATION = '''
+---
+module: foreman_scc_product
+short_description: Manage Foreman SccProduct
+description:
+  - "Manage Foreman SccProduct Entities"
```
evgeni: the same "scc" comments as in the other PR apply here ;)

plugins/modules/foreman_scc_product.py
evgeni: (and given it's a "state", should it be called "subscribed"?)

tests/test_playbooks/scc_product.yml
```diff
@@ -0,0 +1,16 @@
...
+- hosts: tests
+  gather_facts: false
+  vars_files:
+    - vars/server.yml
+  tasks:
+  - include: tasks/scc_product.yml
+    vars:
+      scc_product_name: 'SUSE Linux Enterprise Server'
+      scc_product_friendly_name: 'SUSE Linux Enterprise Server 10 SP1 online i386'
+      # requires synced scc_account named "test"
```
evgeni: I guess we cannot set that up, as it would require a working SUSE account?

But maybe we still can provide a boilerplate w/o a set username/password?

plugins/modules/foreman_scc_product.py
mdellweg: Not having a `state` is what i think make an action module. So i guess we agree there?

evgeni: Yes, also we're not really using the entity spec, as we don't call ensure_entity?

evgeni:
now that #588 is merged, should we move the apidoc json to `scc_manager.json` and symlink `scc_account` and `scc_product` to it?

tests/test_playbooks/scc_product.yml
Manisha15: > I guess we cannot set that up, as it would require a working SUSE account?
> 
> But maybe we still can provide a boilerplate w/o a set username/password?

You mean providing false scc_account and expect the module to fail or record on actual scc_account but only provide it's name in playbook?

evgeni: no, I mean provide a play against `localhost`, that would setup `scc_account` but feed it with wrong/no SCC credentials so that a re-record would fail unless the user configures it properly.

in a way like https://github.com/theforeman/foreman-ansible-modules/blob/29f903edb41ce4b07ea7c336c7e59f34efa84480/tests/test_playbooks/repository_set.yml#L10-L13

the manifest we have commited in the repo is not sufficient for this to work, but someone can provide a working one for re-recording.

plugins/modules/foreman_scc_product.py
```diff
@@ -0,0 +1,108 @@
...
+
+    entity_dict['scc_account'] = module.find_resource_by_name('scc_accounts', name=entity_dict['scc_account'], params=scope, thin=True)
+
+    scope = {'scc_account_id': entity_dict['scc_account']['id']}
+
+    # Try to find the SccProduct to work on
+    # name is however not unique, but friendly_name is
+    entity = None
+    # If we have a friendly_name, search for it
+    if 'friendly_name' in entity_dict and entity_dict['friendly_name'] != '':
```
evgeni: now that you made `friendly_name` required, the `if` can be removed completely?

plugins/modules/foreman_scc_product.py
```diff
@@ -0,0 +1,108 @@
...
+        search_string = 'friendly_name="{0}"'.format(entity_dict['friendly_name'])
+        entity = module.find_resource('scc_products', search_string, params=scope, failsafe=True)
+
+    if module.state == 'subscribe':
```
evgeni: the module in its current state does not support `state=subscribe`

plugins/modules/foreman_scc_product.py
```diff
@@ -0,0 +1,108 @@
...
+'''
+
+RETURN = ''' # '''
+
+
+from ansible.module_utils.foreman_helper import KatelloEntityAnsibleModule
+
+
+def main():
+    module = KatelloEntityAnsibleModule(
```
mdellweg: I think, `KatelloAnsibleModule` should be sufficient.

plugins/modules/foreman_scc_product.py
```diff
@@ -0,0 +1,108 @@
...
+module: foreman_scc_product
+short_description: Manage Foreman Suse Customer Center Account Product
+description:
+  - "Manage Foreman Suse Customer Center Product Entities"
+  - "This module requires the foreman_scc_manager plugin set up in the server"
+  - "See: U(https://github.com/ATIX-AG/foreman_scc_manager)"
+author:
+  - "Manisha Singhal (@manisha15) ATIX AG"
+options:
+  name:
```
mdellweg: As we use `friendly_name` as a key, is it still useful to also specify `name`?

plugins/modules/foreman_scc_product.py
```diff
@@ -0,0 +1,108 @@
...
+    description: Name of the suse customer center account associated with product
+    required: true
+    type: str
+extends_documentation_fragment: foreman
+'''
+
+EXAMPLES = '''
+- name: "Subscribe to suse customer center product"
+  foreman_scc_product:
+    name: "Product1"
```
mdellweg: I believe it must be `friendly_name` here too.

tests/fixtures/apidoc/scc_product.json
```diff
@@ -0,0 +1 @@
+scc_manager.json
```
mdellweg: A similar symlink should also work for `scc_account`

plugins/module_utils/foreman_helper.py
```diff
@@ -220,7 +220,7 @@ def __init__(self, argument_spec, **kwargs):
         if 'verify_ssl' in self._params:
             self.warn("Please use 'validate_certs' instead of deprecated 'verify_ssl'.")
 
-        self.task_timeout = 60
+        self.task_timeout = 240
```
mdellweg: You can adjust that timeout in the module itself.

See https://github.com/theforeman/foreman-ansible-modules/blob/4ab1fbea7140302b818cd2674e8d95cd7612f7b2/plugins/modules/katello_sync.py#L113

tests/test_playbooks/scc_product.yml
```diff
@@ -0,0 +1,38 @@
...
+  gather_facts: false
+  vars_files:
+    - vars/server.yml
+  tasks:
+  - include: tasks/organization.yml
+    vars:
+      organization_state: present
+  - include: tasks/scc_account.yml
+    vars:
+      scc_account_state: present
```
mdellweg: Here, you want to use the
`#Parameter for scc_product test`

tests/test_playbooks/scc_product.yml
```diff
@@ -0,0 +1,38 @@
...
+      expected_change: true
+
+- hosts: localhost
+  gather_facts: false
+  vars_files:
+    - vars/server.yml
+  tasks:
+    - include: tasks/scc_account.yml
+      vars:
+        scc_account_state: absent
```
mdellweg: And here at least the name again.

plugins/modules/foreman_scc_product.py
```diff
@@ -0,0 +1,102 @@
...
+    scc_account_scope = {'scc_account_id': params['scc_account']['id']}
+
+    # Try to find the SccProduct to work on
+    # name is however not unique, but friendly_name is
+
+    search_string = 'friendly_name="{0}"'.format(params['friendly_name'])
+    scc_product = module.find_resource('scc_products', search_string, params=scc_account_scope, failsafe=True)
+
+    if scc_product and not scc_product['product_id']:
+        module.resource_action('scc_products', 'subscribe', {'id': scc_product['id'], 'scc_account_id': scc_account_scope['scc_account_id']})
```
mdellweg: I know, we have a pr pending, in scc_manager, that we can check whether the product is already subscribed.
But i am wondering what happens if we try to subscribe again without this.

Manisha15: It gives Internal server error with error msg : Product already subscribed.

mdellweg: That is good enough for me for now.

plugins/modules/foreman_scc_product.py
```diff
@@ -0,0 +1,102 @@
...
+    elif scc_product and scc_product['product_id']:
+        module.fail_json(msg='{0} is already subscribed'.format(params['friendly_name']))
```
evgeni: I'd argue that enabling an enabled product should not fail but warn, but will not block on this.

plugins/modules/foreman_scc_product.py
```diff
@@ -0,0 +1,102 @@
...
+module: foreman_scc_product
+short_description: Subscribe Foreman Suse Customer Center Account Product
+description:
+  - "Manage Foreman Suse Customer Center Product Entities"
+  - "This module requires the foreman_scc_manager plugin set up in the server"
+  - "See: U(https://github.com/ATIX-AG/foreman_scc_manager)"
+author:
+  - "Manisha Singhal (@manisha15) ATIX AG"
+options:
+  friendly_name:
```
evgeni: I am a tad confused about the API here. You can search on `name` and on `friendly_name` (https://github.com/ATIX-AG/foreman_scc_manager/blob/67dc908192f729b2de90169188def6b4614e5aa0/app/models/scc_product.rb#L21-L22) but only `friendly_name` is unique?

What I am wondering is: would it hurt if we call this Ansible param `name`, even tho it really represents `friendly_name`? It just feels weird to use something else but `name` (which is a very common Ansible pattern) to find something.

mdellweg: Yeah, it is confusing. Even the information that `friendly_name` is unique (enough) is only evidence based. But renaming it to `name` could be even more confusing if someone had a look at the the api output. Keeping `friendly_name` might help to ring the right bells.

plugins/modules/foreman_scc_product.py
mdellweg: Maybe i missed it. This change surely depends on https://github.com/ATIX-AG/foreman_scc_manager/pull/29 ...

plugins/modules/foreman_scc_product.py
evgeni: k, I won't have the pleasure to use the module any time soon, so I'll trust your judgement :)

plugins/modules/foreman_scc_product.py
mdellweg: I would not even warn, but _just_ report no change.
Also if we did `scc_product.get('product_id')` we might get away without waiting for that other PR to arrive in production.

mdellweg:
I believe it is fine now. Thank you very much.

Issue #605, opened 2019-11-21
auth_source_ldaps: not idempotent when account_password is provided


evgeni:
##### SUMMARY

The API doesn't return the password (good), so we always try to set it. This should be documented.

##### ISSUE TYPE
 - Bug Report

##### ANSIBLE VERSION

```

```

##### KATELLO/FOREMAN VERSION

```

```

##### APYPIE VERSION

```

```

##### STEPS TO REPRODUCE

```yaml

```

##### ACTUAL RESULTS

```

```

mdellweg:
I think, we have similar examples in compute_resource and hostgroup.

PR #606, opened 2019-11-21
auth_source_ldap: improve docs and add required_if for onthefly_register

evgeni:
Fixes: #605

### DIFF
--- a/plugins/modules/foreman_auth_source_ldap.py
+++ b/plugins/modules/foreman_auth_source_ldap.py
@@ -192,7 +192,7 @@ def main():
             locations=dict(type='entity_list', flat_name='location_ids'),
             organizations=dict(type='entity_list', flat_name='organization_ids'),
         ),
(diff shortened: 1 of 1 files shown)


plugins/modules/foreman_auth_source_ldap.py
```diff
@@ -181,6 +192,7 @@ def main():
             locations=dict(type='entity_list', flat_name='location_ids'),
             organizations=dict(type='entity_list', flat_name='organization_ids'),
         ),
+        required_if=[['onthefly_register', True, ['attr_login', 'attr_firstname', 'attr_lastname', 'attr_mail']]]
```
mdellweg:
```suggestion
        required_if=[['onthefly_register', True, ['attr_login', 'attr_firstname', 'attr_lastname', 'attr_mail']]],
```

evgeni: picky.

Issue #324, opened 2019-07-05
RFC: "automated" module generation


evgeni:
##### SUMMARY
While reviewing #318 and working on slides for an upcoming talk, I noticed that our modules mostly follow the same structure:

1. write a `name_map` to map Ansible params to Foreman API params
1. create a `ForemanAnsibleModule` instance that receives a `argument_spec` that duplicates some data from `name_map`
1. create an `entity_dict` from the Ansible params
1. `module.connect()`
1. `entity = module.find_resource_by_name(…)`
1. enrich `entity_dict` with data if we need to reference other entities
1. `changed = module.ensure_resource_state(…)`
1. `module.exit_json(changed=changed)`

My crazy idea was to merge the data of `name_map` and `argument_spec` into a single `dict` (that can be later processed by `ForemanAnsibleModule`) and make the module look like this:

```python
foreman_spec = { name_map+argument_spec }
module = ForemanAnsibleModule(resource='resource', foreman_spec=foreman_spec)
module.run()
```

And `module.run()` would be a wrapper like this:

```python
def run(self):
  entity_dict = self.clean_params()
  self.connect()
  entity = self.find_resource_by_name(self.resource, …)`
  for additional_resource in self.foreman_spec:
    entity_dict[additional_resource['name']] = find…
  changed = self.ensure_resource_state(…)
  self.exit_json(changed=changed)
```

This surely won't work for some more complicated modules we have, but for those that "just" ensure entities, this should work.

##### ISSUE TYPE
 - Crazy Idea

ekohl:
I think the `name_map` can still be a separate parameter but I like the idea. We also have very common mappings like `locations -> location` and `organizations -> organization`. Perhaps it should automatically handle those? I also wonder why we have mappings like `name -> name`. That sounds like a noop to me.

How do you see modules which can't do this simple run? Would they override the run method?

mdellweg:
@ekohl I know, but a mapping `name: name` still serves as a filter. Every other parameter will be removed. Also we had the discussion earlier, and my answer was: "Consistency."
I think, we gain something, if those modules look as similar as possible. (At least something to refactor away from them.)

mdellweg:
I started some work on the state and name_map parameters here: #337

evgeni:
With #610 merged, I as the author of the original issue consider this done.

PR #607, opened 2019-11-21
DRY host and host_group a little

mdellweg:
affects #256


bagasse:
Maybe we can dry the documentation as well with a doc fragment ?

evgeni:
@bagasse theoretically, yes, but this would break our current makefile to generate the collection. Evgeni was lazy when writing it :(

bagasse:
This is out of scope of this PR: Maybe we can improve `entity_spec` add a field `resource_type_name` when `type=entity` to automatically get the corresponding entities in the base class if defined and not in each modules ? Maybe we have to think about scope/filter/search for some resources but I think that will dry up a lot of code.

eg (not tested):
in module
```        entity_spec=dict(
            name=dict(required=True),
            hostgroup=dict(type='entity', flat_name='hostgroup_id', resource_type='hostgroups'),
            location=dict(type='entity', flat_name='location_id', , resource_type='locations'),
            organization=dict(type='entity', flat_name='organization_id', resource_type='organizations'),
            enabled=dict(type='bool'),
            managed=dict(type='bool'),
            build=dict(type='bool'),
            **common_host_spec
        ),

[...]

resolve_entities()

```

in base class

```
def resolve_entities(self, entity_dict):
   entities = {key: value for key, value in self.entity_spec.itmes() if value['type'] == 'entity' and 'resource_type' in value}
    for entity_name, entity_opts in  entities.items():
        if entity_name in entity_dict:
            entity_dict[entity_name] = self.find_resource_by_name(entity_opts['resource_type'], name=entity_dict[entity_name], failsafe=False, thin=True)
    
```

mdellweg:
@bagasse I had a similar thought. Please keep on thinking that way. But as you said, it's out of scope here.
BTW: The difficulties i see there are from the fact that we sometimes search with `=` and with `~`.

evgeni:
@bagasse yeah, I also had similar idea and opened https://github.com/theforeman/foreman-ansible-modules/issues/324 a long time ago. we def should explore that, but not in this PR ;)

plugins/modules/foreman_hostgroup.py
```diff
@@ -283,11 +284,6 @@ def main():
             parent=dict(type='entity', flat_name='parent_id'),
             organizations=dict(type='entity_list', flat_name='organization_ids'),
             locations=dict(type='entity_list', flat_name='location_ids'),
-            compute_resource=dict(type='entity', flat_name='compute_resource_id'),
```
evgeni: looking at the APIdoc, the following parameters are allowed for both, host and hostgroup:
```
environment_id
compute_profile_id
compute_resource_id
operatingsystem_id
architecture_id
pxe_loader
medium_id
ptable_id
subnet_id
subnet6_id
domain_id
realm_id
config_group_ids
puppet_proxy_id
puppet_ca_proxy_id
root_pass
```

(the other *_proxy_id probably too, I only looked at foremans apidoc)

so, shall we move all?

plugins/modules/foreman_hostgroup.py
```diff
@@ -340,20 +337,7 @@ def main():
...
-
-        if 'domain' in entity_dict:
-            entity_dict['domain'] = module.find_resource_by_name('domains', name=entity_dict['domain'], failsafe=False, thin=True)
-
-        if 'subnet' in entity_dict:
-            entity_dict['subnet'] = module.find_resource_by_name('subnets', name=entity_dict['subnet'], failsafe=False, thin=True)
-
-        if 'subnet6' in entity_dict:
-            entity_dict['subnet6'] = module.find_resource_by_name('subnets', name=entity_dict['subnet6'], failsafe=False, thin=True)
+        module.handle_common_host_spec(entity_dict)
```
evgeni: I wrote that on IRC, but I'll repeat here so that others can comment too. This modifies `entity_dict` in-place and is kinda not obvious to the user. I wonder if `entity_dict = module.handle_common_host_spec(entity_dict)` would be more explicit?

evgeni:
Oh, and can we add a few tests for foreman_host that the new params are working?

plugins/modules/foreman_hostgroup.py
mdellweg: Probably yes. I started with a small random set to see if it works.

plugins/modules/foreman_host.py
```diff
@@ -150,6 +171,8 @@ def main():
         if 'hostgroup' not in entity_dict and entity_dict.get('managed', True):
             module.fail_json(msg='Hostgroup can be omitted only with managed=False')
 
+        entity_spec = module.handle_common_host_spec(entity_dict)
```
evgeni:
```suggestion
        entity_dict = module.handle_common_host_spec(entity_dict)
```

evgeni: (will also fix the linter error ;))
```
plugins/modules/foreman_host.py:174:9: F841 local variable 'entity_spec' is assigned to but never used
```

evgeni: (and that's why I wanted tests ;P)

mdellweg: (Yes, but the linter saved us. And it's tagged WIP)

evgeni:
*whistle* @mdellweg this needs a rebase after #637, sorriiiiiiiiiiiiiiiiiiiiiiiiiii

plugins/modules/foreman_hostgroup.py
```diff
@@ -370,7 +333,9 @@ def main():
             entity_dict['content_view'] = module.find_resource_by_name('content_views', name=entity_dict['content_view'],
                                                                        params=scope, failsafe=False, thin=True)
 
-    entity_dict = module.handle_taxonomy_params(entity_dict)
+        entity_dict = module.handle_common_host_spec(entity_dict)
```
evgeni: the taxonomy handler has `if not self.desired_absent`, so you don't *have* to move it inside here, but I wonder if we should standardize on doing the if or not in the handler.

mdellweg: Oh right. It was not called `maybe_handle_taxonomy_params` so i just assumed. ;)

I'm really not sure, which one is better.

evgeni:
Overall I'm happy with this!
I posted a question inline, more about how we should design these helper handlers (which we now have two) and I think we should decide whether we want to go on with the current set of params or add all the possible ones?

plugins/modules/foreman_hostgroup.py
evgeni: I slightly prefer to have the "if" in the handler, as this allows some modules to omit the if in their own code completely (those which don't reference anything but taxonomies).

mdellweg:
Rebuilt with a mixin and with the `if` inside.

plugins/modules/foreman_hostgroup.py
evgeni: opened https://github.com/theforeman/foreman-ansible-modules/issues/647 to track this

Issue #608, opened 2019-11-21
Several unicode problems all around


mdellweg:
##### SUMMARY
The modules do not work well for unicode in various ways. (Not yet finished exploring. Will probably add more.)

- [ ] identify *where* we have issues with unicode
  - [ ] which Ansible/Python/Apypie/Requests version
  - [ ] which module/type of request
  - [ ] special characters in passwords work in python2, not in python3 (maybe that was apypie 0.1.0)
  - [ ] special characters in name break idempotence of at least architecture and location in python2
- [ ] research how Ansible *expects* this to be handled (docs are rare)
- [ ] FIX
- [ ] add tests

##### ISSUE TYPE
 - Bug Report

mdellweg:
The only documentation to this i can find is:
https://docs.ansible.com/ansible/latest/dev_guide/developing_python_3.html

evgeni:
So I wondered how much Ansible and how much everything else is at play here, so I tried with plain Python+Apypie:

```
python2.7:
  works:
    apypie.Api(uri='https://foreman.example.com', username='üser', password='changeme', api_version=2, verify_ssl=False)
    apypie.Api(uri='https://cforeman.example.com', username='user', password='pässword', api_version=2, verify_ssl=False)
  fails:
    apypie.Api(uri='https://foreman.example.com', username=u'üser', password='changeme', api_version=2, verify_ssl=False) (500)
    apypie.Api(uri='https://foreman.example.com', username='user', password=u'pässword', api_version=2, verify_ssl=False) (401)

python3.7:
  works:
    apypie.Api(uri='https://foreman.example.com', username='üser'.encode('utf-8'), password='changeme', api_version=2, verify_ssl=False)
    apypie.Api(uri='https://foreman.example.com', username='user', password='pässword'.encode('utf-8'), api_version=2, verify_ssl=False)
  fails:
    apypie.Api(uri='https://foreman.example.com', username='üser', password='changeme', api_version=2, verify_ssl=False) (500)
    apypie.Api(uri='https://foreman.example.com', username='user', password='pässword', api_version=2, verify_ssl=False) (401)
```

the actual testing was done by executing `api.resource('hosts').call('index')` on the resulting API object.

it seems as long as we pass `bytes` (that's what Python2's `str` essentially is), `requests` does the right thing. Otherwise not.

btw, this is with Apypie 0.2.1 and Requests 2.22.0 on Fedora 30.

evgeni:
[...]

PR #609, opened 2019-11-21
Fix some random unicode problems

mdellweg:
Fixes: #608


evgeni:
looking at other modules in `ansible/ansible`, it seems that's the right-ish way to do it.

we also should replace all `str(e)` with `to_native(e)` so that our exceptions are proper too

evgeni:
and reading files should be to_bytes on the filename too.

mdellweg:
Yet still, this does not work everywhere, and i cannot really pin it down.

plugins/module_utils/foreman_helper.py
```diff
@@ -14,6 +14,7 @@
 from functools import wraps
 
 from ansible.module_utils.basic import AnsibleModule
+from ansible.module_utils._text import to_bytes
```
mdellweg: Is this meant seriously? I need a private interface to do the simple things in 'the right way'?

mdellweg:
Not entirely convinced here... But this is the best i could come up with.
If this does not look right as a whole, please consider the individual commits.

plugins/module_utils/foreman_helper.py
```diff
@@ -245,12 +246,12 @@ def check_requirements(self):
         if not HAS_APYPIE:
             self.fail_json(msg='The apypie Python module is required', exception=APYPIE_IMP_ERR)
 
-    @_exception2fail_json(msg="Failed to connect to Foreman server: %s ")
+    @_exception2fail_json(msg="Failed to connect to Foreman server: {0}")
     def connect(self):
         self.foremanapi = apypie.Api(
             uri=self._foremanapi_server_url,
             username=self._foremanapi_username,
```
evgeni: IMHO the username also should be `to_bytes`

plugins/modules/katello_upload.py
```diff
@@ -151,27 +151,24 @@ def main():
...
 
+    b_src = to_bytes(entity_dict['src'])
     filename = os.path.basename(entity_dict['src'])
 
-    checksum = hashlib.sha256()
-    with open(entity_dict['src'], 'rb') as contentfile:
-        for chunk in iter(lambda: contentfile.read(CONTENT_CHUNK_SIZE), b""):
-            checksum.update(chunk)
-    checksum = checksum.hexdigest()
+    checksum = module.sha256(entity_dict['src'])
```
evgeni: "oh!"

evgeni: also interesting how you have to pass the original `src` here, as `sha256 → digest_from_file` does `b_filename = to_bytes(filename, errors='surrogate_or_strict')` itself. but then again, shouldn't `to_bytes(to_bytes())` work?

plugins/module_utils/foreman_helper.py
```diff
@@ -453,8 +454,13 @@ def _update_entity(self, resource, desired_entity, current_entity, params, entit
         desired_entity = _flatten_entity(desired_entity, entity_spec)
         current_entity = _flatten_entity(current_entity, entity_spec)
         for key, value in desired_entity.items():
-            if current_entity.get(key) != value:
-                payload[key] = value
+            # String comparison needs extra care in face of unicode
```
evgeni: can we have a test for this? or is the `architecture_name` below testing this?

plugins/module_utils/foreman_helper.py
```diff
@@ -656,8 +659,18 @@ def _entity_spec_helper(spec):
...
             _dummy, argument_value['options'] = _entity_spec_helper(argument_value.pop('entity_spec'))
             entity_value = None
+        elif 'type' in argument_value:
+            entity_value['type'] = argument_value['type']
+
+        if flat_name:
+            entity_value['flat_name'] = flat_name
+            entity_spec[flat_name] = {}
+            if 'type' in argument_value:
+                entity_spec[flat_name]['type'] = argument_value.get('type', 'str')
```
evgeni: isn't the `or 'str'` in the `get()` superfluous, as you already checked that `type` is in `argument_value`?

plugins/modules/katello_upload.py
mdellweg: > `to_bytes(to_bytes())`

Yes i think that should do, but i think i have seen that failing. Maybe due to the missing `errors='surrogate_or_strict'`. And if they convert it anyway, why should i.

plugins/module_utils/foreman_helper.py
mdellweg: What about the url then? I think, DNS knows about unicode...

plugins/module_utils/foreman_helper.py
mdellweg: Well, we can ask for more, but the strangely appearing new processor architecture i invented there tests it.
To be more specific: On python3 it worked without this, on python2 it failed idempotence, because it compared bytes to unicode.

plugins/module_utils/foreman_helper.py
mdellweg: Right.
```suggestion
                entity_spec[flat_name]['type'] = argument_value['type']
```

plugins/module_utils/foreman_helper.py
evgeni: don't go to far ;)

plugins/modules/katello_upload.py
evgeni: yeah, it's just weird to read.

plugins/module_utils/foreman_helper.py
evgeni: that works for me.

evgeni:
@mdellweg shall this be squashed or rebased?

mdellweg:
It can be merged. The commits were only separate in case we only wanted them selectively.

m-bucher:
I believe https://github.com/psf/requests/pull/3673/files also plays a part in this.
Without this fix I did not manage to get it running

evgeni:
Probably. Seems to be part of requests v2.12.2.

PR #610, opened 2019-11-22
Add resolve mechanism for entities

bagasse:
Trying to address some parts of #324

#### foreman modules
- [x] foreman_architecture
- [x] foreman_auth_source_ldap
- [x] foreman_bookmark
- [x] foreman_compute_attribute
- [x] foreman_compute_profile
- [x] foreman_compute_resource
- [x] foreman_config_group
- [x] foreman_domain
- [x] foreman_environment
- [x] foreman_external_usergroup
- [x] foreman_global_parameter
- [x] foreman_hostgroup
-  foreman_host_power *(make actions, out of scope)*
- [x] foreman_host
- [x] foreman_installation_medium
- foreman_job_template (a lot of custom logic)
- [x] foreman_location
- [x] foreman_operatingsystem
- [x] foreman_organization
- [x] foreman_os_default_template
- [x] foreman_provisioning_template
- [x] foreman_ptable
- [x] foreman_realm
- [x] foreman_role
-  foreman_search_facts (low level, use ForemanAnsible module directly, out of scope)
- [x] foreman_setting
- foreman_snapshot (Needs vmware, no setup to test)
- [x] foreman_subnet
- [x] foreman_usergroup
- [x] foreman_user  (tests OK in my env but KO here)


bagasse:
Weird, test pass on my machine (Python 3.7.5, ansible 2.9.1)...

plugins/module_utils/foreman_helper.py
```diff
@@ -592,6 +594,46 @@ def __init__(self, argument_spec=None, **kwargs):
...
+                    if value.get('resolve', True) and 'type' in value and (value['type'] == 'entity' or value['type'] == 'entity_list')}
+        for entity_name, entity_opts in entities.items():
+            if entity_name in entity_dict:
+                if entity_opts['type'] == 'entity':
+                    entity_dict[entity_name] = self._resolve_entity(entity_name, entity_dict[entity_name], entity_opts)
+                else:
+                    entity_dict[entity_name] = self._resolve_entity_list(entity_name, entity_dict[entity_name], entity_opts)
+
+    def _resolve_entity(self, entity_name, value, entity_opts):
+        p = inflect.engine()
```
evgeni: I'd prefer not to have a dependency on `inflect`, plus apypie has an own inflector: https://github.com/Apipie/apypie/blob/master/apypie/inflector.py

bagasse: I wanted to add a note about this, but didn't make it. Yes, I was thinking about using an static hashmap or something else for pluralization. I've added inflect just to go ahead for the POC, this is realy WIP ;). Didn't checked how apypie managed that but if I can rely on apypie inflector that's nice

evgeni: the apypie one is the "same" (well, python port) as what hammer uses, so it should be what we need :)

bagasse:
I think I'm mostly done (at least on feature POV). Migrated all `foreman` modules, except following:
* `foreman_host_power`: make actions, out of scope
* `foreman_job_template`: make a lot of custom logic
* `foreman_search_facts`: low level, use ForemanAnsible module directly, out of scope
* `foreman_snapshot`: Needs vmware, no setup to record tests

plugins/module_utils/foreman_helper.py
```diff
@@ -590,10 +591,238 @@ def exit_json(self, changed=False, **kwargs):
...
+        if 'parent' in self.sub_entities:
+            self.entity_opts['search_by'] = 'title'
+        elif 'search_by' not in self.entity_opts:
+            self.entity_opts['search_by'] = self.entity_key
+
+        # Convert current class name from CamelCase to snake_case
+        class_name = re.sub(r'(?<=[a-z])[A-Z]|[A-Z](?=[^A-Z])', r'_\g<0>', self.__class__.__name__).lower().strip('_')
+        # Get entity name from snake case class name
+        self.entity_name = '_'.join(class_name.split('_')[1:-1])
+        p = apypie.Inflector()
```
evgeni: this makes the `ansible-test` for imports explode, and I don't totally understand why:

```
ERROR: plugins/modules/foreman_architecture.py:109:0: traceback: NameError: global name 'apypie' is not defined (at plugins/module_utils/foreman_helper.py:628:0)
```

plugins/module_utils/foreman_helper.py
```diff
@@ -590,10 +591,238 @@ def exit_json(self, changed=False, **kwargs):
...
+            self.entity_opts['failsafe'] = True
+        if 'parent' in self.sub_entities:
+            self.entity_opts['search_by'] = 'title'
+        elif 'search_by' not in self.entity_opts:
+            self.entity_opts['search_by'] = self.entity_key
+
+        # Convert current class name from CamelCase to snake_case
+        class_name = re.sub(r'(?<=[a-z])[A-Z]|[A-Z](?=[^A-Z])', r'_\g<0>', self.__class__.__name__).lower().strip('_')
+        # Get entity name from snake case class name
+        self.entity_name = '_'.join(class_name.split('_')[1:-1])
```
evgeni: I wonder if this is too much magic and we should just pass `entity_name` to `__init__`?

plugins/modules/foreman_hostgroup.py
```diff
@@ -267,133 +270,53 @@ def main():
...
 
-    hostgroup = module.ensure_entity('hostgroups', entity_dict, entity)
+    # Remove manualy organization from entity_spec if present because it is just used for katello's scoped resources
+    entity_dict.pop('organization', None)
 
-    if hostgroup:
-        scope = {'hostgroup_id': hostgroup['id']}
-        module.ensure_scoped_parameters(scope, entity, parameters)
+    entity = module.ensure_entity('hostgroups', entity_dict, entity)
+    module.ensure_entity_parameters(entity, entity_dict)
```
evgeni: check_mode is failing with

```
"msg": "Failed to list resource: Can't overwrite existing cassette ('fixtures/hostgroup-0.yml') in your current record mode ('none').\nNo match for the request (<Request (GET) https://foreman.example.com/api/hostgroups/-1/parameters?per_page=4294967296>) was found.\nFound 1 similar requests with 1 different matcher(s) :\n\n1 - (<Request (GET) https://foreman.example.com/api/hostgroups/5/parameters?per_page=4294967296>).\nMatchers succeeded : ['method', 'query_matcher_ignore_proxy', 'body_json_l2_matcher']\nMatchers failed :\npath - assertion failure :\n/api/hostgroups/-1/parameters != /api/hostgroups/5/parameters\n"
```

this is because you call `ensure_entity_parameters` for a "just" created hostgroup, but in check mode you didn't really create it, so the `id` you got back is `-1`.

plugins/module_utils/foreman_helper.py
```diff
@@ -590,10 +591,238 @@ def exit_json(self, changed=False, **kwargs):
...
+             and resource_type is defined. You can also pass 'failsafe' and 'thin' options via entity_spec.
+             Params available for subentities:
+               * resolve (boolean): True by default. If false, this entity will be excluded and not resolved.
+               * failsafe (boolean): False by default.
+               * thin (boolean): True by default.
+               * scope (str): None by default. Provide the scope of the resource (eg: 'organization') (TODO: manage array for 'nested' scope ?)
+                              Defined scope must be a key of entity_spec.
+        """
+        if entity_dict:
+            dict = entity_dict.copy()
```
evgeni: `dict` is a keyword, please name this somehow differently :)

plugins/modules/foreman_hostgroup.py
evgeni: but I also might be wrong here and @mdellweg needs to help :D

plugins/module_utils/foreman_helper.py
```diff
@@ -590,10 +591,238 @@ def exit_json(self, changed=False, **kwargs):
...
+        # Convert current class name from CamelCase to snake_case
+        class_name = re.sub(r'(?<=[a-z])[A-Z]|[A-Z](?=[^A-Z])', r'_\g<0>', self.__class__.__name__).lower().strip('_')
+        # Get entity name from snake case class name
+        self.entity_name = '_'.join(class_name.split('_')[1:-1])
+        p = apypie.Inflector()
+        self._entity_resource_path = p.pluralize(self.entity_name)
+        self.search = ""
+        super(ResolverEntityMixin, self).__init__(argument_spec=argument_spec, **kwargs)
+
+    def run(self, entity_dict=None, entity=None, connect=True, exit=True):
```
evgeni: I think I once talked with either @mdellweg or @ekohl about this (the semi-automatic modules), and we came up with the idea to have `connect()` as a context manager, so the api would be smth like:

```python
with module.connect():
    module.run()
```

this would allow us to avoid the `connect` and `exit` params and easily add steps before and after `run`.

What do you think?

mdellweg: If you meant the "there is no propoper destructor idiom in python" discussion, it was me. And i'd vote for it.

plugins/modules/foreman_hostgroup.py
mdellweg: I can only speculate what `ensure_entity_parameters` is for, but your analysis sounds reasonable.

evgeni:
the above is not a complete review :)
but I fear we managed to produce a few merge conflicts nevertheless and you'll have to rebase :(

plugins/modules/foreman_hostgroup.py
mdellweg: I think, in `ensure_entity_parameters` there should be a safeguard to not search for existing parameters if we already know, we _just_ created the resource.

plugins/module_utils/foreman_helper.py
bagasse: Yes I saw that and didn't understand...

plugins/module_utils/foreman_helper.py
bagasse: As you want, the spirit was "convention over configuration", and it feels redundant with the class name for me.

plugins/module_utils/foreman_helper.py
```diff
@@ -388,21 +388,23 @@ def find_resource(self, resource, search, name=None, params=None, failsafe=False
                 result = self.show_resource(resource, result['id'], params=params)
         return result
 
-    def find_resource_by_name(self, resource, name, **kwargs):
-        search = 'name="{0}"'.format(name)
-        kwargs['name'] = name
+    def find_resource_by(self, resource, search_field, value, **kwargs):
```
ekohl: I'm not sure I think this indirection is a great added value. Foreman is very much based on the search strings. The API is not obvious. Previously it was also possible to search using multiple fields at the same time which this API doesn't allow.

plugins/module_utils/foreman_helper.py
evgeni: I wouldn't block on this, the magic way is fine too :)

ekohl: I'd at least have a helper `get_entity_name_from_class` that you can individually test. You can also allow passing an `entity_name` to `__init__` but fall back on the automatic way.

plugins/module_utils/foreman_helper.py
ekohl: I don't recall talking about this, but it looks very nice.

plugins/module_utils/foreman_helper.py
```diff
@@ -643,10 +644,238 @@ def exit_json(self, changed=False, **kwargs):
...
+
+        if exit:
+            self.exit_json()
+        else:
+            return entity
+
+    def remove_sensitive_fields(self, entity):
+        """ Set fields with 'no_log' option to None """
+        if entity:
+            for blaclisted_field in self.blacklisted_fields:
```
ekohl: Typo (that's at least consistent with the next line):
```suggestion
            for blacklisted_field in self.blacklisted_fields:
```

plugins/modules/foreman_bookmark.py
```diff
@@ -115,15 +119,9 @@ def main():
     )
 
     entity_dict = module.clean_params()
+    module.search = 'name="{0}",controller="{1}"'.format(entity_dict['name'], entity_dict['controller'])
```
ekohl: I think this feels like an odd pattern. Setting properties and then using that later feels hard to follow. Not sure what would be a better design.

bagasse: Yes, moved to method param.

plugins/module_utils/foreman_helper.py
bagasse: Moved the 'magic' to method and added ability to pass it to `__init__`

plugins/module_utils/foreman_helper.py
evgeni: But `find_resource` still exists, this only combines the `find_resource_by_*` helpers, which never accepted more than one field.

ekohl: I was thinking about a `build_search_string` helper so you don't need so many versions. In the end I wasn't sure if it would really simplify things.

plugins/module_utils/foreman_helper.py
evgeni: Yeah, it was Matthias :)

plugins/module_utils/foreman_helper.py
evgeni: Ok, I *think* I know what happens here.

The `ResolverEntityMixin` is used *before* `ForemanAnsibleModule` in `ForemanEntityAnsibleModule`, so its `__init__` gets called first. Thus the call to `check_requirements` didn't happen yet

https://github.com/theforeman/foreman-ansible-modules/blob/efd20737a2542ec9facbf0b0717ae95f060433c0/plugins/module_utils/foreman_helper.py#L214
https://github.com/theforeman/foreman-ansible-modules/blob/efd20737a2542ec9facbf0b0717ae95f060433c0/plugins/module_utils/foreman_helper.py#L300-L302

But that's the line that is responsible to make the `import` test from `ansible-test` succeed - it checks whether the module properly reports failed imports.

Feel free to just call `check_requirements` in the first lines of `__init__`.

plugins/module_utils/foreman_helper.py
```diff
@@ -643,10 +644,241 @@ def exit_json(self, changed=False, **kwargs):
...
+        self.entity_resolve = kwargs.pop('entity_resolve', True)
+        self.entity_opts = kwargs.pop('entity_opts', {})
+        self.entity_name = kwargs.pop('entity_key', self.entity_name_from_class)
+        if 'thin' not in self.entity_opts:
+            self.entity_opts['thin'] = False
+        if 'failsafe' not in self.entity_opts:
+            self.entity_opts['failsafe'] = True
+        if 'parent' in self.sub_entities:
+            self.entity_opts['search_by'] = 'title'
+        elif 'search_by' not in self.entity_opts:
```
evgeni: Should `search_by` be able to override the `parent` behavior? Not sure we'd even need it, but I think this should be the "primary" way to change the magic.

plugins/module_utils/foreman_helper.py
```diff
@@ -643,10 +644,241 @@ def exit_json(self, changed=False, **kwargs):
...
+            self.entity_opts['search_by'] = self.entity_key
```
```suggestion
            self.entity_opts['search_by'] = self.entity_key
        else:
            self.entity_opts['search_by'] = 'name'
```

plugins/module_utils/foreman_helper.py
```diff
@@ -643,10 +644,241 @@ def exit_json(self, changed=False, **kwargs):
         super(ForemanAnsibleModule, self).exit_json(**kwargs)
 
 
-class ForemanEntityAnsibleModule(ForemanAnsibleModule):
+class ResolverEntityMixin(object):
+    """ This mixin add automatic entities resolution based on provided attributes/ sub entities options.
+        It add following options to entity_spec 'entity' and 'entity_list' types:
+        * search_by (str): Field used to search the sub entity. Defaults to 'name'.
```
```suggestion
        * search_by (str): Field used to search the sub entity. Defaults to 'name' unless `parent` was set, in which case it defaults to `title`.
```

plugins/module_utils/foreman_helper.py
```diff
@@ -643,10 +644,241 @@ def exit_json(self, changed=False, **kwargs):
...
+    def _resolve_entity(self, entity_name, value, entity_opts, params=None):
+        p = apypie.Inflector()
+        resource_path = entity_opts.get('resource_type', p.pluralize(entity_name))
+        if resource_path == 'operatingsystems':
+            return self.find_operatingsystem(value, params=params,
+                                             failsafe=entity_opts.get('failsafe', False),
+                                             thin=entity_opts.get('thin', True))
+        else:
+            return self.find_resource_by(resource_path,
+                                         entity_opts.get('search_by', 'name'), value,
```
evgeni: IMHO we can set the `name` default in the `__init__`, see suggestion there.

plugins/module_utils/foreman_helper.py
```diff
@@ -643,10 +644,241 @@ def exit_json(self, changed=False, **kwargs):
...
+                                         search_operator=entity_opts.get('search_operator', '='),
```
evgeni: same, I think `search_operator` should be set in `__init__`.

plugins/module_utils/foreman_helper.py
```diff
@@ -643,10 +644,241 @@ def exit_json(self, changed=False, **kwargs):
...
+                                         failsafe=entity_opts.get('failsafe', False),
+                                         thin=entity_opts.get('thin', True),
```
evgeni: `failsafe` and `thin` have defaults in `__init__`, so should not need it here.

plugins/module_utils/foreman_helper.py
```diff
@@ -643,10 +644,241 @@ def exit_json(self, changed=False, **kwargs):
...
+                                         params=params)
+
+    def _resolve_entity_list(self, entity_name, values, entity_opts, params=None):
```
evgeni: same comments as for `_resolve_entity` apply here too, as the code is very similar.

I'd *love* to de-duplicate this, but I don't see a *good* way to do so w/o risking readability.

plugins/module_utils/foreman_helper.py
```diff
@@ -691,12 +923,16 @@ def exit_json(self, **kwargs):
...
+    def blacklisted_fields(self):
+        return [key for key, value in self.entity_spec.items() if value.get('no_log', False)]
+
 
 class ForemanTaxonomicEntityAnsibleModule(ForemanEntityAnsibleModule):
     def __init__(self, argument_spec=None, **kwargs):
         spec = dict(
             organizations=dict(type='entity_list', flat_name='organization_ids'),
-            locations=dict(type='entity_list', flat_name='location_ids'),
+            locations=dict(type='entity_list', flat_name='location_ids', search_by='title'),
```
evgeni: I wonder if we could/should have a list of entities that should be `search_by=title` by default (locations, hostgroups, operatingsystems come to mind).

Or would that be too  much magic?

plugins/modules/foreman_compute_resource.py
```diff
@@ -230,16 +234,10 @@ def main():
 
     entity_dict = module.clean_params()
 
-    module.connect()
-
-    entity = module.find_resource_by_name('compute_resources', name=entity_dict['name'], failsafe=True)
-
-    entity_dict = module.handle_taxonomy_params(entity_dict)
+    if not module.desired_absent and 'provider' not in entity_dict:
```
evgeni: I think this changes the logic, as before the error was only raised when `entity=None`, so when we're creating a new one, now it's always raised when `state` is not `absent`.

plugins/module_utils/foreman_helper.py
```diff
@@ -643,10 +644,241 @@ def exit_json(self, changed=False, **kwargs):
...
-class ForemanEntityAnsibleModule(ForemanAnsibleModule):
+class ResolverEntityMixin(object):
+    """ This mixin add automatic entities resolution based on provided attributes/ sub entities options.
+        It add following options to entity_spec 'entity' and 'entity_list' types:
+        * search_by (str): Field used to search the sub entity. Defaults to 'name'.
+        * search_operator (str): Operator used to search the su entity. Defaults to '='.
+        * resource_type (str): Resource type used to build API resource PATH. Defaults to pluralized entity key.
+        It add following attributes:
+        * entity_search (str): Defautls to None. If provided, the base entity resolver will use this search query instead to try to build it.
+        * entity_key (str): field used to search current entity. Defaults to 'name'. If entityèspec contains 'parent', defaults to 'title'.
```
```suggestion
        * entity_key (str): field used to search current entity. Defaults to 'name'. If entity_spec contains 'parent', defaults to 'title'.
```

plugins/module_utils/foreman_helper.py
evgeni: Or maybe this should not have a default, and we rely on `entity_key`?

evgeni: I wonder if it would be easier to mark one entry in the `entity_spec` as `primary=True` and then not have the special `entity_key`?

plugins/module_utils/foreman_helper.py
```diff
@@ -643,10 +644,241 @@ def exit_json(self, changed=False, **kwargs):
...
+        entity = self.remove_sensitive_fields(entity)
+        self.ensure_entity_parameters(entity, entity_dict)
+
+        if exit:
+            self.exit_json()
+        else:
+            return entity
+
+    def remove_sensitive_fields(self, entity):
+        """ Set fields with 'no_log' option to None """
```
evgeni: Is that really necessary? I would have expected the api not to return those fields at all, and if it returns them, it's an API security bug?

Also, what's the reasoning behind `None`, instead of deleting this field completely?

plugins/module_utils/foreman_helper.py
```diff
@@ -643,10 +644,241 @@ def exit_json(self, changed=False, **kwargs):
...
+    @property
+    def _unscoped_sub_entities(self):
+        return {key: value for key, value in self.sub_entities.items() if 'scope' not in value}
+
+    @property
+    def _scoped_sub_entities(self):
+        return {key: value for key, value in self.sub_entities.items() if 'scope' in value}
+
+    def _resolve_entity(self, entity_name, value, entity_opts, params=None):
+        p = apypie.Inflector()
```
evgeni: you already used an inflector in `__init__`, should we store it there and re-use instead of creating a new one?

evgeni:
@bagasse posted you roughly 1000 questions :)

plugins/module_utils/foreman_helper.py
bagasse: I was thinking about this too, a list of entities that we don't search by name.

plugins/module_utils/foreman_helper.py
bagasse: I've just moved this logic from https://github.com/theforeman/foreman-ansible-modules/pull/610/files#diff-558c67c5fcc7107ec4a8a64c9f198440L345 to a central place, but didn't think about changing this. I think this was done for the diff output.

plugins/module_utils/foreman_helper.py
bagasse: You are right ! Done. Tell me if it's less readable

plugins/module_utils/foreman_helper.py
bagasse: But yes you are right, for me too it sounds an API bug

plugins/module_utils/foreman_helper.py
bagasse: This is used to resolve sub entities too, and I didn't set defaults for all sub entities for `search_by` `search_operator`, `thin` and `failsafe`

plugins/module_utils/foreman_helper.py
evgeni: oh, okay. if that;s old code, then I shut up, for now ;)

plugins/module_utils/foreman_helper.py
evgeni: hah, this looks good!

evgeni:
```
ERROR: plugins/modules/foreman_architecture.py:109:0: traceback: AttributeError: 'ForemanArchitectureModule' object has no attribute 'fail_json' (at plugins/module_utils/foreman_helper.py:302:0)
```

Huh? Why can't it resolve that?

plugins/module_utils/foreman_helper.py
bagasse: Maybe make this in another PR ? This one is already a big one... I submit a PR with the context manager and rebase this one on it when merged ? What do you think ?

bagasse:
@evgeni I think that's because of https://github.com/theforeman/foreman-ansible-modules/pull/610/commits/c3a3233a0e8fc235bd586b25f6c57dc7016e3bc9, the Mixin is not an Ansible module so self.fail_json don't work.

evgeni:
> @evgeni I think that's because of [c3a3233](https://github.com/theforeman/foreman-ansible-modules/commit/c3a3233a0e8fc235bd586b25f6c57dc7016e3bc9), the Mixin is not an Ansible module so self.fail_json don't work.

but shouldn't the test happen on the final `ForemanSomethingAnsibleModule`?

plugins/module_utils/foreman_helper.py
bagasse: @evgeni, @mdellweg, @ekohl I've added a context manager called `start()` (the name can change if you want, lack of inspiration)

bagasse:
> > @evgeni I think that's because of [c3a3233](https://github.com/theforeman/foreman-ansible-modules/commit/c3a3233a0e8fc235bd586b25f6c57dc7016e3bc9), the Mixin is not an Ansible module so self.fail_json don't work.
> 
> but shouldn't the test happen on the final `ForemanSomethingAnsibleModule`?

I really don't understand what's going on here...

plugins/module_utils/foreman_helper.py
```diff
@@ -228,6 +230,12 @@ def __init__(self, argument_spec, **kwargs):
         self.state = 'undefined'
         self.entity_spec = {}
 
+    @contextmanager
+    def start(self):
```
evgeni: I think I would have called it `api_connection`, so you end up with (eh), `with module.api_connection(): …`

plugins/module_utils/foreman_helper.py
```diff
@@ -655,10 +690,237 @@ def exit_json(self, changed=False, **kwargs):
...
+        # Convert current class name from CamelCase to snake_case
+        class_name = re.sub(r'(?<=[a-z])[A-Z]|[A-Z](?=[^A-Z])', r'_\g<0>', self.__class__.__name__).lower().strip('_')
+        # Get entity name from snake case class name
+        return '_'.join(class_name.split('_')[1:-1])
+
+    def run(self, entity_dict=None, entity=None, search=None):
+        """ Connect, get entity_dict, resolve entities, ensure entity, remove sensitive data, manage parameters and exit.
+            You can provide custom entity_dict and entity if custom workflow is needed before doing most of listed actions.
+            If your module have some custom workflow, you can opt-out some actions:
+            * Connect by setting connect=False
```
```suggestion
```

plugins/module_utils/foreman_helper.py
```diff
@@ -655,10 +690,237 @@ def exit_json(self, changed=False, **kwargs):
...
+            * Exit by setting exit=False. In this case it will return the current entity
```
```suggestion
```

plugins/module_utils/foreman_helper.py
```diff
@@ -655,10 +690,237 @@ def exit_json(self, changed=False, **kwargs):
...
+    def entity_name_from_class(self):
+        # Convert current class name from CamelCase to snake_case
+        class_name = re.sub(r'(?<=[a-z])[A-Z]|[A-Z](?=[^A-Z])', r'_\g<0>', self.__class__.__name__).lower().strip('_')
+        # Get entity name from snake case class name
+        return '_'.join(class_name.split('_')[1:-1])
+
+    def run(self, entity_dict=None, entity=None, search=None):
+        """ Connect, get entity_dict, resolve entities, ensure entity, remove sensitive data, manage parameters and exit.
+            You can provide custom entity_dict and entity if custom workflow is needed before doing most of listed actions.
+            If your module have some custom workflow, you can opt-out some actions:
```
```suggestion
```

plugins/module_utils/foreman_helper.py
```diff
@@ -655,10 +690,237 @@ def exit_json(self, changed=False, **kwargs):
...
+
+    @property
+    def entity_name_from_class(self):
+        # Convert current class name from CamelCase to snake_case
+        class_name = re.sub(r'(?<=[a-z])[A-Z]|[A-Z](?=[^A-Z])', r'_\g<0>', self.__class__.__name__).lower().strip('_')
+        # Get entity name from snake case class name
+        return '_'.join(class_name.split('_')[1:-1])
+
+    def run(self, entity_dict=None, entity=None, search=None):
+        """ Connect, get entity_dict, resolve entities, ensure entity, remove sensitive data, manage parameters and exit.
```
```suggestion
        """ Get entity_dict, resolve entities, ensure entity, remove sensitive data and manage parameters.
```

plugins/module_utils/foreman_helper.py
```diff
@@ -655,10 +690,237 @@ def exit_json(self, changed=False, **kwargs):
...
+            You can provide custom entity_dict and entity if custom workflow is needed before doing most of listed actions.
+            If your module have some custom workflow, you can opt-out some actions:
+            * Connect by setting connect=False
+            * Exit by setting exit=False. In this case it will return the current entity
+        """
+        if not entity_dict:
+            entity_dict = self.clean_params()
+
+        entity, entity_dict = self.resolve_entities(entity_dict, entity, search)
+        new_entity = entity is None
```
evgeni: so after staring at the old `hostgroup` code for too long (and not liking this `new_entity` thing because we pass `entity` to `ensure_scoped_params` exactly for that: know if the entity existed or not), I realized the following (shortened old code):

```python
    entity = module.find_resource_by_title('hostgroups', title=build_fqn(name, parent), failsafe=True)
    parameters = entity_dict.get('parameters')

    hostgroup = module.ensure_entity('hostgroups', entity_dict, entity)

    if hostgroup:
        scope = {'hostgroup_id': hostgroup['id']}
        module.ensure_scoped_parameters(scope, entity, parameters)
```

the `find_resource` result is stored in `entity`, but `ensure_entity` is stored in `hostgroup`. now, if we do `if hostgroup` we know that the hostgroup exists now, but then we pass the old (!) `entity` to `ensure_scoped_params`, which will be `None` if the hostgroup didn't exist before.

so to translate this to your code, we could get away with something like:

```python
entity, entity_dict = self.resolve_entities(entity_dict, entity, search)
new_entity = self.ensure_entity(self._entity_resource_path, self._cleanup_entity_dict(entity_dict), entity)
new_entity = self.remove_sensitive_fields(new_entity)
if new_entity:
    self.ensure_entity_parameters(entity, entity_dict)
return new_entity
```

plugins/module_utils/foreman_helper.py
```diff
@@ -655,10 +690,237 @@ def exit_json(self, changed=False, **kwargs):
         super(ForemanAnsibleModule, self).exit_json(**kwargs)
 
 
-class ForemanEntityAnsibleModule(ForemanAnsibleModule):
+class ResolverEntityMixin(object):
```
evgeni: This mixin is only used in `ForemanEntityAnsibleModule`, so I wonder if we should just merge it into that class, which would also probably solve our import test issue?

plugins/module_utils/foreman_helper.py
```diff
@@ -655,10 +690,237 @@ def exit_json(self, changed=False, **kwargs):
...
+    """ This mixin add automatic entities resolution based on provided attributes/ sub entities options.
+        It add following options to entity_spec 'entity' and 'entity_list' types:
```
evgeni: I think this should be added to https://github.com/theforeman/foreman-ansible-modules/blob/master/docs/developing.md#specification-of-the-entity_spec :)

plugins/module_utils/foreman_helper.py
```diff
@@ -655,10 +690,237 @@ def exit_json(self, changed=False, **kwargs):
...
+            self.entity_opts['search_operator'] = '='
+        if 'parent' in self.sub_entities:
+            self.entity_opts['search_by'] = 'title'
+        elif 'search_by' not in self.entity_opts:
+            self.entity_opts['search_by'] = self.entity_key
+        else:
+            self.entity_opts['search_by'] = 'name'
+
+        self.inflector = apypie.Inflector()
+        self._entity_resource_path = self.inflector.pluralize(self.entity_name)
```
evgeni: why not calling this `resource_name`? I find the `path` in the name weird (but it might just be me ;))

plugins/module_utils/foreman_helper.py
```diff
@@ -655,10 +690,237 @@ def exit_json(self, changed=False, **kwargs):
...
+                                                                    parent, thin=True, failsafe=self.desired_absent)
+                if self.desired_absent and entity_dict['parent'] is None:
+                    # Parent does not exist so just exit here
+                    return entity
+
+            value = build_fqn(entity_dict[self.entity_key], parent)
+        else:
+            value = entity_dict[self.entity_key]
+        # Get current entity with its scope if defined (eg: organization for katello resources)
+        if (self.entity_scope and self.entity_scope in entity_dict and type(entity_dict[self.entity_scope]) is str
```
```suggestion
        if (self.entity_scope and self.entity_scope in entity_dict and isinstance(entity_dict[self.entity_scope], str)
```

but also `str` might be wrong here and you get `unicode` or so. Ansible bundles `six`, so you can do `from ansible.module_utils import six` and then `isinstance(entity_dict[self.entity_scope], six.string_types)`

plugins/module_utils/foreman_helper.py
```diff
@@ -655,10 +690,237 @@ def exit_json(self, changed=False, **kwargs):
...
+           and self.entity_scope in self._unscoped_sub_entities):
+            entity_dict[self.entity_scope] = self._resolve_entity(self.entity_scope, entity_dict[self.entity_scope],
+                                                                  self._unscoped_sub_entities[self.entity_scope],
+                                                                  params={'failsafe': self.desired_absent})
+            if entity_dict[self.entity_scope]:
+                if type(entity_dict[self.entity_scope]) is dict and 'id' in entity_dict[self.entity_scope]:
```
```suggestion
                if isinstance(entity_dict[self.entity_scope], dict) and 'id' in entity_dict[self.entity_scope]:
```

plugins/module_utils/foreman_helper.py
```diff
@@ -655,10 +690,237 @@ def exit_json(self, changed=False, **kwargs):
...
+                    scope_id = entity_dict[self.entity_scope]['id']
+                else:
+                    scope_id = entity_dict[self.entity_scope]
+                scope = "{0}_id = {1}".format(self.entity_scope, scope_id)
+                entity = self._resolve_entity(self.entity_name, value, entity_opts=self.entity_opts, params=scope)
+            else:
+                self.fail_json(msg='Scope {0} not found for {1}'.format(str(self.entity_scope), self.entity_name))
```
```suggestion
                self.fail_json(msg='Scope {0} not found for {1}'.format(to_native(self.entity_scope), self.entity_name))
```

plugins/module_utils/foreman_helper.py
```diff
@@ -655,10 +690,237 @@ def exit_json(self, changed=False, **kwargs):
...
+        elif not self.entity_scope:
+            entity = self._resolve_entity(self.entity_name, value, entity_opts=self.entity_opts)
+        else:
+            self.fail_json(msg='Invalid scope {0} for {1}'.format(str(self.entity_scope), self.entity_name))
```
```suggestion
            self.fail_json(msg='Invalid scope {0} for {1}'.format(to_native(self.entity_scope), self.entity_name))
```

plugins/module_utils/foreman_helper.py
```diff
@@ -655,10 +690,237 @@ def exit_json(self, changed=False, **kwargs):
...
+
+    def _resolve_sub_entities(self, entity, entities, entity_dict):
+        for entity_name, entity_opts in {key: value for key, value in entities.items() if key in entity_dict}.items():
+            if 'scope' in entity_opts:
+                if entity_opts['scope'] == self.entity_name and entity:
+                    scope_id = entity['id']
+                elif entity_opts['scope'] in entity_dict:
+                    scope_id = entity_dict[entity_opts['scope']]['id']
+
+                scope = {'{}_id'.format(entity_opts['scope']): scope_id}
```
```suggestion
                scope = {'{0}_id'.format(entity_opts['scope']): scope_id}
```

plugins/module_utils/foreman_helper.py
```diff
@@ -655,10 +690,237 @@ def exit_json(self, changed=False, **kwargs):
...
+            else:
+                scope = None
+            if entity_opts['type'] == 'entity' and type(entity_dict[entity_name]) is str:
+                entity_dict[entity_name] = self._resolve_entity(entity_name, entity_dict[entity_name], entity_opts, params=scope)
+            elif (entity_opts['type'] == 'entity_list' and type(entity_dict[entity_name]) is list
```
```suggestion
            elif (entity_opts['type'] == 'entity_list' and isinstance(entity_dict[entity_name], list)
```

plugins/module_utils/foreman_helper.py
```diff
@@ -655,10 +690,237 @@ def exit_json(self, changed=False, **kwargs):
...
+                  and len(entity_dict[entity_name]) > 0 and type(entity_dict[entity_name][0]) is str):
```
```suggestion
                  and len(entity_dict[entity_name]) > 0 and isinstace(entity_dict[entity_name][0]), six.string_types):
```

(as above)

plugins/module_utils/foreman_helper.py
```diff
@@ -655,10 +690,237 @@ def exit_json(self, changed=False, **kwargs):
...
+            if 'scope' in entity_opts:
+                if entity_opts['scope'] == self.entity_name and entity:
+                    scope_id = entity['id']
+                elif entity_opts['scope'] in entity_dict:
+                    scope_id = entity_dict[entity_opts['scope']]['id']
+
+                scope = {'{}_id'.format(entity_opts['scope']): scope_id}
+            else:
+                scope = None
+            if entity_opts['type'] == 'entity' and type(entity_dict[entity_name]) is str:
```
```suggestion
            if entity_opts['type'] == 'entity' and isinstance(entity_dict[entity_name]), six.string_types):
```

plugins/modules/foreman_hostgroup.py
```diff
@@ -220,22 +220,39 @@
...
+        if len(current_puppetclasses) > 0:
+            for leftover_puppetclass in current_puppetclasses:
+                module.ensure_entity('hostgroup_classes', {}, {'id': leftover_puppetclass}, {'hostgroup_id': entity['id']}, state='absent', entity_spec={})
+
+
 def main():
-    module = ForemanHostgroupAnsibleModule(
+    module = ForemanHostgroupModule(
         entity_spec=dict(
+            organization=dict(type='entity', flat_name='organization_id', ensure=False),
```
evgeni: this `ensure=False` kinda feels like "we put this into the entity_spec, but it's not really part of the entity" (and that matches the old code, we had org and puppetclasses in the regular argument spec). I wonder if that's really worth the added complexity.

plugins/module_utils/foreman_helper.py
mdellweg: You can make the module itself a context manager so you would have 'with module:'.

plugins/module_utils/foreman_helper.py
bagasse: Yes, I wanted to keep it as a Mixin at least at start, to avoid possible merge conflicts, but that make sense to merge it, and yes it solve import problems.

plugins/module_utils/foreman_helper.py
bagasse: Thanks for pointer ! didn't noticed that

plugins/module_utils/foreman_helper.py
bagasse: Yes, but I wanted to avoid making any connection to the server before validating `entity_dict` params that can be validated before opening connection, like it is in some modules.

evgeni:
tests failing due to vcr, see #668

evgeni:
```
ERROR: Found 6 pylint issue(s) which need to be resolved:
ERROR: plugins/modules/foreman_compute_attribute.py:108:8: blacklisted-name: Black listed name "_"
ERROR: plugins/modules/foreman_external_usergroup.py:96:8: blacklisted-name: Black listed name "_"
ERROR: plugins/modules/foreman_installation_medium.py:135:12: blacklisted-name: Black listed name "_"
ERROR: plugins/modules/foreman_os_default_template.py:112:8: blacklisted-name: Black listed name "_"
ERROR: plugins/modules/foreman_provisioning_template.py:345:12: blacklisted-name: Black listed name "_"
ERROR: plugins/modules/foreman_ptable.py:293:12: blacklisted-name: Black listed name "_"
```

Right, Ansible's pylint config forbids `_` (You can name things `_foo`, just not literal `_`).

plugins/module_utils/foreman_helper.py
```diff
@@ -671,6 +747,186 @@ def __init__(self, argument_spec=None, **kwargs):
...
+        if entity:
+            for blacklisted_field in self.blacklisted_fields:
+                entity[blacklisted_field] = None
+        return entity
+
+    def ensure_entity_parameters(self, entity, entity_dict):
+        if (entity and 'parameters' in self.original_entity_spec
+           and self.original_entity_spec['parameters'].get('type', 'str') == 'nested_list' and 'parameters' in entity_dict):
+            parameters = entity_dict.get('parameters')
+            scope = {self.entity_name + '_id': entity['id']}
```
evgeni: could you please use `.format` instead of `+`?

evgeni:
m)
I produced a (trivial) conflict in `plugins/modules/foreman_realm.py`, sorry

plugins/module_utils/foreman_helper.py
```diff
@@ -671,6 +747,186 @@ def __init__(self, argument_spec=None, **kwargs):
...
+            Returns entity.
+        """
+        if not entity_dict:
+            entity_dict = self.clean_params()
+
+        entity, entity_dict = self.resolve_entities(entity_dict, entity, search)
+        new_entity = self.ensure_entity(self._entity_resource_name, self._cleanup_entity_dict(entity_dict), entity)
+        new_entity = self.remove_sensitive_fields(new_entity)
+
+        if new_entity:
```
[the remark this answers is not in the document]
bagasse: I think it's good to have these checks in this method, if we want to use `ensure_entity_parameters` directly in a module, for a custom workflow, we don't have to handle  these checks in the module another time.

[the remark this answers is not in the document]
evgeni: I'm fine with that :)

plugins/modules/foreman_compute_attribute.py
```diff
@@ -88,29 +88,27 @@
...
-    entity = next((item for item in compute_attributes if item.get('compute_profile_id') == entity_dict['compute_profile']['id']), None)
-
-    module.ensure_entity('compute_attributes', entity_dict, entity)
-
-    module.exit_json()
+    with module.api_connection():
+        _entity, entity_dict = module.resolve_entities(entity_dict)
+        compute_attributes = entity_dict['compute_resource'].get('compute_attributes')
+        entity = next((item for item in compute_attributes if item.get('compute_profile_id') == entity_dict['compute_profile']['id']), None)
+        module.run(entity_dict, entity)
```
mdellweg: It seems like every parameter to `run` is describing an exception. Therefore i would pass them all by name.

plugins/modules/foreman_domain.py
```diff
@@ -83,43 +83,27 @@
...
         argument_spec=dict(
             updated_name=dict(),
         ),
         entity_spec=dict(
             name=dict(required=True),
             description=dict(aliases=['fullname'], flat_name='fullname'),
-            dns_proxy=dict(type='entity', flat_name='dns_id', aliases=['dns']),
+            dns_proxy=dict(type='entity', flat_name='dns_id', aliases=['dns'], resource_type='smart_proxies'),
+            locations=dict(type='entity_list', flat_name='location_ids'),
+            organizations=dict(type='entity_list', flat_name='organization_ids'),
```
mdellweg: shouldn't this be handled by the `ForemanTaxonomicEntityAnsibleModule`?

docs/developing.md
```diff
@@ -29,6 +59,11 @@ The module must handle the entities separately.
 See domain parameters in [`foreman_domain`](plugins/modules/foreman_domain.py) for an example.
 The sub entities must be described by `entity_spec=<sub_entity>_spec`.
 * `type='invisible'` The parameter is available to the API call, but it will be excluded from Ansible's `argument_spec`.
+* `search_by='login'`: Used with `type='entity'` or `type='entity_list'`. Field used to search the sub entity. Defaults to 'name' unless `parent` was set, in which case it defaults to `title`.
```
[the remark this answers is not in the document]
bagasse: the "Defaults to 'name' unless `parent` was set, in which case it defaults to `title`." will be removed, as I implemented `resolve_entity_key()` based on a predefined list as suggested by @evgeni. I forgot to reflect this in the docs.

plugins/modules/foreman_bookmark.py
```diff
@@ -116,14 +120,8 @@ def main():
...
-    module.connect()
-
-    search = 'name="{0}",controller="{1}"'.format(entity_dict['name'], entity_dict['controller'])
-    entity = module.find_resource('bookmarks', search, failsafe=True)
-
-    module.ensure_entity('bookmarks', entity_dict, entity)
-
-    module.exit_json()
+    with module.api_connection():
+        module.run(search='name="{0}",controller="{1}"'.format(entity_dict['name'], entity_dict['controller']))
```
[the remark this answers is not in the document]
bagasse: As this is not related to this change, and I want to avoid to rerecord tests if possible, I prefer to not handle that in this pr.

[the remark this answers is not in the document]
mdellweg: Now i see this is _in_ the search string. We probably don't want that change anyway. Suggestion withdrawn.

plugins/modules/foreman_hostgroup.py
bagasse: Only organization was in the argument spec, puppetclasses was already there and popped before ensure call in the module. For the organization part, I think it was in arg_spec instead of entity_spec because hostgroup module have to handle 'bare' foreman workflow and katello workflow (which needs an org). It's the only module that do that. I understand your point but put it in entity like this will keep on 'declarative way' instead of writing code on module side. I can add a comment before this line to explain or fallback to the 'old way', or keep it 'as is' as you want.

evgeni:
Hmm.

```
commit 5e0de93efc6d5a5c179656f473f8d6f0003cee26 (refs/bisect/bad)
Author: Baptiste Agasse <baptiste.agasse@gmail.com>
Date:   Thu Jan 2 19:26:22 2020 +0100

    fix compute_resource behavior when provider is absent
```

this broke the CR tests :(

plugins/modules/foreman_hostgroup.py
evgeni: Yeah, it's there for Katello-related params (CV etc), which is scoped to a single org.

I think my main source of dislike comes from the fact that `foreman_hostgroup` is the only module that uses `ensure=False` (for puppetclasses and org), which makes me thing this special handling should happen in the module, not in the helper (so "old way" if you want it). But then again the old way was not 100% consistent too (as you correctly pointed out).

Slight preference to put both into `arg_spec` and let the module handle it.

@mdellweg?

evgeni:
And it'd be great if GH would not hide comments, just because they started on old code and are now on "outdated" parts of the PR. See https://github.com/theforeman/foreman-ansible-modules/pull/610#discussion_r362805337

bagasse:
> Hmm.
> 
> ```
> commit 5e0de93efc6d5a5c179656f473f8d6f0003cee26 (refs/bisect/bad)
> Author: Baptiste Agasse <baptiste.agasse@gmail.com>
> Date:   Thu Jan 2 19:26:22 2020 +0100
> 
>     fix compute_resource behavior when provider is absent
> ```
> 
> this broke the CR tests :(
> 
> which is "funny" as it's targeting one of my review comments in [#610 (comment)](https://github.com/theforeman/foreman-ansible-modules/pull/610#discussion_r358175165)
> 
> but probably due to the changed api interactions to get to this point

Yes, and https://github.com/theforeman/foreman-ansible-modules/pull/610/commits/96d84fbc29049d90f194caba57f2efd0c8a33b10 broke ansible sanity check once again (on apypie import)... ansible sanity check will make me crazy...

plugins/modules/foreman_hostgroup.py
mdellweg: Do we need `ensure=False` or can we take it out of the `entity_dict` after all the lookup and pass that to `run`?

bagasse: Reverted that and manage it before calling `run()`

plugins/modules/foreman_compute_resource.py
bagasse: reverted to the previous logic.

bagasse:
Ok, @evgeni, I think I'm mostly done, except for the apypie import error on ansible sanity checks introduced in https://github.com/theforeman/foreman-ansible-modules/commit/96d84fbc29049d90f194caba57f2efd0c8a33b10 I really don't understand why ansible sanity is whining on this...

evgeni:
> Ok, @evgeni, I think I'm mostly done, except for the apypie import error on ansible sanity checks introduced in [96d84fb](https://github.com/theforeman/foreman-ansible-modules/commit/96d84fbc29049d90f194caba57f2efd0c8a33b10) I really don't understand why ansible sanity is whining on this...

This happens because you use `apypie` in `_entity_spec_helper`, which is a class-less function, so the guard we have in the `__init__` of `ForemanAnsibleModule` doesn't catch it.
You could move the function inside the class and then use `self.inflector`? Will require adjusting the tests in `tests/test_entity_spec_helper.py` tho, as they currently test the helper w/o any `AnsibleModule` (which is fine at the moment)

bagasse:
> > Ok, @evgeni, I think I'm mostly done, except for the apypie import error on ansible sanity checks introduced in [96d84fb](https://github.com/theforeman/foreman-ansible-modules/commit/96d84fbc29049d90f194caba57f2efd0c8a33b10) I really don't understand why ansible sanity is whining on this...
> 
> This happens because you use `apypie` in `_entity_spec_helper`, which is a class-less function, so the guard we have in the `__init__` of `ForemanAnsibleModule` doesn't catch it.
> You could move the function inside the class and then use `self.inflector`? Will require adjusting the tests in `tests/test_entity_spec_helper.py` tho, as they currently test the helper w/o any `AnsibleModule` (which is fine at the moment)

I already tried to move `_entity_spec_helper()` to `ForemanAnsibleModule` when I worked on https://github.com/theforeman/foreman-ansible-modules/commit/96d84fbc29049d90f194caba57f2efd0c8a33b10 some days ago with no luck and so reverted that before pushing.

evgeni:
Damn, that would have been too easy

evgeni:
Weeell... This "fixes" it:

```patch
diff --git a/plugins/module_utils/foreman_helper.py b/plugins/module_utils/foreman_helper.py
index 75478b3..12f4259 100644
--- a/plugins/module_utils/foreman_helper.py
+++ b/plugins/module_utils/foreman_helper.py
@@ -1028,7 +1028,8 @@ def _entity_spec_helper(spec):
     """
     entity_spec = {'id': {}}
     argument_spec = {}
-    inflector = apypie.Inflector()
+    if HAS_APYPIE:
+        inflector = apypie.Inflector()
     for key, value in spec.items():
         entity_value = {}
         argument_value = value.copy()
@@ -1041,7 +1042,8 @@ def _entity_spec_helper(spec):
         elif argument_value.get('type') == 'entity_list':
             argument_value['type'] = 'list'
             if not flat_name:
-                flat_name = '{0}_ids'.format(inflector.singularize(key))
+                if HAS_APYPIE:
+                    flat_name = '{0}_ids'.format(inflector.singularize(key))
             entity_value['type'] = 'entity_list'
         elif argument_value.get('type') == 'nested_list':
             argument_value['type'] = 'list'

```

Explanation:

`_entity_spec_helper` is called *before* we call `__init__` of `ForemanAnsibleModule` and thus *before* the `if HAS APYPIE` check happens:
https://github.com/theforeman/foreman-ansible-modules/blob/c171957fd74c7de98f5838c9c160e288a8f23c77/plugins/module_utils/foreman_helper.py#L660-L668

This is absolutely correct, as we need to know the *`argument_spec`* of the *Entity* module before calling stuff from the non-entity part.

Now the next logical solution would be to call `AnsibleModule.fail_json` inside `_entity_spec_helper`, but again, as we didn't call `ForemanAnsibleModule.__init__` yet, we also didn't call `AnsibleModule.__init__` yet and `fail_json` will fail (yuk).

So we need to make `_entity_spec_helper` work when `apypie` is not available, but can't rely on our usual safe nets. Let's see!

1. In which situations can we end up calling `_entity_spec_helper` without having `apypie` available?
    1. when the user calls the module but doesn't have the right Python libraries installed
    1. when Ansible generates docs from the `argument_spec`
1. Do we care if the generated specs are slightly off in those cases?
    1. no, as the spec will never be used anyways, `__init__` will bail out
    1. yes for `argument_spec`, but no for `entity_spec`
1. You use the `inflector` to set `flat_name` which then ends up in `entity_spec`, `argument_spec` can be correctly generated w/o the Inflector.
1. Write the above patch

evgeni:
Well, let's break the world!

Thanks a ton @bagasse. The next round of favorite beverage is on me, erm, @RedHatOfficial ;)

bagasse:
Thanks a lot @evgeni, @mdellweg and @ekohl for reviews/comments !

mdellweg:
Thanks for taking the idea of metaprogrammed modules that far.

ogajduse:
@bagasse You introduced the following line to the docstring, but I can not find it anywhere in the code. Can you please comment on this?
https://github.com/theforeman/foreman-ansible-modules/blob/7a715b0cd4b167bb464bfba410342972256393e1/plugins/module_utils/foreman_helper.py#L858

I hit this while developing a new `foreman_puppet_autosign` module (#770). I'm trying to get it working, but I can not get rid of searching for the `autosign` entity by `search_by` and `search_param`. `foreman_helper` is now forcing using search while searching for entities.

Looking at the Foreman API doc, I can not see `autosign` GET endpoint supports `search` field.
How would you suggest handling this?

bagasse:
@ogajduse, my bad, it was in a previous 'version' of this PR, it was replaced by the 'search' parameter in the `resolve_entities` method, and forgot to remove it. Anyway a new implementation of entity lookup by @mdellweg was merged recently, maybe you should take a look at it instead. Let me know If you have any other question :)

ogajduse:
@bagasse Thanks for the prompt response! I will hack on this a bit by myself and ask @mdellweg for additional help if needed.  Working on it requires a deep dive into `foreman_helper`. :smiley:

PR #614, opened 2019-11-25
use pytest-xdist to spread the load

evgeni:


evgeni:
before the actual test step took 8-9 minutes on GH, now it's more 4-5.

PR #617, opened 2019-11-28
add a release workflow to publish to galaxy

evgeni:


.github/workflows/release.yml
```diff
@@ -0,0 +1,24 @@
+name: release
+
+on:
+  push:
+    tags:
+      - v*
```
evgeni: can we check if this is a number?

evgeni:
@ehelms want to have the honor here? :)

.github/workflows/release.yml
```diff
@@ -0,0 +1,24 @@
+name: release
+
+on:
+  push:
+    tags:
+      - v[0-9]+.[0-9]+.[0-9]+
```
mdellweg: :+1:

ehelms: I never have liked to the little `v` in the tags but looking at releases that ship has sailed.

ehelms:
I don't suppose we could push a version bump and it could do the tag and release for us?

evgeni:
> I don't suppose we could push a version bump and it could do the tag and release for us?

actually, we could by patch matching the galaxy.yml and then doing some magic. but I prefer to do the tags by hand (and sign them).

PR #618, opened 2019-11-29
add kickstart_repository option to hostgroup

### LINKED ISSUE #526: foreman_hostgroup: add kickstart_repository option
##### SUMMARY
kickstart_repository_id of the API is missing in the module.
Without it one cannot install hosts from synced kickstart repos instead of media. 

##### ISSUE TYPE
 - Feature Idea

nesanton:
Fixes #526 

Could not write tests using any of the public mock repos. Need something like RHEL's kickstart to enable "synced content" option on "Operating System" tab. If anyone hints how to - I'd be happy to add tests.
For Satellite/RHEL this would be quite a common use case.

### DIFF
--- a/plugins/modules/foreman_hostgroup.py
+++ b/plugins/modules/foreman_hostgroup.py
@@ -161,7 +161,7 @@ options:
     required: false
     type: str
   kickstart_repository:
-    description: Kickstart repository name
+    description: Kickstart repository name. Only available for katello installations.
     required: false
     type: str
   content_view:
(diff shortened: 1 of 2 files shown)


plugins/modules/foreman_hostgroup.py
```diff
@@ -160,6 +160,10 @@
     description: Katello Lifecycle environment. Only available for katello installations.
     required: false
     type: str
+  kickstart_repository:
+    description: Kickstart repository name
```
```suggestion
    description: Kickstart repository name. Only available for katello installations.
```

Issue #574, opened 2019-11-01
search_facts: add tests for params and org


evgeni:
In #537 and #573 we added params and organization options, but not tests for them. Fix that ;)

PR #619, opened 2019-11-29
add search_facts tests for organization and params

evgeni:
Fixes: #574

### DIFF
--- a/tests/test_playbooks/search_facts.yml
+++ b/tests/test_playbooks/search_facts.yml
@@ -3,10 +3,21 @@
   gather_facts: false
   vars_files:
     - vars/server.yml
+    - vars/search_facts.yml
   tasks:
     - include: tasks/organization.yml
       vars:
+        organization_name: "{{ item.organization }}"
         organization_state: present
+      loop: "{{ test_resources }}"
+    - include: tasks/domain.yml
+      vars:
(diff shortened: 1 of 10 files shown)


tests/test_playbooks/search_facts.yml
```diff
@@ -3,10 +3,21 @@
   gather_facts: false
   vars_files:
     - vars/server.yml
+    - vars/search_facts.yml
   tasks:
     - include: tasks/organization.yml
       vars:
+        organization_name: "{{ item.organization }}"
         organization_state: present
+      loop: "{{ test_resources }}"
```
mdellweg: I am not sure, i understand the purpose of the `test_reseources` variable, but should we add a `when: item.organization` here?

evgeni: oh the purpose was purely that I don't have to write individual create/destroy calls and not to repeat the same data multiple times.

IMHO item.org will always be set, so no need for the when

mdellweg: As the list is rather short, we could just explode it here...

evgeni: I'd prefer to keep it in a vars file. Would a different variable name help in any way?

PR #621, opened 2019-11-29
Use list for extends_documentation_fragment

mdellweg:
This is why sed is not called regex_replace. ;)


evgeni:
It should probably be called "the clever replace" instead. Thanks!

PR #622, opened 2019-12-02
Move nested parameter doc to fragment

### LINKED ISSUE #447: Make common parameter docs fragments?
I was talking to a user on irc today (hi strangeglyph!) and they reported that the use of operating system as a param to other modules is confusing (is it name/description/whatever?). While this might be an issue on its own (strangeglyph promised to submit wording fixes), I thought we might want to move certain params (org, loc, os) to own doc fragments and thus make it easier to update their wording across-the-board?

mdellweg:
fixes #447

### DIFF
--- a/plugins/doc_fragments/foreman.py
+++ b/plugins/doc_fragments/foreman.py
@@ -80,3 +80,28 @@ options:
           - 'json'
         type: str
 '''
+
+    OS_FAMILY = '''
+options:
+  os_family:
+    description:
+      - The OS family the entity shall be assigned with.
+    required: false
+    choices:
+      - AIX
+      - Altlinux
+      - Archlinux
+      - Coreos
+      - Debian
+      - Freebsd
+      - Gentoo
+      - Junos
+      - NXOS
+      - Rancheros
+      - Redhat
+      - Solaris
+      - Suse
+      - Windows
+      - Xenserver
+    type: str
+'''
--- a/plugins/doc_fragments/foreman.py
+++ b/plugins/doc_fragments/foreman.py
@@ -45,3 +45,38 @@ options:
     default: true
     type: bool
 '''
+
+    NESTED_PARAMETERS = '''
+options:
+  parameters:
+    description:
+      - Entity domain specific host parameters
+    required: false
+    type: list
+    elements: dict
+    suboptions:
+      name:
+        description:
+          - Name of the parameter
+        required: true
+        type: str
+      value:
+        description:
+          - Value of the parameter
+        required: true
+        type: raw
+      parameter_type:
+        description:
+          - Type of the parameter
+        default: 'string'
+        choices:
+          - 'string'
+          - 'boolean'
+          - 'integer'
+          - 'real'
+          - 'array'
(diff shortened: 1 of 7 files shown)


evgeni:
I think we can move more :D

```
plugins/modules/foreman_ptable.py:
    os_family:
      description: The OS family the template shall be assigned with.
plugins/modules/foreman_installation_medium.py:
    os_family:
      description: The OS family the template shall be assigned with. If no os_family is set but a operatingsystem, the value will be derived from it.
plugins/modules/foreman_operatingsystem.py:
    family:
      description:
        - distribution family of the Operating System
```

(the one in operatingsystem is probably not foldable into the other two because of the other name)

plugins/doc_fragments/foreman_parameters.py
```diff
@@ -0,0 +1,59 @@
+# (c) 2019, Matthias Dellweg <dellweg@atix.de>
```
evgeni: you can have multiple fragments in one file, which I think we can use here?

https://docs.ansible.com/ansible/latest/dev_guide/developing_modules_documenting.html#documentation-fragments

mdellweg:
> I think we can move more :D

I'm not sure, it is worth the trouble there (os_family). With parameters, that corresponds nicely to the shared code that handles the nested entities.

plugins/doc_fragments/foreman_parameters.py
mdellweg: TIL (and it's still quite early...)

evgeni:
> > I think we can move more :D
> 
> I'm not sure, it is worth the trouble there (os_family). With parameters, that corresponds nicely to the shared code that handles the nested entities.

are you saying we don't have the list of supported OSes shared? ;-)

https://github.com/theforeman/foreman-ansible-modules/blob/87781573d571f7e89eab20feeaddb55a92991878/plugins/module_utils/foreman_helper.py#L747-L763

plugins/doc_fragments/foreman.py
```diff
@@ -45,3 +45,38 @@ class ModuleDocFragment(object):
     default: true
     type: bool
 '''
+
+    NESTED_PARAMETERS = '''
+options:
+  parameters:
+    description:
```
evgeni: I think if you have `description` set here *and* in the individual modules, you end up with both descriptions joined. Which is not what we want.

mdellweg: I tested, and at least ansible-doc only takes the one from the module. So i let it be a fallback description.

mdellweg:
> > > I think we can move more :D
> > 
> > 
> > I'm not sure, it is worth the trouble there (os_family). With parameters, that corresponds nicely to the shared code that handles the nested entities.
> 
> are you saying we don't have the list of supported OSes shared? ;-)

OIC. Well...

plugins/doc_fragments/foreman.py
evgeni: ok, then my memory is playing games with me. thanks for testing.

Issue #493, opened 2019-09-23
cleanly error out when plugin is not available


evgeni:
##### SUMMARY
Currently, when you run a Katello module against a plain Foreman, you'll get a "unknown resource" error from apypie, which is not really helpful for users. I'll suggest teaching our modules (or rather the abstraction layer) to check for a specific resource and cleanly error out when that is not available.

##### ISSUE TYPE
 - Feature Idea

PR #624, opened 2019-12-02
nicer errors when plugin resources are not found

evgeni:
when using a Katello module against a non-Katello setup, this will now
error:

    localhost | FAILED! => {
        "changed": false,
        "msg": "Failed to connect to Foreman server: The server does not seem to have the Katello plugin installed. "
    }

and on any other missing plugin:

    localhost | FAILED! => {
        "changed": false,
        "msg": "Failed to list resource: The server doesn't know about products, is the right plugin installed?"
    }

Fixes: #493


evgeni:
hah, of course, I broke search_facts.

PR #633, opened 2019-12-04
correct required attributes spotted by ansible-test sanity

evgeni:
doc-required-mismatch errors was introduced in https://github.com/ansible/ansible/commit/4be8b2134f0f6ed794ef57a621534f9561f91895


mdellweg:
As this pops up on the devel builds, it seems Ansible has gotten better checks... ;)

PR #635, opened 2019-12-04
add host limits and system purpose to k_activation_key

evgeni:


tests/test_playbooks/activation_key.yml
```diff
@@ -165,6 +165,24 @@
         activation_key_state: absent
         expected_change: true
 
+    - name: create AK with unlimited_hosts=false
```
evgeni: I only added the test for this one as it's a bit special

evgeni:
I have a weird idea how to fix the boilerplate. Need beers ;)

PR #637, opened 2019-12-05
move taxonomy handling to the helper

evgeni:
inspired by #607 and #622 :)

### DIFF
--- a/plugins/doc_fragments/foreman.py
+++ b/plugins/doc_fragments/foreman.py
@@ -105,3 +105,13 @@ options:
       - Xenserver
     type: str
 '''
+
+    TAXONOMY = '''
+options:
+  organizations:
+    description: List of organizations the entity should be assigned to
+    type: list
+  locations:
+    description: List of locations the entity should be assigned to
+    type: list
+'''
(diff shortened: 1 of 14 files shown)


plugins/doc_fragments/foreman.py
```diff
@@ -105,3 +105,13 @@ class ModuleDocFragment(object):
       - Xenserver
     type: str
 '''
+
+    TAXONOMY = '''
+options:
+  organizations:
+    description: List of organizations the entity should be assigned to
```
evgeni: I did not override the description in the various modules, as I find "the entity" exact enough (plus it would make the doc fragment almost useless, as it only contains description and type.)

plugins/module_utils/foreman_helper.py
```diff
@@ -639,6 +639,29 @@ def exit_json(self, **kwargs):
         super(ForemanEntityAnsibleModule, self).exit_json(**kwargs)
 
 
+class ForemanTaxonomicEntityAnsibleModule(ForemanEntityAnsibleModule):
```
evgeni: If anyone has a better name, please!

mdellweg:
Do you want to add a handle_organization to the katelloMixin here?

evgeni:
> Do you want to add a handle_organization to the katelloMixin here?

No, I wanted to do katello-related stuff later.

PR #639, opened 2019-12-05
move state documentation to a fragment

evgeni:


plugins/modules/foreman_installation_medium.py
```diff
@@ -92,6 +88,7 @@ def main():
     module = ForemanTaxonomicEntityAnsibleModule(
         argument_spec=dict(
             updated_name=dict(),
+            state=dict(default='present', choices=['present', 'present_with_defaults', 'absent']),
```
evgeni: yes, the code below has handling for "with_defaults", but it was missing from the definition m)

Issue #453, opened 2019-09-06
katello_repository_set - Won't enable RHEL 8 repo if basearch is set


rhjhunt:
##### SUMMARY
When using the module **katello_repository_set** to enable a RHEL 8 repository if _basearch_ is set to x86_64 no action is performed. If _basearch_ isn't defined the repository is enabled but placed under `noarch`.

![noarch](https://user-images.githubusercontent.com/384643/64403500-00199a00-d036-11e9-9f75-568c7ac1e861.png)

##### ISSUE TYPE
 - Bug Report

##### ANSIBLE VERSION

```console
ansible 2.8.4
  config file = /home/jhunt/git/vagabond/ansible.cfg
  configured module search path = ['/home/jhunt/git/vagabond/library/foreman/plugins/modules']
  ansible python module location = /usr/lib/python3.7/site-packages/ansible
  executable location = /usr/bin/ansible
  python version = 3.7.4 (default, Jul  9 2019, 16:32:37) [GCC 9.1.1 20190503 (Red Hat 9.1.1-1)]
```

##### KATELLO/FOREMAN VERSION

```console
tfm-rubygem-katello-3.10.0.55-1.el7sat.noarch
foreman-1.20.1.35-1.el7sat.noarch
```

##### NAILGUN VERSION

```console
Version: 0.32.0
```

I so far have done `pip install nailgun`.

##### STEPS TO REPRODUCE

I have a simple task to enable a repository set:

```yaml
---
- hosts: release, testing
  become: true
  tasks:
    - name: "Enable RHEL 8 BaseOS RPMs repository"
      katello_repository_set:
        username: "{{ admin_user }}"
        password: "{{ admin_pass }}"
        server_url: "https://{{ ansible_fqdn }}"
        organization: "{{ initial_org }}"
        state: enabled
        validate_certs: false
        label: rhel-8-for-x86_64-baseos-rpms
        repositories:
          - releasever: "8"
            basearch: "x86_64"
```

##### EXPECTED RESULTS

 
I expect that the repository set will be enabled.

##### ACTUAL RESULTS

The repository set was not enabled, the task was skipped with no action to perform.

```console
TASK [config : Enable RHEL 8 BaseOS RPMs repository] ***************************
task path: /home/jhunt/git/vagabond/roles/config/tasks/setup.yml:41
Thursday 05 September 2019  23:23:49 -0600 (0:00:00.781)       0:00:18.429 **** 
[...]

PR #641, opened 2019-12-06
force repository substitutions to be strings

evgeni:
users might enter numbers, but the API will always expect them as strings


evgeni:
found while researching #453

evgeni:
@akofink you get to review this for liking my tweet :P

PR #642, opened 2019-12-09
Fixes #620 - Define and update minimal supported Ansible release

### LINKED ISSUE #620: Define and update minimal supported Ansible release
We currently document that the modules *should* work with Ansible 2.3+, but only really test 2.7+.

Ansible itself also [only supports the last 3 versions](https://docs.ansible.com/ansible/latest/reference_appendices/release_and_maintenance.html) (so 2,7, 2.8 and 2.9 at the time of writing).

We have multiple issues/PRs open that would bump our minimum version to 2.8: https://github.com/theforeman/foreman-ansible-modules/issues/399, https://github.com/theforeman/foreman-ansible-modules/pull/328, https://github.com/theforeman/foreman-ansible-modules/issues/326

While I don't think we should bump the minimal version to 2.8 just yet, I think it is time to document that we only support 2.7+ and follow Ansible's rule to support only the last three stable versions in the future.

This was also briefly mentioned in https://github.com/theforeman/foreman-ansible-modules/issues/599 (even tho it  was off-topic there).

bagasse:


bagasse:
Yes, and replacing with Ansible>=2.8 ;)

PR #643, opened 2019-12-09
Remove asynchronous feature

### LINKED ISSUE #520: Remove async options from modules
##### SUMMARY
Ansible has features to run tasks asynchronous (and later wait for them). I think we should stop supporting that at the module level.

##### ISSUE TYPE
 - Feature Idea

mdellweg:
fixes #520


plugins/module_utils/foreman_helper.py
```diff
@@ -532,24 +532,24 @@ def _delete_entity(self, resource, current_entity, params, synchronous):
...
         resource_payload = self._resource_prepare_params(resource, action, params)
         if options is None:
             options = {}
         try:
             result = None
             if ignore_check_mode or not self.check_mode:
                 result = self._resource_call(resource, action, resource_payload, options=options, data=data, files=files)
-                is_foreman_task = isinstance(result, dict) and 'action' in result and 'state' in result and 'pending' in result
-                if synchronous and is_foreman_task:
+                is_foreman_task = isinstance(result, dict) and 'action' in result and 'state' in result and 'started_at' in result
```
evgeni: why the change from pending to started at?

README.md
```diff
@@ -29,6 +29,9 @@ Starting with Ansible 2.7, Ansible only supports Python 2.7 and 3.5 (and higher)
 
 ### Known issues
 
+* Some modules, mainly `katello_sync`, trigger long running tasks on the server side. It might be beneficial to your playbook to wait for their completion in an asynchronous manner.
```
evgeni: sync and cv_version, should both be mentions?

README.md
```diff
@@ -29,6 +29,9 @@ Starting with Ansible 2.7, Ansible only supports Python 2.7 and 3.5 (and higher)
...
+  As ansible has facilities to do so, the modules will wait unconditionally. See [ansible documentation](https://docs.ansible.com/ansible/latest/user_guide/playbooks_async.html) for putting tasks in the background.
```
evgeni:
```suggestion
  As Ansible has facilities to do so, the modules will wait unconditionally. See the [Ansible documentation](https://docs.ansible.com/ansible/latest/user_guide/playbooks_async.html) for putting tasks in the background.
```

plugins/module_utils/foreman_helper.py
mdellweg: Oh, so you spotted it...
`pending` seems to be added by a special katello renderer `started_at` is available on all tasks i could find.

README.md
mdellweg: Like this?
```suggestion
* Some modules, e.g. `katello_sync` and `katello_content_view_version`, trigger long running tasks on the server side. It might be beneficial to your playbook to wait for their completion in an asynchronous manner.
```

plugins/module_utils/foreman_helper.py
evgeni: WFM. Was just wondering :)

README.md
evgeni: Yes

PR #648, opened 2019-12-09
move the org param to the helper

evgeni:


plugins/module_utils/foreman_helper.py
```diff
@@ -617,6 +617,15 @@ def fail_from_exception(self, exc, msg):
                 fail['error'] = exc.response.text
         self.fail_json(**fail)
 
+    def handle_organization_param(self, entity_dict):
```
evgeni: this is in `ForemanAnsibleModule` because that's what's shared between all the Katello modules and foreman_hostgroup. We could move that to `KatelloAnsibleModule` and not add it in hostgroup.

plugins/module_utils/foreman_helper.py
```diff
@@ -617,6 +617,15 @@ def fail_from_exception(self, exc, msg):
...
+        entity_dict = entity_dict.copy()
+
+        entity_dict['organization'] = self.find_resource_by_name('organizations', name=entity_dict['organization'], thin=True)
```
evgeni: no `if absent` because katello actually needs it always, and hostgroup has it in an if already.

nice when I am so consistent with myself and my reviews.

plugins/module_utils/foreman_helper.py
evgeni: I think doing katello only is cleaner, but wanted to hear your opinion.

evgeni:
I think I will also add a doc_fragment

plugins/module_utils/foreman_helper.py
mdellweg: So an OrganizationMixin would also work. But i am fine with katello only.

plugins/module_utils/foreman_helper.py
mdellweg: Maybe leave a comment for future you.

mdellweg:
Am i right, that you cannot move the arg_spec / ent_spec on organization to the mixin, because we cannot decide which one it will be?

evgeni:
@mdellweg I didn't do that for the same reason I didn't add the doc fragment to hostgroup: Katello needs required=true, while hostgroup is fine with required=false.

PR #649, opened 2019-12-10
add a warning when no desired repos were found for a repo_set

evgeni:
this for example happens when the user entered more substitutions than
the repo provides.

also adjust the error message in the case we desire more than we have as
`name` is not always available

Refs: #355
Refs: #453

### DIFF
--- a/plugins/modules/katello_repository_set.py
+++ b/plugins/modules/katello_repository_set.py
@@ -185,6 +185,8 @@ def main():
         repo_set = module.find_resource_by_name('repository_sets', name=module_params['name'], params=scope)
         record_data['name'] = module_params['name']
 
+    repo_set_identification = ' '.join(['{0}: {1}'.format(k, v) for (k, v) in record_data.items()])
+
     repo_set_scope = {'id': repo_set['id'], 'product_id': repo_set['product']['id']}
     repo_set_scope.update(scope)
 
@@ -197,9 +199,13 @@ def main():
     current_repo_names = set(map(lambda repo: repo['name'], current_repos))
     desired_repo_names = set(map(lambda repo: repo['repo_name'], desired_repos))
 
+    if not desired_repo_names:
+        module.warn("Desired repositories not found for repository set {0}. Available: {1}"
+                    .format(repo_set_identification, available_repo_names))
+
     if len(desired_repo_names - available_repo_names) > 0:
         module.fail_json(msg="Desired repositories are not available on the repository set {0}. Desired: {1} Available: {2}"
-                         .format(module_params['name'], desired_repo_names, available_repo_names))
+                         .format(repo_set_identification, desired_repo_names, available_repo_names))
 
     if module.state == 'enabled':
         for repo in desired_repo_names - current_repo_names:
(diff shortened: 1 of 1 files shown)


plugins/modules/katello_repository_set.py
```diff
@@ -197,9 +199,13 @@ def main():
     current_repo_names = set(map(lambda repo: repo['name'], current_repos))
     desired_repo_names = set(map(lambda repo: repo['repo_name'], desired_repos))
 
+    if not desired_repo_names:
+        module.warn("Desired repositories not found for repository set {0}. Available: {1}"
```
evgeni: I am unsure if this should be a `warn` or `fail`.

evgeni: cc @akofink @sean797

akofink: IMO it should fail because the module cannot move forward. Correct me if I'm wrong, but it looks like if they specify 3 repos and only 1 is found, the module will continue like normal. Maybe that should be where the warning happens - the module can do part of the work requested.

evgeni: We handle the "asked for 3, but only 1 found" case in the next lines:
https://github.com/theforeman/foreman-ansible-modules/blob/aaa72556db39ac82b692da69df4a5e083bab8f0a/plugins/modules/katello_repository_set.py#L200-L202

akofink: Ah, okay :+1: So it seems users would get two similar messages. I don't understand why line 200 isn't sufficient to handle both cases?

evgeni: This tries to handle two different (but related, in a way) cases:
1. the original check is "the user requested more than we can provide"
1. the check I added is "the user didn't request anything (that we could find)"

We could fold both `if` statements into one and just yield one error message

akofink: I think that's the best option, especially since the two error messages are essentially the same.

evgeni: However, now that I've said this, I think the old check was never called, because `desired_repos` is calculated from `available_repos`, so it can never be bigger than that.

akofink: Basically if `desired_repos is not available_repos` we want this error - I'm fine with using this PR to fix this other discovered issue.

evgeni: did you mean "desired is strict subset of available"? because desired can very well be not available when you have 7.7, 7.6 and 7.5 provided as available, but you only want to enable 7.7.

akofink: As we said before, desired will always be a strict subset of available, no?

evgeni: "hah"

evgeni:
So I reworked this to fail if the amount of found "desired" repos does not match the number of repos asked by the user.

Now the module will fail if those don't match, like this (requesting two releaseversions, one of the can't be found):

```
fatal: [localhost]: FAILED! => {
    "changed": false
}

MSG:

Desired repositories are not available on the repository set label: rhel-8-for-x86_64-baseos-rpms.
Searched: [{'releasever': 'lol'}, {'releasever': '8'}]
Found: {'Red Hat Enterprise Linux 8 for x86_64 - BaseOS RPMs 8'}
Available: {'Red Hat Enterprise Linux 8 for x86_64 - BaseOS RPMs 8.0', 'Red Hat Enterprise Linux 8 for x86_64 - BaseOS RPMs 8', 'Red Hat Enterprise Linux 8 for x86_64 - BaseOS RPMs 8.1'}
```

PR #650, opened 2019-12-10
allow to enable all repositories in a repo set

evgeni:
Fixes: #355

### DIFF
--- a/plugins/modules/katello_repository_set.py
+++ b/plugins/modules/katello_repository_set.py
@@ -221,7 +221,7 @@ def main():
     current_repo_names = set(map(lambda repo: repo['name'], current_repos))
     desired_repo_names = set(map(lambda repo: repo['repo_name'], desired_repos))
 
-    if len(module_params['repositories']) != len(desired_repo_names):
+    if not module_params['all_repositories'] and len(module_params['repositories']) != len(desired_repo_names):
         repo_set_identification = ' '.join(['{0}: {1}'.format(k, v) for (k, v) in record_data.items()])
 
         error_msg = "Desired repositories are not available on the repository set {0}.\nSearched: {1}\nFound: {2}\nAvailable: {3}".format(
--- a/plugins/modules/katello_repository_set.py
+++ b/plugins/modules/katello_repository_set.py
@@ -50,8 +50,8 @@ options:
     type: str
   repositories:
     description:
-      - Release version and base architecture of the repositories to enable
-      - Required when I(all_repositories=false)
+      - Release version and base architecture of the repositories to enable.
+      - Required when I(all_repositories) is unset or C(false).
     required: false
     type: list
     suboptions:
@@ -65,10 +65,10 @@ options:
         type: str
   all_repositories:
     description:
-      - Affect all available repositories in the repository set instead of listing them in I(repositories)
+      - Affect all available repositories in the repository set instead of listing them in I(repositories).
+      - Required when I(repositories) is unset or an empty list.
     required: false
     type: bool
-    default: false
   state:
     description:
       - Whether the repositories are enabled or not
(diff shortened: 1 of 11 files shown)


evgeni:
python2 *shakes_fist*

plugins/modules/katello_repository_set.py
```diff
@@ -191,7 +202,10 @@ def main():
     available_repos = module.resource_action('repository_sets', 'available_repositories', params=repo_set_scope, ignore_check_mode=True)
     available_repos = available_repos['results']
     current_repos = repo_set['repositories']
-    desired_repos = get_desired_repos(module_params['repositories'], available_repos)
+    if not module_params['all_repositories']:
+        desired_repos = get_desired_repos(module_params['repositories'], available_repos)
+    else:
+        desired_repos = available_repos.copy()
```
mdellweg:
```suggestion
        desired_repos = available_repos[:]
```

evgeni:
updated for the new logic from #649

evgeni:
@mdellweg @akofink @ogajduse thoughts on this one?

plugins/modules/katello_repository_set.py
```diff
@@ -159,10 +166,14 @@ def main():
...
             product=dict(),
             label=dict(),
-            repositories=dict(required=True, type='list', elements='dict'),
+            repositories=dict(type='list', elements='dict'),
+            all_repositories=dict(type='bool', default=False),
             state=dict(default='enabled', choices=['disabled', 'enabled']),
         ),
         required_one_of=[['label', 'name']],
+        required_if=[
+            ['all_repositories', False, ['repositories']],
```
akofink: I'm a bit rusty, but do we also need one in the opposite direction?

```py
['repositories', [], ['all_repositories']]
```

evgeni: That would still allow the user to pass `all_repositories: False`, but yeah, this might be a good idea. I'll see that I add it.

evgeni: Ah, no, `all_repositories` defaults to `False`, so this never triggers. Let's see...

evgeni: And if I drop the default, it also doesn't work :(

evgeni: so the only way to trick the logic currently is to explicitly pass `repositories: []` which I think is safe enough?

akofink: Hmm... but if you pass `repositories: []` and `enabled: true` will this module ensure no repositories are enabled?

akofink: And if so, that makes me think we don't need an `all_repositories` setting - it'd be the same as `repositories: []` and `disabled: true`

evgeni: You mean state=disabled? Yeah, that'd probably work, but really not intuitive ;)

evgeni: Ah, no, repositories selects the "actionable set", so if you pass an empty list, the module does nothing.

akofink: Ok great - so setting `repositories=dict(type='list', elements='dict', default=[])` should be fine and help solve the validation here, right?

evgeni: So I think I managed to confuse both you and me pretty well :crying_cat_face: 

When we set a default, `required_if` doesn't really work anymore, as it says "when field A is set to value B require field C to be set", but as C has a default, it's always set.

I have now dropped the defaults and added a `required_one_of` to the mix. What do you think?

akofink: Yes, well that makes a lot more sense just looking at it ;) :+1: Fair warning: I didn't test it to be sure

PR #653, opened 2019-12-11
improve the error when searching for resources

evgeni:
before:

    No data found for label="rhel-8-for-x86_64-baseos-rpms"

after:

    Found no results while searching for repository_sets with label="rhel-8-for-x86_64-baseos-rpms

### DIFF
--- a/plugins/module_utils/foreman_helper.py
+++ b/plugins/module_utils/foreman_helper.py
@@ -380,7 +380,11 @@ class ForemanAnsibleModule(AnsibleModule):
         elif failsafe:
             result = None
         else:
-            self.fail_json(msg="No data found for %s" % search)
+            if len(results) > 1:
+                error_msg = "too many"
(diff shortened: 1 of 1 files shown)


plugins/module_utils/foreman_helper.py
```diff
@@ -380,7 +380,11 @@ def find_resource(self, resource, search, name=None, params=None, failsafe=False
         elif failsafe:
             result = None
         else:
-            self.fail_json(msg="No data found for %s" % search)
+            if len(results) > 1:
+                error_msg = "too many"
```
mdellweg: Why not give the number in the error?
It could be a good hint while debugging whether it's 2 or 203410.

evgeni:
```suggestion
                error_msg = "too many ({0})".format(len(results))
```

like this?

Issue #583, opened 2019-11-08
Module outputs list of entities even if only one entity was managed


Fodoj:
##### SUMMARY

Commit in question: https://github.com/theforeman/foreman-ansible-modules/commit/72376b0f42eb74919604c00f4cb89fabb599ba9f

Before that it was possible to register variable, for example usergroup like this:

```
register: usergroup
```

And then use it like this:

```
usergroup['entity']['id']
```

Now it's a bit more verbose:

```
usergroup['entity']['usergroups'][0]['id']
```

Which is a bit too much, given that only one resource is managed at a time.

##### ISSUE TYPE
 - Bug Report

mdellweg:
Can we always report the main entity at `result['entity']['id']` and report the others as subentities / elsewhere?

PR #654, opened 2019-12-11
Document returned entity

mdellweg:


plugins/modules/foreman_architecture.py
```diff
@@ -82,7 +82,12 @@
     state: absent
 '''
 
-RETURN = ''' # '''
+RETURN = '''
+entity:
+  description: Final state of the affected Entity
```
evgeni:
```suggestion
  description: Final state of the affected entity.
```

plugins/modules/foreman_architecture.py
```diff
@@ -82,7 +82,12 @@
...
+  returned: success
+  type: dict
```
evgeni: given it is really a dict with arrays of dicts, can we be more verbose here?
also see https://github.com/theforeman/foreman-ansible-modules/issues/583

mdellweg: Do you mean that level of 'more verbose'?

evgeni: Yes. That looks good. Ans now you can't just copy&paste this to all modules :P

plugins/modules/foreman_architecture.py
```diff
@@ -82,7 +82,26 @@
     state: absent
 '''
 
-RETURN = ''' # '''
+RETURN = '''
+entity:
+  description: Final state of the affected entities grouped by their type
```
evgeni:
```suggestion
  description: Final state of the affected entities grouped by their type.
```

evgeni:
I like!

plugins/modules/foreman_auth_source_ldap.py
```diff
@@ -160,7 +160,108 @@
...
+        attr_mail": "mail",
+        attr_photo": null,
+        base_dn": "uid",
+        external_usergroups": [],
+        groups_base": "cn=groups,cn=accounts,dc=example,dc=com",
+        ldap_filter": "(uid=%s)",
+        onthefly_register": true,
+        type": "AuthSourceLdap",
+        use_netgroups": false,
+        usergroup_sync": true
```
mdellweg: I don't even know what those are for.
Anyone keen to use the suggests feature?

Issue #659, opened 2019-12-12
katello_upload: The following required parameters are missing: size when running against 3.14


evgeni:
##### SUMMARY
`katello_upload` fails when used against Katello 3.14:

```
fatal: [tests]: FAILED! => {
    "changed": false,
    "invocation": {
        "module_args": {
            "organization": "Test Organization",
            "password": "VALUE_SPECIFIED_IN_NO_LOG_PARAMETER",
            "product": "Test Product",
            "repository": "Test Repository",
            "server_url": "https://centos7-katello-3-13.yatsu.example.com",
            "src": "/home/egolov/Devel/theforeman/foreman-ansible-modules/tests/test_playbooks/data/bear-4.1-1.noarch.rpm",
            "username": "admin",
            "validate_certs": false
        }
    },
    "msg": "Error while performing create on content_uploads: The following required parameters are missing: size"
}
```

##### ISSUE TYPE
 - Bug Report

##### ANSIBLE VERSION

```
2.9.2
```

##### KATELLO/FOREMAN VERSION

```
3.14
```

##### APYPIE VERSION

```
0.2.1
```

##### STEPS TO REPRODUCE

```yaml
just run tests
```

##### ACTUAL RESULTS

```

```

evgeni:
related commits:
* katello: https://github.com/Katello/katello/commit/4be7f4c4d513cedaaac15a0e66ebf84df483d865
* hammer-cli-katello: https://github.com/Katello/hammer-cli-katello/commit/9276ae51f22b46ad405ac0783762ca833e9e8d6e

PR #660, opened 2019-12-12
update apidoc for katello 3.14

evgeni:
blocked by #659


evgeni:
acking your commit ;)

Issue #651, opened 2019-12-10
k_repository_set force repositories params to be basearch and releaseversion


evgeni:
##### SUMMARY
Today we accept *any* dict, but Katello actually does not:

https://github.com/Katello/katello/blob/5f7f32ebf1f4a221b6c370ea2a15b733e18527b8/app/controllers/katello/api/v2/repository_sets_controller.rb#L153-L155

PR #667, opened 2019-12-20
k_repository_set: only accept releasever and basearch for repositories

evgeni:
Fixes: #651

### DIFF
--- a/plugins/modules/katello_repository_set.py
+++ b/plugins/modules/katello_repository_set.py
@@ -53,6 +53,15 @@ options:
       - Release version and base architecture of the repositories to enable
     required: true
     type: list
+    suboptions:
+      basearch:
+        description:
+          - Basearch of the repository to enable.
+        type: str
+      releasever:
+        description:
+          - Releasever of the repository to enable.
+        type: str
   state:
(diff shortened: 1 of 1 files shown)


evgeni:
tests failing due to vcr, see https://github.com/theforeman/foreman-ansible-modules/pull/668

plugins/modules/katello_repository_set.py
```diff
@@ -131,13 +140,11 @@
 RETURN = ''' # '''
 
 from ansible.module_utils.foreman_helper import KatelloEntityAnsibleModule
-from ansible.module_utils._text import to_text
 
 
 def get_desired_repos(desired_substitutions, available_repos):
     desired_repos = []
     for sub in desired_substitutions:
-        sub = {to_text(k): to_text(v) for (k, v) in sub.items()}
```
mdellweg: Is this replaced by specifying the suboptions, or was it unneeded anyway? (Sensing some unicode related weirdness here)

evgeni: It's replaced by the suboptions, yeah.

PR #670, opened 2019-12-20
always provide a basearch when enabling a repository_set

evgeni:
Fixes: #453

### DIFF
--- a/plugins/modules/katello_repository_set.py
+++ b/plugins/modules/katello_repository_set.py
@@ -130,6 +130,8 @@ EXAMPLES = '''
 
 RETURN = ''' # '''
 
+_ARCHES = ['x86_64', 's390x', 'ppc64le', 'aarch64', 'multiarch', 'ppc64']
+
 from ansible.module_utils.foreman_helper import KatelloEntityAnsibleModule
 from ansible.module_utils._text import to_text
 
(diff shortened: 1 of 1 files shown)


evgeni:
this is missing tests for the added code, but I'd like to add them only after #650 has been merged as it includes great reshuffling of the repo_set test playbook

evgeni:
@nesanton

nesanton:
looks ok

evgeni:
Interestingly, this breaks idempotency of this module, as the repo name differs.

When listing, I get `Red Hat Enterprise Linux 8 for x86_64 - BaseOS RPMs 8`, but after enabling (both, UI and API with this PR applied), I see `Red Hat Enterprise Linux 8 for x86_64 - BaseOS RPMs x86_64 8`.

And this does not bode well with the rest of the detection logic.

evgeni:
So, after debugging this further, I've opened https://projects.theforeman.org/issues/28555 and https://projects.theforeman.org/issues/28644 and would like not to merge this PR until we have clear guidance from Katello what is going on there.

evgeni:
`basearch` gets pruned from repos that don't have that substitution in katello 3.16, see https://projects.theforeman.org/issues/28644

I am thus closing this PR as "invalid" and the issue it fixes as "notabug"

PR #671, opened 2019-12-23
add taxonomy support to realm module

evgeni:


evgeni:
I didn't feel like writing tests because :woman_shrugging:

PR #672, opened 2020-01-03
purge empty substitutions

evgeni:
ansible sets them to None


evgeni:
that's a regression introduced in #667

@mdellweg that's what happens if you trust me without tests ;)

PR #673, opened 2020-01-07
Add test for organization with unicode-characters

m-bucher:
Yet another encoding fix


m-bucher:
Looks like I have to test my stuff on python3 as well :wink:

plugins/module_utils/foreman_helper.py
```diff
@@ -394,7 +394,7 @@ def find_resource(self, resource, search, name=None, params=None, failsafe=False
         return result
 
     def find_resource_by_name(self, resource, name, **kwargs):
-        search = 'name="{0}"'.format(name)
+        search = 'name="{0}"'.format(to_bytes(name))
```
evgeni: having `bytes` here seems weird to me. shouldn't it be `native`?

evgeni: or rather, text, looking at https://github.com/ansible/ansible/blob/devel/lib/ansible/modules/net_tools/basics/uri.py#L477

evgeni:
Can you provide a test for that please?

plugins/module_utils/foreman_helper.py
m-bucher: Yes, you are right.
However, I currently have to evaluate if the change is needed altogether, because after running my tests it seems to work again without this change as well :man_shrugging:
Encoding problems are weird.

m-bucher:
So apparently using `to_native()` does the same as leaving the variable untouched, so I am reducing this PR to adding some tests with non-ASCII characters :wink:

tests/test_playbooks/organization.yml
```diff
@@ -20,6 +20,24 @@
...
+      organization_description: "アクメ株式会社"
+      expected_change: true
+      expected_diff: true
+      expected_diff_before: ""
+      expected_diff_after: "description['\": ]*\\\\u30a2\\\\u30af\\\\u30e1\\\\u682a\\\\u5f0f\\\\u4f1a\\\\u793e"
+  - include: tasks/organization.yml
+    vars:
+      organization_state: present
+      organization_name: "ÄCMÈ"
+      organization_description: "Because Ä makes it look more METAL"
```
evgeni: :black_heart:

tests/test_playbooks/organization.yml
```diff
@@ -20,6 +20,24 @@
...
       expected_diff_after: "description.*test"
+  - include: tasks/organization.yml
+    vars:
+      organization_state: present
+      organization_name: "会社"
+      organization_description: "アクメ株式会社"
+      expected_change: true
+      expected_diff: true
+      expected_diff_before: ""
+      expected_diff_after: "description['\": ]*\\\\u30a2\\\\u30af\\\\u30e1\\\\u682a\\\\u5f0f\\\\u4f1a\\\\u793e"
```
evgeni: I don't think using `expected_diff` here brings a lot of benefit to the tests (besides ugly unicode nunmbers), but I'd also not block on this.

m-bucher: I agree that the `expected_diff_after` regexp is rather ugly, but I think the main features of this test are that the module does not crash when unicode characters are used and that the result the module creates is also what was filled in.

What came to my mind now is that the description is already being checked in the next test.
So I could change this test to check the resulting `organization_name` instead, which would also reduce the number of unicodes that need to be checked.

evgeni: that sounds good!

PR #680, opened 2020-01-20
use cleaned repository substitutions everywhere

evgeni:
before we did use them for searching, but not in the resulting error
message, which can be confusing

### DIFF
--- a/plugins/modules/katello_repository_set.py
+++ b/plugins/modules/katello_repository_set.py
@@ -152,7 +152,6 @@ from ansible.module_utils.foreman_helper import KatelloEntityAnsibleModule
 def get_desired_repos(desired_substitutions, available_repos):
     desired_repos = []
     for sub in desired_substitutions:
-        sub = {k: v for (k, v) in sub.items() if v is not None}
         desired_repos += filter(lambda available: available['substitutions'] == sub, available_repos)
     return desired_repos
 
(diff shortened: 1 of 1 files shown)


plugins/modules/katello_repository_set.py
```diff
@@ -152,7 +152,6 @@
 def get_desired_repos(desired_substitutions, available_repos):
```
ekohl: Should this now be `return itertools.chain(filter(...) for sub in desired_substitutions)`?

plugins/modules/katello_repository_set.py
```diff
@@ -152,7 +152,6 @@
...
     desired_repos = []
     for sub in desired_substitutions:
-        sub = {k: v for (k, v) in sub.items() if v is not None}
         desired_repos += filter(lambda available: available['substitutions'] == sub, available_repos)
```
ekohl: Does this work in Python 3? Isn't this a `list += generator`?

evgeni: It does work :)

plugins/modules/katello_repository_set.py
evgeni: maybe? ;)

evgeni: `TypeError: 'filter' object is not subscriptable` and I am not in the mood to understand why ;)

Issue #681, opened 2020-01-21
broken subnet6 handling


evgeni:
##### SUMMARY
~~747f59de4594d90900557752148366ab27c87afb accidently removed calls to `handle_common_host_params` and `handle_taxonomy_params`, breaking those features.~~

subnet6 is the only one broken :)

evgeni:
wait, it's not broken, just the special casing is not required anymore

evgeni:
okay, we only broke subnet6 handling, that's fine :)

PR #682, opened 2020-01-21
use subnets resource type for subnet6

evgeni:
Fixes: #681


bagasse:
My bad, I updated all modules files but forgot some parts in `foreman_helper`...

evgeni:
Thanks! And no worries, the actual issue ended up being quite small :)

PR #684, opened 2020-01-22
Fixes #676 - Add smart_class_parameter module

### LINKED ISSUE #676: Add  foreman_smart_class_parameter module
##### SUMMARY

Add foreman_smart_class_parameter to add support for smart class parameters and overrides.

##### ISSUE TYPE
 - Feature Idea

bagasse:
Add smart_class_parameter and override_values support. Only work for string params for now.

### DIFF
--- a/plugins/modules/foreman_smart_class_parameter.py
+++ b/plugins/modules/foreman_smart_class_parameter.py
@@ -1,0 +1,248 @@
+#!/usr/bin/python
+# -*- coding: utf-8 -*-
+# (c) 2020 Baptiste Agasse (@bagasse)
+#
+# This file is part of Ansible
+#
+# Ansible is free software: you can redistribute it and/or modify
+# it under the terms of the GNU General Public License as published by
+# the Free Software Foundation, either version 3 of the License, or
+# (at your option) any later version.
+#
+# Ansible is distributed in the hope that it will be useful,
+# but WITHOUT ANY WARRANTY; without even the implied warranty of
+# MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+# GNU General Public License for more details.
+#
+# You should have received a copy of the GNU General Public License
+# along with Ansible.  If not, see <http://www.gnu.org/licenses/>.
+
+from __future__ import absolute_import, division, print_function
+__metaclass__ = type
+
+
+ANSIBLE_METADATA = {'metadata_version': '1.1',
+                    'status': ['preview'],
+                    'supported_by': 'community'}
+
+DOCUMENTATION = '''
+---
+module: foreman_smart_class_parameter
+short_description: Manage Foreman Smart Class Parameters using Foreman API
+description:
+  - Update Foreman Smart Class Parameters using Foreman API. Smart Class Paramters are created/deleted at on puppet classes import
+  - and cannot be created or deleted via Foreman API.
+author:
+  - "Baptiste Agasse (@bagasse)"
+options:
+  puppetclass_name:
+    description: Name of the puppetclass that own the parameter
+    required: true
+    type: str
+  parameter:
+    description: Name of the parameter
+    required: true
+    type: str
+  description:
+    description: Description of the Smart Class Parameter
+    type: str
+  override:
+    description: Whether the smart class parameter value is managed by Foreman
+    type: bool
+  default_value:
+    description: Value to use by default.
+    type: raw
+  hidden_value:
+    description: When enabled the parameter is hidden in the UI.
+    type: bool
+  omit:
+    description:
+      - Foreman will not send this parameter in classification output.
+      - Puppet will use the value defined in the Puppet manifest for this parameter.
+    type: bool
+  override_value_order:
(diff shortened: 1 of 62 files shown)


plugins/modules/foreman_smart_class_parameter.py
```diff
@@ -0,0 +1,248 @@
...
+    description: Types of validation values.
+    type: str
+    choices:
+      - regexp
+      - list
+  validator_rule:
+    description: Used to enforce certain values for the parameter values.
+    type: str
+  parameter_type:
+    description: Types of variable values. Set an empty to reset the parameter type.
```
evgeni:
```suggestion
    description: Types of variable values. Set to empty string to reset the parameter type.
```

plugins/modules/foreman_smart_class_parameter.py
```diff
@@ -0,0 +1,248 @@
...
+    choices:
+      - string
+      - boolean
+      - integer
+      - real
+      - array
+      - hash
+      - yaml
+      - json
+      - ''
```
evgeni: I wonder if this rather should be "reset" or something. I know the Foreman API wants an empty string (or rather `nil`), but this feels weird from a UX perspective.

plugins/modules/foreman_smart_class_parameter.py
```diff
@@ -0,0 +1,248 @@
...
+    )
+
+    entity_dict = module.clean_params()
+    search = "puppetclass_name={0} and parameter={1}".format(entity_dict['puppetclass_name'], entity_dict['parameter'])
+    override_values = entity_dict.pop('override_values', None)
+    if 'override_value_order' in entity_dict:
+        entity_dict['override_value_order'] = '\n'.join(entity_dict['override_value_order'])
+    with module.api_connection():
+        entity, entity_dict = module.resolve_entities(search=search, entity_dict=entity_dict)
+        # Foreman API returns 'hidden_value?' instead of 'hidden_value' this is a bug ?
```
evgeni: This seems intentional.

Maybe @orrabin remembers? The code was added so long ago:

```
commit 9643a2cb1e67f4f5fb75e47e83729fa9c168fdb8
Author: Ori Rabin <orabin@redhat.com>
Date:   Sun Dec 11 15:18:49 2016 +0200

    Fixes #17653 - Allow all parameters api to unhide values
```

plugins/modules/foreman_smart_class_parameter.py
bagasse: Yes, it will be more readable if we use `reset` or `none` instead of empty string. Any preference between `reset` or `none` ?

evgeni: I think this depends on the behavior of the module/api. I think `none` is fine if the action is idempotent (aka the reply of a show action later will also yield "" or `nil` or an unset parameter). And `reset` should be used if the next show action will show us `string` or whatever. Does it make sense?

Also, how does our param to string helper handle this?

plugins/modules/foreman_smart_class_parameter.py
orrabin: If I remember correctly `hidden_value` was actually the value when hidden (so ********) but `hidden_value?` is meant to understand if you want to see it as hidden or plain text.
I removed the `hidden_value` and in stead added this partial: https://github.com/theforeman/foreman/blob/develop/app/views/api/v2/common/show_hidden.json.rabl that would show the real or hidden value in the same field.

plugins/modules/foreman_smart_class_parameter.py
bagasse: Ok, moved this to `none`, it fallback to string if empty for parm to string helper

tests/test_playbooks/smart_class_parameter.yml
```diff
@@ -0,0 +1,558 @@
...
+        smart_class_parameter_parameter: "{{ item }}"
+        smart_class_parameter_override: false
+        smart_class_parameter_state: present
+      loop:
+        - logfile
+        - config_dir
+        - servers
+        - burst
+        - stepout
+- hosts: tests
```
evgeni:
```suggestion

- hosts: tests
```

tests/test_playbooks/smart_class_parameter.yml
```diff
@@ -0,0 +1,558 @@
+---
```
evgeni: Can you add a comment that the (which?) `ntp` puppet module needs to be present for re-recording this?

Like we do in https://github.com/theforeman/foreman-ansible-modules/blob/923f37f22af15f0b883e286fbb6a2ce1ce66a9c6/tests/test_playbooks/config_group.yml#L2-L5  or https://github.com/theforeman/foreman-ansible-modules/blob/923f37f22af15f0b883e286fbb6a2ce1ce66a9c6/tests/test_playbooks/hostgroup.yml#L2-L8

Or isn't that needed?

bagasse: Yes it's needed. Added comment.

Issue #683, opened 2020-01-21
Found too many results while searching for domains with some.domain.com


rht-jbittner:
##### SUMMARY
In our Satellite we have some third level and forth level domains like:

- us.domain.com
- ny.us.domain.com
- orl.us.domain.com

If I do:

- name: Create subnet
  foreman_subnet:
    name: "US office subnet - 192.168.0.0"
    description: "us.domain.com"
    network: "192.168.0.0"
    mask: "255.255.255.0"
    gateway: "192.168.0.1"
    boot_mode: "Static"
    dhcp_proxy: "someproxy.domain.com"
    dns_proxy: "someproxy.domain.com"
    mtu: 1500
    domains:
       - "us.domain.com"
    organizations: "someorg"
    locations: "US"
    state: present

It fails and returns:
Found too many (3) results while searching for domains with us.domain.com"

##### ISSUE TYPE
 - Bug Report

##### ANSIBLE VERSION

```
2.8
```

##### KATELLO/FOREMAN VERSION

```
tfm-rubygem-katello-3.12.0.29-1.el7sat.noarch
foreman-1.22.0.33-1.el7sat.noarch
```

##### APYPIE VERSION

```
Metadata-Version: 1.1
Version: 0.2.0
```

##### STEPS TO REPRODUCE

Set Foreman/Satellite with domains like:

- us.domain.com
- ny.us.domain.com
- orl.us.domain.com

Run code bellow.

```yaml
- name: Create subnet
  foreman_subnet:
    name: "US office subnet - 192.168.0.0"
    description: "us.domain.com"
    network: "192.168.0.0"
    mask: "255.255.255.0"
    gateway: "192.168.0.1"
    boot_mode: "Static"
    dhcp_proxy: "someproxy.domain.com"
    dns_proxy: "someproxy.domain.com"
    mtu: 1500
    domains:
       - "us.domain.com"
    organizations: "someorg"
    locations: "US"
    state: present
```

##### EXPECTED RESULTS

Configured subnet with right domain

##### ACTUAL RESULTS

```
Found too many (3) results while searching for domains with us.domain.com"

```

evgeni:
Hey, I tried to reproduce this, and couldn't with the latest `master`. Which version of foreman-ansible-modules are you running?

evgeni:
Ah, I can repro with the latest 0.4.0 release.

evgeni:
The following patch should fix it for you:

```patch
[...]

PR #685, opened 2020-01-22
foreman_subnet: search domains by name

evgeni:
Fixes: #683


evgeni:
Confirmed to be working as expected in #683

PR #686, opened 2020-01-22
0.4.1 cherry picks

evgeni:


evgeni:
If we're doing a 0.4.1, these two commits are good candidates too :)

mdellweg:
Nice one!

Issue #690, opened 2020-01-29
k_content_view_version returns `entity: {}`


evgeni:
##### SUMMARY
Seems like change tracking is not working well when promoting a content view version.

`changed` is properly set, so we probably should just omit returning empty `entity` data for now. (or record proper data, but oh well)

##### ISSUE TYPE
 - Bug Report

##### ANSIBLE VERSION

```

```

##### KATELLO/FOREMAN VERSION

```

```

##### APYPIE VERSION

```

```

##### STEPS TO REPRODUCE

```yaml

```

##### ACTUAL RESULTS

```

```

PR #691, opened 2020-01-29
record CV version data when promoting

evgeni:
Fixes: #690


evgeni:
tests should be fixed in https://github.com/theforeman/foreman-ansible-modules/pull/692

PR #692, opened 2020-01-30
argument_spec and documentation updates to please latest sanity checks

evgeni:
Ansible's sanity tests got stricter around suboptions, so let's follow suite :)

### DIFF
--- a/plugins/module_utils/foreman_helper.py
+++ b/plugins/module_utils/foreman_helper.py
@@ -1011,6 +1011,9 @@ def _entity_spec_helper(spec):
     """
     entity_spec = {'id': {}}
     argument_spec = {}
+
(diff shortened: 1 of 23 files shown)


plugins/module_utils/foreman_helper.py
```diff
@@ -1011,6 +1011,9 @@ def _entity_spec_helper(spec):
     """
     entity_spec = {'id': {}}
     argument_spec = {}
+
+    _FILTER_SPEC_KEYS = ['thin', 'scope', 'resource_type', 'search_by', 'search_operator', 'resolve', 'ensure']
```
evgeni: @bagasse @mdellweg is this list complete? :)

mdellweg: Do we need `failsafe`?
I fear, i have lost track about the evolution of entity_spec.

evgeni: @mdellweg it works without 🤷

PR #700, opened 2020-02-06
add foreman_model module

evgeni:


evgeni:
@erickellerek1 this is based on the discussion we had in Gent to close out the gaps between FAM and your foreman-yml :)

evgeni:
@bagasse writing new modules is SO easy :heart_eyes_cat:

Issue #404, opened 2019-08-13
make the entity_spec work for non-entity modules


evgeni:
##### SUMMARY

See https://github.com/theforeman/foreman-ansible-modules/pull/335#discussion_r312357992

TL;DR: entity_spec is currently coupled to the fact that it describes an "entity", while we have action modules that don't operate on an entity as such, but would still benefit from the conversion done in the spec. maybe we should call it foreman_spec?

##### ISSUE TYPE
 - Feature Idea

mdellweg:
...foreman_spec and move it's logic to the base Class.

evgeni:
triage: let's see where the entity_spec is used outside of entity modules and see how this can be streamlined

mdellweg:
In the meantime the benefits also include related entity lookup.

nesanton:
```bash
$ grep -rP 'module = (Foreman|Katello)AnsibleModule' .
./plugins/modules/foreman_template_sync.py:    module = ForemanAnsibleModule(
./plugins/modules/katello_content_view_filter.py:    module = KatelloAnsibleModule(
./plugins/modules/katello_sync.py:    module = KatelloAnsibleModule(
./plugins/modules/katello_upload.py:    module = KatelloAnsibleModule(
./plugins/modules/foreman_search_facts.py:    module = ForemanAnsibleModule(
./plugins/modules/foreman_scc_product.py:    module = KatelloAnsibleModule(
```

nesanton:
also, issue #404 is not found

PR #705, opened 2020-02-11
Add templates_import module wrapping foreman_templates plugin

### LINKED ISSUE #677: Add module for foreman_templates plugin
##### SUMMARY
A lot of people ask how to sync templates from git
Wrapping foreman_templates plugin in a module could be a good idea https://github.com/theforeman/foreman_templates 

##### ISSUE TYPE
 - Feature Idea

nesanton:
Fixes #677


nesanton:
apparently need to do something about `check_mode: true`

plugins/modules/foreman_template_sync.py
```diff
@@ -0,0 +1,222 @@
...
+    required: false
+  associate:
+    description: Associate to OSes, Locations and Organizations based on metadata.
+    required: false
+    choices:
+     - always
+     - new
+     - never
+  verbose:
+    description: Add template diffs to the output.
```
mdellweg: audits! I like that.

mdellweg:
If it's impossible to implement check mode (like katello_manifest) you can choose not to support it.

nesanton:
> If it's impossible to implement check mode (like katello_manifest) you can choose not to support it.

Yeah it's rather hard. Dropping check_mode at least for now.

plugins/modules/foreman_template_sync.py
```diff
@@ -0,0 +1,220 @@
...
+
+    entity_dict = module.clean_params()
+    module.connect()
+
+    if 'template' not in module.foremanapi.resources:
+        raise Exception('The server does not seem to have the foreman_templates plugin installed.')
+
+    all_templates = []
+
+    template_types = ['provisioning_templates', 'report_templates', 'ptables']
```
mdellweg: This gets me thinking, whether we want to add a parameter to only affect a subset of available template types.

plugins/modules/foreman_template_sync.py
```diff
@@ -0,0 +1,220 @@
...
+    if 'job_templates' in module.foremanapi.resources:
+        template_types.append('job_templates')
+    for template_type in template_types:
+        resources = module.list_resource(template_type)
+        for resource in resources:
+            all_templates.append([resource['name'], resource['id']])
+
+    result = module.resource_action('template', entity_dict['direction'], record_change=False, params=entity_dict)
+    changed = False
```
mdellweg: You can keep the changed state in the model if you drop this line and call `module.set_changed()` below.

plugins/modules/foreman_template_sync.py
```diff
@@ -0,0 +1,220 @@
...
+    msg_templates = result['message'].pop('templates', [])
+
+    if entity_dict['direction'] == 'import':
+        diff = {'changed': [], 'new': []}
+    else:
```
mdellweg: I think, the flow of control here is a little bit confusing. Can you move this else-branch to the end of the method?

nesanton: totally agree, doing

plugins/modules/foreman_template_sync.py
nesanton: This might be a good idea, but I have no control over it. The plugin will import whatever it finds. I need this only to compile a list of what's already present.
User might choose to incorporate the types into naming convention and then use filter. Or we fine a feature request to the plugin. I feel like I'll be doing this anyway e.g. for purging

plugins/modules/foreman_template_sync.py
nesanton: done

plugins/modules/foreman_template_sync.py
```diff
@@ -0,0 +1,210 @@
...
+  - Some defaults can be set in TemplateSync settings using M(foreman_setting) or GUI.
+  - Module attempts to be idempotent as much as the plugin allows.
+  - "Known issues:"
+  - I(direction=export) exports all templates even if only one changes, there also could be 500 responses on C(export).
+author:
+  - "Anton Nesterov (@nesanton)"
+options:
+  direction:
+    description:
+      - Direction of the sync.
```
nesanton: I wonder if this should become `state: imported/exported` instead of a new param

plugins/modules/foreman_template_sync.py
mdellweg: It was just a thought. But a feature request to the plugin sounds good anyway!

plugins/modules/foreman_template_sync.py
mdellweg: Maybe it's better to call it an action module (a bit like katello_sync) that has no state at all. This kind of falls in line with the fact that there is no proper check_mode.

nesanton: is there a special way to make it so? I seem to be inheriting `state` in argument spec

mdellweg: I should be possible if you inherit from `ForemanAnsibleModule`.

nesanton: ah right

plugins/modules/foreman_template_sync.py
```diff
@@ -143,12 +143,12 @@
...
+from ansible.module_utils.foreman_helper import ForemanAnsibleModule
 
 
 def main():
-    module = ForemanEntityAnsibleModule(
-        entity_spec=dict(
+    module = ForemanAnsibleModule(
+        argument_spec=dict(
             direction=dict(choices=['import', 'export'], default='import'),
             location=dict(type='entity', flat_name='location_id'),
```
mdellweg: With `argument_spec` you probably loose all the nice additions from our `entity_spec`. :crab: 
We had that discussion earlier. Maybe it's time to revisit.
@evgeni @bagasse ?

nesanton:
@evgeni, CI wants `type: entity` from some `str` thinks like `location`. I can't put my finger on it

plugins/modules/foreman_template_sync.py
nesanton: I thought entity_spec won't work with ForemanAnsibleModule

evgeni: Right. But without entity_spec you can't use the extensions we have (like type:entity, and flat_name etc).
Which leads us to https://github.com/theforeman/foreman-ansible-modules/issues/404 ;)

evgeni:
@nesanton see my inline comment about entity spec ;)

evgeni:
I do wonder if we should split this in two modules, one for import, one for export, as the "direction" sounds weird to me.

nesanton:
> I do wonder if we should split this in two modules, one for import, one for export, as the "direction" sounds weird to me.

let's discuss in the issue #677

mdellweg:
> I do wonder if we should split this in two modules, one for import, one for export, as the "direction" sounds weird to me.

Not a bad idea. It would also underline the action character.

evgeni:
FWIW, the resource name was changed to templateS in git recently: https://github.com/theforeman/foreman_templates/commit/c1c4756a71fb44e274a7b113dfbc57f77a9549e4

So ideally we would support both. Stable apis...

mdellweg:
> So ideally we would support both. Stable apis...

D you mean: "So ideally we would support both stable apis..." ?? :smiling_imp:

nesanton:
Ok, made the changes to support the (logical) rename of the API resource.
Stripped the export functionality and renamed the module.
What do we do with entity_spec?

mdellweg:
I think, you need to resolve the entities vintage style for now, or you file a pr to solve #404 forever.

plugins/modules/foreman_templates_import.py
```diff
@@ -0,0 +1,187 @@
...
+        supports_check_mode=False,
+    )
+
+    module_params = module.clean_params()
+    module.connect()
+
+    if 'template' in module.foremanapi.resources:
+        resource_name = 'template'
+    elif 'templates' in module.foremanapi.resources:
+        resource_name = 'template'
```
mdellweg:
```suggestion
    if 'template' in module.foremanapi.resources or 'templates' in module.foremanapi.resources:
        resource_name = 'template'
```

nesanton: doh!

plugins/modules/foreman_templates_import.py
```diff
@@ -0,0 +1,185 @@
...
+        supports_check_mode=False,
+    )
+
+    module_params = module.clean_params()
+    module.connect()
+
+    if 'template' in module.foremanapi.resources or 'templates' in module.foremanapi.resources:
+        resource_name = 'template'
+    else:
+        raise Exception('The server does not seem to have the foreman_templates plugin installed.')
```
mdellweg: I think, this should be `fail_json`, or do we catch exceptions if we go with `with module.connect()`?

plugins/modules/foreman_templates_import.py
```diff
@@ -0,0 +1,185 @@
...
+    all_templates = []
+
+    template_types = ['provisioning_templates', 'report_templates', 'ptables']
+    if 'job_templates' in module.foremanapi.resources:
+        template_types.append('job_templates')
+
+    for template_type in template_types:
+        resources = module.list_resource(template_type)
+        for resource in resources:
+            all_templates.append([resource['name'], resource['id']])
```
mdellweg: If you like list comprehensions:
```suggestion
        all_templates += [[resource['name'], resource['id']] for resources in module.list_resource(template_type)]
```

mdellweg: Also, i would change the 'name', 'id' pair into a tupel (not a list).

tests/test_playbooks/templates_import.yml
```diff
@@ -0,0 +1,59 @@
...
+        templates_import_prefix: "fam_tsync_test_"
+        templates_import_filter: "^fam_tsync_test_Kickstart default$"
+        expected_change: false
+    - include: tasks/provisioning_template.yml
+      vars:
+        provisioning_template_name: "fam_tsync_test_Kickstart default"
+        provisioning_template_template: "Modified Kickstart default provisioning template"
+        provisioning_template_state: present
+        provisioning_template_kind: provision
+        provisioning_template_locations: '{{ omit }}'
```
mdellweg: There's no need, they default to `omit` anyway.
However, as i fail to see, where you resolve the taxonomies into ids, id love to see them somewhere in the tests.

plugins/modules/foreman_templates_import.py
```diff
@@ -0,0 +1,185 @@
...
+
+    result = module.resource_action(resource_name, 'import', record_change=False, params=module_params)
+    msg_templates = result['message'].pop('templates', [])
```
mdellweg: I had a hard time to understand the purpose of this variable. Maybe call it something with 'report'.

plugins/modules/foreman_templates_import.py
nesanton: I agree, modules should fail the module way

plugins/modules/foreman_templates_import.py
nesanton: right, += is something I was missing. Also +1 to tuple

tests/test_playbooks/templates_import.yml
nesanton: fair point and it's not too hard to add, ok

plugins/modules/foreman_templates_import.py
nesanton: Building a list of templates that are present before import, so we can tell if any are new (because the plugin does not care it just says `imported=true` for every template that either changed or got added. IIRC, the diff is not helping in case of new)

all_existing_templates / all_present_templates / template_report / templates_before_import ?

mdellweg: I think `template_import` would do.

plugins/modules/foreman_templates_import.py
```diff
@@ -0,0 +1,185 @@
+#!/usr/bin/python
+# -*- coding: utf-8 -*-
+# Copyright (c) 2020, Anton Nesterov (@nesanton)
+#
+# This file is part of Ansible
```
evgeni: no it's not, can you please update to the header we (now) use in all the other plugins? :)

plugins/modules/foreman_templates_import.py
```diff
@@ -0,0 +1,185 @@
...
+  - "Anton Nesterov (@nesanton)"
+options:
+  location:
+    description: Scope by location
+    required: false
+    type: str
+  organization:
+    description: Scope by organization
+    required: false
+    type: str
```
evgeni:
```suggestion
```

these are only used for searching in the API and we don't use them at all in the modules

plugins/modules/foreman_templates_import.py
```diff
@@ -0,0 +1,185 @@
...
+    required: false
+    type: list
+    elements: str
+  organizations:
+    description: REPLACE organizations with given list.
+    required: false
+    type: list
+    elements: str
+extends_documentation_fragment:
+  - foreman
```
evgeni:
```suggestion
extends_documentation_fragment:
  - foreman
  - foreman.taxonomy
```

plugins/modules/foreman_templates_import.py
```diff
@@ -0,0 +1,185 @@
...
+RETURN = ''' # '''
+
+from ansible.module_utils.foreman_helper import ForemanAnsibleModule
+
+
+def main():
+    module = ForemanAnsibleModule(
+        foreman_spec=dict(
+            location=dict(type='entity', flat_name='location_id'),
+            organization=dict(type='entity', flat_name='organization_id'),
```
evgeni:
```suggestion
```

plugins/modules/foreman_templates_import.py
```diff
@@ -0,0 +1,185 @@
...
+            branch=dict(),
+            repo=dict(),
+            filter=dict(),
+            dirname=dict(),
+            verbose=dict(type='bool'),
+            force=dict(type='bool'),
+            lock=dict(type='bool'),
+            negate=dict(type='bool'),
+            locations=dict(type='entity_list', flat_name='location_ids'),
+            organizations=dict(type='entity_list', flat_name='organization_ids'),
```
evgeni:
```suggestion
            locations=dict(type='entity_list'),
            organizations=dict(type='entity_list'),
```

our helper should figure out the flat name on its own.

we also have `ForemanTaxonomicEntityAnsibleModule` so you don't have to pass that, but its for entities...

plugins/modules/foreman_templates_import.py
evgeni: this can be `required_plugins=[('templates', ['*'])]` in `__init__` these days

plugins/modules/foreman_templates_import.py
```diff
@@ -0,0 +1,185 @@
...
+        ),
+        supports_check_mode=False,
+    )
+
+    module_params = module.clean_params()
+    module.connect()
+
+    if 'template' in module.foremanapi.resources or 'templates' in module.foremanapi.resources:
+        resource_name = 'template'
```
evgeni: shouldn't it be `resource_name = 'templates'` in the second case?

plugins/modules/foreman_templates_import.py
```diff
@@ -0,0 +1,185 @@
...
+    else:
+        raise Exception('The server does not seem to have the foreman_templates plugin installed.')
+
+    # Build a list of all existing templates of all supported types to check if we are adding any new
+    all_templates = []
+
+    template_types = ['provisioning_templates', 'report_templates', 'ptables']
+    if 'job_templates' in module.foremanapi.resources:
+        template_types.append('job_templates')
```
evgeni: i've seen this before... move to helper?

plugins/modules/foreman_templates_import.py
```diff
@@ -0,0 +1,185 @@
...
+
+    for template_type in template_types:
+        resources = module.list_resource(template_type)
+        for resource in resources:
+            all_templates.append([resource['name'], resource['id']])
+
+    result = module.resource_action(resource_name, 'import', record_change=False, params=module_params)
+    msg_templates = result['message'].pop('templates', [])
+
+    diff = {'changed': [], 'new': []}
```
evgeni: this is not what ansible expects in the diff, it wants two strings (or lists) before and after, like `diff(1)`

if that's too hard to generate, maybe rename this?

plugins/modules/foreman_templates_import.py
mdellweg: Sounds like it should be a Mixin.

plugins/modules/foreman_templates_import.py
mdellweg: Is it `report`?

plugins/modules/foreman_templates_import.py
evgeni: probably, yeah

plugins/modules/foreman_templates_import.py
evgeni: yeah, why not report

plugins/modules/foreman_templates_import.py
nesanton: ack

plugins/modules/foreman_templates_import.py
nesanton: I did not like the descriptions in taxonomy. But fair point, a lot better to reuse than to have a seemingly "better" description

plugins/modules/foreman_templates_import.py
nesanton: oops, true

nesanton: -- this can be required_plugins=[('templates', ['*'])] in __init__ these days
but will it handle the rename of template resource to templates in the newer foreman?

plugins/modules/foreman_templates_import.py
nesanton: ok.
but can't find any similar stuff. Do you recall where?

plugins/modules/foreman_templates_import.py
nesanton: report!

plugins/modules/foreman_templates_import.py
nesanton: Something like this #779 ?

plugins/modules/foreman_templates_import.py
nesanton: @evgeni, Ok, I see, this can be moved to required_plugins (adding the plugin to _PLUGIN_RESOURCES and a little kludge for template/templates somewhere along the lines), but I still need to know which resource to call the resource_action with.

plugins/modules/foreman_templates_import.py
nesanton: @evgeni, I'd still this if-else in the module because I don't know which resource to use with resource_action. Or is there a smarter way? Maybe move the kludge to the helper (`if resource=="blah" in the resource_action func)  and leave the module clean?

plugins/modules/foreman_templates_import.py
evgeni: Yeah, now that you said that, `required_plugins` does not cope with that weird renaming well :(
So maybe, in this case, let the kludge be here and don't bother with `required_plugins` for now?

plugins/modules/foreman_templates_import.py
evgeni: and yeah, the if/else is still needed, sure.

or, we could add a api patch (as we do with other plugins) to always rename it to `templates`?

plugins/modules/foreman_templates_import.py
evgeni: I thought in the other template modules, but can't find, so ignore me please

plugins/modules/foreman_templates_import.py
evgeni: You can add something like this to `ForemanAnsibleModule`

```python
_patch_templates_resource_name(self):
    if 'template' in self.foremanapi.apidoc['docs']['resources']:
        self.foremanapi.apidoc['docs']['resources']['templates'] = self.foremanapi.apidoc['docs']['resources']['template']
```

and then call it here:
https://github.com/theforeman/foreman-ansible-modules/blob/36e6e602ff273f76e11f056737c36f98f4b8f059/plugins/module_utils/foreman_helper.py#L390-L391

then no if/else is needed and `require_plugins` will work too.

evgeni: and probably add a comment referring to https://github.com/theforeman/foreman_templates/commit/c1c4756a71fb44e274a7b113dfbc57f77a9549e4 or https://projects.theforeman.org/issues/28750

plugins/modules/foreman_templates_import.py
```diff
@@ -0,0 +1,163 @@
...
+  - Module attempts to be idempotent as much as the plugin allows.
+author:
+  - "Anton Nesterov (@nesanton)"
+options:
+  prefix:
+    description: Adds specified string to beginning of the template, but only if the template name does not start with the prefix already.
+    required: false
+    type: str
+  associate:
+    description: Associate to OSes, Locations and Organizations based on metadata.
```
evgeni, line 43:
```suggestion
    description: Associate to Operatingsystems, Locations and Organizations based on metadata.
```

plugins/modules/foreman_templates_import.py
```diff
@@ -0,0 +1,163 @@
...
+    required: false
+    type: str
+  filter:
+    description:
+      - Sync only templates with name matching this regular expression, after I(prefix) was applied.
+      - Case-insensitive, snippets are not filtered.
+    required: false
+    type: str
+  negate:
+    description: Negate the prefix.
```
evgeni, line 77: What does that mean? Will it remove the `prefix` param from the names?

plugins/modules/foreman_templates_import.py
```diff
@@ -0,0 +1,163 @@
...
+description:
+  - Sync provisioning templates, report_templates, partition tables and job templates from external git repository and/or file system.
+  - Based on foreman_templates plugin U(https://github.com/theforeman/foreman_templates).
+  - Some defaults can be set in TemplateSync settings using M(foreman_setting) or GUI.
+  - Module attempts to be idempotent as much as the plugin allows.
+author:
+  - "Anton Nesterov (@nesanton)"
+options:
+  prefix:
+    description: Adds specified string to beginning of the template, but only if the template name does not start with the prefix already.
```
evgeni, line 39:
```suggestion
    description:
      - Adds specified string to beginning of the template, but only if the template name does not start with the prefix already.
```

(please always use `description` as an array)

plugins/modules/foreman_templates_import.py
```diff
@@ -0,0 +1,163 @@
...
+            verbose=dict(type='bool'),
+            force=dict(type='bool'),
+            lock=dict(type='bool'),
+            negate=dict(type='bool'),
+        ),
+        supports_check_mode=False,
+        required_plugins=[('foreman_templates', ['*'])],
+    )
+
+    module_params = module.foreman_params
```
evgeni, line 123: this is IMHO not necessary, you only use it once, and can safely use the long name :)

tests/test_playbooks/templates_import.yml
evgeni: Yupp, I think currently the imported templates won't be taxed properly.

evgeni: sounds like a plugin bug to me? document and move on? :)

plugins/modules/foreman_templates_import.py
nesanton: Omg, it negates the filter not the prefix. Wich starts to make sense ;) I'll fix it

plugins/modules/foreman_templates_import.py
nesanton: okay

plugins/modules/foreman_templates_import.py
nesanton: right

tests/test_playbooks/templates_import.yml
nesanton: Ok, there were two problems
1. non-entity modules did not have functionality to lookup entity params and flatten them
I made a mock attempt to do this - will start a new thread when pushed
2. bug https://projects.theforeman.org/issues/29534
made the assertion for the change think "false" for now

plugins/module_utils/foreman_helper.py
```diff
@@ -306,9 +307,23 @@ def api_connection(self):
     def changed(self):
         return self._changed
 
+    def flatten_entity_params(self):
+        """ Flattens taxonimy and other parameters that are of entity or entity_list type.
+             Will work only after connect()
+        """
+        self.auto_lookup_entities()
+        self.foreman_params = _flatten_entity(self.foreman_params, self.foreman_spec)
+
```
nesanton, line 316: The taxonomy (and potentially other entity params) need lookup and flattening for non-entity modules.
This is the cheapest way I could think of.

So now I need to call this from the module after connect()
Is there are smart way of doing this? And should we make this done automatically?

nesanton: cc @evgeni

plugins/module_utils/foreman_helper.py
```diff
@@ -387,6 +402,7 @@ def connect(self):
 
         self.check_required_plugins()
 
+        self._patch_templates_resource_name
```
evgeni, line 405:
```suggestion
        self._patch_templates_resource_name()
```

nesanton: oi

evgeni:
well, look at that, green tests.

nesanton:
> well, look at that, green tests.

one assertion is worked around though https://projects.theforeman.org/issues/29534

plugins/module_utils/foreman_helper.py
nesanton: rename to _lookup_entity_params and call it in connect() for ForemanTaxonomicAnsibleModule?

evgeni: So the problem here is, that we're not in an "entity" module, so `_flatten_entity` is never called (it's called from the various entity create/update funcs we have, but those are not called here for obvious reasons). And additionally, given this is not an entity module, the lookup also does not happen.

I am not sure I like that both these tasks are done in one function with a name only pointing at one part, but I see that those calls are required.

@mdellweg any elegant ideas how to solve that?

plugins/module_utils/foreman_helper.py
```diff
@@ -51,6 +51,7 @@
     'remote_execution': 'remote_execution_features',
     'scc_manager': 'scc_accounts',
     'snapshot_management': 'snapshots',
+    'foreman_templates': 'templates',
```
evgeni, line 54: FWIW, we usually skip the `foreman_` prefix for the other plugins.

plugins/module_utils/foreman_helper.py
mdellweg: Maybe i am not seeing the whole picture, but it seems to be quite special to the templates_import module. You could just call those two lines there.

plugins/module_utils/foreman_helper.py
nesanton: here you go

plugins/modules/foreman_templates_import.py
```diff
@@ -0,0 +1,166 @@
...
+            repo=dict(),
+            filter=dict(),
+            dirname=dict(),
+            verbose=dict(type='bool'),
+            force=dict(type='bool'),
+            lock=dict(type='bool'),
+            negate=dict(type='bool'),
+        ),
+        supports_check_mode=False,
+        required_plugins=[('foreman_templates', ['*'])],
```
evgeni, line 129: you need to change it in both places ;)

plugins/module_utils/foreman_helper.py
evgeni: it's special for every non-entity, non-katello module that does reference entities, we just don't have many of those ;)

but yeah, I think the cleanest solution for now would be not having this function and use

```python
    with module.api_connection():

        module.auto_lookup_entities()

        # Build a list of all existing templates of all supported types to check if we are adding any new
        template_report = []

        template_types = ['provisioning_templates', 'report_templates', 'ptables']
        if 'job_templates' in module.foremanapi.resources:
            template_types.append('job_templates')

        for template_type in template_types:
            template_report += [(resource['name'], resource['id']) for resource in module.list_resource(template_type)]

        result = module.resource_action('templates', 'import', record_change=False, params=_flatten_entity(module.foreman_params, module.foreman_spec))
```

(well, and obviously importing `_flatten_entity` at the top

mdellweg: or build the payload with scope
`payload = module.scope_for('location')`

mdellweg: This will also automatically look up location.

nesanton: Gone evgeni's way for now. scope_for seems longer - it'll need looping thu orgs and locations, right?

plugins/module_utils/foreman_helper.py
```diff
@@ -385,11 +393,12 @@ def connect(self):
 
         self.ping()
 
-        self.check_required_plugins()
-
+        self._patch_templates_resource_name()
         self._patch_location_api()
         self._patch_subnet_rex_api()
 
+        self.check_required_plugins()
```
mdellweg, line 400: :+1: Its a matter of execution order now.

plugins/modules/foreman_templates_import.py
```diff
@@ -0,0 +1,166 @@
...
+  - Sync provisioning templates, report_templates, partition tables and job templates from external git repository and/or file system.
+  - Based on foreman_templates plugin U(https://github.com/theforeman/foreman_templates).
+  - Some defaults can be set in TemplateSync settings using M(foreman_setting) or GUI.
+  - Module attempts to be idempotent as much as the plugin allows.
+author:
+  - "Anton Nesterov (@nesanton)"
+options:
+  prefix:
+    description:
+      - Adds specified string to beginning of the template, but only if the template name does not start with the prefix already.
```
mdellweg, line 40: maybe sthing like:
...of all imported templates that do not yet have that prefix...

If i remember correctly, this could be set in the settings. If that is so, maybe drop a note here, or collect the names of all such options in the note above.

plugins/modules/foreman_templates_import.py
```diff
@@ -0,0 +1,166 @@
...
+    required: false
+    type: str
+  associate:
+    description:
+      - Associate to Operatingsystems, Locations and Organizations based on metadata.
```
mdellweg, line 45: Should we describe the individual choices?
"...for all, only imported, no templates"
Or is it changed ones?

plugins/modules/foreman_templates_import.py
```diff
@@ -0,0 +1,166 @@
...
+    required: false
+    type: bool
+  lock:
+    description:
+      - Lock imported templates.
+    required: false
+    type: bool
+  branch:
+    description:
+      - Branch in Git repo.
```
mdellweg, line 69: maybe: Branch of the git I(repo) to look at.
This can only be used when repo is a git repository, right?

plugins/modules/foreman_templates_import.py
```diff
@@ -0,0 +1,166 @@
...
+module: foreman_templates_import
+short_description: Sync templates from a repository
+description:
+  - Sync provisioning templates, report_templates, partition tables and job templates from external git repository and/or file system.
+  - Based on foreman_templates plugin U(https://github.com/theforeman/foreman_templates).
+  - Some defaults can be set in TemplateSync settings using M(foreman_setting) or GUI.
+  - Module attempts to be idempotent as much as the plugin allows.
+author:
+  - "Anton Nesterov (@nesanton)"
+options:
```
evgeni, line 37:
```suggestion
notes:
  - Due to a bug in the foreman_templates plugin, this module won't report C(changed=true) when the only change is the Organization/Location association of the imported templates. Please see U(https://projects.theforeman.org/issues/29534) for details.
options:
```

plugins/modules/foreman_templates_import.py
nesanton: I tried to keep the description consistent with what they say in plugin docs here https://www.theforeman.org/plugins/foreman_templates/6.0/index.html

For many params I changed it, and I agree that it's too long, so ok!

nesanton: About settings, I'm saying this in line 33
`- Some defaults can be set in TemplateSync settings using M(foreman_setting) or GUI.`

plugins/modules/foreman_templates_import.py
nesanton: idk, they seem to be quite clear to me.
All three are only effective for imported templates. I can add something like:
```
- C(always) - read metadata and make associations for each imported template
- C(new) - read metadata and make associations for new templates only
- C(never) - ignore metadata
```

plugins/modules/foreman_templates_import.py
nesanton: yes,
maybe:
`Branch of the I(repo). Only for git-based repositories.`
Like they say here https://www.theforeman.org/plugins/foreman_templates/6.0/index.html

plugins/modules/foreman_templates_import.py
nesanton: ack!

plugins/modules/foreman_templates_import.py
mdellweg: I was wondering if it's worth beeing more specific about which settings are.
It's probably not "All settings can..."

plugins/modules/foreman_templates_import.py
mdellweg: OK, lets keep it as is.

plugins/modules/foreman_templates_import.py
nesanton: Actually they are all there indeed. Changing to:
`- Default values for all module options can be set using M(foreman_setting) for TemplateSync category or on the settings page in GUI.`

nesanton: @evgeni, do you want this in `notes:`?

evgeni: If it applies to all parameters, then yes, notes seems appropriate.

Oh, and could you please write UI or WebUI, not GUI? ;)

nesanton: there you go

PR #707, opened 2020-02-12
Replace hostname in fixture writing

### LINKED ISSUE #706: Rewrite URL in fixtures
It should be possible to hook into the vcr persister to always rewrite the server url to foreman.example.org.

I have seen people doing this manually, and I don't like those files to be touched by humans.

mdellweg:
Fixes #706

### DIFF
--- a/tests/vcr_python_wrapper.py
+++ b/tests/vcr_python_wrapper.py
@@ -4,6 +4,10 @@ import os
 import sys
 import vcr
 import json
+try:
+    from urlparse import urlparse, urlunparse
+except ImportError:
+    from urllib.parse import urlparse, urlunparse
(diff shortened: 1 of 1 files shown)


mdellweg:
@nesanton ?

nesanton:
oooooh this looks good! No more sed

evgeni:
Do we *really* need this urlparse/urlunparse abomination? But if it works...

mdellweg:
```
request.host = "foreman.example.org"
<...>
"msg": "Failed to connect to Foreman server: can't set attribute"
```
Seems to be read only.

evgeni:
Thanks for trying!

PR #712, opened 2020-02-19
Add image module

hlawatschek:


plugins/module_utils/foreman_helper.py
```diff
@@ -413,6 +413,15 @@ def find_resources_by_title(self, resource, titles, **kwargs):
     def find_resources_by_id(self, resource, obj_ids, **kwargs):
         return [self.find_resource_by_id(resource, obj_id, **kwargs) for obj_id in obj_ids]
 
+    def find_image(self, name, params=None, failsafe=False, thin=None):
+        images = self.list_resource('images', 'name="{0}"'.format(name), params=params)
+        # foreman API returns too many results (BUG #29066).
```
```suggestion
        # foreman API returns too many results, see https://projects.theforeman.org/issues/29066
```

evgeni: (that makes it easier to grep for workarounds in the code ;))

evgeni: also, see my comments in that redmine on why I think it fails for you :)

hlawatschek:
thanks for the review @evgeni. I walked through the code and it seems that i had already "fixed" the "issue" by accident during a development cycle. I reverted the change in foreman_helper.

evgeni:
Cool. I'll have a second look in the next few days!

plugins/modules/foreman_image.py
```diff
@@ -0,0 +1,128 @@
+#!/usr/bin/python
+# -*- coding: utf-8 -*-
+# (c) 2020 Mark Hlawatschek (ATIX AG)
+#
+# This file is part of Ansible
+#
```
evgeni:
```suggestion
```

plugins/modules/foreman_image.py
```diff
@@ -0,0 +1,128 @@
...
+# Ansible is free software: you can redistribute it and/or modify
```
evgeni:
```suggestion
# This program is free software: you can redistribute it and/or modify
```

plugins/modules/foreman_image.py
```diff
@@ -0,0 +1,128 @@
...
+# it under the terms of the GNU General Public License as published by
+# the Free Software Foundation, either version 3 of the License, or
+# (at your option) any later version.
+#
+# Ansible is distributed in the hope that it will be useful,
```
evgeni:
```suggestion
# This program is distributed in the hope that it will be useful,
```

plugins/modules/foreman_image.py
```diff
@@ -0,0 +1,128 @@
...
+# but WITHOUT ANY WARRANTY; without even the implied warranty of
+# MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+# GNU General Public License for more details.
+#
+# You should have received a copy of the GNU General Public License
+# along with Ansible.  If not, see <http://www.gnu.org/licenses/>.
```
evgeni:
```suggestion
# along with this program.  If not, see <http://www.gnu.org/licenses/>.
```

plugins/modules/foreman_image.py
```diff
@@ -0,0 +1,128 @@
...
+
+    entity_dict = module.clean_params()
+
+    with module.api_connection():
+        entity_dict['username'] = entity_dict['image_username']
+        if 'image_password' in entity_dict:
+            entity_dict['password'] = entity_dict['image_password']
+        operatingsystem_id = module.find_operatingsystem(entity_dict['operatingsystem'], thin=True)['id']
+        entity, entity_dict = module.resolve_entities(entity_dict)
+        scope = {'operatingsystem_id': operatingsystem_id, 'compute_resource_id': entity_dict['compute_resource']['id']}
```
evgeni: I think passing the OS in`scope` here is unnecessary, as we've seen in https://projects.theforeman.org/issues/29066 that passing *both* in the URL is not possible anyways.

plugins/modules/foreman_image.py
```diff
@@ -0,0 +1,128 @@
...
+    )
+
+    entity_dict = module.clean_params()
+
+    with module.api_connection():
+        entity_dict['username'] = entity_dict['image_username']
+        if 'image_password' in entity_dict:
+            entity_dict['password'] = entity_dict['image_password']
+        operatingsystem_id = module.find_operatingsystem(entity_dict['operatingsystem'], thin=True)['id']
+        entity, entity_dict = module.resolve_entities(entity_dict)
```
evgeni: I *think* this will fail if you try to add an image with the same name but different OS to the same CR, as it's uniqueness is defined on the (CR, OS) scope:
https://github.com/theforeman/foreman/blob/26dda8dbbf32d357657c3d22018c9d247fb0cb4e/app/models/image.rb#L12

Can you please add a test for that?

hlawatschek: Imo the expected behavior would be a modification of the existing image with the same name and cr/name is the identifier.

evgeni: is it? in the UI, I can create an image called "testimage" on CR1 with OS1, and another one with OS2.

hlawatschek: oh, you're right. This makes it ugly. identifier is cr/name-os then?

evgeni: yeah, I think you need "scope:CR, search:name+OS" to get an unique restult

hlawatschek: ok, I think I found a solution for this. Could you please re-review ;-) ?

evgeni: Will do. btw, no need to explicitly request another review in this case ;)

plugins/modules/foreman_image.py
```diff
@@ -0,0 +1,128 @@
...
+            name=dict(required=True),
+            username=dict(type='invisible'),
+            uuid=dict(required=True, aliases=['image_uuid']),
+            password=dict(type='invisible', no_log=True),
+            compute_resource=dict(type='entity', required=True),
+            architecture=dict(type='entity', required=True),
+            operatingsystem=dict(type='entity', required=True),
+            user_data=dict(type='bool')
+        ),
+        entity_scope="compute_resource"
```
evgeni:
```suggestion
        entity_scope="compute_resource",
        entity_resolve=False,
```

plugins/modules/foreman_image.py
```diff
@@ -0,0 +1,128 @@
...
+    )
+
+    entity_dict = module.clean_params()
+
+    with module.api_connection():
+        entity_dict['username'] = entity_dict['image_username']
+        if 'image_password' in entity_dict:
+            entity_dict['password'] = entity_dict['image_password']
+        operatingsystem_id = module.find_operatingsystem(entity_dict['operatingsystem'], thin=True)['id']
+        entity, entity_dict = module.resolve_entities(entity_dict)
```
evgeni:
```suggestion
        _entity, entity_dict = module.resolve_entities(entity_dict)
```

before my `entity_resolve=False`, this line would have searched for the *wrong* entity (without taking in account the OS). not bad per se, as you do the right search in L122, but also a waste of resources. with e_r=False, this does not happen anymore and `resolve_entities` will return `entity` as `None`, just for the sake of not polluting the `entity` var before L122, let's save it to a dumpster.

tests/test_playbooks/image.yml
```diff
@@ -0,0 +1,176 @@
...
+          vars:
+            image_name: "MYOS1"
+            image_uuid: "MY2NDOStemplate"
+            image_username: "myosadmin"
+            image_operatingsystem: "MY2NDOS"
+            image_compute_resource: "MYCR"
+            image_architecture: "MY1024"
+            image_state: absent
+            expected_change: true
+- hosts: localhost
```
evgeni:
```suggestion

- hosts: localhost
```

tests/test_playbooks/image.yml
```diff
@@ -0,0 +1,176 @@
...
+            image_name: "MYOS1"
+            image_uuid: "MYOStemplate-new"
+            image_username: "myosadmin"
+            image_operatingsystem: "MYOS SP1"
+            image_compute_resource: "MYCR"
+            image_architecture: "MY1024"
+            image_user_data: true
+            image_state: present
+            expected_change: true
+        - name: Create MY2NDIMAGE with same name
```
evgeni: can you please add the following two steps after this:
- "Do not change MY2NDIMAGE" (basically this same task, but expected_change: false)
- "Do not change MYIMAGE" (the same task as above, expected_change: false, ensuring that MYOS1 for MYOS SP1 still exists and was not updated by accident).

tests/test_playbooks/image.yml
```diff
@@ -0,0 +1,176 @@
...
+
+- hosts: tests
+  gather_facts: false
+  vars_files:
+    - vars/server.yml
+  tasks:
+    - name: 'MYIMAGE'
+      block:
+        - name: Create MYIMAGE
+          include: tasks/image.yml
```
evgeni:
```suggestion
          include_tasks: tasks/image.yml
```

(this applies to the whole test playbook :))

tests/test_playbooks/fixtures/image-0.yml
```diff
@@ -0,0 +1,335 @@
...
+- request:
+    body: null
+    headers:
+      Accept: [application/json;version=2]
+      Accept-Encoding: ['gzip, deflate']
+      Connection: [keep-alive]
+      Cookie: [_session_id=59d2159a88f34dc648baf654b5fd8477]
+      User-Agent: ['apypie (https://github.com/Apipie/apypie)']
+    method: GET
+    uri: https://foreman.example.org/api/compute_resources/46/images?per_page=4294967296&search=name%3D%22MYOS1%22
```
evgeni: here is the wrong search I mentioned in my earlier comment :)

tests/test_playbooks/fixtures/image-0.yml
```diff
@@ -0,0 +1,335 @@
...
+- request:
+    body: null
+    headers:
+      Accept: [application/json;version=2]
+      Accept-Encoding: ['gzip, deflate']
+      Connection: [keep-alive]
+      Cookie: [_session_id=59d2159a88f34dc648baf654b5fd8477]
+      User-Agent: ['apypie (https://github.com/Apipie/apypie)']
+    method: GET
+    uri: https://foreman.example.org/api/compute_resources/46/images?per_page=4294967296&search=name%3DMYOS1%2Coperatingsystem%3D33
```
evgeni: and here is the correct one you generated "manually" :)

plugins/modules/foreman_image.py
hlawatschek: after adding the changes, somehow entity_dict['compute_resource']['id'] was not set. Therefore I added a manual search for the compute resource.

hlawatschek:
Thanks @evgeni for the review and the explanations. I really appreciate that!
I tried to fulfill all your requested changes, but had to add a manual cr search.
Unfortunately the tests fail now which I cannot reproduce in my dev environment. Could you please have a look?

evgeni:
The failures are due to the merge of https://github.com/theforeman/foreman-ansible-modules/pull/724 yesterday, and you probably didn't rebase your local branch ontop of master yet.

I'll post inline comments how to fix that, but essentially:
* `s/entity_spec/foreman_spec/g`
* `s/entity_dict/module_params/g`
in your favorite editor should do it

plugins/modules/foreman_image.py
```diff
@@ -0,0 +1,130 @@
...
+    pass
+
+
+def main():
+    module = ForemanImageModule(
+        argument_spec=dict(
+            image_username=dict(required=True),
+            image_password=dict(no_log=True),
+        ),
+        entity_spec=dict(
```
evgeni:
```suggestion
        foreman_spec=dict(
```

plugins/modules/foreman_image.py
```diff
@@ -0,0 +1,130 @@
...
+            compute_resource=dict(type='entity', required=True),
+            architecture=dict(type='entity', required=True),
+            operatingsystem=dict(type='entity', required=True),
+            user_data=dict(type='bool')
+        ),
+        entity_scope="compute_resource",
+        entity_resolve=False,
+    )
+
+    entity_dict = module.clean_params()
```
evgeni:
```suggestion
    module_params = module.clean_params()
```

plugins/modules/foreman_image.py
```diff
@@ -0,0 +1,130 @@
...
+
+    with module.api_connection():
+        entity_dict['username'] = entity_dict['image_username']
```
evgeni:
```suggestion
        module_params['username'] = module_params['image_username']
```

plugins/modules/foreman_image.py
```diff
@@ -0,0 +1,130 @@
...
+        if 'image_password' in entity_dict:
```
evgeni:
```suggestion
        if 'image_password' in module_params:
```

plugins/modules/foreman_image.py
```diff
@@ -0,0 +1,130 @@
...
+            entity_dict['password'] = entity_dict['image_password']
```
evgeni:
```suggestion
            module_params['password'] = module_params['image_password']
```

plugins/modules/foreman_image.py
```diff
@@ -0,0 +1,130 @@
...
+        operatingsystem_id = module.find_operatingsystem(entity_dict['operatingsystem'], thin=True)['id']
```
evgeni:
```suggestion
        operatingsystem_id = module.find_operatingsystem(module_params['operatingsystem'], thin=True)['id']
```

plugins/modules/foreman_image.py
```diff
@@ -0,0 +1,130 @@
...
+        compute_resource_id = module.find_resource_by_name('compute_resources', entity_dict['compute_resource'], thin=True)['id']
```
evgeni:
```suggestion
        compute_resource_id = module.find_resource_by_name('compute_resources', module_params['compute_resource'], thin=True)['id']
```

plugins/modules/foreman_image.py
```diff
@@ -0,0 +1,130 @@
...
+        _entity, entity_dict = module.resolve_entities(entity_dict)
```
evgeni:
```suggestion
        _entity, module_params = module.resolve_entities(module_params)
```

plugins/modules/foreman_image.py
```diff
@@ -0,0 +1,130 @@
...
+        scope = {'compute_resource_id': compute_resource_id}
+        entity = module.find_resource('images', search="name={0},operatingsystem={1}".format(entity_dict['name'],
```
evgeni:
```suggestion
        entity = module.find_resource('images', search="name={0},operatingsystem={1}".format(module_params['name'],
```

plugins/modules/foreman_image.py
```diff
@@ -0,0 +1,130 @@
...
+                                      operatingsystem_id), name=entity_dict['name'], params=scope, failsafe=True)
```
evgeni:
```suggestion
                                      operatingsystem_id), name=module_params['name'], params=scope, failsafe=True)
```

plugins/modules/foreman_image.py
```diff
@@ -0,0 +1,130 @@
...
+        module.ensure_entity('images', entity_dict, entity, params=scope)
```
evgeni:
```suggestion
        module.ensure_entity('images', module_params, entity, params=scope)
```

hlawatschek:
Thanks! I made a rebase to the latest master, squashed some commits and applied your suggested changes. Tests are running successful now.

evgeni:
Thanks a ton Mark! And sorry for the inconvenience from the refactoring.

I still find it weird that you had to search for the CR yourself, I'll look into that a bit later.

evgeni:
Okay, so you need to search for CR/OS because resolve_entity() does not actually resolve in the case of state:absent.

hlawatschek:
Here you are :-) Thanks for your review and support!

evgeni:
Thanks for the contribution! You clearly didn't pick the easiest one for your first own module :)

PR #713, opened 2020-02-20
add support for foreman_azure_rm v2

hlawatschek:
In foreman_azure_rm V2 some API endpoints have been changed.
tested with foreman-1.24.2  and foreman_azure_rm 2.0.7

### DIFF
--- a/plugins/modules/foreman_compute_resource.py
+++ b/plugins/modules/foreman_compute_resource.py
@@ -2,6 +2,7 @@
 # -*- coding: utf-8 -*-
 # (c) Philipp Joos 2017
 # (c) Baptiste Agasse 2019
+# (c) Mark Hlawatschek 2020
 #
 # This file is part of Ansible
 #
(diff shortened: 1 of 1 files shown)


mdellweg:
Is it still working with the old one then?

hlawatschek:
No, unfortunately not. But imo the "old" one has not been delivered as package by upstream anyway.

plugins/modules/foreman_compute_resource.py
```diff
@@ -243,7 +248,7 @@ def get_provider_info(provider):
         return 'EC2', ['user', 'password', 'region']
 
     elif provider_name == 'azurerm':
-        return 'AzureRM', ['url', 'user', 'password', 'tenant', 'region']
+        return 'AzureRm', ['url', 'user', 'password', 'tenant', 'region', 'app_ident']
```
mdellweg: should this 'url' then be removed?

mdellweg: Also the capitalization should be adjusted in the other places as well.

PR #721, opened 2020-03-03
Run sanity checks with installed collection

mdellweg:
This will
* build a collection and install it in order to do linting and sanity checks
* bring test playbooks to a nice (ansible like) yaml linting standard
* remove the (presumably broken) debugging facilities


evgeni:
the last commit broke it :)

mdellweg:
@evgeni It becomes green again, once you sed the right files. >_<

Makefile
```diff
@@ -1,33 +1,48 @@
+NAMESPACE := $(shell python -c 'import yaml; print(yaml.safe_load(open("galaxy.yml"))["namespace"])')
+NAME := $(shell python -c 'import yaml; print(yaml.safe_load(open("galaxy.yml"))["name"])')
+VERSION := $(shell python -c 'import yaml; print(yaml.safe_load(open("galaxy.yml"))["version"])')
+MANIFEST := build/collections/ansible_collections/$(NAMESPACE)/$(NAME)/MANIFEST.json
+
+PLUGIN_TYPES := filter modules module_utils doc_fragments
```
mdellweg: Should i add 'inventory' and 'callback' now?

evgeni: Can't you just dynamically gather that with `$(wildcard plugins/*)` or similar?

I think the following should do it:
```make
```

/me goes to wash his hands

evgeni: (yes, this is, again dependant on the fact that one does not pass the trailing slash, shoot me)

Makefile
```diff
@@ -1,33 +1,48 @@
...
 	@echo "  test_<test>    to run a specific unittest"
 	@echo "  record_<test>  to (re-)record the server answers for a specific test"
-
-lint:
+	@echo "  clean_<test>   to run a specific test playbook with the teardown and cleanup tags"
+	@echo "  dist           to build the collection artifact"
+
+info:
+	@echo Building collection $(NAMESPACE)-$(NAME)-$(VERSION)
+	@echo   filter: $(notdir $(_filter))
```
evgeni: wonder if a loop over `PLUGIN_TYPES` would be cool or horrible here

Makefile
```diff
@@ -50,56 +65,45 @@ record_%: FORCE
...
-	# fix the imports to use the collection namespace
-	sed -i '/ansible.module_utils.foreman_helper/ s/ansible.module_utils/ansible_collections.theforeman.foreman.plugins.module_utils/g' $(COLLECTION_TMP)/plugins/modules/*.py
-	sed -i -e '/extends_documentation_fragment/{:1 n; s/- foreman/- theforeman.foreman.foreman/; t1}' $(COLLECTION_TMP)/plugins/modules/*.py
+$(MANIFEST): $(NAMESPACE)-$(NAME)-$(VERSION).tar.gz
+	ansible-galaxy collection install -p build/collections $< --force
 
-	# adjust README.md not to point to files that we don't ship in the collection
-	sed -i '/Documentation how to/d' $(COLLECTION_TMP)/README.md
+# fix the imports to use the collection namespace
+build/src/plugins/modules/%.py: plugins/modules/%.py | build
```
evgeni: what's `| build`?

Makefile
mdellweg: That is if we rely on the fact that each directory here is in fact a plugin type.
But yes, why not. I will try this.
We can also use the filter-out trick in the other wildcard thingies.

Makefile
mdellweg: I had the same question. As a loop, this will end up almost unreadable methinks.

Makefile
```diff
@@ -1,33 +1,48 @@
...
+	@echo   doc_fragments: $(notdir $(_doc_fragments))
+
+lint: tests/test_playbooks/vars/server.yml $(MANIFEST)
+	yamllint -f parsable tests/test_playbooks
+	ansible-playbook --syntax-check tests/test_playbooks/*.yml | grep -v '^$$'
 	flake8 --ignore=E402,W503 --max-line-length=160 plugins/ tests/
 
+sanity: $(MANIFEST)
+	# Fake a fresh git repo for ansible-test
+	cd $(<D) ; git init ; echo tests > .gitignore ; ansible-test sanity --local --python $(PYTHON_VERSION)
```
evgeni: Why the switch from `--venv` (in the old playbook) to `--local`?

Makefile
mdellweg: Build is a directoy, whose mtime will always be that of the last file changed in it.
So by declearing it a 'order-only-prerequisite' we can assure it is only called if the directory does not yet exist.

Makefile
mdellweg: It should be faster, because it does not create another venv, and in the CI we are already in a fine environment for that specific python version.

Makefile
evgeni: 🤯

Makefile
evgeni: I would consider every directory under `plugins` to be a plugin type, yes.

Makefile
evgeni: ack

Makefile
evgeni: Correct, but that will pollute my local development virtualenv when not run in CI.

Should we set that dynamically, based on [`GITHUB_ACTIONS=true`](https://help.github.com/en/actions/configuring-and-managing-workflows/using-environment-variables#default-environment-variables)?

Makefile
mdellweg: Turns out to be not too bad. ;)

Makefile
mdellweg: As i mostly care about the speed on my machine (with this specific task), it is more like should we do one or the other?

mdellweg: Switched it back to --venv.
Turned out not to be a big issue.

evgeni: Sanity does weird things to the env it's running in, so I'd prefer if it kept it hands from my system.

That said, we could do:

```make
SANITY_TARGET='local'

sanity:
    ansible-test sanity --$(SANITY_TARGET) --python $(PYTHON_VERSION)
```

and I can use SANITY_TARGET=venv :)

PR #723, opened 2020-03-04
Refactor mixins wrt organizations

mdellweg:
while doing it, i realized, no katello module was using resolve_entities.

@bagasse maybe have a look?

### DIFF
--- a/plugins/modules/foreman_scc_account.py
+++ b/plugins/modules/foreman_scc_account.py
@@ -117,8 +117,12 @@ RETURN = ''' # '''
 from ansible.module_utils.foreman_helper import KatelloEntityAnsibleModule
 
 
+class KatelloSccAccountModule(KatelloEntityAnsibleModule):
+    pass
+
+
 def main():
-    module = KatelloEntityAnsibleModule(
+    module = KatelloSccAccountModule(
         entity_spec=dict(
             name=dict(required=True),
             updated_name=dict(),
@@ -134,16 +138,12 @@ def main():
         ),
     )
 
-    entity_dict = module.clean_params()
-
     module.task_timeout = 4 * 60
 
     with module.api_connection():
-        entity_dict, scope = module.handle_organization_param(entity_dict)
-
-        failsafe = (module.state != 'synced')
-
-        entity = module.find_resource_by_name('scc_accounts', name=entity_dict['name'], params=scope, failsafe=failsafe)
+        module.entity_opts['failsafe'] = (module.state != 'synced')
+        entity, entity_dict = module.resolve_entities()
+        scope = {'organization_id': entity_dict['organization']['id']}
 
         if not module.desired_absent:
             if not entity:
--- a/plugins/modules/katello_activation_key.py
+++ b/plugins/modules/katello_activation_key.py
@@ -192,14 +192,18 @@ def override_to_boolnone(override):
     return value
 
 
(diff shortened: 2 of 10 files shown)


bagasse:
I like it ! :) Yes I didn't move katello modules on resolve_entities, the PR introducing it was already too big... Katello modules migration was on my todo-list but have to postpone due to some other priorities... Thanks !

mdellweg:
I just realised repository is giving me a serious problem, as its scope (product) is scoped again. We might need a proper dependency solution via topological ordering.

bagasse:
Yes. I missed katello repository case when I did 'scope resolution'.

evgeni:
Yeah, repositories are super weird in regard of scoping. I'm perfectly fine if we come to a 80% katello fix, ignoring repositories :)

mdellweg:
To not interfere too much with the efforts in #724, i am unmarking this one WIP now.
Please review.

plugins/modules/foreman_scc_account.py
```diff
@@ -119,6 +119,7 @@
 
 def main():
     module = KatelloEntityAnsibleModule(
+        entity_name='scc_account',
```
evgeni: Can't this be avoided if the class is `KatelloSccAccountModule`? I read the helper code that it should work?

https://github.com/theforeman/foreman-ansible-modules/blob/923f37f22af15f0b883e286fbb6a2ce1ce66a9c6/plugins/module_utils/foreman_helper.py#L770-L786

plugins/modules/katello_activation_key.py
```diff
@@ -194,12 +194,13 @@ def override_to_boolnone(override):
 
 def main():
     module = KatelloEntityAnsibleModule(
+        entity_name='activation_key',
```
evgeni: Same q as above ;)

plugins/modules/katello_activation_key.py
```diff
@@ -344,7 +334,7 @@ def main():
                     current_host_collection_ids = set(activation_key['host_collection_ids'])
                 else:
                     current_host_collection_ids = set(item['id'] for item in activation_key['host_collections'])
-                desired_host_collections = module.find_resources_by_name('host_collections', host_collections, params=scope, thin=True)
+                desired_host_collections = host_collections
                 desired_host_collection_ids = set(item['id'] for item in desired_host_collections)
```
evgeni:
```suggestion
                desired_host_collection_ids = set(item['id'] for item in host_collections)
```

plugins/modules/katello_content_credential.py
```diff
@@ -74,19 +74,16 @@
 
 def main():
     module = KatelloEntityAnsibleModule(
+        entity_name='content_credential',
```
evgeni: You know the drill ;)

plugins/modules/foreman_scc_account.py
mdellweg: Sure it can, but i ended up subclassing KatelloEntityAnsibleModule (3 loc) just to trigger that extra magic, while this is one concice line to describe the entity in question.
I prefer this one.

evgeni: But it makes it inconsistent with the foreman modules :(

evgeni: I won't block this, but highly prefer consistency over 2 lines save.

PR #724, opened 2020-03-05
Rename entity_spec and make it work in non-entity modules

### LINKED ISSUE #404: make the entity_spec work for non-entity modules

nesanton:
* Rename entity_spec to foreman_spec
* Make foreman_spec usable by ForemanAnsibleModule class

I'm sure this is too rookie of an attempt to fix #404, but I wanted to begin


tests/test_entity_spec_helper.py
```diff
@@ -1,11 +1,11 @@
-from plugins.module_utils.foreman_helper import _entity_spec_helper
```
mdellweg: That file could be renamed, too.

nesanton: oh right

nesanton:
Idk if `entity_dict` still makes sense in `entity_dict = module.clean_params()`

nesanton:
> Idk if `entity_dict` still makes sense in `entity_dict = module.clean_params()`

when I look at it it makes less and less sense. Should we call it something like `module_params` everywhere?

nesanton:
@evgeni, what do you think of ^^^

nesanton:
This will totally screw all unmerged PRs ;)

evgeni:
So bear with me while I read the diff, this PR does essentially the following:

* move spec initialization/parsing from ForemanEntityAnsibleModule to ForemanAnsibleModule
* `s/entity_spec/foreman_spec/g`
* `s/entity_dict/module_params/g`

So now non-Entity modules could use `foreman_spec`, but they so far don't yet, correct? Not saying it has to be in this PR, just making sure I did not overlook anything.

That all said, the changes look sane to me.

@bagasse and @mdellweg your opinions as heavy cerators/users of the spec are appreciated :)

nesanton:
> * move spec initialization/parsing from ForemanEntityAnsibleModule to ForemanAnsibleModule
> * `s/entity_spec/foreman_spec/g`
> * `s/entity_dict/module_params/g`

100% precision.

I've tried merging it with templates_sync and replacing arg_spec to forman_spec. Seems to work

nesanton:
Here you go

bagasse:
It's ok for me too :)

mdellweg:
Maybe a little late to the party, but i like it!

PR #726, opened 2020-03-06
Fix sanity tests

ehelms:


ehelms:
Oops, I did not mean to keep the getting rid of that break. I put it back.

PR #734, opened 2020-03-13
also pass params when fetching full details

evgeni:
otherwise you can't fetch full details from a katello resource


mdellweg:
Should we have a test here?

evgeni:
We would need to test f_s_f against Katello, which I'd rather not implement?

evgeni:
I swear I tested it (against a Satellite 6.6 :see_no_evil:)

Issue #728, opened 2020-03-11
Error while performing update on subnets


ephracis:
##### SUMMARY

I am still seeing #428 on 071e59c1cc8a52574a9ab45d675b2bbd486e67c6.

##### ISSUE TYPE
 - Bug Report

##### ANSIBLE VERSION

```
ansible 2.9.5
  config file = /Users/chbr/repos/work/components/foreman/ansible.cfg
  configured module search path = ['/Users/chbr/.ansible/plugins/modules', '/usr/share/ansible/plugins/modules']
  ansible python module location = /usr/local/lib/python3.7/site-packages/ansible
  executable location = /usr/local/bin/ansible
  python version = 3.7.6 (default, Dec 30 2019, 19:38:26) [Clang 11.0.0 (clang-1100.0.33.16)]
```

##### KATELLO/FOREMAN VERSION

```
tfm-rubygem-katello-3.14.1-1.el7.noarch
foreman-1.24.2-1.el7.noarch
```

##### APYPIE VERSION

```
Version: 0.2.1
```

##### STEPS TO REPRODUCE

Run the task `foreman_subnet` twice. First time is fine but second time, and every retry gives the same error.

Task file inside role:
```yaml
- name: Create subnets
  foreman_subnet:
    username: admin
    password: "{{ foreman_admin_password }}"
    server_url: "{{ foreman_url }}"
    validate_certs: no
    state: "{{ item.state | default(omit) }}"
    name: "{{ item.name }}"
    network_type: "{{ item.network_type | default(omit) }}"
    dns_primary: "{{ item.dns_primary | default(omit) }}"
    dns_secondary: "{{ item.dns_secondary | default(omit) }}"
    domains: "{{ item.domains | default(omit) }}"
    gateway: "{{ item.gateway | default(omit) }}"
    network: "{{ item.network | default(omit) }}"
    cidr: "{{ item.cidr | default(omit) }}"
    mask: "{{ item.mask | default(omit) }}"
    from_ip: "{{ item.from_ip | default(omit) }}"
    to_ip: "{{ item.to_ip | default(omit) }}"
    boot_mode: "{{ item.boot_mode | default(omit) }}"
    ipam: "{{ item.ipam | default(omit) }}"
    dhcp_proxy: "{{ item.dhcp_proxy | default(omit) }}"
    tftp_proxy: "{{ item.tftp_proxy | default(omit) }}"
[...]

PR #738, opened 2020-03-16
add required_plugins to ForemanAnsibleModule

evgeni:
this PR adds `required_plugins` as a parameter to `ForemanAnsibleModule`, which allows individual modules to express a dependency on Foreman plugins, either for the whole module, or for individual parameters.

the syntax is:
```python
required_plugins=[
    ('plugin_name', ['list', 'of', 'params']),
    ('plugin_name', ['*']),
]
```

Fixes: #728

### DIFF
--- a/plugins/module_utils/foreman_helper.py
+++ b/plugins/module_utils/foreman_helper.py
@@ -696,7 +696,9 @@ class ForemanAnsibleModule(AnsibleModule):
         missing_plugins = []
         for (plugin, params) in self.required_plugins:
             for param in params:
-                if param in self.clean_params() and not self.has_plugin(plugin):
+                if (param in self.clean_params() or param == '*') and not self.has_plugin(plugin):
+                    if param == '*':
+                        param = 'the whole module'
                     missing_plugins.append("{0} (for {1})".format(plugin, param))
         if missing_plugins:
             missing_msg = "The server is missing required plugins: {0}.".format(', '.join(missing_plugins))
--- a/plugins/module_utils/foreman_helper.py
+++ b/plugins/module_utils/foreman_helper.py
@@ -40,6 +40,15 @@ parameter_foreman_spec = dict(
     parameter_type=dict(default='string', choices=['string', 'boolean', 'integer', 'real', 'array', 'hash', 'yaml', 'json']),
 )
 
+_PLUGIN_RESOURCES = {
+    'discovery': 'discovery_rules',
+    'katello': 'subscriptions',
+    'openscap': 'scap_contents',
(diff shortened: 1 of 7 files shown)


evgeni:
@ephracis for your specific problem, this would have generated the following output:

```
The server is missing required plugins: discovery (for discovery_proxy).
```

Would that have been helpful? :)

evgeni:
(I'd like to hear back from Christoffer before merging this if it gets approved before that)

plugins/module_utils/foreman_helper.py
```diff
@@ -261,7 +273,7 @@ def _patch_subnet_rex_api(self):
             see https://projects.theforeman.org/issues/19086
         """
 
-        if 'remote_execution_features' not in self.foremanapi.apidoc['docs']['resources']:
+        if not self.has_plugin('remote_execution'):
```
mdellweg: We could move this patching to the place where we check for the availability of the plugin, right? No need to patch, if it is not used...

plugins/module_utils/foreman_helper.py
```diff
@@ -79,7 +88,7 @@ class KatelloMixin(OrganizationMixin):
     def connect(self):
         super(KatelloMixin, self).connect()
 
-        if 'subscriptions' not in self.foremanapi.resources:
+        if not self.has_plugin('katello'):
```
mdellweg: Would it be possible to just add 'katello' to the `required_plugins`?

evgeni: Yeah, I considered that, would add ~5 lines and remove 2 XD

plugins/module_utils/foreman_helper.py
evgeni: You mean only patch if the param is used?

mdellweg: I meant only patch if the plugin is required.

ephracis:
Yes, that message would have been plenty helpful.

plugins/module_utils/foreman_helper.py
mdellweg: This is not going to be a blocker for me.

plugins/module_utils/foreman_helper.py
evgeni: implemented

plugins/module_utils/foreman_helper.py
evgeni: I'd like to leave that out for now. Feel free to open a follow up ;)

evgeni:
now with :100: more docs!

PR #744, opened 2020-03-18
implement inventory plugin testing

evgeni:
draft for now, as this is a bit complicated.

the plugin is called `theforeman.foreman.foreman` in the repo, but that name cannot be used outside a collection context, and that is where we run our tests right now (running tests with collections would require Ansible 2.9, but we support 2.7+).

there is a "small hack" commit in the history here, which actually renames the plugin back to `foreman` and uses that -- still fails on Ansible 2.7 for some reason but IMHO what we can go with that for now?


evgeni:
@ehelms @mdellweg for your pleasure :)

ansible.cfg
```diff
@@ -3,6 +3,9 @@ library = plugins/modules
 module_utils = plugins/module_utils
 doc_fragment_plugins = plugins/doc_fragments
 filter_plugins = plugins/filter
+inventory_plugins = plugins/inventory
+callback_plugins = plugins/callback
```
mdellweg: You should not need that if you use it from the installed collection.

tests/test_playbooks/inventory_plugin.yml
```diff
@@ -0,0 +1,117 @@
...
+          production:
+            adapter: sqlite3
+            database: db/production.sqlite3
+            pool: 10
+            timeout: 5000
+
+    - name: fetch Foreman container
+      docker_image:
+        name: quay.io/foreman/foreman:2.0-stable
+        # not setting source for now, this needs Ansible 2.8+
```
mdellweg: I think, we should only test that plugin with ansible >= 2.9 anyway.

ansible.cfg
mdellweg: honestly i have been thinking of having one ansible.cfg for use with and one without the collection, respectively.
Mixing them is quite fuzzy about where the modules/plugins are looked up.

tests/test_playbooks/inventory_plugin.yml
kdelee: yeah this plugin installed w/ collections would only be used w 2.9+ if I understand correctly

evgeni: Yepp, just the playbook gets currently executed on 2.7+, will fix that ;)

ansible.cfg
evgeni: agree, but I need that if I want to test outside the collection -- which I think I want for now because otherwise I can't test on 2.7 (and 2.8)

tests/test_playbooks/inventory_plugin.yml
evgeni: fixed

evgeni:
Ideally, we will revert "[temp] pretend that we have the "foreman" inventory plugin" when we can switch to collection-only testing with 2.9+

ehelms:
Why do we want to test outside the collection?

evgeni:
@ehelms because we want to test against 2.7 which does not support collections at all, and 2.8 which can consume but not build collections. and by "test" I mean "run all the tests", not "test this one plugin"

ehelms:
Might be controversial, but we built these and created a collection. Master is a collection. For me, we only need to support Ansible's that support collections in master. What edge cases am I not accounting for in that view point?

evgeni:
Actually, master is not a collection, it only becomes one when we build it:
https://github.com/theforeman/foreman-ansible-modules/blob/83181d145ac70708d00471148baa3a874cf006f7/Makefile#L83-L86

Today, we say we support Ansible 2.7+ (= all Ansible versions that are under active community support right now) and document how to use this repo w/o a collection: https://github.com/theforeman/foreman-ansible-modules/blob/master/README.md#run-from-git.

.github/workflows/main.yml
```diff
@@ -50,6 +50,8 @@ jobs:
         run: make test-setup
       - name: fake redhat-uep.pem for redhat_manifest module
         run: sudo mkdir -p /etc/rhsm/ca/ && sudo touch /etc/rhsm/ca/redhat-uep.pem
+      - name: adjust inventory name
+        run: sed -i '/NAME/ s/theforeman\.foreman\.//' plugins/inventory/foreman.py
```
mdellweg: Can we have this in the other direction?
- Change the file for real to make it work outside of the collection
- Compile it in the Makefile for inclusion into the collection

tests/test_playbooks/inventory_plugin.yml
```diff
@@ -0,0 +1,162 @@
...
+        name: foreman
+        image: quay.io/foreman/foreman:2.0-stable
+        command: bundle exec bin/rails server -b 0.0.0.0
+        published_ports:
+          - "0.0.0.0:3000:3000"
+        volumes:
+          - "/tmp/database.yml:/home/foreman/config/database.yml"
+
+    - name: hack the gibson
+      command: "docker exec foreman sed -i 's/FactImporter.importer_for(type)/PuppetFactImporter/' /home/foreman/app/models/host/base.rb"
```
evgeni: this is a terrible hack, but for some reason `FactImporter.importer_for(type)` does return `nil` inside the container, whereas it should return `PuppetFactImporter` and I have no idea why

.github/workflows/main.yml
evgeni: Yes, we could do that. I've been avoiding it for now for the sake of "we don't actually want to support that outside the collection".

mdellweg: But this step would also be necessary to run the test locally, yet it is hidden deeply in the gh workflow.

tests/test_playbooks/inventory_plugin.yml
```diff
@@ -0,0 +1,141 @@
...
+        name: foreman
+        image: quay.io/foreman/foreman:2.0-stable
+        command: bash -c "bundle exec bin/rails db:create db:migrate && bundle exec bin/rails db:seed && bundle exec bin/rails server -b 0.0.0.0"
+        env:
+          SEED_ADMIN_PASSWORD: changeme
+          RAILS_ENV: production
+        published_ports:
+          - "0.0.0.0:3000:3000"
+        volumes:
+          - "/tmp/database.yml:/home/foreman/config/database.yml"
```
ohadlevy: you can define yhe database as an env variable instead of a file, e.g.
```
ENV DATABASE_URL=sqlite3:tmp/bootstrap-db.sql
```

tests/test_playbooks/inventory_plugin.yml
```diff
@@ -0,0 +1,128 @@
...
+          hostvars['{{ item.key }}']['foreman_domain_name']  == 'example.com'
+      with_dict: "{{ foreman_hosts }}"
+
+    - name: test that all hosts have the OS fact set
+      assert:
+        that: >
+          hostvars['{{ item.key }}']['foreman_operatingsystem_name']  == 'CentOS 7.7'
+      with_dict: "{{ foreman_hosts }}"
+
+    # FIXME: add some cache tests
```
ehelms: Switch this to a FAM issue?

tests/test_playbooks/inventory_plugin.yml
```diff
@@ -0,0 +1,128 @@
...
+            famtesthost: true
+            operatingsystem: CentOS
+            operatingsystemrelease: "7.7"
+        force_basic_auth: true
+        status_code: 201
+        body_format: json
+      with_dict: "{{ foreman_hosts }}"
+      ignore_errors: true
+
+    - command: "docker logs foreman"
```
ehelms: Just for debugging test output?

evgeni: eh, yeah, that can go

tests/test_playbooks/inventory_plugin.yml
evgeni: yupp

evgeni: #755

ehelms:
Looks good by me to drop it from draft mode.

evgeni:
@mdellweg @ehelms non draft now, made the plugin listen to the "foreman" name in the repo and modify it to the collection name in the Makefile.

ansible.cfg
mdellweg: Do we test on 2.7/2.8 now?

evgeni: no, 2.9+

evgeni: the inventory on 2.9+, the other on 2.7+
and because of that the test suite itself runs outside the collection.

evgeni:
I squashed some commits to make the history nicer and going to merge this as soon as CI is green

PR #745, opened 2020-03-18
move common params from hostgroup to HostMixin

evgeni:


evgeni:
I've not added tests to host yet

plugins/module_utils/foreman_helper.py
```diff
@@ -175,10 +175,33 @@ def __init__(self, foreman_spec=None, **kwargs):
...
+            openscap_proxy=dict(type='entity', resource_type='smart_proxies'),
+            content_source=dict(type='entity', scope='organization', resource_type='smart_proxies'),
+            lifecycle_environment=dict(type='entity', scope='organization'),
+            kickstart_repository=dict(type='entity', scope='organization', resource_type='repositories'),
+            content_view=dict(type='entity', scope='organization'),
         )
         if foreman_spec:
             args.update(foreman_spec)
-        super(HostMixin, self).__init__(foreman_spec=args, **kwargs)
+        required_plugins = kwargs.pop('required_plugins', []) + [
```
mdellweg: It works for Mixins? :roller_coaster:

evgeni:
Yes, just not for every param

plugins/module_utils/foreman_helper.py
evgeni: yes

evgeni:
added tests and dropped the "needs HG" requirement as that's not true anymore

tests/test_playbooks/vars/host.yml
```diff
@@ -5,3 +5,9 @@ host:
   hostgroup: "test_group"
   domain: "example.net"
   subnet: Test subnet4
+  arch: superARCH
```
mdellweg: Is this a 42 bit architecture?

evgeni: 192!

It's a reference to https://en.wikipedia.org/wiki/SuperH ;-)

PR #746, opened 2020-03-19
added 'architecture' to 'content_view_filter'

dericcrago:
It appears that this was just missed since the `content_filter_rule_rpm_spec` already included `architecture`.


evgeni:
thanks! this seems indeed like an oversight.

Issue #737, opened 2020-03-13
resolve_entities does not resolve the scope when entity_resolve=False


evgeni:
##### SUMMARY
This came out when looking at #712 and I am not 100% sure we want to fix that or ignore the corner case.

What happened?
* images are unique per compute_resource AND operating_system
* to find the unique image you want to manipulate, you need to scope by CR and then add OS to the search (ugh)
* as this is not easily supported by resolve_entities, we used `entity_resolve=False`
* when we do that, resolve_entities also does not resolve the CR scope that we passed with `entity_scope=compute_resource`

should it tho? or should it not? I am not sure.

@bagasse I'd love your thoughts on that.

see https://github.com/theforeman/foreman-ansible-modules/blob/827acd8232c154a5f075e9d53112ba96114fa4b3/plugins/modules/foreman_image.py#L120-L125 for an example of code that maybe could be simplified, but at which cost?

##### ISSUE TYPE
 - Bug Report

##### ANSIBLE VERSION

```

```

##### KATELLO/FOREMAN VERSION

```

```

##### APYPIE VERSION

```

```

##### STEPS TO REPRODUCE

```yaml

```

##### ACTUAL RESULTS

```

```

bagasse:
Yes, it should resolve compute_resource, for me it's a bug. When `entity_resolve=False` is set, only the related entity should not be resolved.

evgeni:
Thanks! I think @mdellweg has an (accidental) fix for that after we merge his last refactoring pr ;)

mdellweg:
With automatic scope dependency resolution, this should indeed be fixed.

Issue #736, opened 2020-03-13
We want to use `resolve_entities` in non entity modules


mdellweg:
`resolve_entites` should be able to run with and without a 'central' entity.
It might be best to resolve the 'central entity' in much the same way as any other.

This might be closely related to #735

mdellweg:
`lookup_entity` is now part of `ForemanAnsibleModule`.

Issue #735, opened 2020-03-13
Dependencies in resolve_entities


mdellweg:
For `katello_repository`, the _entity_ is scoped by `product`, which in turn is scoped by `organization`.

So `resolve_entities` should work in topologically sorted order to handle those cases.

mdellweg:
Did you mean `katello_repository`?

mdellweg:
The new `lookup_entity` solves this problem.

PR #747, opened 2020-03-20
Resolve entities

mdellweg:
~~Far from finished, but i think it is time to share my efforts.~~

This ~~will most probably~~ adress(es) #735, #736 and #737.


plugins/module_utils/foreman_helper.py
```diff
@@ -829,13 +901,41 @@ def run(self, module_params=None, entity=None, search=None):
...
+
+        return new_entity
+
+    def cycle(self):
+        """ lookup entities, ensure entity, remove sensitive data, manage parameters.
+            Like 'run', just faster and more convenient...
+        """
+        module_params = self.foreman_params
+        if not self.desired_absent:
+            self.auto_lookup_entities()
```
mdellweg: I keep thinking, that we would not even need this, if we did the entity lookup completely lazy.

mdellweg:
I think, this is the first version of it with all the precious features we love,
including auto_lookup and entities with parents.

plugins/modules/foreman_bookmark.py
```diff
@@ -114,12 +114,20 @@ def main():
...
         ),
+        entity_resolve=False,
     )
 
-    module_params = module.clean_params()
+    module_params = module.foreman_params
 
     with module.api_connection():
-        module.run(search='name="{0}",controller="{1}"'.format(module_params['name'], module_params['controller']))
+        # TODO figure out, how to map this kind of search to lookup_entity
```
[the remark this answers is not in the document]
evgeni: Do we *need* to fix that for that one (or two) modules that need it?

plugins/modules/foreman_bookmark.py
```diff
@@ -114,12 +114,20 @@ def main():
             ['state', 'present', ['query']],
             ['state', 'present_with_defaults', ['query']],
         ),
+        entity_resolve=False,
```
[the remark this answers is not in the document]
evgeni: Maybe, yeah.

plugins/modules/foreman_compute_attribute.py
```diff
@@ -100,13 +100,14 @@ def main():
...
     with module.api_connection():
-        _entity, module_params = module.resolve_entities(module_params)
-        compute_attributes = module_params['compute_resource'].get('compute_attributes')
-        entity = next((item for item in compute_attributes if item.get('compute_profile_id') == module_params['compute_profile']['id']), None)
-        module.run(module_params=module_params, entity=entity)
+        compute_attributes = module.lookup_entity('compute_resource').get('compute_attributes')
+        compute_profile_id = module.lookup_entity('compute_profile').get('id')
+        entity = next((item for item in compute_attributes if item.get('compute_profile_id') == compute_profile_id), None)
+        module.foreman_params['entity'] = entity
+        module.foreman_spec['entity']['resolved'] = True
```
evgeni: that feels like exposing too much internals to the user. `module.set_entity(entity)` that wraps those two?

plugins/modules/foreman_external_usergroup.py
```diff
@@ -91,7 +91,9 @@ def main():
             if external_usergroup['name'] == module_params['name']:
                 entity = external_usergroup
 
-        _entity, module_params = module.resolve_entities(module_params)
+        module.foreman_params['entity'] = entity
+        module.foreman_spec['entity']['resolved'] = True
```
evgeni: yeah, `module.set_entity(entity)`

plugins/modules/foreman_location.py
```diff
@@ -107,14 +107,14 @@ def main():
     module = ForemanLocationModule(
         foreman_spec=dict(
             name=dict(required=True),
-            parent=dict(type='entity'),
+            parent=dict(type='entity', resource_type='locations'),
```
evgeni: can't `parent` be guessed from `entity` as it's the same resource type?

plugins/modules/foreman_provisioning_template.py
```diff
@@ -326,8 +326,9 @@ def main():
         if module.state == 'present_with_defaults':
             module.fail_json(msg="'state: present_with_defaults' and 'name: *' cannot be used together")
         if module.desired_absent:
-            if len(module_params.keys()) != 1:
-                module.fail_json(msg="When deleting all templates, there is no need to specify further parameters.")
+            further_params = set(module_params.keys()) - {'name', 'entity'}
```
evgeni: scope creep, but this check exists in many modules and should be in the helper

plugins/modules/foreman_setting.py
```diff
@@ -87,30 +81,34 @@ class ForemanSettingModule(ForemanAnsibleModule):
...
-        argument_spec=dict(
+        foreman_spec=dict(
             name=dict(required=True),
             value=dict(type='raw'),
         ),
     )
 
-    module_params = module.clean_params()
+    # TODO Maybe refactor this into a EntityMixin
+    module.foreman_spec.update(_foreman_spec_helper(dict(
```
evgeni: calling a private helper here feels weird. and the code is hard to read. the normal magic can't help here as it's not an entity module?

plugins/modules/foreman_bookmark.py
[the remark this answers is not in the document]
mdellweg: Absolutely. I've had the same thought and then forgot about it.

plugins/modules/foreman_compute_attribute.py
mdellweg: Right.
I hoped to be abe to avoid it altogether.

plugins/modules/foreman_location.py
mdellweg: As we do some guesswork around parent anyway, why not.

plugins/modules/foreman_setting.py
mdellweg: Well, it could be an Entity module if that would not imply state.

There could be a lot more separation of concerns...

mdellweg:
@bagasse, @evgeni
I see you both worrying about the name `cycle`. I didn't want to break all modules at once, so the plan was to build something beneth it (needing a new name, and `ng_run` didn't stick and `perform_lifecycle` i didnt like either).
I am happy with renaming it back, once i have crawled through all modules. Replacing the 'old' code was the plan all along.
If i do it in one PR, it will probably break all other PR's immediately. But we have seen that before...

plugins/modules/foreman_bookmark.py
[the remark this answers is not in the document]
bagasse: Idem when I introduced it ;)

plugins/modules/foreman_bookmark.py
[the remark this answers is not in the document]
bagasse: You are right, if only needed for one or two it's not needed at all.

plugins/modules/foreman_bookmark.py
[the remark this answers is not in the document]
mdellweg: Should we rename `entity_opts` to `entity_spec` on the way?

plugins/modules/foreman_provisioning_template.py
mdellweg: The whole affects multiple should be.
That's my exercise to the reader these days.

plugins/modules/foreman_bookmark.py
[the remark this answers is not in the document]
bagasse: And then rename it `foreman_spec` :smile:. Didn't think about this but `entity_spec` would be more consistent with others "sub entities" specs.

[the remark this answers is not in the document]
mdellweg: I actually used it that way internally:
https://github.com/theforeman/foreman-ansible-modules/pull/747/files#diff-dad222080f86e9bcc4e9132d0d8aa70fR497

plugins/modules/foreman_provisioning_template.py
evgeni: You mean the reader of https://github.com/theforeman/foreman-ansible-modules/issues/244?

mdellweg: My new take on 'foreman_installation_medium' might help there.

plugins/module_utils/foreman_helper.py
```diff
@@ -471,6 +483,77 @@ def find_puppetclass(self, name, environment=None, params=None, failsafe=False,
...
+                        failsafe=failsafe, thin=thin, params=params,
+                    ) for value in self.foreman_params[key]]
+        self.set_entity(key, result)
+        return result
+
+    def auto_lookup_entities(self):
+        return [
+            self.lookup_entity(key)
+            for key, entity_spec in self.foreman_spec.items()
+            if entity_spec.get('resolve', True) and entity_spec.get('type') in {'entity', 'entity_list'}
```
evgeni: Out of curiosity, why are you using a set (`in {a, b}`) and not a list (`in [a, b]`) or a tuple (`in (a, b)`). Is set access quicker?

plugins/module_utils/foreman_helper.py
```diff
@@ -788,23 +852,47 @@ def __init__(self, **kwargs):
...
+
+        self.foreman_spec.update(_foreman_spec_helper(dict(
+            entity=dict(
+                type='entity',
+                flat_name='id',
+                ensure=False,
+                **self.entity_opts
+            ),
+        ))[0])
+        if 'parent' in self.foreman_spec and self.foreman_spec['parent'].get('type') == 'entity':
```
```suggestion

        if 'parent' in self.foreman_spec and self.foreman_spec['parent'].get('type') == 'entity':
```

evgeni: (empty line for visual break)

plugins/modules/foreman_provisioning_template.py
evgeni: You mean "pull it out of the with api section so it can be moved to a helper"? yes, looks like it.

evgeni:
Naaaaaaaaaaaaaaaaaaais! :)

Given you have now ported all modules from run to cycle, shall we recycle (lol) the run name for the new way of life? Also dis needs docs update, but you knew that already :)

plugins/module_utils/foreman_helper.py
mdellweg: Superstition. In theory lookup in a set should be quicker. Building it might not be, and for two elements, it may not be worth discussing.

mdellweg:
Now i really have all modules (i hope)

evgeni:
test errors unrelated, seems the inventory container is not 100% stable

plugins/modules/katello_content_view_filter.py
```diff
@@ -172,9 +172,10 @@
 
 RETURN = ''' # '''
 
-from ansible.module_utils.foreman_helper import KatelloAnsibleModule
+from ansible.module_utils.foreman_helper import KatelloAnsibleModule, _foreman_spec_helper
 
 content_filter_spec = {
+    'id': {},
```
evgeni: shouldn't these be `invisible`?

plugins/modules/katello_content_view_filter.py
```diff
@@ -240,32 +250,39 @@ def main():
         ),
     )
 
-    module_params = module.clean_params()
-    filter_state = module_params.pop('filter_state')
-    rule_state = module_params.pop('rule_state')
+    # TODO Maybe refactor this into a EntityMixin
```
evgeni: can you open an issue for that please

plugins/modules/katello_sync.py
```diff
@@ -103,25 +103,22 @@
 def main():
     module = KatelloAnsibleModule(
         foreman_spec=dict(
-            product=dict(required=True),
-            repository=dict(),
+            product=dict(type='entity', scope=['organization'], required=True),
+            repository=dict(type='entity', scope=['product'], failsafe=True),
+            # This should be scoped more explicit for better serch performance, but needs rerecording
```
evgeni: can you open an issue for this please

plugins/modules/foreman_job_template.py
```diff
@@ -350,78 +350,82 @@ def main():
...
-            extra_params = {}
+            # The name could have been determined to late, so copy it again
+            module.foreman_params['entity'] = module.foreman_params['name']
+            entity = module.lookup_entity('entity')
+            # TemplateInputs need to be added as separate entities later
+            template_inputs = module.foreman_params.get('template_inputs')
 
-        if not affects_multiple:
-            job_template = module.ensure_entity('job_templates', module_params, entity, params=extra_params)
+            # TODO Use module.run _with_ extra_params
```
evgeni: issue please? :)

evgeni:
There are a few todo comments that should be converted to issues IMHO, otherwise looks good. Will have another look on monday with a fresh brain, but otherwise this is going in!

plugins/modules/katello_content_view_filter.py
mdellweg: Maybe if they were used in the `foreman_spec` that gives raise to the `argument_spec`, but they aren't.

plugins/modules/katello_sync.py
mdellweg: Does it improve search performance?

plugins/modules/katello_content_view_filter.py
mdellweg:
```suggestion
```

plugins/modules/foreman_job_template.py
mdellweg: no issue needed anymore

evgeni:
Interesting that the 3.7/devel failure is consistent.

plugins/modules/foreman_bookmark.py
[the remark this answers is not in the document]
mdellweg: It turns out, this is a little bit harder. So i'd like to postpone it to another contribution.

After my first try, i suspect `entity_key` and `search_by` not to be the same.

plugins/modules/katello_sync.py
evgeni: probably

evgeni:
> Interesting that the 3.7/devel failure is consistent.

narrator: it wasn't

plugins/module_utils/foreman_helper.py
```diff
@@ -220,25 +216,26 @@ def __init__(self, **kwargs):
...
 
         self.required_plugins = kwargs.pop('required_plugins', [])
 
         super(ForemanAnsibleModule, self).__init__(argument_spec=argument_spec, supports_check_mode=supports_check_mode, **kwargs)
 
-        self._params = self.params.copy()
+        if 'verify_ssl' in self.params:
+            self.warn("Please use 'validate_certs' instead of deprecated 'verify_ssl'.")
+
+        aliases = {alias for arg in argument_spec.values() for alias in arg.get('aliases', [])}
```
evgeni: TIL: you can have multiple `for` in a list comprehension 🤯

mdellweg: The order of the expressions is unintuitive though.

plugins/module_utils/foreman_helper.py
```diff
@@ -471,6 +465,77 @@ def find_puppetclass(self, name, environment=None, params=None, failsafe=False,
...
         return [self.find_puppetclass(name, **kwargs) for name in names]
 
+    def scope_for(self, key):
+        return {'{0}_id'.format(key): self.lookup_entity(key)['id']}
+
+    def set_entity(self, key, entity):
+        self.foreman_params[key] = entity
+        self.foreman_spec[key]['resolved'] = True
+
+    def lookup_entity(self, key):
```
```suggestion
    def lookup_entity(self, key='entity'):
```

I think we can default to looking up the "main" entity?

plugins/module_utils/foreman_helper.py
```diff
@@ -471,6 +465,77 @@ def find_puppetclass(self, name, environment=None, params=None, failsafe=False,
...
+                        failsafe=failsafe, thin=thin, params=params,
+                    )
+                else:
+                    result = [self.find_resource_by(
+                        resource=resource_type,
+                        value=value,
+                        search_field=entity_spec.get('search_by', ENTITY_KEYS.get(resource_type, 'name')),
+                        search_operator=entity_spec.get('search_operator', '='),
+                        failsafe=failsafe, thin=thin, params=params,
+                    ) for value in self.foreman_params[key]]
```
evgeni: I find the "handle entity vs entity_list" in this block very repetitive, but don't have a nice solution for it just now. Maybe something we can improve in the future, tho.

plugins/module_utils/foreman_helper.py
```diff
@@ -1060,9 +1022,8 @@ def _foreman_spec_helper(spec):
     #
     # So in conclusion, we only have to verify that apypie is available before using it.
     # Lazy evaluation helps there.
-    inflector = HAS_APYPIE and apypie.Inflector()
     for key, value in spec.items():
-        foreman_value = {k: v for (k, v) in value.items() if k in ['type', 'ensure']}
+        foreman_value = {k: v for (k, v) in value.items() if k in ['ensure', 'type']}
```
evgeni: Should `['ensure', 'type']` be in some `_STATIC`?

plugins/module_utils/foreman_helper.py
```diff
@@ -1072,14 +1033,14 @@ def _foreman_spec_helper(spec):
             if not flat_name:
                 flat_name = '{0}_id'.format(key)
             foreman_value['resource_type'] = value.get('resource_type', HAS_APYPIE and inflector.pluralize(key))
-            foreman_value.update({k: v for (k, v) in value.items() if k in ['thin', 'scope', 'search_by', 'search_operator', 'resolve']})
+            foreman_value.update({k: v for (k, v) in value.items() if k in ['failsafe', 'resolve', 'scope', 'search_by', 'search_operator', 'thin']})
```
evgeni: `['failsafe', 'resolve', 'scope', 'search_by', 'search_operator', 'thin']` should be in a `_STATIC` ;)

plugins/module_utils/foreman_helper.py
```diff
@@ -1060,9 +1022,8 @@ def _foreman_spec_helper(spec):
     #
     # So in conclusion, we only have to verify that apypie is available before using it.
     # Lazy evaluation helps there.
```
evgeni: The whole comment is not to be here anymore, as we initialiaze the Inflector somewhere else, and don't need the `HAS_APYPIE` hack there.

plugins/module_utils/foreman_helper.py
mdellweg: even if at this level of module inheritance 'entity' isn't even a thing?

plugins/module_utils/foreman_helper.py
mdellweg: Maybe by embedding the `if resource_type == 'operatingsystems':` into `find_resource_by`.
But i think the operatingsystem module needs an unenhanced version of the latter.

plugins/module_utils/foreman_helper.py
evgeni: :woman_shrugging: 

We can also override it in `ForemanEntityAnsibleModule`, it just felt weird to read `lookup_entity('entity')` so often in the diff. But you can also just call me crazy and I'll shut up :)

plugins/module_utils/foreman_helper.py
mdellweg: Actually it is only initialized, if `HAS_APYPIE`, so the comment is still valid.
The improvement is that it is only initialized at most once in a central place.

plugins/module_utils/foreman_helper.py
mdellweg: It is now.

Issue #149, opened 2018-07-17
smart-proxy - module doesn't exist


thomasmckay:
No smart proxy module

evgeni:
this will be fixed by #314

cmeissner:
@evgeni The attached [PR#314](https://github.com/theforeman/foreman-ansible-modules/pull/314) is closed. Are there any ambitions to develop a smart_proxy module?

evgeni:
Do we want a module? Yes. Do we have bandwidth to write one (based on the existing PR) not sure

cmeissner:
I'll not promise much but I try to have a look for the code too. Is there a way to get help, if I decide to do some work on it? It's because I never do such development with apypie and you sophisticated code.

evgeni:
Easiest? Pop up on IRC/Matrix (#theforeman-dev on freenode) during Europe day and we can chat (I'm Zhenech there).

mdellweg:
ping me as x9c4 on #theforeman-dev.

mdellweg:
@cmeissner Sorry, but you reminding me that this good work is lying around and me having a lonely train ride made me give it a spin. It felt like back in the times when we refactored all those modules over and over...
Expect a pr soon. If you still want to help, i encourage you to test it.

cmeissner:
@mdellweg oh great to hear. It could have been the start of direct collaboration, but of course I am very happy to test it. Give me a ping, if I can start testing, maybe Monday morning.

mdellweg:
@cmeissner It is ready to test.
Also there is #869 which might be an opportunity for contribution.

cmeissner:
@mdellweg works fine. Thank you. Waiting for publishing the new version.

PR #753, opened 2020-03-25
[WIP] Add foreman_anislbe_roles module

branic:
This PR adds a foreman_ansible_roles module.

This module will allow working with Foreman Ansible Roles.

Marking this PR as WIP, but I wanted to get it open to get feedback on the current (although limited) functionality.


evgeni:
Is state=fetch used to list the available roles on a proxy?

If so, I'd prefer if we could have one module for listing, and one for performing actions, as that's what current ansible best practice is.

branic:
> Is state=fetch used to list the available roles on a proxy?
> 
> If so, I'd prefer if we could have one module for listing, and one for performing actions, as that's what current ansible best practice is.

@evgeni The fetch action lists the Ansible Roles that have been placed in `/etc/ansible/roles` and are available to be imported.

I don't mind making that action a separate module if that's the desired direction. I think the `foreman_search_facts` module will be able to be used for listing roles that have already been imported.

Would you like fetch to be its own module?

plugins/modules/foreman_ansible_roles.py
```diff
@@ -0,0 +1,100 @@
...
+class ForemanAnsibleRolesModule(ForemanEntityAnsibleModule):
+    pass
+
+
+def main():
+
+    module = ForemanAnsibleRolesModule(
+        foreman_spec = dict(
+            state = dict(default='fetch', choices=['fetch']),
+            proxy_id = dict(),
```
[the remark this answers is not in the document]
branic: Thanks for the suggestion @mdellweg . I'll make the change.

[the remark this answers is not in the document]
branic: @mdellweg When implementing either one of the suggestions I'm getting the below error:

```
{"msg": "Unsupported parameters for (basic.py) module: proxy_id Supported parameters include: password, proxy, server_url, state, username, validate_certs", "failed": true, "invocation": {"module_args": {"username": "admin", "password": "VALUE_SPECIFIED_IN_NO_LOG_PARAMETER", "server_url": "https://satellite.example.com", "validate_certs": "false", "state": "fetch", "proxy_id": "1"}}}
SystemExit: 1
```

The only difference in the error message between the two suggestions is that `smart_proxy` or `proxy` is listed in the `Supported parameters` list in the error message. The proxy_id is appearing in the `module_args` as expected.

Any ideas on how to fix the error are appreciated.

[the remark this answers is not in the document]
mdellweg: Well i should have explained better. In order to be able to write playbooks that cannot know any foreman internal database id's, our modules describe all entities (including smart_proxies) by their name. So with either of the suggestions, it is expected that the module only accepts the proxy in the (smart_)proxy field, and you should specify it's fqdn there. That is why the module than (in turn) needs to do some lookup to figure out its id at runtime. Have a look at (i.e.) the domain module.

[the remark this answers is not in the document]
branic: It looked to me like using `entity` type syntax you provided would automatically resolve to the ID, but had to use `find_resource_by_name` to find the id that needed to be provided to the `fetch` API call.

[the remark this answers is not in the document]
evgeni: The automatic resolution only happens if you call `module.run()` or `_entity, module_params = module.resolve_entities(module_params)`. The later is probably what you're looking for here.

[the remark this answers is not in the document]
branic: Thanks @evgeni. I've modified the code to call `resolve_entities`.

branic:
I've renamed the module to `foreman_anisble_roles_facts` and implemented not only `fetch` but also `list` and `show`.

I'll limit this PR to just the fact gathering module and open another PR to implement the `import`, `delete`, and `obsolete`.

I'll start working on adding test cases.

branic:
@evgeni What is the recommended way to get files on the filesystem when running tests? 

I need to place a test role in `/etc/ansible/roles` so that fetch will have data to report. I'll need this in the future for testing import functionality as well.

branic:
@evgeni One other question. During the test run I'm creating the Organization and Location, but I need to assign the location to the smart proxy. I'm not finding any functionality to do that. Is there functionality to do that?

evgeni:
> @evgeni One other question. During the test run I'm creating the Organization and Location, but I need to assign the location to the smart proxy. I'm not finding any functionality to do that. Is there functionality to do that?

Nope, sorry. We have https://github.com/theforeman/foreman-ansible-modules/issues/149 open, and I think there even was a PR for it, but never got merged :(

evgeni:
> @evgeni What is the recommended way to get files on the filesystem when running tests?
> 
> I need to place a test role in `/etc/ansible/roles` so that fetch will have data to report. I'll need this in the future for testing import functionality as well.

Sadly not. We have the same issue with Puppet modules too. In the past, I've been using the following snippet
https://github.com/theforeman/foreman-ansible-modules/blob/41d0a7aa3f35bff406af9e90e3cf68b08f21ba01/tests/test_playbooks/host.yml#L2-L10

and adding a `foreman` host to the inventory that is my real Foreman installation, not nice, but at least documented. you can also do it by hand and write it in the header of the test playbook:

https://github.com/theforeman/foreman-ansible-modules/blob/41d0a7aa3f35bff406af9e90e3cf68b08f21ba01/tests/test_playbooks/hostgroup.yml#L2-L8

plugins/modules/foreman_ansible_roles_facts.py
```diff
@@ -0,0 +1,186 @@
...
+
+
+def main():
+
+    module = ForemanAnsibleRolesFactsModule(
+        foreman_spec = dict(
+            action = dict(default='fetch', choices=['fetch', 'list', 'show']),
+            proxy = dict(type='entity', flat_name='proxy_id', aliases=['smart_proxy'], resource_type='smart_proxies'),
+            search=dict(default=""),
+            full_details=dict(type='bool', aliases=['info'], default='false'),
```
evgeni, line 134:
```suggestion
            full_details=dict(type='bool', aliases=['info'], default=False),
```

plugins/modules/foreman_ansible_roles_facts.py
```diff
@@ -0,0 +1,186 @@
...
+            organization=dict(type='entity', flat_name='organization_id', resource_type='organizations'),
+            location=dict(type='entity', flat_name='location_id', resource_type='locations'),
```
evgeni, line 136:
```suggestion
            organization=dict(type='entity'),
            location=dict(type='entity'),
```

our helper should autodetect these

plugins/modules/foreman_ansible_roles_facts.py
```diff
@@ -0,0 +1,186 @@
...
+class ForemanAnsibleRolesFactsModule(ForemanEntityAnsibleModule):
+    pass
+
+
+def main():
+
+    module = ForemanAnsibleRolesFactsModule(
+        foreman_spec = dict(
+            action = dict(default='fetch', choices=['fetch', 'list', 'show']),
+            proxy = dict(type='entity', flat_name='proxy_id', aliases=['smart_proxy'], resource_type='smart_proxies'),
```
evgeni, line 132:
```suggestion
            action=dict(default='fetch', choices=['fetch', 'list', 'show']),
            proxy=dict(type='entity', flat_name='proxy_id', aliases=['smart_proxy'], resource_type='smart_proxies'),
```

the linter is picky ;)

plugins/modules/foreman_ansible_roles_facts.py
```diff
@@ -0,0 +1,186 @@
...
+    resource = 'ansible_roles'
+    search = module_params['search']
+    params = {}
+
+    if module_params['action'] == 'show':
+        module_params['action'] = 'list'
+        module_params['full_details'] = True
+
+    with module.api_connection():
+        _entity, module_params = module.resolve_entities(module_params, 'ansible_roles')
```
evgeni, line 154:
```suggestion
        module.auto_lookup_entities()
        params = _flatten_entity(module.foreman_params, module.foreman_spec)
```

I'm afraid we managed to leave this PR around for too long and merged a huge refactoring since then. I'll provide suggestions how to update.

plugins/modules/foreman_ansible_roles_facts.py
```diff
@@ -0,0 +1,186 @@
...
+
+        if 'proxy' in module_params:
+            params['proxy_id'] = module_params['proxy']['id']
+
+        if 'organization' in module_params:
+            params['organization_id'] = module_params['organization']['id']
+
+        if 'location' in module_params:
+            params['location_id'] = module_params['location']['id']
```
evgeni, line 163:
```suggestion
```

plugins/modules/foreman_ansible_roles_facts.py
```diff
@@ -0,0 +1,186 @@
...
+            response = module.list_resource(resource, search, params)
+
+            if module_params['full_details']:
+                resources = []
+                for found_resource in response:
+                    resources.append(module.show_resource(resource, found_resource['id'], params))
+            else:
+                resources = response
+
+            module.exit_json(ansible_roles=resources)
```
evgeni, line 183:
```suggestion
        if module.foreman_params['action'] == 'fetch':
            resources = module.fetch_resource('ansible_roles', params)
            resources = resources['results']['ansible_roles']
        elif module.foreman_params['action'] == 'list':
            response = module.list_resource(resource, module.foreman_params.get('search'), params)

            if module.foreman_params['full_details']:
                resources = []
                for found_resource in response:
                    resources.append(module.show_resource(resource, found_resource['id'], params))
            else:
                resources = response

        module.exit_json(ansible_roles=resources)
```

plugins/modules/foreman_ansible_roles_facts.py
```diff
@@ -0,0 +1,186 @@
...
+
+        required_if = [
+            ['action', 'fetch', ['proxy']]
+        ],
+    )
+
+    module_params = module.clean_params()
+    resource = 'ansible_roles'
+    search = module_params['search']
+    params = {}
```
evgeni, line 147:
```suggestion
    resource = 'ansible_roles'
```

plugins/modules/foreman_ansible_roles_facts.py
```diff
@@ -0,0 +1,186 @@
...
+  description: Roles that are able to be imported on the proxy
+  returned: state = fetched
+  type: list
+'''
+
+from ansible.module_utils.foreman_helper import ForemanEntityAnsibleModule
+
+
+class ForemanAnsibleRolesFactsModule(ForemanEntityAnsibleModule):
+    pass
```
evgeni, line 124:
```suggestion
from ansible.module_utils.foreman_helper import ForemanAnsibleModule, _flatten_entity


class ForemanAnsibleRolesFactsModule(ForemanAnsibleModule):
    pass
```

plugins/modules/foreman_ansible_roles_facts.py
```diff
@@ -0,0 +1,186 @@
...
+    )
+
+    module_params = module.clean_params()
+    resource = 'ansible_roles'
+    search = module_params['search']
+    params = {}
+
+    if module_params['action'] == 'show':
+        module_params['action'] = 'list'
+        module_params['full_details'] = True
```
evgeni, line 151:
```suggestion
    if module.foreman_params['action'] == 'show':
        module.foreman_params['action'] = 'list'
        module.foreman_params['full_details'] = True
```

plugins/modules/foreman_ansible_roles_facts.py
branic: Fixed.

plugins/modules/foreman_ansible_roles_facts.py
branic: Fixed.

plugins/modules/foreman_ansible_roles_facts.py
branic: Fixed.

plugins/modules/foreman_ansible_roles_facts.py
branic: Fixed.

plugins/modules/foreman_ansible_roles_facts.py
branic: Fixed.

plugins/modules/foreman_ansible_roles_facts.py
branic: Fixed.

plugins/modules/foreman_ansible_roles_facts.py
branic: Fixed.

plugins/modules/foreman_ansible_roles_facts.py
branic: Fixed.

plugins/modules/foreman_ansible_roles_facts.py
branic: Fixed.

plugins/modules/foreman_ansible_roles_facts.py
```diff
@@ -0,0 +1,180 @@
...
+        params = _flatten_entity(module.foreman_params, module.foreman_spec)
+
+        if module.foreman_params['action'] == 'fetch':
+            resources = module.fetch_resource(resource, params)
+            resources = resources['results']['ansible_roles']
+
+        elif module.foreman_params['action'] == 'list':
+            # According to the APIDoc organization_id and location_id are valid
+            # parameters, but a "500 Internal Server Error" error is returned
+            # when an organization_id or location_id are present in the request
```
evgeni, line 161: The error raised is "Association not found for organization", and I've seen that before, just can't remember where.

evgeni: Ah, here: https://projects.theforeman.org/issues/25277

Will need to talk to @ares about that

plugins/modules/foreman_ansible_roles_facts.py
```diff
@@ -0,0 +1,180 @@
...
+                    'supported_by': 'community'}
+
+DOCUMENTATION = '''
+---
+module: foreman_ansible_roles_facts
+short_description: Gather Ansible Roles facts
+description:
+  - Gather facts about Foreman Ansible Roles
+author:
+  - "Brant Evans (@branic)
```
evgeni, line 33:
```suggestion
  - "Brant Evans (@branic)"
```

plugins/modules/foreman_ansible_roles_facts.py
```diff
@@ -0,0 +1,180 @@
...
+
+class ForemanAnsibleRolesFactsModule(ForemanAnsibleModule):
+    pass
+
+
+def main():
+
+    module = ForemanAnsibleRolesFactsModule(
+        foreman_spec = dict(
+            action=dict(default='fetch', choices=['fetch', 'list', 'show']),
```
evgeni, line 131: this contradicts with the documented default above ;)

plugins/modules/foreman_ansible_roles_facts.py
evgeni: I've opened https://projects.theforeman.org/issues/29583 -- could you please refer to that in the comment?

plugins/modules/foreman_ansible_roles_facts.py
```diff
@@ -0,0 +1,180 @@
...
+    default: list
+    choices:
+      - list
+      - show
+      - fetch
+    required: false
+    type: str
+  proxy:
+    description:
+      - The Smart Porxy name to use for fetching Ansible Roles
```
evgeni, line 47:
```suggestion
      - The Smart Proxy name to use for fetching Ansible Roles
```

plugins/modules/foreman_ansible_roles_facts.py
```diff
@@ -0,0 +1,180 @@
...
+    required: false
+    type: str
+    alias: smart_proxy
```
evgeni, line 50:
```suggestion
    aliases:
      - smart_proxy
```

plugins/modules/foreman_ansible_roles_facts.py
```diff
@@ -0,0 +1,180 @@
...
+
+
+class ForemanAnsibleRolesFactsModule(ForemanAnsibleModule):
+    pass
+
+
+def main():
+
+    module = ForemanAnsibleRolesFactsModule(
+        foreman_spec = dict(
```
evgeni, line 130:
```suggestion
        foreman_spec=dict(
```

plugins/modules/foreman_ansible_roles_facts.py
```diff
@@ -0,0 +1,180 @@
...
+            if module.foreman_params['full_details']:
+                resources = []
+                for found_resource in response:
+                    resources.append(module.show_resource(resource, found_resource['id'], params))
+            else:
+                resources = response
+
+        module.exit_json(ansible_roles=resources)
+
+if __name__ == '__main__':
```
evgeni, line 179:
```suggestion

if __name__ == '__main__':
```

plugins/modules/foreman_ansible_roles_facts.py
```diff
@@ -0,0 +1,180 @@
...
+    main()
```
evgeni, line 180:
```suggestion
    main()

```

mdellweg: Oh, let's hope, this only introduces a single line feed at the end...

evgeni:
Thanks @branic! And once more, sorry for keeping you busy by catching up on changes we made to our helpers.

I've tried your module, and it works as advertised. However, I am a bit confused about the exposed workflow(s).

1. The `list`/`show` actions effectively duplicate the functionality of the `foreman_search_facts` module. I do agree that "one module per resource" is easier for the users (and would even go so far to say we should implement that for all our resources), but at the same time it feels weird to have it in the same module as `fetch` which is about a completely different dataset.
2. The `fetch` action is the really interesting feature here, as it allows us to find modules that are available for import.
3. Having an `action` parameter in a "facts" module is rather confusing, as facts are informational not actionable and IMHO exposing the API action name to the user is too much.

So if I'd be king (I am not, lucky you), I'd name this module `foreman_ansible_importable_roles_info`¹ (or `…available_roles`? or switch roles and the term? not sure) and only make it implement the `fetch` action (thus dropping the parameter altogether). And then I'd add a second module `foreman_ansible_installed_roles` (or whatever) that will again has no `action` param, and will do either list or show, depending on `full_details` or not.

¹: there has been a huge movement in Ansible to move away from the "facts" term to "info" - see https://docs.ansible.com/ansible/latest/modules/list_of_all_modules.html how all _facts modules are marked as deprecated in favor of their _info siblings

evgeni:
Oh, and I didn't see any difference between list and show in my tests. What extra info should show expose?

plugins/modules/foreman_ansible_roles_facts.py
evgeni: Oh yeah. Newline at the end of the file is all that's needed here.

branic:
> Oh, and I didn't see any difference between list and show in my tests. What extra info should show expose?

I haven't seen any difference in list and show in my tests either. The API has both a list and a show endpoint which was why I implemented both.

evgeni:
Yeah. I think list and show are just default methods of the api and thus exist here. Show is cheaper if you only need to see one item. But in our specific case we only need the list IMHO.

branic:
For the workflow, I've been conflicted as well. The list/show functionality does overlap with the `foreman_search_facts` module. I implemented them in with the fetch as I was looking at the module as returning available roles that either could be imported or that already were.

I agree it does make sense to split the fetch and list/show into their own modules. As for having a dedicated list/show module is that a direction you would like the project to go or should users just rely on the `foreman_search_facts` to gather that information?

evgeni:
Long term I'd love to see us having separate "info" modules for the various resources, but it's far from a priority as long foreman_search_facts covers the same functionality with a slightly more complicated interface.

branic:
Great. I'll work on splitting the module into two modules.

Thanks for all the great feedback.

plugins/modules/foreman_ansible_roles_facts.py
branic: Fixed.

plugins/modules/foreman_ansible_roles_facts.py
branic: Fixed.

plugins/modules/foreman_ansible_roles_facts.py
branic: Fixed.

plugins/modules/foreman_ansible_roles_facts.py
branic: Fixed.

plugins/modules/foreman_ansible_roles_facts.py
branic: Fixed.

plugins/modules/foreman_ansible_roles_facts.py
branic: Fixed.

plugins/modules/foreman_ansible_roles_facts.py
branic: Fixed.

plugins/modules/foreman_ansible_roles_facts.py
branic: Fixed.

plugins/modules/foreman_ansible_roles_importable_info.py
```diff
@@ -0,0 +1,106 @@
...
+  - Gather names of Ansible Roles that can be imported
+author:
+  - "Brant Evans (@branic)"
+options:
+  proxy:
+    description:
+      - The Smart Proxy name to use for fetching Ansible Roles
+    required: false
+    type: str
+    aliases:
```
evgeni, line 40:
```suggestion
    aliases:
```

(the linter is unhappy)

plugins/modules/foreman_ansible_roles_importable_info.py
```diff
@@ -0,0 +1,106 @@
...
+short_description: Gather names of Ansible Roles that can be imported
+description:
+  - Gather names of Ansible Roles that can be imported
+author:
+  - "Brant Evans (@branic)"
+options:
+  proxy:
+    description:
+      - The Smart Proxy name to use for fetching Ansible Roles
+    required: false
```
evgeni, line 38:
```suggestion
    required: true
```

plugins/modules/foreman_ansible_roles_importable_info.py
```diff
@@ -0,0 +1,106 @@
...
+__metaclass__ = type
+
+
+ANSIBLE_METADATA = {'metadata_version': '1.1',
+                    'status': ['preview'],
+                    'supported_by': 'community'}
+
+DOCUMENTATION = '''
+---
+module: foreman_ansible_roles_facts
```
evgeni, line 28:
```suggestion
module: foreman_ansible_roles_importable_info
```

plugins/modules/foreman_ansible_roles_importable_info.py
```diff
@@ -0,0 +1,106 @@
...
+    pass
+
+
+def main():
+
+    module = ForemanAnsibleRolesImportableInfoModule(
+        foreman_spec=dict(
+            proxy=dict(type='entity', flat_name='proxy_id', aliases=['smart_proxy'], resource_type='smart_proxies', required=True),
+            organization=dict(type='entity'),
+            location=dict(type='entity'),
```
evgeni, line 89: are those two actually used for anything? I know you can pass them in the request, but I don't see any use for doing so. the data on disk (which this action lists) is not org/loc aware.

I *think* the API just allows to pass those, as it allows this on every request.

evgeni: @ares is that right? 'fetch' should not be scoped?

plugins/modules/foreman_ansible_roles_imported_info.py
```diff
@@ -0,0 +1,137 @@
...
+
+class ForemanAnsibleRolesImportedInfoModule(ForemanAnsibleModule):
+    pass
+
+
+def main():
+
+    module = ForemanAnsibleRolesImportedInfoModule(
+        foreman_spec=dict(
+            search=dict(default=""),
```
evgeni, line 101:
```suggestion
            search=dict(),
```

no need for a default

plugins/modules/foreman_ansible_roles_imported_info.py
```diff
@@ -0,0 +1,137 @@
...
+            full_details=dict(type='bool', aliases=['info'], default=False),
```
evgeni, line 102: given the API does not return more info, shall we drop that param?

plugins/modules/foreman_ansible_roles_imported_info.py
```diff
@@ -0,0 +1,137 @@
...
+__metaclass__ = type
+
+
+ANSIBLE_METADATA = {'metadata_version': '1.1',
+                    'status': ['preview'],
+                    'supported_by': 'community'}
+
+DOCUMENTATION = '''
+---
+module: foreman_ansible_roles_facts
```
evgeni, line 28:
```suggestion
module: foreman_ansible_roles_imported_info
```

Issue #758, opened 2020-03-26
`foreman_host` is missing provisioning related attributes


evgeni:
##### SUMMARY
The [`hosts` api](https://theforeman.org/api/1.24/apidoc/v2/hosts/create.html) allows passing `provision_method`, `image` and `capabilities`. Our module does not support that.

While I don't understand the `capabilities` one, `p_m` and `image` are required for image based provisioning. (`image` is needed when `p_m=image` ;)).

##### ISSUE TYPE
 - Feature Idea

PR #759, opened 2020-03-26
add missing provisioning params to foreman_host

evgeni:
Fixes: #758

### DIFF
--- a/plugins/modules/foreman_host.py
+++ b/plugins/modules/foreman_host.py
@@ -98,6 +98,21 @@ options:
       - Mutually excluside with I(owner).
     type: str
     required: false
+  provision_method:
+    description:
+      - The method used to provision the host.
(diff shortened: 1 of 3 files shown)


evgeni:
@hlawatschek as I think you're using image based provisioning, mind taking a look here?

evgeni:
after reading foreman source code, I don't think capabilities should be writeable and thus dropped it for now

evgeni:
and I opened https://projects.theforeman.org/issues/29420 to fix that in foreman :)

plugins/modules/foreman_host.py
```diff
@@ -98,6 +98,21 @@
       - Mutually excluside with I(owner).
     type: str
     required: false
+  provision_method:
```
mdellweg: And this is not available on host_groups?

evgeni: Correct

Issue #749, opened 2020-03-20
Missing module for tailoring files


ogajduse:
##### SUMMARY

FAM currently does not provide a module to manage tailoring files. (`/api/compliance/tailoring_files`)
##### ISSUE TYPE
 - Feature Idea

PR #761, opened 2020-03-26
 add tailoring_file module

evgeni:
Fixes: #749

### DIFF
--- a/plugins/doc_fragments/foreman.py
+++ b/plugins/doc_fragments/foreman.py
@@ -276,3 +276,19 @@ options:
     required: true
     type: str
 '''
+
+    SCAP_DATASTREAM = '''
+options:
+  scap_file:
+    description:
+      - File containing XML DataStream content.
+      - Required when creating a new DataStream.
+    required: false
+    type: path
+  original_filename:
+    description:
+      - Original file name of the XML file.
+      - If unset, the filename of I(scap_file) will be used.
+    required: false
+    type: str
+'''
--- a/plugins/module_utils/foreman_helper.py
+++ b/plugins/module_utils/foreman_helper.py
@@ -5,7 +5,9 @@ from __future__ import absolute_import, division, print_function
 __metaclass__ = type
 
 
+import hashlib
 import json
+import os
 import re
 import time
 import traceback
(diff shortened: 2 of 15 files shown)


evgeni:
no tests, no docs, but that's what I meant by inheritance

cc @ogajduse

evgeni:
I wonder if this should be `foreman_scap_tailoring_file` instead. Or maybe even both modules `foreman_openscap_…`? We've been using the resource name so far as a guideline, but outside of Foreman and Katello, things get wild.

evgeni:
That said, this is ready for review.

@jameerpathan111 thoughts on this?

jameerpathan111:
> I wonder if this should be `foreman_scap_tailoring_file` instead. Or maybe even both modules `foreman_openscap_…`? 

`foreman_scap_tailoring_file` seems good to me.

plugins/module_utils/foreman_helper.py
```diff
@@ -1017,6 +1019,39 @@ def __init__(self, **kwargs):
...
+class ForemanDataStreamModule(ForemanTaxonomicEntityAnsibleModule):
+    def __init__(self, **kwargs):
+        foreman_spec = dict(
+            original_filename=dict(type='str'),
+            scap_file=dict(type='path'),
+        )
+        foreman_spec.update(kwargs.pop('foreman_spec', {}))
+        super(ForemanDataStreamModule, self).__init__(foreman_spec=foreman_spec, **kwargs)
+
+    def datastream_run(self):
```
mdellweg: Why don't you override the actual 'run' method?

evgeni: because I'm a lazy bum and hate Python's `super`.

plugins/module_utils/foreman_helper.py
```diff
@@ -1028,7 +1028,7 @@ def __init__(self, **kwargs):
         foreman_spec.update(kwargs.pop('foreman_spec', {}))
         super(ForemanDataStreamModule, self).__init__(foreman_spec=foreman_spec, **kwargs)
 
-    def datastream_run(self):
+    def run(self):
```
mdellweg: I know lazy n stuff...

But should it not support the same interface as the old one?

evgeni: I don't see a benefit implementing this at this point, tbh :)

mdellweg: Good for you, that the 'new' run does not accept parameters anyway.

evgeni:
I'll merge this *after* #747 and handle the needed rebasing.

evgeni:
FTR, the rebased version of this ontop of #747 can be found in https://github.com/evgeni/foreman-ansible-modules/tree/tailoring_entity, will replace this PR with it as soon as 747 lands.

plugins/doc_fragments/foreman.py
```diff
@@ -276,3 +276,19 @@ class ModuleDocFragment(object):
     required: true
     type: str
 '''
+
+    DATASTREAM = '''
```
mdellweg: Should this be named `SCAP_DATASTREAM` as it is rather specific?

evgeni: yes

evgeni: I've also renamed the helper class to match

evgeni:
also squashed the commits to make it nicer :)
merging on green ci

PR #765, opened 2020-03-28
Search facts documentation

### LINKED ISSUE #762: Ability to return a list of items
##### SUMMARY
With FAM it is currently possible to create and delete items such as domains, subnets, hosts etc but I haven't found a way to return a list of the current domains registered in a foreman instance for example.

The use case I'm looking at is maintaining an up to date foreman instance based on information from an ansible inventory (the inventory data is from a custom plugin that queries an asset register)..if any of the domains for hosts in the inventory are no longer required, how do I know which to delete without comparing with the existing list?
If deleting a domain, the domain would need to be unassigned from the subnet first, so I'd need to find out which subnets have the domain association and remove those before deleting the domain.

This may already be possible but I haven't found anything in the documentation to suggest it

##### ISSUE TYPE
 - Feature Idea

pmgration:
Update examples for foreman_search_facts module to:

- Fix `with_items` variable format
- Highlight which examples relate specifically to Katello facts
- Changes first example to search for a fact that doesn't default to potentially null value (http_proxy) in a fresh foreman installation.

Fixes #762

### DIFF
--- a/plugins/modules/foreman_search_facts.py
+++ b/plugins/modules/foreman_search_facts.py
@@ -86,7 +86,7 @@ EXAMPLES = '''
   register: result
 - debug:
     var: item.name
-  with_items: result.resources
+  with_items: "{{ result.resources }}"
 
 - name: "Read all Organizations with full details"
   foreman_search_facts:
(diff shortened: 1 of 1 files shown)


evgeni:
test errors  unrelated

PR #771, opened 2020-03-31
build docs via gh actions into gh pages

evgeni:


.github/workflows/docs.yml
```diff
@@ -0,0 +1,41 @@
...
+      - name: Build docs
+        run: make doc
+      - name: commit docs
+        run: |
+          git checkout gh-pages
+          rm -rf $(basename ${GITHUB_REF})
+          mv docs/_build/html $(basename ${GITHUB_REF})
+          dirname */index.html | sort --version-sort | xargs -I@@ -n1 echo '<a href="@@/">@@</a><br/>' > index.html
+          git add $(basename ${GITHUB_REF}) index.html
+          git commit -a -m "update docs"
```
ehelms: Is it possible to include the tag in the commit message?

evgeni: sure

PR #774, opened 2020-03-31
generate nicer html by including a header and a footer

evgeni:


evgeni:
@ehelms mind having another look, now I don't need stuff in #773 anymore

PR #781, opened 2020-04-02
Update documentation for operatingsystem and os_default_template modules

pmgration:
- Updates operatingsystem creation examples as they can't contain spaces.
- Description of operatingsystem attribute for os_default_template which can be an operating system `name` or `title` (title being name and major version). Using just the name when multiple operating systems exist with the same name (but different versions) results in an error when trying to create a template association.

### DIFF
--- a/plugins/modules/foreman_os_default_template.py
+++ b/plugins/modules/foreman_os_default_template.py
@@ -34,7 +34,7 @@ author:
 options:
   operatingsystem:
     description:
-      - Title of the Operating System (name, or name and major version, that uniquely identify the OS)
+      - Title of the Operating System (name, or name and major version, that uniquely identifies the OS)
     required: true
     type: str
   template_kind:
--- a/plugins/modules/foreman_os_default_template.py
+++ b/plugins/modules/foreman_os_default_template.py
@@ -34,7 +34,7 @@ author:
 options:
   operatingsystem:
     description:
-      - Title of the Operating System (name or name and major version)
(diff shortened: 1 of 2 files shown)


plugins/modules/foreman_os_default_template.py
```diff
@@ -34,7 +34,7 @@
 options:
   operatingsystem:
     description:
-      - Name of the Operating System
+      - Title of the Operating System (name or name and major version)
```
mdellweg, line 37: Can you add that this information must be able to uniquely identify the os? That would be great.

pmgration:
@mdellweg @evgeni
Going one level deeper on this issue, it doesn't seem possible to target a specific operating system if multiple operating systems exist with the same name and major version, but different minor versions e.g.

CentOS 7.7.1908
CentOS 7.8.2003

Is there something I'm missing here? Should we add the ability to target the minor version too as part of the OS title or is the assumption that you should only have one major version of an OS?

evgeni:
Can you open anew issue please. Otherwise this will be lost :(

I fear this is an API limitation and one needs to search differently for minor versions, but would need to confirm later at a computer (and not my phone ;))

pmgration:
I just put together an example playbook to create a new issue but found I had it working by including the minor version in the title for the `operatingsystem` attribute of `foreman_os_default_template` module tasks for example, I'm sure I tried this yesterday which prompted me to add the comment above but can't reproduce now - if I come across this again I'll put together a new issue though.

Thanks!

PR #790, opened 2020-04-10
Added 'solve_dependencies' to 'content_view'

russianguppie:
Added 'solve_dependencies' to 'content_view'

### DIFF
--- a/tests/test_playbooks/tasks/content_view.yml
+++ b/tests/test_playbooks/tasks/content_view.yml
@@ -15,6 +15,7 @@
     composite: "{{ composite | default(omit) }}"
     components: "{{ components | default(omit) }}"
     auto_publish: "{{ auto_publish | default(omit) }}"
+    solve_dependencies: "{{ solve_dependencies | deafult(omit) }}"
   register: result
 
 - assert:
--- a/tests/test_playbooks/tasks/content_view.yml
+++ b/tests/test_playbooks/tasks/content_view.yml
@@ -15,7 +15,7 @@
     composite: "{{ composite | default(omit) }}"
     components: "{{ components | default(omit) }}"
     auto_publish: "{{ auto_publish | default(omit) }}"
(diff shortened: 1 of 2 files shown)


tests/test_playbooks/tasks/content_view.yml
```diff
@@ -15,6 +15,7 @@
     composite: "{{ composite | default(omit) }}"
     components: "{{ components | default(omit) }}"
     auto_publish: "{{ auto_publish | default(omit) }}"
+    solve_dependencies: "{{ solve_dependencies | deafult(omit) }}"
```
mdellweg, line 18:
```suggestion
    solve_dependencies: "{{ solve_dependencies | default(omit) }}"
```

russianguppie:
Could someone point me in the right direction towards resolving these errors? The most useful message I could find was:
`assertion failure :\n{'name': 'Test Content View', 'composite': False, 'repository_ids': [25], 'auto_publish': False, 'solve_dependencies': False} != {'composite': False, 'repository_ids': [25], 'auto_publish': False, 'name': 'Test Content View'}\n"'
`

mdellweg:
Sure. Please read https://github.com/theforeman/foreman-ansible-modules/blob/master/docs/testing.md

evgeni:
In this particular case, I would suggest dropping the `default=False`. Then the request won't include the parameter unless the user explicitly set it, thus falling back to the server default. And thus will lead to passing tests, as they won't set it (as of now) and the request will match the recorded one.

mdellweg:
I absolutely agree on dropping the `default=false` (not only for the tests, but for functionality).
And i'd say it's optional, whether you want to explicitly include  the new parameter in the tests.

russianguppie:
Thank you for the suggestion!

evgeni:
This looks good to me! And I agree with @mdellweg, no tests required for this. Unless you want, of course ;)

Issue #791, opened 2020-04-14
Links to modules broken


Fobhep:
##### SUMMARY
The links to modules in the devel guide
https://github.com/theforeman/foreman-ansible-modules/blob/master/docs/developing.md
are broken.

I think there is always a `/docs/` layer too much included.

##### ISSUE TYPE
 - Bug Report

PR #794, opened 2020-04-15
fix doc links after moving developing.md into the docs folder

evgeni:
Fixes: #791


evgeni:
This makes it still wrongly linked on https://theforeman.github.io/foreman-ansible-modules/, so I wonder if we should use absolute links to github here?

evgeni:
absolute links result in documentation referring the "wrong" branch, so I'm going to merge this, as it's an improvement, and see how we go and fix the sphinx result too.

PR #796, opened 2020-04-20
fix notes rendering in templates_import module

evgeni:
with the | it just renders the first line as a note

### DIFF
--- a/plugins/modules/foreman_templates_import.py
+++ b/plugins/modules/foreman_templates_import.py
@@ -33,8 +33,7 @@ description:
 author:
   - "Anton Nesterov (@nesanton)"
 notes:
-  - |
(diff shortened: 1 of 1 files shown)


evgeni:
@nesanton did I mention I hate YAML? :)

evgeni:
before:
Screenshot from 2020-04-20 14-32-02

after:
Screenshot from 2020-04-20 14-31-53

nesanton:
oh I'm sorry. I felt like making it one line (per your initial suggestion) then found that it hurts my eyes, then grep -r 'notes:'  to see if any other module has such issues, and after finding out that I'm alone landed with the |.

TIL, LGTM

evgeni:
no worries, YAML is just that horrible :)

PR #799, opened 2020-04-24
Allow foreman inventory to exclude hosts

lukepafford:
The `exclude_patterns` option will take a list of regular expressions that filter out hosts you don't want to be in your Ansible inventory.

This is useful for certain host objects that aren't actual hosts, such as 'virt-who' objects that are added by integrating with VMware.

### DIFF
--- a/plugins/inventory/foreman.py
+++ b/plugins/inventory/foreman.py
@@ -226,9 +226,9 @@ class InventoryModule(BaseInventoryPlugin, Cacheable, Constructable):
                 # Skip any hosts that match exclude_patterns
                 exclude_patterns = self.get_option('exclude_patterns')
                 if exclude_patterns:
-                    if any([ re.search(exclude_pattern, host['name']) for exclude_pattern in exclude_patterns]):
+                    if any([re.search(exclude_pattern, host['name']) for exclude_pattern in exclude_patterns]):
                         continue
-                        
+                
                 host_name = self.inventory.add_host(host['name'])
 
                 # create directly mapped groups
--- a/plugins/inventory/foreman.py
+++ b/plugins/inventory/foreman.py
@@ -65,7 +65,7 @@ DOCUMENTATION = '''
             - Places hostvars in a dictionary with keys `foreman`, `foreman_facts`, and `foreman_params`
         type: boolean
         default: False
-      exclude_patterns:
+      host_filters:
         description: List of regular expressions that will filter out hosts that match
         type: list
         required: False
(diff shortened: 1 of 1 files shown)


evgeni:
@ares you're more familiar with the inventory, could you please have a look?

sean797:
I feel like this would be better implemented inside the report template this plugin consumes? There should be a small performance improvement and something like excluding virt-who hosts I would suggest should be the default?

https://github.com/theforeman/community-templates/blob/develop/report_templates/ansible_-_ansible_inventory.erb

evgeni:
@sean797 I don't think the inventory uses that report

jainnikhil30:
@lukepafford can we call the parameter as "host_filters" ? That way it matches with script used to have.

https://github.com/ansible-collections/community.general/blob/ce9701f813c3b8d8171ce7272ae15be5dbb7fd37/scripts/inventory/foreman.ini#L133

plugins/inventory/foreman.py
```diff
@@ -215,6 +223,11 @@ def _populate(self):
         for host in self._get_hosts():
 
             if host.get('name'):
+                # Skip any hosts that match exclude_patterns
+                exclude_patterns = self.get_option('exclude_patterns')
```
mdellweg, line 227: Can we move that lookup out of the for loop?

lukepafford:
`exclude_patterns` has been renamed to `host_filters`, and the `host_filters` lookup has been moved to outside of the for loop.

As for whether this logic belongs somewhere else, I can't really say. I originally needed this feature from the perspective of an Ansible end user who wanted to use the `foreman.py` inventory plugin, and needed to filter out host results.

plugins/inventory/foreman.py
```diff
@@ -211,10 +219,14 @@ def _get_hostvars(self, host, vars_prefix='', omitted_vars=()):
...
     def _populate(self):
+        host_filters = self.get_option('host_filters')
 
         for host in self._get_hosts():
-
             if host.get('name'):
+                # Skip any hosts that match host_filters
+                if host_filters:
+                    if any([re.search(host_filter, host['name']) for host_filter in host_filters]):
+                        continue
```
evgeni, line 229: the old script used the same config file option to pass a "search" to foreman:
https://github.com/ansible/ansible/blob/5192f68f3a056765bbdb6614aea478e4bd977335/contrib/inventory/foreman.py#L204-L206

using the same name but a completely different filtering seems odd to me. plus this will still fetch the host information from the API and discard it later, instead of not fetching it at all.

IMHO the change in #823 is cleaner and should be prefered.

evgeni:
I've merged #823 now, so this at least conflicts.

lukepafford:
Functionality was implemented in request #823

PR #806, opened 2020-05-01
Port support for rich_params to inventory plugin

jladdjr:
Porting support for rich_params from https://github.com/ansible/ansible/pull/26180

### DIFF
--- a/plugins/inventory/foreman.py
+++ b/plugins/inventory/foreman.py
@@ -65,6 +65,9 @@ DOCUMENTATION = '''
             - Places hostvars in a dictionary with keys `foreman`, `foreman_facts`, and `foreman_params`
         type: boolean
         default: False
+      rich_params:
+        description:
+            - Toggle, if true parameters will be interpreted if possible as JSON
 '''
 
 EXAMPLES = '''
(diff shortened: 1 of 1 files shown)


jladdjr:
cc @willtome

mdellweg:
Parameters in foreman have type and value these days. So i think, there should be no trying, but knowing.

jladdjr:
@mdellweg - added test to ensure that `loads` is getting a string, which prevents the TypeError from happening. Updated the ValueError to refer to  `json.JSONDecodeError` specifically, which definitely does get raised if the string doesn't represent a json object.

mdellweg:
I might have been to brief here.
Your code as is will attempt to decode whatever comes along as json. But maybe the user provides a string that _looks_ like an empty dictionary, but intents to be literal `{}`?
Foreman provides types along with the value of a parameter. So we know exactly, whether a specific parameter is a string, boolean, integer, array, dictionary, json or even yaml. We should decode it accordingly.
Have a look at how the `foreman_global_parameter` module handles types.

jladdjr:
Learned today that `rich_params` is limited to sat 6.5 and earlier - with 6.5 set to EOL soon. Closing as this seems like it won't be relevant for very long.

PR #813, opened 2020-05-05
correct install statement dropping "git"

evgeni:


mdellweg:
Dropping the use "from git" leaves us with only using it as a collection, right?
iirc the rpm is the packaged version of exactly that collection.
So this would effectively drop support for ansible < 2.8

evgeni:
We already dropped the instructions how to do so from the readme, this only corrects the heading.

And yes, both rpm and galaxy use the collection style distribution which needs 2.8+.

But the code still works with 2.7, we just don't advertise how to use it anymore.

mdellweg:
Yeah, i was wandering, whether this means we are about to drop all that convert into collection mechanics.

evgeni:
Not before 2.7 is EOL/2.10 is released.

PR #729, opened 2020-03-11
Check for missing commits vs devel for inventory plugin

willtome:
diff against ansible-collection-migration/theforeman.foreman for `plugins/inventory/foreman.py` taking into account changes from #591

PR #823, opened 2020-05-26
add host_filters and want_ansible_ssh_host like script used to have

jainnikhil30:
This PR adds two features:

- host_filters: This is backward compatible with the script for filtering hosts
- want_ansible_ssh_host: This is backward compatible with script to add ip address of the node as a ansible_ssh_host  hostvars.

Both of these were present in the script:

https://github.com/ansible-collections/community.general/blob/ce9701f813c3b8d8171ce7272ae15be5dbb7fd37/scripts/inventory/foreman.ini#L133

https://github.com/ansible-collections/community.general/blob/ce9701f813c3b8d8171ce7272ae15be5dbb7fd37/scripts/inventory/foreman.ini#L193

### DIFF
--- a/plugins/inventory/foreman.py
+++ b/plugins/inventory/foreman.py
@@ -71,6 +71,7 @@ DOCUMENTATION = '''
         default: False
       host_filters:
         description: This can be used to restrict the list of returned host
+        type: string
 '''
 
 EXAMPLES = '''
@@ -80,6 +81,7 @@ url: http://localhost:2222
 user: ansible-tester
 password: secure
 validate_certs: False
+host_filters: 'organization="Web Engineering"'
 '''
 
 from distutils.version import LooseVersion
--- a/plugins/inventory/foreman.py
+++ b/plugins/inventory/foreman.py
@@ -144,7 +144,7 @@ class InventoryModule(BaseInventoryPlugin, Cacheable, Constructable):
             s = self._get_session()
             if params is None:
                 params = {}
-            params['page'] = 1 
+            params['page'] = 1
             params['per_page'] = 250
             while True:
                 ret = s.get(url, params=params)
@@ -188,7 +188,6 @@ class InventoryModule(BaseInventoryPlugin, Cacheable, Constructable):
         if self.get_option('host_filters'):
             params['search'] = self.get_option('host_filters')
         return self._get_json(url, params=params)
-
 
     def _get_all_params_by_id(self, hid):
         url = "%s/api/v2/hosts/%s" % (self.foreman_url, hid)
--- a/plugins/inventory/foreman.py
+++ b/plugins/inventory/foreman.py
@@ -65,6 +65,12 @@ DOCUMENTATION = '''
             - Places hostvars in a dictionary with keys `foreman`, `foreman_facts`, and `foreman_params`
         type: boolean
         default: False
+      want_ansible_ssh_host:
+        description: Toggle, if true the plugin will populate the ansible_ssh_host variable to explicitly specify the connection target
+        type: boolean
+        default: False
+      host_filters:
+        description: This can be used to restrict the list of returned host
 '''
 
 EXAMPLES = '''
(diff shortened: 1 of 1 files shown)


jainnikhil30:
@AlanCoding @jladdjr @evgeni 

Can we please check this ? We need both of these params, because its breaking things in awx and Tower both.

plugins/inventory/foreman.py
```diff
@@ -65,6 +65,13 @@
             - Places hostvars in a dictionary with keys `foreman`, `foreman_facts`, and `foreman_params`
         type: boolean
         default: False
+      want_ansible_ssh_host:
```
AlanCoding, line 68: I don't know if _this_ parameter is necessary any more since you could add the option

```yaml
compose:
  ansible_ssh_host: foreman['ip6'] | default(foreman['ip'], true)
```

but looking at this now, I'm unclear on if that would handle all the ip, ipv4, ipv6 cases. So maybe this is absolutely necessary.

plugins/inventory/foreman.py
```diff
@@ -274,6 +289,17 @@ def _populate(self):
                             except ValueError as e:
                                 self.display.warning("Could not create groups for host collections for %s, skipping: %s" % (host_name, to_text(e)))
 
+                # put ansible_ssh_host as hostvar
+                if self.get_option('want_ansible_ssh_host'):
+                    for key in ('ip', 'ipv4', 'ipv6'):
+                        if host.get(key):
+                            try:
+                                self.inventory.set_variable(host_name, 'ansible_ssh_host', host[key])
```
AlanCoding, line 297: isn't ansible_host preferred? I'm not particular, I just want to raise the question early.

AlanCoding:
It seems like this should be preferred over https://github.com/theforeman/foreman-ansible-modules/pull/799 because that's client-side filtering? Did I read that right?

AlanCoding:
👍  from me, everything here looks necessary and correct.

plugins/inventory/foreman.py
evgeni: this is also the reasoning @willtome gave in https://github.com/theforeman/foreman-ansible-modules/pull/729#issuecomment-602625977, so I'd *prefer* not to add this, unless there is really a scenario where `compose` does not help us.

jainnikhil30: @evgeni @AlanCoding  I did look at that, but it seems like Satellite has 3 different values that it can populate: ipv6,ipv4,ip. And I am not sure if the compose handles all of that.

evgeni: IMHO this should only be `ip` and `ip6`, at least that's what Foreman provides. Not sure where `ip4` is coming from.

jainnikhil30: @evgeni got it. If that's the case I can may be remove this and let this be handled via compose. On Tower there will be few customers for whom this will be backward incompatible since script had that parameter "want_ansible_ssh_host" and now they will have to change that to compose.

evgeni: While I like a bit of backward compatibility (see also my comment in https://github.com/theforeman/foreman-ansible-modules/pull/799#discussion_r433803264) I don't think it should come at the cost of carrying around code from an old implementation that's not supported anymore.

jainnikhil30: @evgeni looking at this reply here https://github.com/ansible/awx/issues/2916#issuecomment-584834026 looks there was a case when ipv4 was there ?

evgeni: "was there"? :)
The API has `ip` and `ip6`, no `ipv4` and no `ipv6` (note the missing *v* in the name) (see https://github.com/theforeman/foreman/blob/develop/app/models/host/base.rb#L92-L101)

jainnikhil30: @evgeni oh that's interesting. May be we need to check the compose too then. But I have removed this param from this PR. Now this only covers host_filters.
Thanks a lot for looking into it.

AlanCoding:
👍  I'm on the same page about removing the ssh host option, and just getting in the host_filters.

PR #824, opened 2020-05-28
don't fail recording new composite content view components

evgeni:


evgeni:
Python 3.5 tests fail on devel with `name 'ModuleNotFoundError' is not defined`, which does not look related to this change

evgeni:
seems to be broken by https://github.com/ansible/ansible/commit/f7dfa817ae6542509e0c6eb437ea7bcc51242ca2#diff-cb51550dcfdbd27a3f224be2de15e2e8R474

evgeni:
CI is green, GH just thinks it's cool not to invalidate the older results 🤷‍♀️

PR #834, opened 2020-06-10
don't directly pull in sanity deps

evgeni:
Ansible will do this automatically now, and the file does not exist
anymore


evgeni:
well, that didn't work.

PR #839, opened 2020-06-17
Use 'missing_required_lib' to report import errors

### LINKED ISSUE #326: There is a convention for importing libraries, we should conform to.

evgeni:
Fixes #326
Replaces #328

### DIFF
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -15,25 +15,18 @@ jobs:
       matrix:
         python: ["2.7", "3.5", "3.6", "3.7"]
         ansible:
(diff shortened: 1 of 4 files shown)


mdellweg:
there is another place:
`module.fail_json(missing_required_lib("PyYAML"), exception=PYYAML_IMP_ERR)`

evgeni:
@mdellweg you're right -- and even more places :)

PR #840, opened 2020-06-18
drop more Katello/Foreman mentions

evgeni:


plugins/modules/resource_info.py
```diff
@@ -77,7 +77,7 @@
 - debug:
     var: result.resources[0].value
 
-- name: "Read all Registries (Katello)"
+- name: "Read all Registries"
```
mdellweg, line 80:
```suggestion
- name: "Read all Container Registries"
```
Certainly only if that would be correct...

evgeni: To be honest, I have no idea, and thus would not touch it :D

PR #841, opened 2020-06-18
pure collection

evgeni:


mdellweg:
I think this is blocked by out tests not using the modules from the collection. I am trying to close that gap with:
https://github.com/theforeman/foreman-ansible-modules/pull/842

PR #853, opened 2020-06-24
drop metadata, it's not needed inside a collection

evgeni:


mdellweg:
I think you forgot the filter plugin here.

evgeni:
> I think you forgot the filter plugin here.

ofcourse I did

evgeni:
For @gundalow: https://github.com/ansible-collections/overview/issues/45#issuecomment-627922948

gundalow:
Thank you :)

PR #855, opened 2020-06-24
add make record-crud

pondrejk:
makefile option to re-record responses for all crud tests

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -26,6 +26,7 @@ help:
 	@echo "  test-setup     to install test dependencies"
 	@echo "  test_<test>    to run a specific unittest"
 	@echo "  record_<test>  to (re-)record the server answers for a specific test"
+	@echo "  record-crud    to (re-)record the server answers for crud tests"
 	@echo "  clean_<test>   to run a specific test playbook with the teardown and cleanup tags"
 	@echo "  dist           to build the collection artifact"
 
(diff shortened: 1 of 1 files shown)


evgeni:
Hey,

what's the use case here? Do far I never had to re-record all in one go.

Do you rather want to run the tests against a live instance, ignoring the records?

pondrejk:
> 
> Do you rather want to run the tests against a live instance, ignoring the records?

Yes, basically I want that, though I thought having the records from the run might come handy in the future, for example if the upstream/downstream test sets start to diverge.

evgeni:
So I think I would prefer a `live` fixture that works similarly as our current `record` one instead. But:
- I'm not opposed to having both in (as I see a certain use of having the newly recorded results if one is analysing some issue)
- I'm not going to implement `live` just now.

cc @mdellweg who wrote a big chunk of the whole VCR test setup and might have opinions.

mdellweg:
I am not against it per se, but i fear people will open pr's with _all_ fixtures regenerated without the need for it.
With a small bash loop, you can automatically regenerate all tests that fail.
The `test_live` sounds like a valueable addition regardless.

pondrejk:
Closing as this approach causes clashes between tests in setup and teardown phases, using loop with record_[item] instead

PR #857, opened 2020-06-27
drop prefixes from the examples

evgeni:
also don't use local_action, that's confusing


evgeni:
Yeah, but we don't do that in any example.

PR #859, opened 2020-06-29
Cleanup api

mdellweg:
While thinking about the `ForemanStatelessEntityAnsibleModule` i found some cruft to cleanup...

### DIFF
--- a/plugins/module_utils/foreman_helper.py
+++ b/plugins/module_utils/foreman_helper.py
@@ -876,14 +876,12 @@ class ForemanEntityAnsibleModule(ForemanAnsibleModule):
         * entity_name (str): name of the current entity.
           By default deduce the entity name from the class name (eg: 'ForemanProvisioningTemplateModule' class will produce 'provisioning_template').
         * entity_scope (list): List of the entity used to build search scope. Defaults to None.
-        * entity_resolve (boolean): Set it to False to disable base entity resolution.
         * entity_opts (dict): Dict of options for base entity. Same options can be provided for subentities described in foreman_spec.
     """
 
     def __init__(self, **kwargs):
         self.entity_key = kwargs.pop('entity_key', 'name')
         self.entity_scope = kwargs.pop('entity_scope', None)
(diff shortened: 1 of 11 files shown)


mdellweg:
CI failure unrelated

evgeni:
the `__init__.py` drop breaks tests on python 2.7:

```
______________ ERROR collecting tests/test_foreman_spec_helper.py ______________
ImportError while importing test module '/home/runner/work/foreman-ansible-modules/foreman-ansible-modules/tests/test_foreman_spec_helper.py'.
Hint: make sure your test modules/packages have valid Python names.
Traceback:
tests/test_foreman_spec_helper.py:1: in <module>
    from plugins.module_utils.foreman_helper import _foreman_spec_helper
E   ImportError: No module named plugins.module_utils.foreman_helper
===================== 106 passed, 1 error in 2.77 seconds ======================
```

Ansible itself is fine with that tho.

mdellweg:
Oh right. That is where we still test the original files.

plugins/module_utils/foreman_helper.py
```diff
@@ -971,10 +962,11 @@ def run(self, **kwargs):
                 self.foreman_params[self.entity_key] = self.foreman_params.pop(updated_key)
 
         params = kwargs.get('params', {})
-        if self.entity_scope:
-            for scope in self.entity_scope:
+        entity_scope = self.foreman_spec['entity'].get('scope')
+        if entity_scope:
+            for scope in entity_scope:
```
evgeni, line 967: you could write `for scope in self.foreman_spec['entity'].get('scope', []):` here… but I don't think it's needed

mdellweg: Do you think, it's more readable? Because that's the only reason why i would do it.

evgeni: It saves two lines, whether this is more readable depends on the reader, I guess. So probably just leave it as it is.

PR #860, opened 2020-06-29
use foreman.example.com in doc examples

evgeni:


evgeni:
Yeah. Katello is a plugin, and IMHO should be handled like one. ;)

PR #862, opened 2020-06-30
Entity mixin

mdellweg:


plugins/module_utils/foreman_helper.py
```diff
@@ -499,10 +509,14 @@ def find_operatingsystems(self, names, **kwargs):
...
 
     def find_puppetclass(self, name, environment=None, params=None, failsafe=False, thin=None):
+        if thin is None:
+            thin = self._thin_default
         if environment:
             scope = {'environment_id': environment}
         else:
-            scope = None
+            scope = {}
+        if params is not None:
```
evgeni, line 518: funny, this was never used before? :)

mdellweg: No, it wasn't. Also noone ever specified `thin` in the `entity_opts`. But you _can_ do that now...

plugins/modules/content_view_filter.py
```diff
@@ -168,7 +168,7 @@
 
 RETURN = ''' # '''
 
-from ansible_collections.theforeman.foreman.plugins.module_utils.foreman_helper import KatelloAnsibleModule, _foreman_spec_helper
+from ansible_collections.theforeman.foreman.plugins.module_utils.foreman_helper import KatelloMixin, ForemanStatelessEntityAnsibleModule
```
evgeni, line 171: Should we have a `KatelloStatelessEntityAnsibleModule`?

mdellweg: Maybe if we use it more than once?

evgeni: Yeah, I somehow read the diff as if we'd do that. But the other module is `setting`, so no.

PR #864, opened 2020-07-02
job invocation module

pondrejk:
A first take on the job invocation module, I'll appreciate any feedback :)
The module is limited to two command templates (ssh and ansible providers), the plan is to include others gradually. More extensive tests pending.


plugins/modules/job_invocation.py
```diff
@@ -0,0 +1,223 @@
...
+  bookmark_id:
+    description:
+      - Bookmark ID to infer the search query from
+    choices:
+      - 6 (active)
+      - 9 (disabled)
+      - 8 (error)
+      - 1 (list hypervisors)
+      - 10 (ok hosts)
+      - 7 (out of sync)
```
mdellweg: Where do the numbers come from? I guess they are special to your system?
As a rule of thumb, we do not expose any id's in the user facing part of the ansible module.
Instead we ask the user for the name of the bookmark and do the id-lookup internally.
I'll keep the bookmarks as an example, but with job_templates it is the same deal.

plugins/modules/job_invocation.py
```diff
@@ -0,0 +1,223 @@
...
+
+from ansible_collections.theforeman.foreman.plugins.module_utils.foreman_helper import (
+    ForemanEntityAnsibleModule,
+)
+
+ssh_foreman_spec = {
+    'effective_user': dict(),
+}
+
+recurrnece_foreman_spec = {
```
mdellweg:
```suggestion
recurrence_foreman_spec = {
```

plugins/modules/job_invocation.py
```diff
@@ -0,0 +1,223 @@
...
+
+class ForemanJobInvocationModule(ForemanEntityAnsibleModule):
+    pass
+
+
+def main():
+    module = ForemanJobInvocationModule(
+        foreman_spec=dict(
+            search_query=dict(),
+            bookmark_id=dict(),
```
mdellweg: This is where the `foreman_spec` magic starts (Have a look at the docs...).
I guess it will be something like:
```suggestion
            bookmark=dict(type='entity'),
```
however for bookmarks you will need to specify the controller, too.

plugins/modules/job_invocation.py
```diff
@@ -0,0 +1,223 @@
...
+            job_template_id=dict(required=True),
+            targeting_type=dict(default='static_query', choices=['static_query', 'dynamic_query']),
+            randomized_ordering=dict(type='bool', default=False),
+            feature=dict(),
+            command=dict(),
+            inputs=dict(),
+            execution_timeout_interval=dict(),
+            ssh=dict(type='nested_list', foreman_spec=ssh_foreman_spec),
```
mdellweg: I do not think, `nested_list` is appropriate here.
maybe `type='dict'`?

plugins/modules/job_invocation.py
```diff
@@ -0,0 +1,223 @@
...
+            recurrence=dict(type='nested_list', foreman_spec=recurrnece_foreman_spec),
+            scheduling=dict(type='nested_list', foreman_spec=scheduling_foreman_spec),
+            concurrency_control=dict(type='nested_list', foreman_spec=concurrency_control_foreman_spec),
+        ),
+        entity_opts={'search_by': 'description'}
+    )
+
+    # make sure we have a search query
+    if 'search_query' not in module.foreman_params.keys() and 'bookmark_id' not in module.foreman_params.keys():
```
mdellweg: Look for `required_one_of` and `required_if` (ansible features).

plugins/modules/job_invocation.py
```diff
@@ -0,0 +1,223 @@
...
+        module.fail_json(
+            msg='No command specified for command template.')
+
+    # command input required by api
+    if 'command' in module.foreman_params.keys():
+        module.foreman_params['inputs'] = {"command": f"{module.foreman_params['command']}"}
+        module.foreman_params.pop('command')
+
+    with module.api_connection():
+        module.run()
```
mdellweg: Because Bookmark lookup is not that easy, you may want to call something like
```
if 'bookmark' in module.foreman_params:
    module.set_entity('bookmark', module.find_resource('bookmark', search='name="{0}",controller="{1}"'.format(module.foreman_params['name'], module.foreman_params['controller']),
            failsafe=False,
        ))
```

plugins/modules/job_invocation.py
```diff
@@ -0,0 +1,223 @@
...
+        description:
+          - Distribute tasks over given number of seconds
+        type: int
+      concurrency_level:
+        description:
+          - Maximum jobs to be executed at once
+        type: int
+extends_documentation_fragment:
+  - theforeman.foreman.foreman
+  - theforeman.foreman.foreman.entity_state_with_defaults
```
mdellweg:
```suggestion
  - theforeman.foreman.foreman.entity_state
```
You don't have `present_with_defaults`.

plugins/modules/job_invocation.py
pondrejk: Is this really required? Right now, bookmarks resolve just fine for me having `bookmark=dict(type='entity', controller=1)`

pondrejk:
> Is there a human readable natural key to those job_invocations, with which we can identify them, that we can put in a playbook before the object is even created?

Job invocation gets a numeric id when it is executed but I'm not sure it is predictable if this is what you aim at

plugins/modules/job_invocation.py
mdellweg: Bookmarks are identified by the combination of their name and the controller.
It may work in your setup, because you have no duplicate names there atm.
I believe for this module, the controller will always be the host controller, but we cannot rely on it having the db-id 1.
Also `controller` is afaik not part of the `foreman_spec` specification, and i am wondering why it is simply ignored.

plugins/modules/job_invocation.py
```diff
@@ -0,0 +1,222 @@
...
+        required_if=[
+            ['job_template', 'Run Command - SSH Default', ['command']],
+            ['job_template', 'Run Command - Ansible Default', ['command']],
+        ],
+        entity_opts={'search_by': 'description'}
+    )
+
+    # command input required by api
+    if 'command' in module.foreman_params.keys():
+        module.foreman_params['inputs'] = {"command": f"{module.foreman_params['command']}"}
```
evgeni:
```suggestion
        module.foreman_params['inputs'] = {"command": module.foreman_params.pop('command')}
```

`f` formats are not  compatible with python 2, and in this case not required anyways :)

and if we just use pop here, we won't need it in the next line :)

plugins/modules/job_invocation.py
```diff
@@ -0,0 +1,222 @@
...
+        required_one_of=[['search_query', 'bookmark']],
+        required_if=[
+            ['job_template', 'Run Command - SSH Default', ['command']],
+            ['job_template', 'Run Command - Ansible Default', ['command']],
+        ],
+        entity_opts={'search_by': 'description'}
+    )
+
+    # command input required by api
+    if 'command' in module.foreman_params.keys():
```
evgeni:
```suggestion
    if 'command' in module.foreman_params:
```

`in` searches for `keys` by default

plugins/modules/job_invocation.py
```diff
@@ -0,0 +1,222 @@
...
+        module.foreman_params['inputs'] = {"command": f"{module.foreman_params['command']}"}
+        module.foreman_params.pop('command')
+
+    with module.api_connection():
+        if 'bookmark' in module.foreman_params.keys():
```
evgeni:
```suggestion
        if 'bookmark' in module.foreman_params:
```

plugins/modules/job_invocation.py
```diff
@@ -0,0 +1,221 @@
...
+            targeting_type=dict(default='static_query', choices=['static_query', 'dynamic_query']),
+            randomized_ordering=dict(type='bool'),
+            feature=dict(),
+            command=dict(),
+            inputs=dict(type='dict'),
+            execution_timeout_interval=dict(type='int'),
+            ssh=dict(type='dict', foreman_spec=ssh_foreman_spec),
+            recurrence=dict(type='dict', foreman_spec=recurrence_foreman_spec),
+            scheduling=dict(type='dict', foreman_spec=scheduling_foreman_spec),
+            concurrency_control=dict(type='dict', foreman_spec=concurrency_control_foreman_spec),
```
evgeni: some of these were `type=dict`, some `type=nested_list.` IMHO (by looking at https://github.com/theforeman/foreman_remote_execution/blob/master/app/controllers/api/v2/job_invocations_controller.rb#L40-L60) they all should be `type=dict`, but that actually does not support `foreman_spec`. we could use `options=…` instead, or teach the f_spec helper to understand `type=dict`?

@mdellweg what do you think?

evgeni: (fwiw, the tests include `concurrency_control`, but the recorded api tape does not)

tests/test_playbooks/job_invocation.yml
```diff
@@ -0,0 +1,34 @@
...
+        password: "{{ foreman_password }}"
+        server_url: "{{ foreman_server_url }}"
+        command: 'pwd'
+        bookmark: 'ok hosts'
+        job_template: 'Run Command - Ansible Default'
+        validate_certs: "{{ foreman_validate_certs }}"
+        recurrence:
+          cron_line: "30 2 * * *"
+        concurrency_control:
+          concurrency_level: "{{ 2 }}"
```
evgeni:
```suggestion
          concurrency_level: 2
```

tests/test_playbooks/job_invocation.yml
```diff
@@ -0,0 +1,34 @@
...
+      job_invocation:
+        username: "{{ foreman_username }}"
+        password: "{{ foreman_password }}"
+        server_url: "{{ foreman_server_url }}"
+        search_query: "{{ foreman_host }}"
+        command: 'ls'
+        job_template: 'Run Command - SSH Default'
+        validate_certs: "{{ foreman_validate_certs }}"
+
+    - name: "Run Ansible command on active hosts once a day"
```
evgeni, line 32: our tests usually follow the following pattern:
1. create an entity (ensure `changed=True`)
2. invoke the module with identical parameters, as the entity already exists, the module should return `changed=False`
3. update the entity (ensure `changed=True)
4. invoke the module with identical parameters, as the entity is already updated, the module should return `changed=False`
5. delete the entity (ensure changed=True)
6. invoke the module with identical parameters, as the entity is already deleted, the module should return `changed=False`

is that something that can be done with job invocations too?

plugins/modules/job_invocation.py
```diff
@@ -0,0 +1,221 @@
...
+# You should have received a copy of the GNU General Public License
+# along with this program.  If not, see <http://www.gnu.org/licenses/>.
+
+from __future__ import absolute_import, division, print_function
+__metaclass__ = type
+
+DOCUMENTATION = '''
+---
+module: job_invocation
+short_description: Manage Job Invocations in Foreman
```
evgeni:
```suggestion
short_description: Manage Job Invocations
version_added: 1.2.0
```

tests/test_playbooks/job_invocation.yml
pondrejk: Since the entity is job invocation itself, I'd say most of the points don't apply. It is perfectly ok to invoke the job with same parameters, and once it is completed, there's no way to update or delete the record of it. 

For future-scheduled jobs, abortion is possible, I'll look into how to test it.

Recurring jobs can be updated and deleted, though I suppose as a part of the recurring job endpoint we didn't implement yet.

evgeni: you have `recurrence` param, isn't that what creates a recurring job?

plugins/modules/job_invocation.py
mdellweg: do you mean `nested_dict`? interesting idea.

evgeni: It's not nested as such. it's just a dict that needs to be passed to the API.

tests/test_playbooks/job_invocation.yml
pondrejk: Yes that is how you create it, but if you want to revisit it, it's called `recurring logics` in the api. This recurring logic will trigger job invocations in the future. If you list job invocations, you can see those that already run and the next one queued by the rec. logic. So you can cancel the next planned job, but to update or cancel all the future repetitions you need to talk to recurring logics (seems that UI only supports cancelling logics, API can also update them)

evgeni:
If I'm reading https://github.com/theforeman/foreman_remote_execution/blob/master/app/controllers/api/v2/job_invocations_controller.rb correctly, then:
* Job invocations are not taxonomic
* There is no update or destroy method
* There is a cancel method (as you mentioned above) which will require special handling
* There is a rerun method which we probably don't need just now

What I was actually looking for: you set search_by description, which is rather unusual, but I could find anything better yet

tests/test_playbooks/job_invocation.yml
evgeni: Oh Chtulhu. It goes via https://github.com/theforeman/foreman-tasks/blob/master/app/controllers/foreman_tasks/api/recurring_logics_controller.rb? "Nice"

The I think this module should not have state (absent) support at all for now, as it only can create invocations.

plugins/modules/job_invocation.py
pondrejk: @evgeni @mdellweg  so I changed from `foreman_spec=` to `options=` and re-recorded the tapes. Is there anything more I should do to get this in?

evgeni: I'll have a look on Monday

plugins/modules/job_invocation.py
```diff
@@ -0,0 +1,222 @@
...
+# along with this program.  If not, see <http://www.gnu.org/licenses/>.
+
+from __future__ import absolute_import, division, print_function
+__metaclass__ = type
+
+DOCUMENTATION = '''
+---
+module: job_invocation
+short_description: Manage Job Invocations
+version_added: 1.2.0
```
evgeni:
```suggestion
version_added: 1.4.0
```

sorry, we released 1.2.0 in the meantime ;-)

plugins/modules/job_invocation.py
```diff
@@ -0,0 +1,222 @@
...
+            recurrence=dict(type='dict', options=recurrence_foreman_spec),
+            scheduling=dict(type='dict', options=scheduling_foreman_spec),
+            concurrency_control=dict(type='dict', options=concurrency_control_foreman_spec),
+        ),
+        required_one_of=[['search_query', 'bookmark']],
+        required_if=[
+            ['job_template', 'Run Command - SSH Default', ['command']],
+            ['job_template', 'Run Command - Ansible Default', ['command']],
+        ],
+        entity_opts={'search_by': 'description'}
```
evgeni: We don't have a description parameter, so the search you executed looks like:
```
https://foreman.example.org/api/job_invocations?search=description%3D%22None%22&per_page=4294967296
```

`description=None` looks wrong?

tests/test_playbooks/job_invocation.yml
```diff
@@ -0,0 +1,34 @@
+# registered host expected
```
evgeni, line 1: Your recorded fixtures contain an APIdoc dump, because we have a bug in the logic that throws it away.

Could you add the following to the beginning of the test file? It should get rid of the huge API doc by pre-warming the cache.

```yaml
---
- hosts: localhost
  collections:
    - theforeman.foreman
  gather_facts: false
  vars_files:
    - vars/server.yml
  tasks:
    - name: fetch status
      include_tasks: tasks/status_info.yml
```

plugins/modules/job_invocation.py
mdellweg:
```suggestion
version_added: 1.3.0
```
as 1.2.0 has been released

plugins/modules/job_invocation.py
```diff
@@ -0,0 +1,222 @@
...
+- name: "Run ansible commad on active hosts once a day"
+  job_invocation:
+    bookmark: 'active'
+    command: 'pwd'
+    job_template: "Run Command - Ansible Default"
+    recurrence:
+      cron_line: "30 2 * * *"
+    concurrency_control:
+      concurrency_level: 2
+'''
```
mdellweg, line 145: I think, this module has a lot of complexity. Would you mind adding more examples, using `feature` and `inputs`?

plugins/modules/job_invocation.py
```diff
@@ -0,0 +1,222 @@
...
+
+RETURN = ''' # '''
```
mdellweg: There should be something it returns. Probably a list of one job-invoaction.

plugins/modules/job_invocation.py
```diff
@@ -0,0 +1,222 @@
...
+    'start_before': dict(),
+}
+
+concurrency_control_foreman_spec = {
+    'time_span': dict(type='int'),
+    'concurrency_level': dict(type='int'),
+}
+
+
+class ForemanJobInvocationModule(ForemanTaxonomicEntityAnsibleModule):
```
evgeni: My comment in https://github.com/theforeman/foreman-ansible-modules/pull/864/files#r473665687 still stands. I think this module should not support `state=absent` or updating an existing invocation because the API does not support that.
At the same time having an *entity* module helps here, as we do *create* an entity when we create the invocation.
We should just not support deletion or updating. The following is probably a good start: use `ForemanTaxonomicAnsibleModule` here and rewrite the `module.run()` to smth like:

```python
entity = module.lookup_entity('entity')
if entity is None:
    module.ensure_entity('job_invocations', module.foreman_params, entity)
else:
    module.fail_json(msg="You cannot update an existing job_invocation")
```

That still requires `lookup_entity` to actually work, which it currently does not. See my comment about `search_by`

evgeni (changes requested):

plugins/modules/job_invocation.py
```diff
@@ -0,0 +1,222 @@
...
+        required_if=[
+            ['job_template', 'Run Command - SSH Default', ['command']],
+            ['job_template', 'Run Command - Ansible Default', ['command']],
+        ],
+        entity_opts={'search_by': 'description'}
+    )
+
+    # command input required by api
+    if 'command' in module.foreman_params:
+        module.foreman_params['inputs'] = {"command": module.foreman_params.pop('command')}
```
mdellweg, line 213: Would it be a big burden on the user if we required, that a command must be specified in the `inputs` section?
It would take away some complexity of this module.
To get even more complexity back, we could add some generic input validation here. (<- Thinking out loud)

plugins/modules/job_invocation.py
```diff
@@ -0,0 +1,222 @@
...
+
+    with module.api_connection():
+        if 'bookmark' in module.foreman_params:
+            module.set_entity('bookmark', module.find_resource('bookmarks', search='name="{0}",controller="hosts"'.format(
+                module.foreman_params['bookmark']),
+                failsafe=False,
+            ))
```
mdellweg, line 220: :+1:

tests/test_playbooks/vars/server.yml.example
```diff
@@ -10,6 +10,9 @@ foreman_proxy: "foreman.example.com"
 # Parameter for snapshot test
 snapshot_host_name: "test_host"
 
+# Parameter for job invocation  test
+foreman_host: "foreman.example.com"
```
mdellweg, line 14: I think this variable needs something like this: https://github.com/theforeman/foreman-ansible-modules/blob/develop/tests/vcr_python_wrapper.py#L128

So configuring the value to another host does not break the test.

tests/test_playbooks/job_invocation.yml
mdellweg: Should it then change to be imperative (invoke_job) instead of declarative (job_invocation) to mark it fit for use in a handler without sacrificing idempotence of your playbook?

mdellweg (commented):
I have the impression, this is getting very close.
My biggest concern however is that you cannot identify existing invocations. So maybe this is only suitable for use in handlers, if you want idempotent playbooks. In that case we should at least leave a note about this in the description part of the docs.

plugins/modules/job_invocation.py
evgeni: welp, #968 -- so this should be 1.4.0 now ;)

plugins/modules/job_invocation.py
pondrejk: @evgeni I battle with this because the default value (name) fails "Field not recognized for searching!", but from recognized fields (description, host) none is unique (descrption is basically string composed of "Run" + command). I filed bz#1881383 to get id into search, I assume the search needs to lead to a unique result to be used in find_resource_by.

evgeni: Yeah, which underlines my thoughts that this is not an entity module

plugins/modules/job_invocation.py
pondrejk: I don't know, I tried to mimic the UI experience as much as possible. I guess it's ok to have it under inputs, if that's the majority opinion :) Not sure how generic we can get with the input validation as every additional template would require different inputs

tests/test_playbooks/job_invocation.yml
pondrejk: @mdellweg  what would this entail? Is it just about renaming the file?

evgeni: And setting the right name in the documentation metadata, but yeah, that's all that's needed.

FWIW, I am fine with the current name (if you need another opinion 😅).

Just for completeness, what's the UI element called where you can schedule jobs?

plugins/modules/job_invocation.py
mdellweg: My idea was to fetch the plugin in question from the server to get that info. But we should consider that a future improvement.

tests/test_playbooks/job_invocation.yml
pondrejk: Heh, to run job inv in UI you usually go through the Hosts page (select host(s), pick Schedule Remote Job); or you can go through Job Templates (click Run). To list invocations you go to Job Invocations page (go to Monitor > Jobs) -- from there you can run jobs as well. Maybe there are other starting points I don't know of yet :)

plugins/modules/job_invocation.py
mdellweg: We have `ForemanStatelessEntityAnsibleModule` for those cases.

plugins/modules/job_invocation.py
pondrejk: I decided to remove feature for now, basically it is an another way of selecting a template, the available features can be listed under Administer > Remote Execution Features in the UI. There's just a few of them and none binds to the two templates the module currently supports.
I'll add inputs example based on the result of discussion https://github.com/theforeman/foreman-ansible-modules/pull/864#discussion_r491867661

plugins/modules/job_invocation.py
pondrejk: @evgeni, @mdellweg  uh oh, went with Evgeni's suggestion already, should I rewrite with ForemanStatelessEntityAnsibleModule ?

plugins/modules/job_invocation.py
pondrejk: @evgeni opinion on this?

tests/test_playbooks/vars/server.yml.example
pondrejk: thanks for the point, I added a matcher into vcr_python_wrapper

plugins/modules/job_invocation.py
evgeni: I can see the beauty of mimicking the UI, but also of keeping it simple.

that said, I don't think these additional two lines are adding much complexity?

plugins/modules/job_invocation.py
evgeni: StatelessEntity can be updated, which a job invocation can not, so I'd think the current way is more correct.

But I also forgot about StatelessEntity when originally writing the above comment. :woman_shrugging:

plugins/modules/job_invocation.py
evgeni: okay, 5 lines, I forgot the `required_if` above…

plugins/modules/job_invocation.py
```diff
@@ -215,7 +218,12 @@ def main():
                 module.foreman_params['bookmark']),
                 failsafe=False,
             ))
-        module.run()
+        module.auto_lookup_entities()
+        entity = module.lookup_entity('entity')
+        if entity is None:
+            module.ensure_entity('job_invocations', module.foreman_params, entity, state='present')
+        else:
+            module.fail_json(msg="You cannot update an existing job_invocation")
```
mdellweg: Can this ever happen? I assumed, this module would just trigger the same command a second time (which when used as a handler would be the desired outcome, right?).
It would look something like:

```suggestion
        module.auto_lookup_entities()
        module.ensure_entity('job_invocations', module.foreman_params, None, state='present')
```

evgeni: Yepp, that's I also wanted to suggest.

I suggested the if/else in one of my previous comments, but given we can't (and don't) search for an entity anymore, this is more correct

mdellweg: And all the `record_before/after[full]` will be handled magically this way, right?

evgeni: `ensure_entity` handles all the before/after recording, yes.

but the reporting is only part of `ForemanEntityAnsibleModule`, :thinking:

evgeni: but we can fix that in a separate PR I think

evgeni: opened https://github.com/theforeman/foreman-ansible-modules/issues/996 for tracking it

plugins/modules/job_invocation.py
```diff
@@ -0,0 +1,226 @@
...
+# You should have received a copy of the GNU General Public License
+# along with this program.  If not, see <http://www.gnu.org/licenses/>.
+
+from __future__ import absolute_import, division, print_function
+__metaclass__ = type
+
+DOCUMENTATION = '''
+---
+module: job_invocation
+short_description: Manage Job Invocations
```
evgeni:
```suggestion
short_description: Invoke Remote Execution Jobs
```

plugins/modules/job_invocation.py
```diff
@@ -0,0 +1,226 @@
...
+version_added: 1.4.0
+description:
+  - "Manage Remote Execution Job Invocations"
```
evgeni:
```suggestion
  - "Invoke and schedule Remote Execution Jobs"
```

plugins/modules/job_invocation.py
```diff
@@ -0,0 +1,226 @@
...
+
+- name: "Run remote command on a single host once"
+  job_invocation:
+    search_query: "name ^ (foreman.example.com)"
+    command: 'ls'
+    job_template: "Run Command - SSH Default"
+    ssh:
+      effective_user: "tester"
+
+- name: "Run ansible commad on active hosts once a day"
```
evgeni:
```suggestion
- name: "Run ansible command on active hosts once a day"
```

evgeni (approved):
Three tiny doc comments from me, ACK otherwise.

@mdellweg also happy?

mdellweg:
:fireworks: 
@pondrejk Thanks for keeping the momentum on this PR!

PR #865, opened 2020-07-03
Ability to control the number of hosts returned per page

tchellomello:
Ability to control the number of hosts returned per page which can be definitely useful some users.

image

Result on Foreman:
```
192.168.169.112 - - [03/Jul/2020:18:00:37 -0400] "GET //api/v2/hosts/3?page=1&per_page=500 HTTP/1.1" 200 2814 "-" "python-requests/2.24.0"
192.168.169.112 - - [03/Jul/2020:18:00:37 -0400] "GET //api/v2/hosts/3?page=1&per_page=500 HTTP/1.1" 200 2814 "-" "python-requests/2.24.0"
192.168.169.112 - - [03/Jul/2020:18:00:38 -0400] "GET //api/v2/hosts/2?page=1&per_page=500 HTTP/1.1" 200 2821 "-" "python-requests/2.24.0"
192.168.169.112 - - [03/Jul/2020:18:00:38 -0400] "GET //api/v2/hosts/2?page=1&per_page=500 HTTP/1.1" 200 2821 "-" "python-requests/2.24.0"
192.168.169.112 - - [03/Jul/2020:18:00:38 -0400] "GET //api/v2/hosts/5?page=1&per_page=500 HTTP/1.1" 200 2796 "-" "python-requests/2.24.0"
```

### DIFF
--- a/plugins/inventory/foreman.py
+++ b/plugins/inventory/foreman.py
@@ -68,6 +68,10 @@ DOCUMENTATION = '''
       host_filters:
         description: This can be used to restrict the list of returned host
         type: string
+      per_page:
+        description: Determine the number of hosts returned by Foreman per page
+        type: integer
+        default: 250
 '''
 
 EXAMPLES = '''
@@ -143,7 +147,7 @@ class InventoryModule(BaseInventoryPlugin, Cacheable, Constructable):
             if params is None:
                 params = {}
             params['page'] = 1
-            params['per_page'] = 250
+            params['per_page'] = self.get_option('per_page')
             while True:
                 ret = s.get(url, params=params)
                 if ignore_errors and ret.status_code in ignore_errors:
--- a/plugins/inventory/foreman.py
+++ b/plugins/inventory/foreman.py
@@ -68,8 +68,8 @@ DOCUMENTATION = '''
       host_filters:
         description: This can be used to restrict the list of returned host
         type: string
-      per_page:
-        description: Determine the number of hosts returned by Foreman per page
+      batch_size:
+        description: Determine the number of hosts that will be returned by the Foreman API per call
         type: integer
         default: 250
 '''
@@ -147,7 +147,7 @@ class InventoryModule(BaseInventoryPlugin, Cacheable, Constructable):
             if params is None:
                 params = {}
             params['page'] = 1
-            params['per_page'] = self.get_option('per_page')
+            params['per_page'] = self.get_option('batch_size')
             while True:
                 ret = s.get(url, params=params)
                 if ignore_errors and ret.status_code in ignore_errors:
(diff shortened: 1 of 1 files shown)


plugins/inventory/foreman.py
```diff
@@ -68,6 +68,10 @@
       host_filters:
         description: This can be used to restrict the list of returned host
         type: string
+      per_page:
+        description: Determine the number of hosts returned by Foreman per page
```
evgeni, line 72: IMHO "per page" doesn't make sense in this context, as the user never sees any "pages". For them, this parameter is more of a "batch size". I'm fine with calling the parameter itself so, as that's also what the API receives, but I think the description text should explain what implications setting this would have.

mdellweg:
You could introduce a "total limit" parameter here.
But in order to get _all_ the hosts for the inventory, you'd need to cycle through the pagination.

evgeni:
@mdellweg the code *does* fetch all hosts: 
https://github.com/theforeman/foreman-ansible-modules/blob/8ebda3f5dcafeea277e38c58d903614659f5b776/plugins/inventory/foreman.py#L176-L177

This is just controlling how big the individual batches are.

plugins/inventory/foreman.py
tchellomello: Yes, makes sense.. I will address that

tchellomello:
> You could introduce a "total limit" parameter here.
> But in order to get _all_ the hosts for the inventory, you'd need to cycle through the pagination (the code already seems to do that, so i do not see the what kind of user visible difference your patch introduces).

I like the idea.. I'll send a different PR for that

mdellweg:
The `batch_size` @evgeni proposed might be a performance thing. So it would be interesting to get some numbers on that.

plugins/inventory/foreman.py
```diff
@@ -143,7 +147,7 @@ def _get_json(self, url, ignore_errors=None, params=None):
             if params is None:
                 params = {}
             params['page'] = 1
-            params['per_page'] = 250
+            params['per_page'] = self.get_option('batch_size')
```
mdellweg, line 150: I guess, you need to introduce the default value somewhere.
I cannot believe, ansible is taking that from the `DOCUMENTATION`.

evgeni: Afaik, for inventories it does

plugins/inventory/foreman.py
```diff
@@ -68,6 +68,10 @@
       host_filters:
         description: This can be used to restrict the list of returned host
         type: string
+      batch_size:
+        description: Determine the number of hosts that will be returned by the Foreman API per call
```
mdellweg, line 72:
```suggestion
        description: Number of hosts per batch that will be retrieved from the Foreman API per individual call
```
I think we should try to express this from the viewpoint of the ansible inventory.

plugins/inventory/foreman.py
mdellweg: weird!

Issue #837, opened 2020-06-17
allow for environment fallback for url/user/pass


evgeni:
##### SUMMARY
Ansible supports automatic environment lookup for parameters (see https://docs.ansible.com/ansible/latest/dev_guide/developing_program_flow_modules.html#fallback) which I think can be useful for url/user/pass

##### ISSUE TYPE
 - Feature Idea

ptoal:
It would help to co-ordinate with awx/tower, as well: https://github.com/ansible/awx/issues/2192

PR #866, opened 2020-07-08
Use environment variables for authentication

willtome:
Updates doc fragment to use environment variables consistent with inventory plugin.
Updates inventory plugin to use doc fragment rather than its own options.

Resolves Issue #837

### DIFF
--- a/plugins/doc_fragments/foreman.py
+++ b/plugins/doc_fragments/foreman.py
@@ -33,10 +33,14 @@ options:
     description: Username accessing the Foreman server
     required: true
     type: str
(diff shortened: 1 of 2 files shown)


evgeni:
Doesn't this require to add `fallback` to the argument spec as described in https://docs.ansible.com/ansible/latest/dev_guide/developing_program_flow_modules.html#fallback to work in the modules too?

willtome:
🤦‍♂️ you're right. missed that. going to close for now and come back when I've got it right for plugins AND modules.

PR #868, opened 2020-07-10
[WIP] submit nested parameters at entity creation

wbclark:
Nested parameters are currently updated by separate requests after the initial entity is created. This causes a problem in Host provisioning where the parameters will not be present at the time when templates are rendered. The problem could probably be worked around in some instances by initially creating the Host with `build: false` and then re-declaring it in a later task with `build: true` but that option isn't available for image based provisioning which immediately creates the Host on the external compute resource.

I opened this WIP to get feedback on the approach of including the nested parameters in the initial request to create the entity. While this change is incomplete, please let me know if you think this is the right approach to solving this problem or if some other strategy would be better. Thanks!

### DIFF
--- a/plugins/module_utils/foreman_helper.py
+++ b/plugins/module_utils/foreman_helper.py
@@ -219,7 +219,7 @@ class NestedParametersMixin(object):
                         'parameters', None, current_parameter, state="absent", foreman_spec=parameter_foreman_spec, params=scope)
 
 
-class HostMixin(NestedParametersMixin):
+class HostMixin(object):
     def __init__(self, **kwargs):
         foreman_spec = dict(
             compute_resource=dict(type='entity'),


evgeni:
From Bugzilla: https://bugzilla.redhat.com/show_bug.cgi?id=1855008

wbclark:
After researching this further, I think I have arrived at a much better solution than my original attempt to modify the `NestedParametersMixin`.

Currently, `NestedParametersMixin` is used by the modules: `operatingsystem`, `organization`, `subnet`, `domain`, and `location`; and also via `class HostMixin(NestedParametersMixin)` the modules: `host`, and `hostgroup`.

I think it's not actually necessary to change the way this is working for `operatingsystem`, `organization`, `subnet`, `domain`, and `location`. So to change the behavior for `host` and (collaterally) `hostgroup` it should be sufficient to make `HostMixin` no longer inherit from `NestedParametersMixin`.

If I am understanding the code correctly, then since `HostMixin` already contains in its `foreman_spec`: `parameters=dict(type='nested_list', foreman_spec=parameter_foreman_spec),` simply disabling the `NestedParametersMixin` should be sufficient.

evgeni:
The tests disagree, but that's because they expect different API requests than you send now.

I am trying to remember why we do params in such a complicated way. Maybe @mdellweg remembers?
My gut said "because a simple GET doesn't show them", but that's not true.

mdellweg:
The handling them separately was at least the way we introduced it at first. We were able to add / delete parameters even if we didn't need to touch the existing entity. Once it worked, i guess nobody wanted to look for a simpler solution.
If we can "just" inject all the parameters into the single POST/PUT/PATCH call that should be even better.

evgeni:
So I looked at the API a bit further, and think the naïve approach won't work as the API expects the parameters in `host_parameters_attributes` (well, `${resource}_parameters_attributes`) when creating/updating, but reports them as `parameters`. Nevertheless some translating is probably easier than what we have now.

evgeni:
okay, I think I got the `subnet` module to do what I want with a very similar approach!

evgeni:
welp, I know why we did it that way: Organizations and Locations *can* have params, but don't support the direct api.

mdellweg:
`ForemanTaxonomicWithNestedAttachedParametersPossiblyOnKatelloEndpointEntity`

But yes, you were right Org/Loc were the first ones. ;)

wbclark:
Beautiful. Thanks @evgeni

PR #870, opened 2020-07-13
Add reporter to report logs sent by foreman callback

### LINKED ISSUE #836: ansible-callback plugin should indicate type
##### SUMMARY
Currently foreman_ansible-module struggles to identify reports from foreman ansible-callback plugin
This needs to be fixed in sync with https://github.com/theforeman/foreman_ansible (see also https://projects.theforeman.org/issues/28069).

##### ISSUE TYPE
 - Bug Report

##### ANSIBLE VERSION

```
ansible 2.9.7
  configured module search path = [u'/root/.ansible/plugins/modules', u'/usr/share/ansible/plugins/modules']
  python version = 2.7.5 (default, Apr  2 2020, 13:16:51) [GCC 4.8.5 20150623 (Red Hat 4.8.5-39)]
```

##### KATELLO/FOREMAN VERSION

```
tfm-rubygem-katello-3.14.1
foreman-1.24.3
```

##### APYPIE VERSION

```
Metadata-Version: 1.1
Version: 0.2.1
```

##### STEPS TO REPRODUCE

Attach an ansible role to a host and click `Run Ansible roles`
The report is not detected to come from Ansible, because foreman_ansible looks for `_ansible_parsed` within the report.

##### EXPECTED RESULTS

foreman_ansible should be able to clearly detect the report as coming from ansible.

##### ACTUAL RESULTS

foreman_ansible does not find `_ansible_parsed` in the report and therefore foreman does not indicate the Report as being from ansible (by showing the Ansible Logo in the report-overview).

m-bucher:
When ansible fails to connect to remote host,
callback sends a report to Foreman without
any indication that the report comes from
Ansible. As a result, Foreman is unable
to determine the report origin.

This adds a 'reporter' attribute into
the logs that are sent.

Fixes #836

### DIFF
--- a/plugins/callback/foreman.py
+++ b/plugins/callback/foreman.py
@@ -208,6 +208,7 @@ class CallbackModule(CallbackBase):
                     "metrics": metrics,
                     "status": status,
                     "logs": log,
+                    "reporter": "ansible",
                 }
             }
             try:


ezr-ondrej:
Would it make sense to move this to the `report` https://github.com/theforeman/foreman-ansible-modules/pull/870/files#diff-0d0e9e5981cc9e47d5e0fa9be86cebf6R205

instead of duplicating this for every log?

m-bucher:
@ezr-ondrej I agree that this would be better.

However, `foreman_ansible` seems to only get the list of `logs` for detecting the reporter:

https://github.com/theforeman/foreman_ansible/blob/344333df2905cdb7c8affbb9a1dbb03f6ef7619c/app/services/foreman_ansible/ansible_report_scanner.rb#L8

ezr-ondrej:
For now I'd agree, but that can be changed quite "easily".
I'd say it's worth the effort. I'm willing to help as much as I can to get it done properly.
Given the `ReportScanner` only job is to recognize the report origin, I think it should get all info we see fit.
Take a look at my proposal: https://github.com/theforeman/foreman/pull/7819

evgeni:
today, `foreman_ansible` looks for `_ansible_parsed`, not for `reporter = ansible`, right? So the PR in the current state wouldn't work right now with the currently released `foreman_ansible`?

m-bucher:
@evgeni correct, but I regard it as the first step to add a change here, before we adopt it in foreman_ansible.
If we can intertwine it with @ezr-ondrej proposed change in foreman that would be good, too.

evgeni:
Cool, then I think Foreman and Foreman Ansible should agree on a way forward and then we can implement the right tag in the callback :)

ezr-ondrej:
If we won't move forward with my suggestion soon, I'd say we can role with what is here now and change just ansible implementation and once we agree get it done in core, we can change the implementation here.

evgeni:
Now that https://github.com/theforeman/foreman/pull/7819 is merged, I guess we need the matching change in `foreman_ansible`?

evgeni:
and then probably set it like in https://github.com/theforeman/foreman-ansible-modules/pull/896/files to the global data, and not on every log object?

ezr-ondrej:
@m-bucher could you update this to set it to the global data, please?
I believe this could go even first as the path is now given by foreman core and `foreman-ansible` needs to follow.

For the record, @xprazak2 is already working on it in https://github.com/theforeman/foreman_ansible/pull/261

ezr-ondrej:
hmm what is the release process for this? :thinking:

evgeni:
This == this collection? We release new stuff here on GH, it gets pushed to Galaxy, the next time Ansible decides to pull new stuff (aka: next is for 2.10.1) it gets in.

ezr-ondrej:
Yeah, I meant this collection :) thanks, I was just checking about the backwards compatibility we will need to get.
That case, I'd say we are ready for this, we will just need to have a talk about the backwards compatibility in `foreman_ansible`, but it's unrelated to this.

evgeni:
The collection is usable by 2.8 and newer, but in the case of 2.8/2.9 the user needs to install it explicitly to get the callback, instead of getting it automatically.

ezr-ondrej:
`foreman_ansible` is in and waiting for this, could we get it in now? :upside_down_face:

evgeni:
up

Issue #867, opened 2020-07-09
Missing module to create HTTP Proxy


knumskull:
##### SUMMARY
While it's possible to set a default proxy for the foreman instance, it's not possible to create a HTTP Proxy for that purpose.

Currently iit's possible to workaround the issue by following snippet, but it's not really idempotent.

```
  - name: Check for proxy existence
    shell: hammer http-proxy info --name "{{ default_http_proxy }}"
    register: result
    failed_when: result.rc == 123412341234
    changed_when: result.rc == 345678

  - name: Update proxy if needed
    shell: hammer http-proxy update --name "{{ default_http_proxy }}"  --username "{{ http_proxy_user }}" --password "{{ http_proxy_passwd }}" --url "{{ http_proxy }}" --location "{{ location }}"
    when: result.rc == 0

  - name: Define http proxy
    shell: hammer http-proxy create --name "{{ default_http_proxy }}" --username "{{ http_proxy_user }}" --password "{{ http_proxy_passwd }}" --url "{{ http_proxy }}" --location "{{ location }}"
    when: result.rc != 0

  - name: Set default https proxy
    setting:
      username: "{{ sat_user }}"
      password: "{{ sat_passwd }}"
      server_url: "{{ sat_url }}"
      name: "content_default_http_proxy"
      value: "{{ default_http_proxy }}"
    when: default_http_proxy is defined
```

##### ISSUE TYPE
 - Feature Idea

PR #874, opened 2020-07-15
add http_proxy module

evgeni:
Fixes: #867

### DIFF
--- a/plugins/modules/http_proxy.py
+++ b/plugins/modules/http_proxy.py
@@ -1,0 +1,107 @@
+#!/usr/bin/python
+# -*- coding: utf-8 -*-
+# (c) 2020 Evgeni Golov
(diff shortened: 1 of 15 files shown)


evgeni:
Actually, no, the username handling is broken

evgeni:
now with 100% more correct tests :)

evgeni:
that ci issue is ansible, not us :(

evgeni:
the 3.7 devel failure should be fixed by https://github.com/theforeman/foreman-ansible-modules/pull/875

mdellweg:
Seems like the pipelines for 3.6 also installs ansible from pypi. It just does not fail with it.

evgeni:
failures unrelated :see_no_evil:

PR #875, opened 2020-07-16
don't pass egg name when installing ansible

evgeni:
ansible devel is now called ansible-base and that seems to confuse some
pythons

### DIFF
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -38,7 +38,7 @@ jobs:
         with:
           python-version: ${{ matrix.python }}
       - name: Install Ansible
-        run: pip install git+https://github.com/ansible/ansible.git@${{ matrix.ansible }}#egg=ansible
+        run: pip install git+https://github.com/ansible/ansible.git@${{ matrix.ansible }}
       - name: Set Environment to use mazer
         run: |
           echo "::set-env name=COLLECTION_COMMAND::mazer"
--- a/Makefile
+++ b/Makefile
@@ -92,7 +92,7 @@ build/src/%: % | build/src
 	cp $< $@
 
 build/src:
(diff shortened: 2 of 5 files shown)


evgeni:
At least now devel fails consistently 🤷‍♀️

.github/workflows/main.yml
```diff
@@ -38,7 +38,7 @@ jobs:
         with:
           python-version: ${{ matrix.python }}
       - name: Install Ansible
-        run: pip install git+https://github.com/ansible/ansible.git@${{ matrix.ansible }}#egg=ansible
+        run: pip install git+https://github.com/ansible/ansible.git@${{ matrix.ansible }}
```
mdellweg, line 41: How about:
```suggestion
        run: pip install git+https://github.com/ansible/ansible.git@${{ matrix.ansible }}#egg=ansible-base
```

according to https://pip.pypa.io/en/stable/reference/pip_install/#vcs-support the egg is needed for dependency resolution before actually downloading anything. That could sure be a problem...

evgeni: It's only called ansible-base in devel, so that'd break the stable branches. And nothing depends on ansible in our requirements, so it should be fine. Emphasis on should.

mdellweg:
I now believe, we didn't really tests against ansible devel for a long time anymore... Because we always installed a released version afterwards.

mdellweg:
This line may need adjustment too: https://github.com/theforeman/foreman-ansible-modules/blob/master/tests/test_module_state.py#L7

evgeni:
Or deleting, as we don't have metadata at all anymore

evgeni:
next try

evgeni:
Damn lint, one day I'll just make black a requirement and be done with it.

evgeni:
Look ma, it's 🍏 and it only took me a million commits and a bazillion tries!

PR #876, opened 2020-07-17
update foreman apidoc to 2.1

evgeni:


evgeni:
well, that was unexpected.

evgeni:
this is due to this change: https://github.com/theforeman/foreman/commit/8695fba1c77b27389405d542e1bdc524d49300dc which seems fair to me.

mdellweg:
Is this change affecting, how we should handle the cidr notation?

evgeni:
@mdellweg no, well, maybe a bit ;)

Previously, the API required the `mask` parameter and we either passed what the user entered or calculated it from `cidr`. The API never documented the `cidr` param so we never sent it. Now it does and so we send both. That's fine.

However, now both `mask` and `cidr` are optional, as they are only required for IPv4 subnets, but not for v6 ones. We still have `required_one_of=[mask,cidr]` -- there is no `required_one_of_if=['type=IPv4', [mask,cidr]` so we would need to wrap that somehow ourself, or keep mask/cidr required for now, which is what I landed on.

evgeni:
we now do.

PR #877, opened 2020-07-17
use foreman 2.1 container in tests

evgeni:


evgeni:
that would have been too easy

mdellweg:
Looks like the container starts successfully one in six times. ;)

evgeni:
0/6 -- we don't start it on 2.8 ;)

evgeni:
this is now using a PostgreSQL container as 2.1 dropped SQLite "support", which should also have the benefit of being more stable as the issues we've seen in the past were usually because of SQLite locks

Issue #878, opened 2020-07-17
make cidr/netmask optional for IPv6 subnets


evgeni:
Previously, the API required the `mask` parameter and we either passed what the user entered or calculated it from `cidr`. The API never documented the `cidr` param so we never sent it. Now it does and so we send both. That's fine.

However, now both `mask` and `cidr` are optional, as they are only required for IPv4 subnets, but not for v6 ones. We still have `required_one_of=[mask,cidr]` -- there is no `required_one_of_if=['type=IPv4', [mask,cidr]` so we would need to wrap that somehow ourself, or keep mask/cidr required for now, which is what I landed on.

_Originally posted by @evgeni in https://github.com/theforeman/foreman-ansible-modules/pull/876#issuecomment-659936104_

Issue #345, opened 2019-07-12
foreman_subnet: requires cidr or mask even for absent


evgeni:
##### SUMMARY
foreman_subnet: requires cidr or mask even for absent

##### ISSUE TYPE
 - Bug Report

evgeni:
cc @sthirugn

PR #880, opened 2020-07-17
Require mask/cidr only on ipv4

mdellweg:
fixes: #878

Additional benefit: You can absent the subnet without mask or cidr.

### DIFF
--- a/plugins/modules/subnet.py
+++ b/plugins/modules/subnet.py
@@ -67,10 +67,10 @@ options:
     required: true
     type: str
   cidr:
-    description: CIDR prefix length; Required if no mask provided
+    description: CIDR prefix length; Required if I(network_type=ipv4) and no I(mask) provided
     type: int
   mask:
-    description: Subnet netmask. Required if no cidr prefix length provided
+    description: Subnet netmask. Required if I(network_type=ipv4) and no I(cidr) prefix length provided
     type: str
   from_ip:
     description: First IP address of the host IP allocation pool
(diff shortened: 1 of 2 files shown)


evgeni:
the additional benefit had an issue: #345 -- added to linked and ack!

plugins/modules/subnet.py
```diff
@@ -67,10 +67,10 @@
     required: true
     type: str
   cidr:
-    description: CIDR prefix length; Required if no mask provided
+    description: CIDR prefix length; Required if I(network_type=ipv4) and no I(mask) provided
```
evgeni, line 70:
```suggestion
    description: CIDR prefix length; Required if I(network_type=IPv4) and no I(mask) provided
```

plugins/modules/subnet.py
```diff
@@ -67,10 +67,10 @@
...
     type: int
   mask:
-    description: Subnet netmask. Required if no cidr prefix length provided
+    description: Subnet netmask. Required if I(network_type=ipv4) and no I(cidr) prefix length provided
```
evgeni, line 73:
```suggestion
    description: Subnet netmask. Required if I(network_type=IPv4) and no I(cidr) prefix length provided
```

plugins/modules/subnet.py
```diff
@@ -237,6 +236,8 @@ def main():
 
     if not module.desired_absent:
         if module_params['network_type'] == 'IPv4':
+            if 'mask' not in module_params and 'cidr' not in module_params:
+                module.fail_json('When specifying IPv4 networks, either "mask" or "cidr" is required.')
```
evgeni, line 240:
```suggestion
                module.fail_json(msg='When specifying IPv4 networks, either "mask" or "cidr" is required.')
```

This is not strictly needed, but I always feel better having named params to this super overloaded func :)

evgeni:
failure unrelated

PR #881, opened 2020-07-17
filter some request and response headers to make diffs smaller

evgeni:


tests/vcr_python_wrapper.py
```diff
@@ -9,6 +9,9 @@
 except ImportError:
     from urllib.parse import urlparse, urlunparse
 
+FILTER_REQUEST_HEADERS = ['Cookie']
+FILTER_RESPONSE_HEADERS = ['Apipie-Checksum', 'Date', 'ETag', 'Server', 'Set-Cookie', 'Via', 'X-Powered-By', 'X-Request-Id', 'X-Runtime']
```
mdellweg, line 13: what about removing everything starting with "X-"?

evgeni: I thought about that, but that would require a different filter approach (loop over existing headers and drop the ones with X-) so I didn't for now.

evgeni: Or did you mean add all the available X headers to that list?

tests/vcr_python_wrapper.py
```diff
@@ -81,15 +84,23 @@ def host_body_matcher(r1, r2):
...
+        response['headers'].pop(header, None)
     return response
 
 
-def filter_request_uri(request):
+def filter_request(request):
+    # always use foreman.example.org as the name of the system
     request.uri = urlunparse(urlparse(request.uri)._replace(netloc="foreman.example.org"))
+
+    for header in FILTER_REQUEST_HEADERS:
```
evgeni, line 99: Hmm, technically those could go to `filter_headers` of VCR directly. That sounds cleaner.

evgeni: yeah, that's better.

tests/vcr_python_wrapper.py
mdellweg: By nature of the X-headers, that would be impossible.
I thought about `if header in FILTER or header.startswith("X-"):`

evgeni: Yeah I meant all currently used ;)

evgeni: Anyways, let's merge this and think about more filtering later ;)

PR #885, opened 2020-07-17
include non-module plugins in docs

evgeni:


evgeni:
draft because the makefile is ugly and the generated page is ugly too (too many headings)

docs/index.rst
```diff
@@ -11,7 +11,14 @@ Welcome to Foreman Ansible Modules's documentation!
    :caption: User documentation
 
    README
+
+.. toctree::
+   :maxdepth: 1
+   :caption: Contained plugins
```
evgeni, line 17: I'm wondering whether this should have a caption or not.

Screenshot from 2020-07-20 13-16-52
Screenshot from 2020-07-20 13-17-03

It has to be an own `toctree` so that we can set `maxdepth: 1`, as otherwise Sphinx will add the subsections of the callback and inventory plugins, which are not useful in this TOC, while having `maxdepth: 2` for the `README` above *is* useful.

evgeni:
closing in favor of #895

PR #886, opened 2020-07-20
use stable-2.10 ansible branch for building docs

evgeni:
in devel the tooling got revamped and can't be abused for our needs
currently


evgeni:
unrelated failure is unrelated

PR #888, opened 2020-07-20
cleanup docs of the callback plugin

evgeni:


evgeni:
unrelated failure is unrelated

PR #892, opened 2020-07-21
Cleanup requirements

evgeni:


evgeni:
the test failure is unrelated, I really need to get around fixing those inventory tests

PR #893, opened 2020-07-21
drop dead code

evgeni:


evgeni:
@mdellweg I'd like your opinion on this

PR #894, opened 2020-07-21
sort tests

evgeni:
this was omited when we dropped the prefix


evgeni:
@mdellweg says go for it

PR #895, opened 2020-07-22
generate docs using antsibull

evgeni:
This is an alternative to/superseeds #885 

preview: https://people.redhat.com/evgeni/fam-antsibull/

~~I'd like to wait with merging this until https://github.com/ansible-community/antsibull/issues/116 is fixed~~ it's fixed


evgeni:
@melcorr @mdellweg might also be interesting for squeezer!

docs/Makefile
```diff
@@ -16,12 +16,12 @@ help:
...
-
-generate-docs: $(ANSIBLEGIT)
-	rm -rf ./modules/
-	bash -c "source $(ANSIBLEGIT)/hacking/env-setup && ANSIBLE_COLLECTIONS_PATHS=../build/collections python3 $(ANSIBLEGIT)/hacking/build-ansible.py document-plugins --module-dir ../plugins/modules --template-dir ./_templates --template-dir $(ANSIBLEGIT)/docs/templates --type rst --output-dir ./modules/"
+generate-docs:
+	mkdir -p ./plugins/
+	rm -rf ./plugins/*
+	ANSIBLE_COLLECTIONS_PATHS=../build/collections antsibull-docs collection --use-current --dest-dir ./plugins/ theforeman.foreman
+	mv ./plugins/collections/theforeman/foreman/* ./plugins/
+	rm -rf ./plugins/collections/
```
mdellweg: Is there a way we can do without all this copying, deleting, moving around?

docs/Makefile
```diff
@@ -16,12 +16,12 @@ help:
...
 
-$(ANSIBLEGIT):
-	git clone --depth=1 https://github.com/ansible/ansible/ -b stable-2.9 $(ANSIBLEGIT)
-
-generate-docs: $(ANSIBLEGIT)
-	rm -rf ./modules/
-	bash -c "source $(ANSIBLEGIT)/hacking/env-setup && ANSIBLE_COLLECTIONS_PATHS=../build/collections python3 $(ANSIBLEGIT)/hacking/build-ansible.py document-plugins --module-dir ../plugins/modules --template-dir ./_templates --template-dir $(ANSIBLEGIT)/docs/templates --type rst --output-dir ./modules/"
+generate-docs:
+	mkdir -p ./plugins/
+	rm -rf ./plugins/*
```
mdellweg:
```suggestion
	rm -rf ./plugins/
	mkdir -p ./plugins/
```
That way we even remove .* files.

docs/requirements.txt
```diff
@@ -1,3 +1,5 @@
--r https://github.com/ansible/ansible/raw/devel/docs/docsite/requirements.txt
+# we need 0.17.0 at least, and that's not released yet
+git+https://github.com/ansible-community/antsibull
```
mdellweg: Maybe they find a better name until then...
Isn't there an `ansibullbot` around somewhere?

docs/Makefile
evgeni: Only if we accept to have those (for us) useless subdirs.

antsibull currently focuses on building docs for *multiple* collections, so that structure is useful there.

mdellweg: If we can make them appear in `build`, i don't care. They would also be subject to `make clean`.

docs/Makefile
evgeni: Interesting. I had it like this before and then thought "it feels wrong to delete and create the main directory all the time". But of course hidden files are a good point.

docs/requirements.txt
evgeni: Yes, I don't like the name too much either, but it's not for us to decide and I'm really happy they split out the tooling in the first place.

docs/Makefile
evgeni: Yeah, we can dump that into build

docs/Makefile
mdellweg: See my other comment. To me it feels wrong to delete files at all to recreate them in the same Make snippet.

felixfontein:
> I'd like to wait with merging this until ansible-community/antsibull#116 is fixed

That's a good idea (it's the most annoying bug antsibull-docs has right now IMO). I've submitted a PR which fixes it (ansible-community/antsibull#146).

docs/requirements.txt
felixfontein: We had long discussions about that name, and in the end we decided on this one since a) the name should not be `ansible-*` (that's reserved to ansible core projects), and b) it should not make tab completion harder (originally we had ansibulled, and that made tab completing both ansible and ansibulled harder). antsibull satisfies both, and is also similar enough to ansible (depending on your English pronounciation, it's very similar ;) ).

Out of curiosity: why do you dislike the name antsibull?

evgeni: I type ansibull all the time (https://github.com/ansible/ansibullbot and https://twitter.com/ansibull)

felixfontein: That sometimes happens to me too. But I prefer this a lot over having something that actually starts with ansibull. During the time the tool was called ansibulled I found that the more complicated tab completion was a much larger PITA than having to remember the `t` :)

Makefile
```diff
@@ -110,8 +110,11 @@ clean:
 doc-setup:
 	pip install -r docs/requirements.txt
 doc: $(MANIFEST)
+	mkdir -p ./docs/plugins/ ./build/plugin_docs
+	antsibull-docs collection --use-current --squash-hierarchy --dest-dir ./build/plugin_docs theforeman.foreman
+	cp -a ./build/plugin_docs/* ./docs/plugins/
```
evgeni: @mdellweg is that OK with you? It generates the docs in `build`, so we don't need to clean them up, but we need them in `docs/plugins` for Sphinx to pick them up later.

docs/requirements.txt
mdellweg: Thinking about it, this sounds a bit like a reverse engineered argument. If i were able to type the first three letters of a command accident free, i would only have three letter commands and tab completion would have never been invented in the first place.
Actually `ls`, `mv`, `cp`, `rm`, `ln`, `du`, `df`, `dd`, `cc`, `sh`, `su` ... Basic Unix operations are not meant to use tab-completion.
Too bad it's not OK to use `ansible-gendoc`.

Makefile
mdellweg: Maybe we should add `docs/plugins` to the `clean` target.

For me to understand better: Docs are only rebuilt, if MANIFEST changed, which means something actually changed. I think, that's OK.

evgeni: Right, and then we can just build into that folder instead of copying things around.

mdellweg: I guess

mdellweg (approved):
Sorry for the detour with copying stuff around

evgeni:
no worries, thanks for the review!

cleaned up history a bit and going to merge this now.

Issue #882, opened 2020-07-17
Rename master branch to develop


evgeni:
To follow Foreman standards

(Will require a rename of the doc folder in gh-pages to preserve history)

evgeni:
done

PR #898, opened 2020-07-24
rename default branch to develop

evgeni:


evgeni:
this and https://github.com/theforeman/foreman-ansible-modules/pull/899 will together close #882 

I have created a develop branch matching master, set it as default and updated all PRs that were opened in 2020 to point at that branch now.

mdellweg (approved):
Have you configured all the necessary branch protection too?

evgeni:
No, there weren't any, but good point.

PR #901, opened 2020-07-25
Document return values

mdellweg:
Supersedes #654

### DIFF
--- a/plugins/modules/activation_key.py
+++ b/plugins/modules/activation_key.py
@@ -168,7 +168,17 @@ EXAMPLES = '''
     service_level: Standard
 '''
 
-RETURN = ''' # '''
+RETURN = '''
+entity:
+  description: Final state of the affected entities grouped by their type.
+  returned: success
+  type: dict
+  contains:
+    activation_keys:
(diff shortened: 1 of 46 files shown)


evgeni:
Thanks for picking this up again!

Two questions:
1. can we use doc fragments here (I guess no and we should file an RFE against Ansible)
2. Can we somehow automate the generation? Or did you gather all the data manually?

mdellweg:
I didn'd gather all the data. I just looked at the ones suspicious enough. So it might be incomplete. Also i refrained from detailing it in every PR. Trying to do so put the first PR on hold... So i'd rather see this as an improvement, and leave more details to future PRs. At least it's not lying about the return value anymore, maybe just not telling the whole truth.
Can we generate it? Maybe we can `debug` the structure of the return values to a file, but we'd need to invent all the descriptions anyway.

evgeni:
No, fragments in return don't work: https://github.com/ansible/ansible/issues/70910

And yeah, this is totally an improvement!

evgeni (approved):

PR #902, opened 2020-07-27
unify wording and formatting of plugin (short) descriptions

evgeni:


mdellweg:
Nice! Not yet consistent in the long description...
Is there a place where you see the long one without the short? I'm wondering if the repetition is useful.
With the help of doc_fragments, we could drop some lines in the long description to explain the three involved hosts (ansible controller; ansible host, foreman server).

evgeni:
Yeah, I didn't touch the long descriptions much yet, unless it was waaaay off. Small steps.

No, I don't think there is really a place where you see long without short. But I think it's nice to have the short one being "manage <entity>" and the long one telling that you can create, update and delete them. or in some cases only update, or refresh, etc?

mdellweg:
Will a description from doc_fragment be merged or hidden?

evgeni:
My memory says they are merged.

mdellweg (approved):
Definitely an improvement.

PR #905, opened 2020-07-28
use vanilla rtd theme and don't configure non-existing lexer

evgeni:


mdellweg (approved):
:zombie: must remove dead code...

:+1:

PR #908, opened 2020-08-05
Improve CI tooling

mdellweg:
* Simplify directory handling in Makefile
* Add support for roles in Makefile
* Limit release workflow to trunk repository

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -7,18 +7,17 @@ on:
 
 jobs:
   release:
-
     runs-on: ubuntu-latest
-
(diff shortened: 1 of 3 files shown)


.github/workflows/release.yml
```diff
@@ -7,18 +7,17 @@ on:
 
 jobs:
   release:
-
     runs-on: ubuntu-latest
-
+    if: github.repository == 'theforeman/foreman-ansible-modules'
```
evgeni, line 11: the same can/should be done in docs.yml

evgeni (approved):
inline comment which should not block the PR

Issue #910, opened 2020-08-05
comply with the Ansible Collections Checklist


evgeni:
there is a checklist at https://github.com/ansible-collections/overview/blob/master/collection_requirements.rst which we should check and try to comply with

known todos:
- [x] run sanity against latest stable (#909)
- [x] use FQCN in all module references (docs, examples, etc) (#913)
- [x] add minimum supported Ansible version to meta/runtime.yml (but I need to talk to @gundalow how to handle 2.8 support) (#914)
- [x] add changelog (#917)
- [x] add version_added to all plugins that were not in ansible 2.9 (#922)

evgeni:
I think we're done.

@ehelms @mdellweg would one of you please read the linked document and briefly verify we're compliant?

mdellweg:
I don't understand and also don't see anything for:
- [ ] galaxy.yml : "dependencies MUST be set to '>=1.0.0'"
- [ ] DOCUMENTATION : "Use collection version numbers for version_added (otherwise version_added_collection must be provided, but this is discouraged)"

Everything else is fine. And wrt the unittest section we are miles ahead of that document.

evgeni:
1. Collections can depend on other collections, and I read that requirement as "you're only allowed to depend on stable (not 0.x) versions. We don't have any dependencies, so "done".
2. Hmm, I guess we could add a "version_added: 1.1.0" to the http_proxy module, as that's new in the next release?

mdellweg:
See, i read that different:
https://github.com/ansible-collections/overview/issues/106

felixfontein:
New modules and plugins should have version_added. Otherwise neither the changelog generator nor users know when the module/plugin was added.

mdellweg:
The question is whether _all_ modules / plugins _need_ to specify something, or only the ones added after 1.0.0.

felixfontein:
There's no need to add `version_added` for anything that already was in Ansible 2.9. Some collections also add version_added for these, but others don't (my personal opinion is not to). In general, `version_added` is optional, but it is good practice to use it correctly - it makes the documentation a lot more useful, and your users will thank you.

mdellweg:
Almost nothing here has been in ansible 2.9. So i guess, adding a version_added: 1.0.0 to everything (but http_proxy which gets 1.1.0) is in order?

felixfontein:
Sounds good to me.

evgeni:
👍

evgeni:
[...]

PR #913, opened 2020-08-10
use FQCN in EXAMPLES

evgeni:
refs #910 and #911

### DIFF
--- a/plugins/modules/activation_key.py
+++ b/plugins/modules/activation_key.py
@@ -145,7 +145,7 @@ extends_documentation_fragment:
 
 EXAMPLES = '''
 - name: "Create client activation key"
-  activation_key:
+  theforeman.foreman.activation_key:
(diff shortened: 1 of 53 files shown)


mdellweg (approved):
I fail to see, why this is especially desirable, but there seems to be no notion of "in the scope of the collection" in ansible.

evgeni:
to make them copy-paste-able, I guess

mdellweg:
I was thinking it makes it hard to maintain a closely following fork of the repository, but we have lost that with the crazy import statements already.

evgeni:
> I was thinking it makes it hard to maintain a closely following fork of the repository, but we have lost that with the crazy import statements already.

nah, it's not that hard ;)

https://github.com/RedHatSatellite/satellite-ansible-collection/blob/master/Makefile#L115-L117

PR #914, opened 2020-08-10
add requires_ansible: '>=2.8' to runtime.yml

evgeni:
refs #910/#911

### DIFF
--- a/meta/runtime.yml
+++ b/meta/runtime.yml
@@ -1,3 +1,5 @@
+---
+requires_ansible: '>=2.8'
 plugin_routing:
   modules:
     foreman_architecture:


meta/runtime.yml
```diff
@@ -1,3 +1,7 @@
+---
+# Ansible versions prior to 2.9.10 don't support parsing runtime.yml,
+# however the collection will still work with those.
+requires_ansible: '>=2.9.10'
```
mdellweg: Maybe i am missing some context, but the comments says the first version to ever actually read that file will already satisfy that requirement.
So, what is the point here?

evgeni: comply with https://github.com/ansible-collections/overview/blob/master/collection_requirements.rst#metaruntimeyml :)

mdellweg: What if we state '>=2.8' here just in case some human being reads this?
And will it fail ansible's requirements? `Version("2.9.10") >= Version("2.8")` still holds true, right?

edit: Well you just did that. ;)

evgeni: updated to read >= 2.8 after discussion on #ansible-community

mdellweg (approved):

PR #918, opened 2020-08-11
upload collection artifact

evgeni:


ehelms (approved):
Where does it upload it to?

evgeni:
If you open up the checks tab in GH, there is a button on the right upper corner saying "artifacts".

The underlying storage is something something cloud, deleted after 90 days. Good enough for peeking into PR builds.

PR #921, opened 2020-08-12
fix want_params when the host has more than one param

### LINKED ISSUE #847: Inventory plugin want_params is broken
##### SUMMARY
When using Foreman as an inventory plugin with "want_params" set to True it seems as only few parameters are available.

##### ISSUE TYPE
 - Bug Report

##### ANSIBLE VERSION
```
2.9
```

##### KATELLO/FOREMAN VERSION

```
2.0
```

##### STEPS TO REPRODUCE

```yaml
Use "want_params" set to True/yes
```

##### EXPECTED RESULTS

Parameters included in host inventory variables

##### ACTUAL RESULTS

```
nothing.
```

If using "legacy_hostvars" so foreman_params is available.
Looking at "plugin/inventory/foreman.py" line 264:
```
self.inventory.set_variable(host_name, p['name'], p['value'])
```
When it should be:
```
self.inventory.set_variable(host_name, k, v)
```

A small suggestion, why not use the Host "ENC" api to get the parameters?
It seems like this is a better idea, it would fetch both the parameters and smart class parameters (ansible, puppet, etc..)

evgeni:
Fixes: #847

### DIFF
--- a/plugins/inventory/foreman.py
+++ b/plugins/inventory/foreman.py
@@ -268,7 +268,7 @@ class InventoryModule(BaseInventoryPlugin, Cacheable, Constructable):
                                 self.inventory.set_variable(host_name, k, v)
                             except ValueError as e:
                                 self.display.warning("Could not set hostvar %s to '%s' for the '%s' host, skipping:  %s" %
-                                                     (p['name'], to_native(p['value']), host, to_native(e)))
+                                                     (k, to_native(v), host, to_native(e)))
 
                 # set host vars from facts
                 if self.get_option('want_facts'):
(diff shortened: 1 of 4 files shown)


plugins/inventory/foreman.py
```diff
@@ -265,7 +265,7 @@ def _populate(self):
                     else:
                         for k, v in filtered_params.items():
                             try:
-                                self.inventory.set_variable(host_name, p['name'], p['value'])
+                                self.inventory.set_variable(host_name, k, v)
                             except ValueError as e:
                                 self.display.warning("Could not set hostvar %s to '%s' for the '%s' host, skipping:  %s" %
                                                      (p['name'], to_native(p['value']), host, to_native(e)))
```
mdellweg:
```suggestion
                                                     (k, to_native(v), host, to_native(e)))
```

mdellweg (approved):

PR #922, opened 2020-08-17
add version_added to all modules

evgeni:
one more for #910/#911


mdellweg:
Looks good.
The reason, the filter plugin does not get it is that it doesn't support DOCUMENTATION, right?

evgeni:
Correct. Filter plugins don't support DOCUMENTATION stanzas :(

mdellweg (approved):

PR #926, opened 2020-08-17
sort docs descending by version

evgeni:
noone cares about super old 0.x docs today


evgeni:
@ehelms I think you requested that once a loong time ago

ehelms (approved):

PR #929, opened 2020-08-19
use the pygments lexer from the ansible theme

### LINKED ISSUE #928: use the real sphinx_ansible_theme
Ansible has split out their theme at https://github.com/ansible-community/sphinx_ansible_theme, so we can use it again without doing any crude hacks.

The benefit of the Ansible theme over RTD is the pygments parser that supports Jinja+YAML in the same code block.

evgeni:
Fixes: #928

### DIFF
--- a/docs/conf.py
+++ b/docs/conf.py
@@ -42,6 +42,7 @@ extensions = [
     'recommonmark',
     'sphinx.ext.autodoc',
     'sphinx.ext.intersphinx',
+    'sphinx_ansible_theme.ext.pygments_lexer',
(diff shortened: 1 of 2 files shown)


evgeni:
contrary to the discussion in #928 I've only configured the lexer from the theme, not the whole theme, as it contains too much things that are specific for docs.ansible.com (including Google trackers :cry:, see https://github.com/ansible-community/sphinx_ansible_theme/issues/27)

mdellweg (approved):

PR #930, opened 2020-08-25
update list of possible template kinds in provisioning_template

evgeni:
Fixes: https://bugzilla.redhat.com/show_bug.cgi?id=1871978


evgeni:
sanity checks are failing, but that's unrelated. looks like we're using wrong pylint version.

mdellweg (approved):
Looks right to me.
Do we want to add a check that plugins are available for certain template kinds?

evgeni:
> Looks right to me.
> Do we want to add a check that plugins are available for certain template kinds?

We could, but then I'd need to know that, and besides "bootdisk", I don't ;)

PR #931, opened 2020-08-25
Include blank parent groups

### LINKED ISSUE #851: Improving Inventory Foreman
_From @selmison on Feb 16, 2018 12:36_

##### ISSUE TYPE
 - Feature Idea

##### COMPONENT NAME
- contrib/inventory/foreman.py

##### ANSIBLE VERSION
```
ansible 2.4.2.0
  configured module search path = [/root/.ansible/plugins/modules, /usr/share/ansible/plugins/modules]
  ansible python module location = /usr/lib/python2.7/dist-packages/ansible
  python version = 2.7.9 (default, Jun 29 2016, 13:08:31) [GCC 4.9.2]
```

##### CONFIGURATION
```
DEFAULT_LOG_PATH(/etc/ansible/ansible.cfg) = /var/log/ansible.log
HOST_KEY_CHECKING(/etc/ansible/ansible.cfg) = False
```
##### OS / ENVIRONMENT
 - Debian 8

##### SUMMARY
Why does the `_get_host_data_by_id` method in `contrib/inventory/foreman.py` use `api/hosts/:id` request instead of using `hosts/:id/enc request`?

If `_get_host_data_by_id` used the `hosts/:id/enc` request, the inventory provided by the Foreman would contain much more information and the parameters would accept ERB codes as in the Foreman/Puppet integration.

_Copied from original issue: ansible/ansible#36300_

### LINKED ISSUE #919: Foreman inventory does not sync empty groups
_From @rht-jbittner on Mar 27, 2020 11:19_

##### SUMMARY
Foreman dynamic inventory script does not sync empty host groups. 

##### ISSUE TYPE
- Bug Report

##### COMPONENT NAME

Foreman/Red Hat Satellite dynamic inventory

##### ANSIBLE VERSION

```paste below
All
```

##### STEPS TO REPRODUCE

Create empty host group in Foreman/Red Hat Satellite and sync inventory. Groups without hosts are not synced.

##### EXPECTED RESULTS

All groups synced. Syncing empty host groups is important due to group inheritance. Like in Ansible you can have host group which contains multiple sub groups and define variables on parent group level. This is not possible with current state of the inventory script.

##### ACTUAL RESULTS

Only groups with host synced.

Problem with foreman inventory script is that it does not call Foreman hostgroup API and construct group and membership from hosts instead.

_Copied from original issue: ansible/ansible#68508_

ezr-ondrej:
Fixes #919

### DIFF
--- a/plugins/inventory/foreman.py
+++ b/plugins/inventory/foreman.py
@@ -235,9 +235,16 @@ class InventoryModule(BaseInventoryPlugin, Cacheable, Constructable):
                 # create directly mapped groups
                 group_name = host.get('hostgroup_title', host.get('hostgroup_name'))
                 if group_name:
-                    group_name = to_safe_group_name('%s%s' % (self.get_option('group_prefix'), group_name.lower().replace(" ", "")))
-                    group_name = self.inventory.add_group(group_name)
-                    self.inventory.add_child(group_name, host_name)
+                    parent_name = None
+                    group_label_parts = []
+                    for part in group_name.split('/'):
+                        group_label_parts.append(part.lower().replace(" ", ""))
+                        gname = to_safe_group_name('%s%s' % (self.get_option('group_prefix'), '/'.join(group_label_parts)))
+                        result_gname = self.inventory.add_group(gname)
+                        if parent_name:
+                            self.inventory.add_child(parent_name, result_gname)
+                        parent_name = result_gname
+                    self.inventory.add_child(result_gname, host_name)
 
                 if self.get_option('legacy_hostvars'):
                     hostvars = self._get_hostvars(host)
--- a/tests/test_playbooks/inventory_plugin.yml
+++ b/tests/test_playbooks/inventory_plugin.yml
@@ -2,11 +2,14 @@
 - hosts: container
   vars:
     foreman_groups:
-      - group_a
-      - group_b
+      - name: group_a
+      - name: group_b
+      - name: group_c
+        label: group_b/group_c
+        parent: group_b
     foreman_hosts:
       testhost1.example.com: group_a
-      testhost2.example.com: group_b
+      testhost2.example.com: group_b/group_c
   collections:
     - community.general
     - theforeman.foreman
@@ -64,7 +67,8 @@
         username: admin
         password: changeme
         server_url: http://127.0.0.1:3000/
-        name: "{{ item }}"
+        name: "{{ item.name }}"
+        parent: "{{ item.parent | default(omit) }}"
         organizations:
           - Default Organization
         locations:
@@ -116,7 +120,7 @@
     - name: test that all groups are present
       assert:
         that: >
-          'foreman_{{ item }}' in groups
+          'foreman_{{ item.label | default(item.name) | regex_replace('/', '_') }}' in groups
       with_items: "{{ foreman_groups }}"
 
     - name: test that all hosts are in the "all" group
@@ -128,7 +132,7 @@
     - name: test that all hosts are in the correct hostgroup
       assert:
         that: >
-          '{{ item.key }}' in groups['foreman_{{ item.value }}']
+          '{{ item.key }}' in groups['foreman_{{ item.value | regex_replace('/', '_') }}']
       with_dict: "{{ foreman_hosts }}"
 
     - name: test that all hosts have the domain fact set
--- a/changelogs/fragments/919-include-empty-parent-groups.yml
+++ b/changelogs/fragments/919-include-empty-parent-groups.yml
@@ -1,0 +1,2 @@
+bugfixes:
+  - inventory plugin - include empty parent groups in the inventory (https://github.com/theforeman/foreman-ansible-modules/issues/919)


evgeni:
shouldn't that fix #919 instead?

ezr-ondrej:
> shouldn't that fix #919 instead?

Of course it should :facepalm: Thanks

ezr-ondrej:
Can I do something about the failures?

evgeni:
The sanity ones? Well, you can debug them, but it's not your fault ;)

ezr-ondrej:
> The sanity ones? Well, you can debug them, but it's not your fault ;)

Yeah, those ones. Well I'd rather not :thinking:  xD

evgeni:
Haha, yeah. I think the issue is pylint 2.6.0 -- ansible pins an older version, but it seems not to work here for us (the pinning).

mdellweg:
> Haha, yeah. I think the issue is pylint 2.6.0 -- ansible pins an older version, but it seems not to work here for us (the pinning).

Even worse: The version pinned by ansible doesn't lint our code for other (more stupid) reasons (like syntax error and complaints about our use of ContextManagers). So the only reason we passed sanity in the past was that the pinning didn't work.
pylint==2.5.2 seems to work however.

mdellweg:
honestly i do not see, where ansible is pinning the version:
https://github.com/ansible/ansible/blob/stable-2.9/test/lib/ansible_test/_data/requirements/sanity.txt#L4
https://github.com/ansible/ansible/blob/devel/test/lib/ansible_test/_data/requirements/sanity.pylint.txt#L1
But still when run with the `--venv` option it uses `pylint==2.3.1`. No idea where this is cached.

evgeni:
The pinning happens by using a constraints file: https://github.com/ansible/ansible/blob/stable-2.9/test/lib/ansible_test/_data/requirements/constraints.txt

Maybe we should just run sanity on all Pythons at once by using --docker and not passing --python? In a separate step that is.

mdellweg:
That old version of pylint is not going to make us happy.
I have a workaround however: #933

evgeni:
@ezr-ondrej could you have a look at the tests (https://github.com/theforeman/foreman-ansible-modules/blob/develop/tests/test_playbooks/inventory_plugin.yml#L62) and add another level of groups and maybe a parameter that gets set via an "empty" group?

ansible.cfg
```diff
@@ -1,4 +1,4 @@
 [defaults]
-collections_paths = build/collections
+collection_paths = build/collections
```
evgeni: please don't, this breaks 2.9

tests/inventory/tests.foreman.yml
```diff
@@ -1,6 +0,0 @@
-plugin: theforeman.foreman.foreman
```
evgeni: if you delete that file, the inventory won't be enabled.

tests/test_playbooks/inventory_plugin.yml
```diff
@@ -2,8 +2,10 @@
...
     foreman_groups:
-      - group_a
-      - group_b
+      - name: group_a
+      - name: group_b_parent
+      - name: group_b
+        parent: group_b_parent
     foreman_hosts:
       testhost1.example.com: group_a
       testhost2.example.com: group_b
```
evgeni:
```suggestion
      testhost2.example.com: group_b_parent/group_b
```

plugins/inventory/foreman.py
```diff
@@ -235,9 +235,15 @@ def _populate(self):
                 # create directly mapped groups
                 group_name = host.get('hostgroup_title', host.get('hostgroup_name'))
                 if group_name:
-                    group_name = to_safe_group_name('%s%s' % (self.get_option('group_prefix'), group_name.lower().replace(" ", "")))
-                    group_name = self.inventory.add_group(group_name)
-                    self.inventory.add_child(group_name, host_name)
+                    parent_name = None
+                    path = group_name.split('/')
+                    for i in range(1, len(path) + 1):
+                        gname = to_safe_group_name('%s%s' % (self.get_option('group_prefix'), '/'.join(path[0:i]).lower().replace(" ", "")))
```
evgeni: I had a bit of a hard time reading this loop and would propose an alternative writing:

```suggestion
                    group_label_parts = []
                    for part in group_name.split('/'):
                        group_label_parts.append(part.lower().replace(" ", ""))
                        gname = to_safe_group_name('%s%s' % (self.get_option('group_prefix'), '/'.join(group_label_parts)))
```

evgeni:
besides the comment about the readability of the loop, I am happy with this!

evgeni:
Oh, and could you please add a changelog fragment? Something like `changelogs/fragments/919-include-empty-parent-groups.yml`:

```yaml
bugfixes:
  - inventory plugin - include empty parent groups in the inventory (https://github.com/theforeman/foreman-ansible-modules/issues/919)
```

ezr-ondrej:
Thanks for the readability improvement, looks much better now :)

> ```yaml
> bugfixes:
>   - inventory plugin - include empty parent groups in the inventory (https://github.com/theforeman/foreman-ansible-modules/issues/919)
> ```

Or exactly that? ^_^

evgeni (approved):

PR #932, opened 2020-08-25
host: add compute_attributes parameter

evgeni:


evgeni:
~~I'd like to add some examples, but don't have any yet.~~

ehelms (approved):

nward:
Hi - how does one add interfaces with this parameter? Or do we need interfaces_attributes as well?

evgeni:
For that you'll need interfaces_parameters too, we had that request multiple times now, so it's rather high on the todo list

PR #934, opened 2020-08-26
possibility to pass flags to test_ and record_ makefile entries

pondrejk:
needed for harvesting junit.xml files from pytest

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -13,6 +13,7 @@ PYTHON_VERSION = $(shell python -c 'import sys; print("{}.{}".format(sys.version
 COLLECTION_COMMAND ?= ansible-galaxy
 SANITY_OPTS = --venv
 TEST =
+FLAGS =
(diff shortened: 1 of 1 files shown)


mdellweg (commented):
I would have expected to see this a the "test" target. Not so much at the "record_%".
Can you set the variable to empty? I think there is a strict mode where Make fails on undefined references.

pondrejk:
> I would have expected to see this a the "test" target. Not so much at the "record_%".
> Can you set the variable to empty? I think there is a strict mode where Make fails on undefined references.

Empty declaration added. I need this on the record for live mode

evgeni:
3.5 failure should be "fixed" by https://github.com/theforeman/foreman-ansible-modules/pull/937

evgeni:
"it's fine"

evgeni (approved):

PR #938, opened 2020-08-31
image module: fix quoting for search values

### LINKED ISSUE #927: Image module can only create but not update or delete vmware images
##### SUMMARY
It is currently not possible to update / modify or delete vmware images created by the Image module  

```
# rpm -q ansible-collection-theforeman-foreman
ansible-collection-theforeman-foreman-0.7.0
```

##### ISSUE TYPE
 - Bug Report

##### ANSIBLE VERSION
```
ansible 2.9.10
  configured module search path = [u'/root/.ansible/plugins/modules', u'/usr/share/ansible/plugins/modules']
  python version = 2.7.5 (default, Apr  2 2020, 13:16:51) [GCC 4.8.5 20150623 (Red Hat 4.8.5-39)]
```
##### KATELLO/FOREMAN VERSION
```
tfm-rubygem-katello-3.14.1
foreman-1.24.3
```

##### STEPS TO REPRODUCE
Generating an image on the first run works fine, a second run ends in failed state

```yaml
- hosts: all
  tasks:
  - name: define default variables
    set_fact:
      tfm_connection: &tfm_connection
      username: "admin"
      password: "secret"
      server_url: "https://localhost"
      validate_certs: "false"
  - name: centos 7 image creation in ESX1
    theforeman.foreman.image:
      <<: *tfm_connection
      state: present
      name: "CentOS 7"
      uuid: "502f263f-6f70-a568-d0c9-7e25f4993ae0"
      image_username: "ansible"
      operatingsystem: "CentOS 7"
      compute_resource: "ESX1"
      architecture: "x86_64"
      user_data: true
```

##### EXPECTED RESULTS
Running the task a second time would update the image

##### ACTUAL RESULTS
Running the task a second time throws an error

```
TASK [centos 7 image creation in ESX1] *************************************************************************************************************************************************************************************************** 
fatal: [localhost]: FAILED! => {"changed": false, "error": {"errors": {"name": ["has already been taken"], "uuid": ["has already been taken"]}, "full_messages": ["Name has already been taken", "UUID has already been taken"], "id": null}, "msg": "Error while performing create on images: 422 Client Error: Unprocessable Entity"}

PLAY RECAP *********************************************************************************************************************************************************************************************************************************** 
localhost : ok=2 changed=0 unreachable=0 failed=1 skipped=0 rescued=0 ignored=0
```

According to the foreman api docs POST should be used for creating an image and PUT to update it  
https://theforeman.org/plugins/katello/3.14/api/apidoc/v2/images/update.html

The error might be caused by the module using POST instead of PUT on both runs

Additionally changing `state: present` to `absent` does not delete the image

hlawatschek:
add quotes to search values (name, operatingsystem).
fixes #927

### DIFF
--- a/plugins/modules/image.py
+++ b/plugins/modules/image.py
@@ -125,7 +125,7 @@ def main():
         operatingsystem_id = module.lookup_entity('operatingsystem')['id']
         module.set_entity('entity', module.find_resource(
             'images',
-            search="name={0},operatingsystem={1}".format(module.foreman_params['name'], operatingsystem_id),
+            search='name="{0}",operatingsystem="{1}"'.format(module.foreman_params['name'], operatingsystem_id),
             params=scope,
             failsafe=True,
         ))
--- a/tests/test_playbooks/fixtures/image-0.yml
+++ b/tests/test_playbooks/fixtures/image-0.yml
@@ -2,352 +2,370 @@ interactions:
 - request:
     body: null
     headers:
-      Accept: [application/json;version=2]
-      Accept-Encoding: ['gzip, deflate']
-      Connection: [keep-alive]
-      User-Agent: ['apypie (https://github.com/Apipie/apypie)']
+      Accept:
+      - application/json;version=2
+      Accept-Encoding:
+      - gzip, deflate
+      Connection:
+      - keep-alive
+      User-Agent:
+      - apypie (https://github.com/Apipie/apypie)
     method: GET
     uri: https://foreman.example.org/api/status
   response:
-    body: {string: !!python/unicode '{"result":"ok","status":200,"version":"1.24.2","api_version":2}'}
+    body:
+      string: !!python/unicode '{"result":"ok","status":200,"version":"1.24.2","api_version":2}'
     headers:
-      cache-control: ['max-age=0, private, must-revalidate']
-      connection: [Keep-Alive]
-      content-length: ['63']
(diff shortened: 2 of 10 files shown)


plugins/modules/image.py
```diff
@@ -125,7 +125,7 @@ def main():
         operatingsystem_id = module.lookup_entity('operatingsystem')['id']
         module.set_entity('entity', module.find_resource(
             'images',
-            search="name={0},operatingsystem={1}".format(module.foreman_params['name'], operatingsystem_id),
+            search='name="{0}",operatingsystem="{1}"'.format(module.foreman_params['name'], operatingsystem_id),
```
evgeni, line 128: Technically, we don't have to quote the operatingsystem, as it's an integer already (operatingsystem_*id*). But I don't mind too much.

evgeni (approved):
Thanks @hlawatschek! Please tell if you want to change the OS quoting or keep it as is. We can merge then!

plugins/modules/image.py
hlawatschek: OK, for me it looks more consistent quoting both variables. If you don't mind, let's keep it.

evgeni: ACK, thanks!

evgeni:
The 3.5 failure is unrelated (and a workaround is posted in #937), merging.

PR #939, opened 2020-08-31
added option "caching_enabled" to vmware

IncredibleRichie:
Added option "caching_enabled" to vmware


plugins/modules/compute_resource.py
```diff
@@ -173,6 +173,7 @@
       - ACME
     provider: vmware
     provider_params:
+      caching_enabled: 0
```
evgeni: If it's `bool`, it's probably better to have the example say `caching_enabled: False`

plugins/modules/compute_resource.py
```diff
@@ -360,6 +362,7 @@ def main():
                 tenant=dict(),
                 app_ident=dict(),
                 datacenter=dict(),
+                caching_enabled=dict(type='bool'),
```
evgeni, line 369: You'll have to add that to the DOCUMENTATION string in the suboptions of `provider_params`:
https://github.com/theforeman/foreman-ansible-modules/blob/15db065de93584b561aec4bee3990aaf3341ac23/plugins/modules/compute_resource.py#L54-L58

See `use_v4` or `ssl_verify_peer` for examples how to do this :)

evgeni (changes requested):

plugins/modules/compute_resource.py
IncredibleRichie: You´re right. Applied

plugins/modules/compute_resource.py
IncredibleRichie: Added

mdellweg:
I believe the new process requires this to be escorted by a changelog entry.
You should find examples here: https://github.com/theforeman/foreman-ansible-modules/tree/develop/changelogs/fragments

@evgeni Is this even correct? Can we provide a contributing section in the docs?

evgeni:
Correct. And yes, I need to do docs and also a check in CI (I have a discussion about that running upstream currently)

evgeni:
Thanks @IncredibleRichie! The diff looks good to me!

As Matthias pointed out, this should get an changelog entry. Could you please add a file in `changelogs/fragments` (I'd suggest smth like `changelogs/fragments/939-compute_resource-caching_enabled.yaml`) that describes the change? The following is probably a good start:

```yaml
minor_changes:
  - compute_resource - added ``caching_enabled`` option for VMware compute resources
```

IncredibleRichie:
@evgeni The comment came just in time, i have added the suggestions to the fragment. Thanks for the link @mdellweg

evgeni (approved):

PR #941, opened 2020-09-03
manage entity parameters in a single API call

evgeni:
this obsoletes #868 and makes *all* entity parameters (but not Organization/Location, as those are taxonomies) being managed in a single call to the endpoint of the managed entity, not the special parameters endpoint

### DIFF
--- a/plugins/module_utils/foreman_helper.py
+++ b/plugins/module_utils/foreman_helper.py
@@ -184,8 +184,9 @@ class TaxonomyMixin(object):
 class ParametersMixin(object):
     def __init__(self, **kwargs):
         self.entity_name = kwargs.pop('entity_name', self.entity_name_from_class)
+        parameters_flat_name = getattr(self, "PARAMETERS_FLAT_NAME", None) or '{0}_parameters_attributes'.format(self.entity_name)
         foreman_spec = dict(
-            parameters=dict(type='list', elements='dict', options=parameter_ansible_spec, flat_name='{0}_parameters_attributes'.format(self.entity_name)),
+            parameters=dict(type='list', elements='dict', options=parameter_ansible_spec, flat_name=parameters_flat_name),
         )
         foreman_spec.update(kwargs.pop('foreman_spec', {}))
         super(ParametersMixin, self).__init__(foreman_spec=foreman_spec, **kwargs)
--- a/plugins/module_utils/foreman_helper.py
+++ b/plugins/module_utils/foreman_helper.py
@@ -11,6 +11,7 @@ __metaclass__ = type
 import hashlib
 import json
 import os
+import operator
 import re
 import time
 import traceback
@@ -46,6 +47,8 @@ parameter_foreman_spec = dict(
     value=dict(type='raw', required=True),
     parameter_type=dict(default='string', choices=['string', 'boolean', 'integer', 'real', 'array', 'hash', 'yaml', 'json']),
 )
+
+parameter_ansible_spec = {k: v for (k, v) in parameter_foreman_spec.items() if k != 'id'}
 
 _PLUGIN_RESOURCES = {
     'discovery': 'discovery_rules',
@@ -178,6 +181,27 @@ class TaxonomyMixin(object):
         super(TaxonomyMixin, self).__init__(foreman_spec=foreman_spec, **kwargs)
 
 
+class ParametersMixin(object):
+    def __init__(self, **kwargs):
+        self.entity_name = kwargs.pop('entity_name', self.entity_name_from_class)
+        foreman_spec = dict(
+            parameters=dict(type='list', elements='dict', options=parameter_ansible_spec, flat_name='{0}_parameters_attributes'.format(self.entity_name)),
+        )
+        foreman_spec.update(kwargs.pop('foreman_spec', {}))
+        super(ParametersMixin, self).__init__(foreman_spec=foreman_spec, **kwargs)
+
+    def run(self, **kwargs):
+        entity = self.lookup_entity('entity')
+        if not self.desired_absent:
+            if entity and 'parameters' in entity:
+                entity['parameters'] = parameters_list_to_str_list(entity['parameters'])
+            parameters = self.foreman_params.get('parameters')
+            if parameters is not None:
+                self.foreman_params['parameters'] = parameters_list_to_str_list(parameters)
+
+        return super(ParametersMixin, self).run(**kwargs)
+
+
 class NestedParametersMixin(object):
     def __init__(self, **kwargs):
         foreman_spec = dict(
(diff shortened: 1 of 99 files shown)


evgeni:
@wbclark this should fix the issue you had with the one user who needed hosts with params. **huge** thanks for all the investigation you did in #868 around this!

plugins/modules/operatingsystem.py
```diff
@@ -187,6 +188,8 @@ def main():
             ptables=dict(type='entity_list'),
             provisioning_templates=dict(type='entity_list'),
             password_hash=dict(choices=['MD5', 'SHA256', 'SHA512', 'Base64', 'Base64-Windows']),
+            # parameters is already in the HostMixin, but the flat_name detection does not work for operatingsystems
```
mdellweg:
```suggestion
            # parameters is already in the ParametersMixin, but the flat_name detection does not work for operatingsystems
```

plugins/module_utils/foreman_helper.py
```diff
@@ -178,6 +181,28 @@ def __init__(self, **kwargs):
...
+    def run(self, **kwargs):
+        entity = self.lookup_entity('entity')
+        if not self.desired_absent:
+            if entity and 'parameters' in entity:
+                entity['parameters'] = parameters_list_to_str_list(entity['parameters'])
+            parameters = self.foreman_params.get('parameters')
+            if parameters is not None:
+                self.foreman_params['parameters'] = parameters_list_to_str_list(parameters)
+
+        self.set_entity('entity', entity)
```
mdellweg: This looks superfluous. The lookup sets the entity, and from that on, we change it in place.

plugins/module_utils/foreman_helper.py
```diff
@@ -178,6 +181,28 @@ def __init__(self, **kwargs):
         super(TaxonomyMixin, self).__init__(foreman_spec=foreman_spec, **kwargs)
 
 
+class ParametersMixin(object):
+    def __init__(self, **kwargs):
+        self.entity_name = kwargs.pop('entity_name', self.entity_name_from_class)
+        foreman_spec = dict(
+            parameters=dict(type='list', elements='dict', options=parameter_ansible_spec, flat_name='{0}_parameters_attributes'.format(self.entity_name)),
```
mdellweg: `'{0}_parameters_attributes'` This naming convention is like 0.6 accurate?

how about
```python
parameters_flat_name = getattr(self, "PARAMETERS_FLAT_NAME") or '{0}_parameters_attributes'.format(self.entity_name)
```
And you can add that as a class constant when needed.

plugins/module_utils/foreman_helper.py
```diff
@@ -47,6 +48,8 @@
     parameter_type=dict(default='string', choices=['string', 'boolean', 'integer', 'real', 'array', 'hash', 'yaml', 'json']),
 )
 
+parameter_ansible_spec = {k: v for (k, v) in parameter_foreman_spec.items() if k != 'id'}
```
mdellweg, line 51: Is this what we used to call `foreman_spec_helper` for?

plugins/module_utils/foreman_helper.py
```diff
@@ -697,13 +721,23 @@ def _update_entity(self, resource, desired_entity, current_entity, params, forem
...
-                if to_native(current_entity.get(key)) != to_native(value):
-                    payload[key] = value
-            else:
-                if current_entity.get(key) != value:
-                    payload[key] = value
+            if foreman_type == 'str':
+                old_value = to_native(current_entity.get(key))
+                new_value = to_native(value)
+            elif foreman_type == 'list':
+                if value and isinstance(value[0], dict):
```
mdellweg: Is this something, the `foreman_spec` could tell us?

mdellweg (changes requested):
A piece of art...
I like, how the size of the fixtures drops with this. That makes me think this is actually good for performance.

plugins/module_utils/foreman_helper.py
evgeni: Kinda, but it's not filtering out the `id` (and I really don't want to have that, as the API returns it, but we never set it, so it would always trigger a change).

plugins/module_utils/foreman_helper.py
evgeni: you mean via `elements='dict'`? yeah, maybe, good idea.

plugins/module_utils/foreman_helper.py
mdellweg: My idea is that we can change other aspects of the parameters specification without the need to revisit the two modules overriding this.

plugins/module_utils/foreman_helper.py
evgeni: Well, no, not easily. *tries to wrap my head around that*

evgeni: So the `elements` that Ansible sees are not the same we send to the API. At least not for things like `entity_list` :(

plugins/module_utils/foreman_helper.py
```diff
@@ -725,16 +725,15 @@ def _update_entity(self, resource, desired_entity, current_entity, params, forem
...
-                    if 'name' in value[0]:
-                        sort_key = 'name'
-                    else:
-                        sort_key = list(value[0].keys())[0]
-                    new_value = sorted(value, key=operator.itemgetter(sort_key))
-                    old_value = sorted(current_entity.get(key), key=operator.itemgetter(sort_key))
+                old_value = to_native(old_value)
+                new_value = to_native(new_value)
+            elif foreman_type == 'list' and foreman_spec[key].get('elements') == 'dict' and value:
+                if 'name' in value[0]:
```
mdellweg, line 735: At this point, the list of repositories in the `content_view` module has been flattened into a list of `repository_ids` already.
But also i am wondering, why the `foreman_type` is not `entity_list`.

evgeni: because it's ansible type at this point, mhhh, maybe I need to revisit my logic once more

mdellweg: We should stop calling the variable `foreman_spec` in that case. It creates wrong expectations.
Is it `ansible_spec`, `parameter_spec` or `compiled_foreman_spec`?

evgeni: Ah, when I put `entity_list` there (or `entity`, any *foreman*-specific type), `_flatten_entity` will bail out when trying to parse that.

evgeni: so I think my old logic was better ;)

mdellweg: Better as in "It worked".
I'd still like to use the spec eventually to make that decision. Can you leave a comment or add an issue?

mdellweg (dismissed):

plugins/module_utils/foreman_helper.py
evgeni: I've flipped back to the old `isinstance` type check and added a comment!

mdellweg (approved):

PR #943, opened 2020-09-04
fall back to environment variables for credentials

### LINKED ISSUE #837: allow for environment fallback for url/user/pass

evgeni:
Fixes: #837

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -78,7 +78,7 @@ tests/test_playbooks/vars/server.yml:
 	@echo "For recording, please adjust $@ to match your reference server."
 
 dist-test: $(MANIFEST)
-	ansible -m $(NAMESPACE).$(NAME).organization -a "username=admin password=changeme server_url=https://foreman.example.test name=collectiontest" localhost | grep -q "Failed to connect to Foreman server"
+	FOREMAN_SERVER_URL=https://foreman.example.test ansible -m $(NAMESPACE).$(NAME).organization -a "username=admin password=changeme name=collectiontest" localhost | grep -q "Failed to connect to Foreman server.*foreman.example.test"
 	ansible-doc $(NAMESPACE).$(NAME).organization | grep -q "Manage Organization"
 
 $(MANIFEST): $(NAMESPACE)-$(NAME)-$(VERSION).tar.gz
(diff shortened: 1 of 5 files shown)


evgeni:
@willtome you had originally used `env` params in `DOCUMENTATION` in https://github.com/theforeman/foreman-ansible-modules/pull/866, but when I do this, `antsibull` fails to generate documentation with the following error:

```
4 validation errors for ModuleDocSchema
doc -> options -> password -> env
  extra fields not permitted (type=value_error.extra)
doc -> options -> server_url -> env
  extra fields not permitted (type=value_error.extra)
doc -> options -> username -> env
  extra fields not permitted (type=value_error.extra)
doc -> options -> validate_certs -> env
  extra fields not permitted (type=value_error.extra)
```

And I also didn't find `env` on https://docs.ansible.com/ansible/latest/dev_guide/developing_modules_documenting.html#documentation-block -- so I went with normal doc lines for now.

I also used slightly different variable names than the ones used in the inventory plugin, as I liked them to match the module parameter name. We could add both to the fallback if I read the docs right?

mdellweg (approved):

willtome:
@evgeni can the new variable names be added to the inventory plugin as well so that modules and inventory are consistent? I believe the `env` option in DOCUMENTATION only works for inventory plugins. It will accept a list of env var names. See https://github.com/ansible-collections/vmware/blob/main/plugins/inventory/vmware_vm_inventory.py as an example.

evgeni:
Thanks @willtome! The example was what I was missing.

I've now added the old names to the new ones, and the new ones to the old ones. So whatever the user sets, it will work :D

evgeni:
@willtome / @mdellweg would one of you mind ACK-ing the latest changes? :)

willtome:
@evgeni looks good to me!

mdellweg (approved):

PR #944, opened 2020-09-04
unify password and username examples

evgeni:


mdellweg (approved):
sed foo :medal_sports:

PR #950, opened 2020-09-11
sync inventory credential doc strings with the rest

evgeni:


mdellweg (changes requested):
I'm ok with making that change, but it is a breaking change , so it needs a changelog.

evgeni:
the default drop is breaking? yeah, somehow that's right. but that'd mean a 2.x which I'd like to avoid for now.

pondrejk (approved):

mdellweg (approved):

PR #951, opened 2020-09-11
make the callback plugin accept other URL env vars

evgeni:


plugins/callback/foreman.py
```diff
@@ -20,16 +20,20 @@
       - requests (python library)
     options:
       url:
-        description: URL to the Foreman server
+        description:
+          - URL of the Foreman server.
         env:
           - name: FOREMAN_URL
+          - name: FOREMAN_SERVER_URL
+          - name: FOREMAN_SERVER
```
mdellweg, line 28: This looks like we'd want all three combinations in the other places too, right?

evgeni: preferably not? but I can be convinced ;)

mdellweg: :cyclone: _convince_ :cyclone:

evgeni: very convincing tornado

mdellweg: for consistency!

mdellweg (approved):

PR #952, opened 2020-09-14
remove CVs from lifecycle environments before deleting them

evgeni:


mdellweg (approved):

sean797:
Shouldn't Katello automatically do this when the user deletes a CVV? Else wouldn't there be orphaned CCVs that the user cannot delete?

evgeni:
> Shouldn't Katello automatically do this when the user deletes a CVV? Else wouldn't there be orphaned CCVs that the user cannot delete?

I agree that Katello should just do this automatically, but I don't follow the "orphaned" part?

sean797:
Turns out Katello doesn't allow the removal, so no orphans \o/

sean797 (approved):

PR #953, opened 2020-09-14
scope image lookups in host module by the CR

evgeni:


evgeni:
see https://bugzilla.redhat.com/show_bug.cgi?id=1878693 for details

evgeni:
technically we can also scope images by OS and ARCH, but the `image` module doesn't either. only CR.

mdellweg (approved):

PR #955, opened 2020-09-15
external_usergroup: Rename the auth_source_ldap parameter to auth_source

Thulium-Drake:
I have renamed auth_source_ldap to auth_source as this also works with the module (tested against Foreman 2.0) and better covers the available options.

### DIFF
--- a/plugins/modules/external_usergroup.py
+++ b/plugins/modules/external_usergroup.py
@@ -88,7 +88,7 @@ def main():
         foreman_spec=dict(
             name=dict(required=True),
             usergroup=dict(required=True),
-            auth_source=dict(required=True, type='entity', flat_name='auth_source_id', resource_type='auth_sources'),
+            auth_source=dict(required=True, aliases=['auth_source_ldap'], type='entity', flat_name='auth_source_id', resource_type='auth_sources'),
         ),
     )
 
--- a/plugins/modules/external_usergroup.py
+++ b/plugins/modules/external_usergroup.py
@@ -39,7 +39,7 @@ options:
       - Name of the linked usergroup
     required: true
     type: str
-  auth_source_ldap:
+  auth_source:
     description:
       - Name of the authentication source to be used for this group
     required: true
@@ -53,7 +53,13 @@ EXAMPLES = '''
 - name: Create an external user group
   theforeman.foreman.external_usergroup:
     name: test
-    auth_source_ldap: "My LDAP server"
+    auth_source: "My LDAP server"
+    usergroup: "Internal Usergroup"
+    state: present
+- name: Link a group from FreeIPA
+  theforeman.foreman.external_usergroup:
+    name: ipa_users
+    auth_source: External
     usergroup: "Internal Usergroup"
     state: present
 '''
@@ -82,7 +88,7 @@ def main():
         foreman_spec=dict(
             name=dict(required=True),
             usergroup=dict(required=True),
-            auth_source_ldap=dict(required=True, type='entity', flat_name='auth_source_id', resource_type='auth_sources'),
+            auth_source=dict(required=True, type='entity', flat_name='auth_source_id', resource_type='auth_sources'),
         ),
     )
 
--- a/tests/test_playbooks/tasks/external_usergroup.yml
+++ b/tests/test_playbooks/tasks/external_usergroup.yml
@@ -9,7 +9,7 @@
     server_url: "{{ foreman_server_url }}"
     validate_certs: "{{ foreman_validate_certs }}"
     name: "{{ external_usergroup_name }}"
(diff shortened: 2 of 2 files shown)


evgeni:
This needs at least to retain the old name as an alias, as otherwise you'll break user playbooks

Thulium-Drake:
Yes, you are absolutely right, forgot about that :-)

Should be in now.

mdellweg:
This needs rerecording, and a changelog.
What do you mean with "better covers the available options"? Has it been renamed in foreman?

evgeni:
It shouldn't *need* re-recording, as the API interaction should be identical?

Thulium-Drake:
> It shouldn't _need_ re-recording, as the API interaction should be identical?

Correct, the API call still links an external usergroup to an existing internal usergroup

> This needs rerecording, and a changelog.
> What do you mean with "better covers the available options"? Has it been renamed in foreman?

No, but now it suggests only LDAP auth_sources can be used, while Foreman also has the External auth_source (those can be used with IPA/Kerberos authentication)

mdellweg:
I sense an underlying api change in foreman here. So we'd need a bit more information since when "Foreman also has the External auth_source". We don't want to break existing installations with reasonably current versions.

evgeni:
Okay, I think I know what's happening here (and also partially explains the problem you describe in #956).

I poked a Foreman I have here with plain LDAP auth, and when listing external groups of a group I have setup (`GET /api/v2/usergroups/1/external_usergroups), the reply looks like this:

```json
[
  {
    "id":2,
    "name":"devs",
    "auth_source_ldap":
      {"id":3,"type":"AuthSourceLdap","name":"LDAP"}
  }
]
```

See how it contains `auth_source_ldap` as the pointer to the LDAP group?

By changing the parameter name to `auth_source` you (accidentally) broke the conversion, and now the module can't properly compare the API reply with the generated data and tries to update things (failing the tests).

Thulium-Drake:
That I can't tell, from what I've seen it's more a naming thing on the Ansible module side then a change on the Foreman side that inspired this change.

It's been around for ages AFAIK, see procedure 8.6 on the link below, which comes from the Satellite documentation:

https://access.redhat.com/documentation/en-us/red_hat_satellite/6.2/html/server_administration_guide/sect-red_hat_satellite-server_administration_guide-creating_user_groups-configuring_external_user_groups

You should be able to replicate this on any Foreman installation as well. And besides configuring any LDAP source, you can also pick EXTERNAL as a source (provided IPA authentication is enabled with ```foreman-installer --foreman-ipa-authentication=true```)

Thulium-Drake:
> Okay, I think I know what's happening here (and also partially explains the problem you describe in #956).
> 
> I poked a Foreman I have here with plain LDAP auth, and when listing external groups of a group I have setup (`GET /api/v2/usergroups/1/external_usergroups), the reply looks like this:
> 
> ```json
> [
>   {
>     "id":2,
>     "name":"devs",
>     "auth_source_ldap":
>       {"id":3,"type":"AuthSourceLdap","name":"LDAP"}
>   }
> ]
> ```
> 
> See how it contains `auth_source_ldap` as the pointer to the LDAP group?
> 
> By changing the parameter name to `auth_source` you (accidentally) broke the conversion, and now the module can't properly compare the API reply with the generated data and tries to update things (failing the tests).

Ah, right-o.

I am developing a role that can configure an external group, coming from IPA (but can also be AD) and I first thought I couldn't use the External auth_source. Which prompted me to run the following playbook:

```
---
- hosts: 'foreman'
  gather_facts: false
  tasks:
    - theforeman.foreman.resource_info:
        username: admin
        password: changeme
        validate_certs: false
        organization: 'Default Organization'
        resource: auth_sources
        search: ''
      register: stuff
    - debug:
        msg: "{{ stuff }}"
```

Which shows (this is a fresh install, which I didn't configure for IPA just yet) there's also other sources then LDAP:

```
ok: [sat.infra.damen.lab] => 
  msg:
    changed: false
    failed: false
    resources:
    - created_at: 2020-09-16 09:42:14 UTC
      id: 1
      locations:
      - description: null
        id: 2
        name: Default Location
        title: Default Location
      name: Internal
      organizations:
      - description: null
        id: 1
        name: Default Organization
        title: Default Organization
      type: AuthSourceInternal  <<<<<< Right here
      updated_at: 2020-09-16 09:42:14 UTC
```

The ones I've seen so far are:
* AuthSourceInternal
* AuthSourceExternal
* AuthSourceLdap

evgeni:
Let's see if https://github.com/theforeman/foreman-ansible-modules/pull/959 is any better.

Thulium-Drake:
obsoleted by #959

PR #962, opened 2020-09-17
pcfe nitpick 20200917

pcfe:
387aa0c to ensure new visitors can hopefully not miss the examples

c8b23b8 is a note to users looking to replace katello-cvmanager

f3bf071 are two minor thing that the linter on my workstation was nagging at

feel free to only take 387aa0c (just seemed silly to make 3 separeate PRs)

### DIFF
--- a/README.md
+++ b/README.md
@@ -32,6 +32,7 @@ Starting with Ansible 2.7, Ansible only supports Python 2.7 and 3.5 (and higher)
 * `compute_resource` can leak sensitive data if used within a loop. According to [Ansible documentation](https://docs.ansible.com/ansible/latest/user_guide/playbooks_loops.html), using loop over Ansible resources can leak sensitive data. You can prevent this by using `no_log: yes` on the task.
   
   eg:
+
    ```yaml
    - name: Create compute resources
      compute_resource:
@@ -83,7 +84,7 @@ But be sure to check out the other plugins too, this collection can do much much
 
 ## Dependencies
 
-These dependencies are required for the Ansible controller, not the Foreman server. 
+These dependencies are required for the Ansible controller, not the Foreman server.
 
 * `PyYAML`
 * [`apypie`](https://pypi.org/project/apypie/)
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -66,9 +67,24 @@ The collection is also available as `ansible-collection-theforeman-foreman` from
 
 After installing the RPM, you can use the modules in the same way as when they are installed directly from Ansible Galaxy.
 
+## Using the modules
+
+Please see [the Plugin Index](https://theforeman.org/plugins/foreman-ansible-modules/) for the list of plugins in the theforeman.foreman collection.
```
evgeni, line 72: how is that different to what we have in https://github.com/theforeman/foreman-ansible-modules/blob/387aa0c23f3ab523b25a24e87047a55ea5566463/README.md#documentation? quoting:

```
A list of all modules and their documentation can be found at theforeman.org/plugins/foreman-ansible-modules.
```

README.md
```diff
@@ -66,9 +67,24 @@ The collection is also available as `ansible-collection-theforeman-foreman` from
...
+You'll find examples on the individual plugin pages.
+
+## Switching from katello-cvmanager
```
evgeni, line 75: I'd prefer not to have this so prominently here. But have  migration page in the docs might be a good idea.

README.md
pcfe: I know it's a repetition, I wanted to drive the point home since "how to use this" does came up from time to time when I pointed people to the main page.

README.md
pcfe: Agreed, a separate page would be nicer. Once #958 is done I can whip something up since I will permanently switch from it to the collection then.
Until we have one maybe keep the section? While it was never supported, a lot of Satellite admins I met used it (because it was so much more pleasant than them having to write shells scripts with hammer commands).

evgeni:
given we have cvmanager migration docs, and I don't want the docs repetition, I'm going to pick only the markdown fix and close out the rest :)

evgeni:
see https://github.com/theforeman/foreman-ansible-modules/pull/1102

PR #964, opened 2020-09-18
add status_info module

evgeni:


mdellweg (commented):
Very nice!
This deserves a changelog.
Also did the fixtures hack not work? The tests/test_playbooks/fixtures/status_info-0.yml seems a bit large.

evgeni:
New modules are automatically added to the changelog, no fragment needed.

I'll look into the fixtures.

evgeni:
Hah, so somehow it doesn't ignore the checksum on record and fetches it once. it just never showed up, as we always have a `localhost` task in the other modules.

plugins/modules/status_info.py
```diff
@@ -0,0 +1,76 @@
...
+# You should have received a copy of the GNU General Public License
+# along with this program.  If not, see <http://www.gnu.org/licenses/>.
+
+from __future__ import absolute_import, division, print_function
+__metaclass__ = type
+
+
+DOCUMENTATION = '''
+---
+module: status_info
```
evgeni, line 24: I still wonder if that should be `status` or `status_info`. It's not like it's the information of the status entity!?

evgeni: After talking to gundalow on IRC, if it never changes state, `_info`.

mdellweg (approved):

PR #967, opened 2020-09-21
add external ipam to subnet module

### LINKED ISSUE #966: Handle subnets thats use external IPAM like phpIPAM
We configure our whole foreman server via ansible and we use phpIPAM as external IP address management. Is there a way to configure subnet via ansible which use an external IPAM. We don't found any hints in documentation. On the contrary we found this for IPAM:

```
DHCP
Internal DB
Random DB
EUI-64
None
```

Where `DHCP` is the default. No idea whether the plugin (in this case smart_proxy_ipam) have to/should/can add the needed fields to API.

evgeni:
Fixes: #966

### DIFF
--- a/changelogs/fragments/966-subnet-external_ipam.yaml
+++ b/changelogs/fragments/966-subnet-external_ipam.yaml
@@ -1,0 +1,2 @@
+minor_changes:
+  - subnet - add support for external IPAM (https://github.com/theforeman/foreman-ansible-modules/issues/966)
--- a/plugins/modules/subnet.py
+++ b/plugins/modules/subnet.py
@@ -96,6 +96,7 @@ options:
       - "Internal DB"
       - "Random DB"
       - "EUI-64"
+      - "External IPAM"
       - "None"
     type: str
   dhcp_proxy:
@@ -131,6 +132,10 @@ options:
     required: false
     type: list
     elements: str
+  externalipam_proxy:
+    description: External IPAM proxy for this subnet
+    required: false
+    type: str
   vlanid:
     description: VLAN ID
     required: false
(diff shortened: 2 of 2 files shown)


mdellweg (approved):
I give it a :+1: codewise.
A confirmation it solves #966 would be good.

cmeissner:
@mdellweg not sure if the change is enough. If I select `External IPAM` I get an extra field `External IPAM Group`. This is not known with your code change i guess.

evgeni:
what can you select there? in the API, I only see "external ipam proxy" which is the smart proxy running the ipam I guess.

here is the api documentation I am looking at: https://theforeman.org/api/2.1/apidoc/v2/subnets/create.html

evgeni:
Reading the code, it seems the group is not exposed via the API :(

cmeissner:
Here you see a screenshot of the UI.

image

Not so good that the field isn't exposed. So the work is harder than expected, I guess.

evgeni:
@cmeissner how useful (or not?) is it to set the subnet to "external" but not being able to select a group? the UI says "optional", so I'd guess it's still a bit useful?

cmeissner:
For our use case we grouping dozen of ip networks to a given site in this groups aka sections in phpIPAM. So we unfortunately need this field. :-(

evgeni:
but in a simpler setup, this would work without, right?

I'll open an issue against Foreman to expose that via the API, but don't expect a too quick turn around.

cmeissner:
Maybe it would work but we can't test it this way. 

Thanks for opening the issue against Foreman but I am afraid that we need to find a workaround for that.

How long will it last that the above change is in a new ansible-modules version available? So we can set configure the subnets for external IPAM and its smart-proxy and only have to set the IPAM group for each by bio script.

evgeni:
I am planning a release (to galaxy) probably tomorrow.
Can you by any chance test before the release? Otherwise I'd merge what we have and you ping us when you installed the new version.

mdellweg:
Is it possible that the optional option is only missing from the apipie docs? (We had that before...)

Edit: Too sad, but i found nothing in the api controller.

evgeni:
@mdellweg no, the controller doesn't know about it.

I've opened https://projects.theforeman.org/issues/30890

cmeissner:
> Can you by any chance test before the release? Otherwise I'd merge what we have and you ping us when you installed the new version.

I can not promise anything, but I'll try it tomorrow morning to test it locally if I found time.

cmeissner:
@evgeni I've tested your change, it works as expected. 
Thanks for quick response.

As long as we don't use same subnets in different sections we can live without the group field otherwise we have to add the value by hand.

Can you give me a ping please if the new version is published to galaxy?

evgeni:
@cmeissner cool, thanks a ton for testing this!

I'll ping you on the release PR and you can expect it to land on galaxy something like 10 minutes after the PR is merged (yay, automation!)

PR #978, opened 2020-09-24
add a role to manage subscription manifest

wbclark:
The idea here is to have a simple role that can be used to upload a subscription manifest to an organization on Foreman.

This will be used as part of a larger effort to automate common post installation setup.


roles/manifest/defaults/main.yml
```diff
@@ -0,0 +1,22 @@
...
+
+# Foreman username and password
+foreman_username: admin
+foreman_password: changeme
+
+# Foreman server URL
+foreman_server_url: https://foreman.example.com
+
+# Foreman organization
+foreman_organization: "Default Organization"
```
ehelms: I'd like to consider introducing a role for common variables that is a dependency of other roles so as to define the variables in a common place. Naming is hard for such a role:

 * server_variables
 * api_variales
 * common_variables
 * ?!?!!

roles/manifest/tasks/upload_manifest.yml
```diff
@@ -0,0 +1,9 @@
+---
+- name: Upload Subscription Manifest to Foreman
+  theforeman.foreman.subscription_manifest:
```
ehelms: Are we able to avoid declaring the collection explicitly here to allow re-use across downstream collections?

ehelms:
Thanks for kick starting this @wbclark . Looking to @evgeni a bit here as to what level of testing we should introduce for roles.

mdellweg:
Not evgeni here, but it should be easy enough to drop a playbook in tests/test_playbooks for roles too. And they would get the same amount of vcr recording. Just be careful to not generate name collisions with existing tests.
So should the test-playbook be role_manifest.yml or manifest_role.yml?

mdellweg:
About testing, here you can see what it would be needed:

https://github.com/mdellweg/foreman-ansible-modules/tree/dummy_role

edit: Tests should fail there on purpose to show they actually run.

ehelms:
Do we want any use of native Ansible testing methods for roles? like molecule?

mdellweg:
> Do we want any use of native Ansible testing methods for roles? like molecule?

Not sure we can integrate them with vcr recording that easily. So they may need a "real" server running. Other than that i'd be happy to see molecule handle this.

roles/manifest/tasks/download_manifest.yml
```diff
@@ -0,0 +1,9 @@
+---
+- name: Download Subscription Manifest from Red Hat Customer Portal
+  get_url:
```
evgeni: we have a module for that :)

https://theforeman.github.io/foreman-ansible-modules/develop/plugins/redhat_manifest_module.html#ansible-collections-theforeman-foreman-redhat-manifest-module

roles/manifest/tasks/download_manifest.yml
```diff
@@ -0,0 +1,9 @@
...
+    url: https://subscription.rhsm.redhat.com/subscription/consumers/{{ manifest_uuid }}/export
+    username: "{{ rhsm_user }}"
+    password: "{{ rhsm_password }}"
+    dest: "{{ manifest_path }}"
+    validate_certs: false
```
evgeni: the module validates certs ;)

roles/manifest/tasks/upload_manifest.yml
```diff
@@ -0,0 +1,9 @@
...
+    username: "{{ foreman_username }}"
+    password: "{{ foreman_password }}"
+    server_url: "{{ foreman_server_url }}"
+    organization: "{{ foreman_organization }}"
```
evgeni: have a look at the discussion in https://github.com/theforeman/foreman-ansible-modules/pull/958 around naming common variables. we should agree on something common ;)

evgeni:
Yeah, I fear molecule, while really cool, will be hard with VCR.

roles/manifest/meta/main.yml
```diff
@@ -0,0 +1,3 @@
+---
+dependencies:
+  - role: common_variables
```
ehelms: I *think* cause remember is hard, we landed on not going the common_variables design route but instead opting to define the common variables through an example vars file e.g. `vars.yaml.example` or some such at the root of the repository. Possibly even documenting them in the README as well.

@evgeni can you weigh in what you recall

wbclark: Yes, it doesn't help that I immediately went out on vacation for most of the week right after this discussion -- I'm struggling to recall as well the exact final design that we landed on. I do recall that it was considered important to have a single file containing common variables to support ease of use.

ehelms: @mdellweg may also recall what we landed on specifically, I just recall we decided not to use a common role and to use this to include an example vars file

evgeni: Right…

So I think we landed on the following "rules":
- if there is no sensible default, don't add it to `defaults`, but document in `README.md` as required (see e.g. https://github.com/theforeman/foreman-ansible-modules/pull/958/files where I did that for the cleanup role)
  - it might be beneficial to have the *common* variables (also?) documented in a more prominent central place, but we don't have such one yet.
- if the variable is only useful for that specific role, prefix it with the role name
- if it's "global" (like username, or organization) no prefix

roles/manifest/README.md
```diff
@@ -0,0 +1,69 @@
+theforeman.foreman.manifest
+===============================================
+
+Upload Subscription Manifest to Katello
+
+Requirements
+------------
+
+- `theforeman.forman` collection
```
phess, line 9: typo alert. :-)   s/forman/foreman/  as fixed below:

```suggestion
- `theforeman.foreman` collection
```

phess: Typo alert. Fix below.

```suggestion
- `theforeman.foreman` collection
```

phess (commented):
Found a typo in line 9 of README.md.
s/forman/foreman/

roles/manifest/README.md
wbclark: Thanks Pablo!! I'll fix it shortly

roles/manifest/README.md
```diff
@@ -0,0 +1,69 @@
...
+
+Role Variables
+--------------
+
+- `server_url`: URL of the Foreman server.
+- `username`: Username accessing the Foreman server.
+- `password`: Password of the user accessing the Foreman server.
+- `validate_certs`: Whether or not to verify the TLS certificates of the Foreman server.
+- `organization`: Organization to upload Manifest to.
```
ehelms: Given this PR introduces them, could we add and document the variables above here also here -- https://github.com/theforeman/foreman-ansible-modules/blob/develop/README.md

roles/manifest/README.md
```diff
@@ -0,0 +1,69 @@
...
+- `manifest_path`: Where to find the Manifest on Foreman server. When using download_manifest, it is first downloaded to this location.
+- `download_manifest`: Whether to first download the Manifest from the Red Hat Customer Portal.
```
ehelms: I think this being specific to the role would be `manifest_download_manifest` or `manifest_download`

roles/manifest/README.md
```diff
@@ -0,0 +1,69 @@
...
+- `manifest_uuid`: UUID of the Manifest to download. Most correspond to a Subscription Allocation defined at https://access.redhat.com/management/subscription_allocations
+- `rhsm_username`: Your username for the Red Hat Customer Portal. Required when download_manifest is True.
+- `rhsm_password`: Your password for the Red Hat Customer Portal. Required when download_manifest is True.
```
ehelms: These two are probably also valid as common variables that can go in the top level README

roles/manifest/README.md
```diff
@@ -0,0 +1,69 @@
+theforeman.foreman.manifest
+===============================================
```
ehelms: Can we use the same markdown formatting at the top level README or is this style something common for Ansible roles?

roles/manifest/README.md
wbclark: What other role would use these, @ehelms ? (cloud_connector?)

roles/manifest/README.md
wbclark: :+1:

roles/manifest/README.md
wbclark: It's a common format that is created when doing `ansible-galaxy init rolename`:

```
$ ansible-galaxy init test-role-1
- Role test-role-1 was created successfully
$ cat test-role-1/README.md
```
Which creates the skeleton:

``` 
Role Name
=========

A brief description of the role goes here.

Requirements
------------

Any pre-requisites that may not be covered by Ansible itself or the role should be mentioned here. For instance, if the role uses the EC2 module, it may be a good idea to mention in this section that the boto package is required.

Role Variables
--------------

A description of the settable variables for this role should go here, including any variables that are in defaults/main.yml, vars/main.yml, and any variables that can/should be set via parameters to the role. Any variables that are read from other roles and/or the global scope (ie. hostvars, group vars, etc.) should be mentioned here as well.

Dependencies
------------

A list of other roles hosted on Galaxy should go here, plus any details in regards to parameters that may need to be set for other roles, or variables that are used from other roles.

Example Playbook
----------------

Including an example of how to use your role (for instance, with variables passed in as parameters) is always nice for users too:

    - hosts: servers
      roles:
         - { role: username.rolename, x: 42 }

License
-------

BSD

Author Information
------------------

An optional section for the role authors to include contact information, or a website (HTML is not allowed).
```

ehelms: Great, thank you for sharing that.

evgeni: FWIW, I don't think we need to adhere to that template by the word. Like license and author seems overly duplicated with other metadata we already have.

roles/manifest/README.md
```diff
@@ -0,0 +1,69 @@
+theforeman.foreman.manifest
+===========================
+
+Upload Subscription Manifest to Katello
```
ehelms: We could just opt for a `Upload a Subscription Manifest`

roles/manifest/README.md
```diff
@@ -0,0 +1,69 @@
...
+
+Requirements
+------------
+
+- `theforeman.foreman` collection
```
ehelms: Is this a requirement given this role is part of the collection?

roles/manifest/README.md
```diff
@@ -0,0 +1,69 @@
...
+
+Role Variables
+--------------
+
+- `server_url`: URL of the Foreman server.
+- `username`: Username accessing the Foreman server.
+- `password`: Password of the user accessing the Foreman server.
+- `validate_certs`: Whether or not to verify the TLS certificates of the Foreman server.
+- `organization`: Organization to upload Manifest to.
```
ehelms: Should we drop these now that they are global? More just enter a line that refers to the global README? I'm not sure, just trying to think and establish a pattern that reduces overhead and possibility of having outdated docs due to multiple places.

roles/manifest/README.md
```diff
@@ -0,0 +1,69 @@
...
+
+License
+-------
+
+GPLv3+
+
+Author Information
+------------------
+
+The Foreman Project
```
ehelms: I think i agree with @evgeni we could drop License and Author INfo here since they are covered at a collection level.

roles/manifest/README.md
ehelms: I'm thinking we could have a line with a link like:

```
This role supports the common role variables - (link to https://github.com/theforeman/foreman-ansible-modules/pull/978/files#diff-04c6e90faac2675aa89e2176d2eec7d8R87)

Comment (wbclark) [reply roles/manifest/README.md]: It's now updated with a link that should become working once the PR is merged.

**Review by ehelms — APPROVED** [review]

File: roles/manifest/tasks/main.yml [hunk 16]
```diff
@@ -0,0 +1,7 @@
+---
+- name: Download Manifest
+  import_tasks: download_manifest.yml
```
evgeni: What's the benefit of having multiple task files each containing one task?

roles/manifest/tasks/main.yml
```diff
@@ -0,0 +1,7 @@
...
+  when: manifest_download == True
```
evgeni: `== True` seems unnecessary?

Is ansible-lint running on the roles? It should have yelled at that.

roles/manifest/README.md
```diff
@@ -0,0 +1,51 @@
...
+
+Upload Subscription Manifest
+
+Role Variables
+--------------
+
+This role supports the [Common Role Variables](https://github.com/theforeman/foreman-ansible-modules/blob/develop/README.md#common-role-variables).
+
+- `manifest_path`: Where to find the Manifest on Foreman server. When using manifest_download, it is first downloaded to this location.
+- `manifest_download`: Whether to first download the Manifest from the Red Hat Customer Portal.
```
evgeni: Given the user needs to provide the UUID, isn't that being present enough of a marker that a download should happen?

roles/manifest/README.md
```diff
@@ -0,0 +1,51 @@
...
+===========================
+
+Upload Subscription Manifest
+
+Role Variables
+--------------
+
+This role supports the [Common Role Variables](https://github.com/theforeman/foreman-ansible-modules/blob/develop/README.md#common-role-variables).
+
+- `manifest_path`: Where to find the Manifest on Foreman server. When using manifest_download, it is first downloaded to this location.
```
evgeni: Not on the foreman server, but wherever the module is executed.

roles/manifest/README.md
```diff
@@ -0,0 +1,51 @@
...
+- `manifest_download`: Whether to first download the Manifest from the Red Hat Customer Portal.
+- `manifest_uuid`: UUID of the Manifest to download. Most correspond to a Subscription Allocation defined at https://access.redhat.com/management/subscription_allocations
```
evgeni: Can I also provide a name instead?

roles/manifest/README.md
```diff
@@ -0,0 +1,51 @@
+theforeman.foreman.manifest
+===========================
+
+Upload Subscription Manifest
+
+Role Variables
```
evgeni, line 6: Which of those are required and which optional?

roles/manifest/README.md
```diff
@@ -0,0 +1,51 @@
...
+- hosts: localhost
+  roles:
+    - role: theforeman.foreman.manifest
+      vars:
+        server_url: https://foreman.example.com
+        username: "admin"
+        password: "changeme"
+        organization: "Default Organization"
+        manifest_path: "/root/manifest.zip"
+        manifest_download: False
```
evgeni: The user should not have to set that, it should be a default.

roles/manifest/README.md
```diff
@@ -0,0 +1,51 @@
...
+- `manifest_path`: Where to find the Manifest on Foreman server. When using manifest_download, it is first downloaded to this location.
+- `manifest_download`: Whether to first download the Manifest from the Red Hat Customer Portal.
+- `manifest_uuid`: UUID of the Manifest to download. Most correspond to a Subscription Allocation defined at https://access.redhat.com/management/subscription_allocations
+- `rhsm_username`: Your username for the Red Hat Customer Portal. Required when manifest_download is True.
+- `rhsm_password`: Your password for the Red Hat Customer Portal. Required when manifest_download is True.
+
+Example Playbooks
+-----------------
+
+Use a Subscription Manifest which is already stored on Foreman server at `/root/manifest.zip`:
```
evgeni: As noted before the path is not on the foreman server

evgeni (changes requested):

roles/manifest/tasks/main.yml
ehelms: I rather like that. What's the downside?

evgeni: More output from Ansible. (As the include is an own task and reported as such)

roles/manifest/README.md
wbclark: Yes, internally. I think the price is that the behavior is less obvious to the user.

roles/manifest/tasks/main.yml
wbclark: Do you prefer a single file `tasks/main.yml` like:

```yaml
---
- name: Download Subscription Manifest from Red Hat Customer Portal
  theforeman.foreman.redhat_manifest:
    uuid: "{{ manifest_uuid }}"
    username: "{{ rhsm_username }}"
    password: "{{ rhsm_password }}"
    path: "{{ manifest_path }}"
  when: manifest_download

- name: Upload Subscription Manifest to Foreman
  theforeman.foreman.subscription_manifest:
    username: "{{ username }}"
    password: "{{ password }}"
    server_url: "{{ server_url }}"
    organization: "{{ organization }}"
    state: present
    manifest_path: "{{ manifest_path }}"
```

evgeni: Personally? Yes. I wouldn't block the PR on that, though. 

The only reasons I ticked "changes required" was the == True part (and the ansible lint run) and the path documentation.

ehelms: That's fair, single file makes sense to me!

wbclark:
Facing this issue when attempting to record tests:

```
fatal: [localhost]: FAILED! => {
    "changed": false,
    "invocation": {
        "module_args": {
            "server_url": "https://centos7-katello-nightly.mercury.example.com",
            "username": "admin",
            "validate_certs": false
        }
    },
    "msg": "Failed to connect to Foreman server: JSONDecodeError: Expecting value: line 1 column 1 (char 0)"
}
```

tests/test_playbooks/vars/server.yml.example
```diff
@@ -4,6 +4,7 @@ foreman_username: admin
 foreman_password: "changeme"
 foreman_server_url: "https://foreman.example.com"
 foreman_validate_certs: false
+foreman_organization: "Default Organization"
```
wbclark: @evgeni my understanding is that `make test-setup` will copy this to `tests/test_playbooks/vars/server.yml`.

but, do you prefer that I define this variable in my test playbook instead?

evgeni: We've been using "Test Organization" in other tests and hard coding there.

wbclark:
Facing:

`"msg": "Error while performing upload on subscriptions: Can't overwrite existing cassette ('fixtures/manifest_role-0.yml') in your current record mode ('none')`

The interesting thing, the issue appears to be the same even when I `git rm fixtures/manifest_role-0.yml` and re-record. Is there something I am missing here @evgeni ?

evgeni:
@wbclark so the problem is that your playbook/role generates a new manifest from the portal and uploads that.
but when the tests run, they try to upload the manifest that we have stored as a fixture, and given that is different, stuff explodes. let me see if we can fix that nicely.

tests/test_playbooks/vars/server.yml.example
evgeni: given you don't use that anymore, mind yanking that change?

tests/test_playbooks/manifest_role.yml
```diff
@@ -0,0 +1,32 @@
...
+- hosts: tests
+  collections:
+    - theforeman.foreman
+  gather_facts: false
+  vars_files:
+    - vars/server.yml
+  roles:
+    - role: manifest
+      vars:
+        manifest_download: false
```
evgeni, line 24: hmm, but you have download false... sure you don't have an "updated" manifest in https://github.com/theforeman/foreman-ansible-modules/tree/develop/tests/test_playbooks/data?

evgeni: oh, I think I know…

roles/manifest/README.md
```diff
@@ -0,0 +1,50 @@
...
+Role Variables
+--------------
+
+This role supports the [Common Role Variables](https://github.com/theforeman/foreman-ansible-modules/blob/develop/README.md#common-role-variables).
+
+- `manifest_path`: Path to subscription Manifest file on Ansible target host. When using manifest_download, it is first downloaded to this location from the Red Hat Customer Portal before being uploaded to Foreman server.
+- `manifest_download`: Whether to first download the Manifest from the Red Hat Customer Portal. Defaults to `False`.
+- `manifest_uuid`: UUID of the Manifest to download, corresponding to a [Subscription Allocation](https://access.redhat.com/management/subscription_allocations) defined on your Red Hat account. Required when manifest_download is True.
+- `rhsm_username`: Your username for the Red Hat Customer Portal. Required when manifest_download is True.
+- `rhsm_password`: Your password for the Red Hat Customer Portal. Required when manifest_download is True.
```
evgeni:
```suggestion
- `rhsm_username`: Your username for the Red Hat Customer Portal. Required when `manifest_download` is `True`.
- `rhsm_password`: Your password for the Red Hat Customer Portal. Required when `manifest_download` is `True`.
```

roles/manifest/README.md
```diff
@@ -0,0 +1,50 @@
...
+Upload Subscription Manifest
+
+Role Variables
+--------------
+
+This role supports the [Common Role Variables](https://github.com/theforeman/foreman-ansible-modules/blob/develop/README.md#common-role-variables).
+
+- `manifest_path`: Path to subscription Manifest file on Ansible target host. When using manifest_download, it is first downloaded to this location from the Red Hat Customer Portal before being uploaded to Foreman server.
+- `manifest_download`: Whether to first download the Manifest from the Red Hat Customer Portal. Defaults to `False`.
+- `manifest_uuid`: UUID of the Manifest to download, corresponding to a [Subscription Allocation](https://access.redhat.com/management/subscription_allocations) defined on your Red Hat account. Required when manifest_download is True.
```
evgeni:
```suggestion
- `manifest_uuid`: UUID of the Manifest to download, corresponding to a [Subscription Allocation](https://access.redhat.com/management/subscription_allocations) defined on your Red Hat account. Required when `manifest_download` is `True`.
```

roles/manifest/README.md
```diff
@@ -0,0 +1,50 @@
...
+===========================
+
+Upload Subscription Manifest
+
+Role Variables
+--------------
+
+This role supports the [Common Role Variables](https://github.com/theforeman/foreman-ansible-modules/blob/develop/README.md#common-role-variables).
+
+- `manifest_path`: Path to subscription Manifest file on Ansible target host. When using manifest_download, it is first downloaded to this location from the Red Hat Customer Portal before being uploaded to Foreman server.
```
evgeni:
```suggestion
- `manifest_path`: Path to subscription Manifest file on Ansible target host. When using `manifest_download`, it is first downloaded to this location from the Red Hat Customer Portal before being uploaded to the Foreman server.
```

roles/manifest/defaults/main.yml
```diff
@@ -0,0 +1,3 @@
+---
+manifest_download: false
+validate_certs: true
```
evgeni:
```suggestion
```

This is the default in the modules anyways.

roles/manifest/tasks/main.yml
```diff
@@ -0,0 +1,18 @@
...
+    password: "{{ rhsm_password }}"
+    path: "{{ manifest_path }}"
+  when: manifest_download
+
+- name: Upload Subscription Manifest to Foreman
+  theforeman.foreman.subscription_manifest:
+    username: "{{ username }}"
+    password: "{{ password }}"
+    server_url: "{{ server_url }}"
+    validate_certs: "{{ validate_certs }}"
```
evgeni:
```suggestion
    validate_certs: "{{ validate_certs | default(omit) }}"
```

tests/test_playbooks/manifest_role.yml
```diff
@@ -0,0 +1,32 @@
+---
+- hosts: localhost
+  collections:
+    - theforeman.foreman
+  gather_facts: false
+  vars_files:
+    - vars/server.yml
+  tasks:
+    - name: fetch status
+      include_tasks: tasks/status_info.yml
```
evgeni:
```suggestion
```

This is not needed if you have any other tasks running against localhost.

tests/test_playbooks/vars/server.yml.example
evgeni:
```suggestion
```

Makefile
```diff
@@ -41,7 +41,7 @@ info:
 	@echo " $(foreach PLUGIN_TYPE,$(PLUGIN_TYPES), $(PLUGIN_TYPE):\n $(foreach PLUGIN,$(basename $(notdir $(_$(PLUGIN_TYPE)))),   - $(PLUGIN)\n)\n)"
 
 lint: $(MANIFEST) | tests/test_playbooks/vars/server.yml
-	yamllint -f parsable tests/test_playbooks
+	yamllint -f parsable tests/test_playbooks roles
```
evgeni, line 44:
```suggestion
	yamllint -f parsable tests/test_playbooks roles
	ansible-lint -v roles/*
```

this might require adding `ansible-lint` to `requirements-lint.txt`

README.md
```diff
@@ -75,3 +75,19 @@ These dependencies are required for the Ansible controller, not the Foreman serv
...
 * `rpm` for the RPM support in the `content_upload` module
 * `debian` for the DEB support in the `content_upload` module
+
+# Foreman Ansible Roles
+
+Roles using the Foreman Ansible Modules to configure Foreman and its plugins.
+
+## Documentation
+
+For individual role documentation, check the README defined at `roles/rolename/README.md`.
```
evgeni, line 85: I'd make our doc generator make a nice overview page for that -- just remind me to actually do it once this is merged ;)

evgeni (changes requested):
a chunk of nitpicks, please have a look.

IMHO good to go after that.

evgeni:
#1040 contains a few changes that will conflict with this PR, so either one will need a revisit when the other is merged ;)

README.md
evgeni: actually, done.

wbclark:
Thanks @evgeni !

PR is updated. I greatly appreciate the troubleshooting and additional suggestions.

wbclark:
Tests are :green_apple: :green_apple: :green_apple: !

evgeni (approved):

Issue #788, opened 2020-04-09
add compute_attributes to foreman_host


evgeni:
##### SUMMARY
@lengoldenstein requested this on twitter: https://twitter.com/lengoldenstein/status/1247501504631549954

I tried adding it, but faced a few interesting bumps, which I don't know how to handle properly:
- foreman does not return these attributes on a normal "show" of a host and you need an additional API request to get them
- not all compute resources allow updating these for existing machines
- different CRs support different options here, so validating them is a tad hard (we'd have just to accept "everything")
- testing this requires a working compute resource, which is cumbersome (but that's only a developer concern)

So my proposal would be to add it without tests and without the idempotency test (so not polling the API endpoint) for now, and fix the fine details later.

cc @Fobhep as he was part of the discusion with Len

##### ISSUE TYPE
 - Feature Idea

Fobhep:
awesome - thanks for your efforts @evgeni :)
I agree with your proposal - seems to be the best way for now

lengoldenstein:
Hi - Thanks for looking in to this.  The equivalent functionality with 'hammer' looks like:
hammer -v host create \
--name "servername.org.com" \
--organization "ORG" \
--location "LOC1" \
--hostgroup "RHEL8-GP" \
--compute-resource "ssvsavcp8.org.com" \
--provision-method "bootdisk" \
--build true \
--ip "192.168.1.2" \
--subnet "Dev-10.33.118.0-EPG" \
--domain "org.com" \
--compute-attributes "cluster=SS-DEV-HXCluster01, cpus=2, corespersocket=1, memory_mb=8196, annotation='hammer poc', guest_id=rhel7_64Guest, path=Sandbox, scsi_controller_type=ParaVirtualSCSIController, cpuHotAddEnabled=1, memoryHotAddEnabled=1, firmware=bios, hardware_version=vmx-13, start=1" \
--interface "compute_type=VirtualVmxnet3, compute_network=Dev-10.33.118.0-EPG" \
--volume "datastore=HX01DataStore01, size_gb=1, thin=0, eager_zero=0" \
--volume "datastore=HX01DataStore01, size_gb=50, thin=0, eager_zero=0" \
--volume "datastore=HX01DataStore01, size_gb=10, thin=0, eager_zero=0" 

This gives a way to dynamically provision systems without using a static compute profile template.  Useful for very large environments.

evgeni:
[...]

PR #979, opened 2020-09-25
host: add interfaces_attributes parameter

### LINKED ISSUE #757: allow host interfaces management

achevalet:
I've just tested this and it looks to address a part of #757 (via the `host` API). I'm not sure but I think we could not make it idempotent, as for the compute_attributes.

### DIFF
--- a/plugins/module_utils/foreman_helper.py
+++ b/plugins/module_utils/foreman_helper.py
@@ -766,6 +766,11 @@ class ForemanAnsibleModule(AnsibleModule):
             # Already looked up or not an entity(_list) so nothing to do
             return self.foreman_params[key]
 
+        result = self._lookup_entity(self.foreman_params[key], entity_spec, params)
+        self.set_entity(key, result)
+        return result
+
+    def _lookup_entity(self, identifier, entity_spec, params=None):
         resource_type = entity_spec['resource_type']
         failsafe = entity_spec.get('failsafe', False)
         thin = entity_spec.get('thin', True)
@@ -782,26 +787,26 @@ class ForemanAnsibleModule(AnsibleModule):
                 if entity_spec.get('type') == 'entity':
                     result = None
                 else:
-                    result = [None for value in self.foreman_params[key]]
+                    result = [None for value in identifier]
             else:
                 self.fail_json(msg="Failed to lookup scope {0} while searching for {1}.".format(entity_spec['scope'], resource_type))
         else:
             # No exception happend => scope is in place
             if resource_type == 'operatingsystems':
                 if entity_spec.get('type') == 'entity':
-                    result = self.find_operatingsystem(self.foreman_params[key], params=params, failsafe=failsafe, thin=thin)
+                    result = self.find_operatingsystem(identifier, params=params, failsafe=failsafe, thin=thin)
                 else:
-                    result = [self.find_operatingsystem(value, params=params, failsafe=failsafe, thin=thin) for value in self.foreman_params[key]]
+                    result = [self.find_operatingsystem(value, params=params, failsafe=failsafe, thin=thin) for value in identifier]
             elif resource_type == 'puppetclasses':
                 if entity_spec.get('type') == 'entity':
-                    result = self.find_puppetclass(self.foreman_params[key], params=params, failsafe=failsafe, thin=thin)
+                    result = self.find_puppetclass(identifier, params=params, failsafe=failsafe, thin=thin)
                 else:
-                    result = [self.find_puppetclass(value, params=params, failsafe=failsafe, thin=thin) for value in self.foreman_params[key]]
+                    result = [self.find_puppetclass(value, params=params, failsafe=failsafe, thin=thin) for value in identifier]
             else:
                 if entity_spec.get('type') == 'entity':
                     result = self.find_resource_by(
                         resource=resource_type,
-                        value=self.foreman_params[key],
+                        value=identifier,
                         search_field=entity_spec.get('search_by', ENTITY_KEYS.get(resource_type, 'name')),
                         search_operator=entity_spec.get('search_operator', '='),
                         failsafe=failsafe, thin=thin, params=params,
@@ -813,16 +818,26 @@ class ForemanAnsibleModule(AnsibleModule):
                         search_field=entity_spec.get('search_by', ENTITY_KEYS.get(resource_type, 'name')),
                         search_operator=entity_spec.get('search_operator', '='),
                         failsafe=failsafe, thin=thin, params=params,
-                    ) for value in self.foreman_params[key]]
-        self.set_entity(key, result)
+                    ) for value in identifier]
         return result
 
     def auto_lookup_entities(self):
+        self.auto_lookup_nested_entities()
         return [
             self.lookup_entity(key)
             for key, entity_spec in self.foreman_spec.items()
             if entity_spec.get('resolve', True) and entity_spec.get('type') in {'entity', 'entity_list'}
         ]
+
+    def auto_lookup_nested_entities(self):
+        for key, entity_spec in self.foreman_spec.items():
+            if entity_spec.get('type') in {'nested_list'}:
+                for nested_key, nested_spec in self.foreman_spec[key]['foreman_spec'].items():
+                    for item in self.foreman_params.get(key, []):
+                        if (nested_key in item and nested_spec.get('resolve', True) and not nested_spec.get('resolved', False)
+                                and nested_spec.get('type') in {'entity', 'entity_list'}):
+                            item[nested_key] = self._lookup_entity(item[nested_key], nested_spec)
+                            nested_spec['resolved'] = True
 
     def record_before(self, resource, entity):
         self._before[resource].append(entity)
@@ -1451,7 +1466,7 @@ def _foreman_spec_helper(spec):
         elif foreman_type == 'nested_list':
             argument_value['type'] = 'list'
             argument_value['elements'] = 'dict'
-            _dummy, argument_value['options'] = _foreman_spec_helper(value['foreman_spec'])
+            foreman_value['foreman_spec'], argument_value['options'] = _foreman_spec_helper(value['foreman_spec'])
             foreman_value['ensure'] = value.get('ensure', False)
         elif foreman_type:
             argument_value['type'] = foreman_type
@@ -1489,6 +1504,8 @@ def _flatten_entity(entity, foreman_spec):
                     result[flat_name] = None
             elif property_type == 'entity_list':
                 result[flat_name] = sorted(val['id'] for val in value)
+            elif property_type == 'nested_list':
+                result[flat_name] = [_flatten_entity(ent, foreman_spec[key]['foreman_spec']) for ent in value]
             else:
                 result[flat_name] = value
     return result
@@ -1658,9 +1675,9 @@ interfaces_spec = dict(
     ip6=dict(),
     type=dict(choices=['interface', 'bmc', 'bond', 'bridge']),
     name=dict(),
-    subnet_id=dict(type='int'),
-    subnet6_id=dict(type='int'),
-    domain_id=dict(type='int'),
+    subnet=dict(type='entity'),
+    subnet6=dict(type='entity', resource_type='subnets'),
+    domain=dict(type='entity'),
     identifier=dict(),
     managed=dict(type='bool'),
     primary=dict(type='bool'),
@@ -1685,4 +1702,3 @@ interfaces_spec = dict(
     bond_options=dict(),
     compute_attributes=dict(type='dict'),
 )
-interfaces_foreman_spec, interfaces_ansible_spec = _foreman_spec_helper(interfaces_spec)
--- a/plugins/module_utils/foreman_helper.py
+++ b/plugins/module_utils/foreman_helper.py
@@ -1649,3 +1649,40 @@ TEMPLATE_KIND_LIST = [
     'user_data',
     'ZTP',
 ]
+
+# interface specs
+interfaces_spec = dict(
+    id=dict(invisible=True),
+    mac=dict(),
+    ip=dict(),
+    ip6=dict(),
+    type=dict(choices=['interface', 'bmc', 'bond', 'bridge']),
+    name=dict(),
+    subnet_id=dict(type='int'),
+    subnet6_id=dict(type='int'),
+    domain_id=dict(type='int'),
+    identifier=dict(),
+    managed=dict(type='bool'),
+    primary=dict(type='bool'),
+    provision=dict(type='bool'),
+    username=dict(),
+    password=dict(no_log=True),
+    provider=dict(choices=['IPMI', 'SSH']),
+    virtual=dict(type='bool'),
+    tag=dict(),
+    mtu=dict(type='int'),
+    attached_to=dict(),
+    mode=dict(choices=[
+        'balance-rr',
+        'active-backup',
+        'balance-xor',
+        'broadcast',
+        '802.3ad',
+        'balance-tlb',
+        'balance-alb',
+    ]),
+    attached_devices=dict(type='list', elements='str'),
+    bond_options=dict(),
+    compute_attributes=dict(type='dict'),
+)
+interfaces_foreman_spec, interfaces_ansible_spec = _foreman_spec_helper(interfaces_spec)
--- a/plugins/modules/host.py
+++ b/plugins/modules/host.py
@@ -150,18 +150,18 @@ options:
           - Interface's DNS name
           - You need to set one of I(identifier), I(name) or I(mac) to be able to update existing interfaces and make execution idempotent.
         type: str
-      subnet_id:
+      subnet:
         description:
-          - Foreman subnet ID of IPv4 interface
-        type: int
-      subnet6_id:
+          - Foreman subnet of IPv4 interface
+        type: str
+      subnet6:
         description:
-          - Foreman subnet ID of IPv6 interface
-        type: int
-      domain_id:
+          - Foreman subnet of IPv6 interface
+        type: str
+      domain:
         description:
-          - Foreman domain ID of interface. Required for primary interfaces on managed hosts.
-        type: int
+          - Foreman domain of interface. Required for primary interfaces on managed hosts.
+        type: str
       identifier:
         description:
           - Device identifier, e.g. eth0 or eth1.1
@@ -356,7 +356,7 @@ entity:
 
 from ansible_collections.theforeman.foreman.plugins.module_utils.foreman_helper import (
     ensure_puppetclasses,
-    interfaces_foreman_spec, interfaces_ansible_spec,
+    interfaces_spec,
     ForemanEntityAnsibleModule,
     HostMixin,
 )
@@ -395,13 +395,15 @@ def ensure_host_interfaces(module, entity, interfaces):
         updated_interface = (existing_interface or {}).copy()
         updated_interface.update(interface)
 
-        module.ensure_entity('interfaces', updated_interface, existing_interface, params=scope, state='present', foreman_spec=interfaces_foreman_spec)
+        module.ensure_entity('interfaces', updated_interface, existing_interface, params=scope, state='present',
+                             foreman_spec=module.foreman_spec['interfaces_attributes']['foreman_spec'])
 
         if 'id' in updated_interface:
             expected_interfaces_ids.add(updated_interface['id'])
 
     for leftover_interface in current_interfaces_ids - expected_interfaces_ids:
-        module.ensure_entity('interfaces', {}, {'id': leftover_interface}, params=scope, state='absent', foreman_spec=interfaces_foreman_spec)
+        module.ensure_entity('interfaces', {}, {'id': leftover_interface}, params=scope, state='absent',
+                             foreman_spec=module.foreman_spec['interfaces_attributes']['foreman_spec'])
 
 
 class ForemanHostModule(HostMixin, ForemanEntityAnsibleModule):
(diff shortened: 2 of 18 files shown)


mdellweg:
Thank you for addressing that.
Why is it not idempotent? Don't we have enough information to see whether something needs change? I agree that it might not "just work" with the tools we have in place, but not that it is generally impossible.
Also i'd like to see the foreman_spec to specify the subitems in the interface list.

achevalet:
I think that @evgeni 's comment in #788 also applies to interfaces_attributes:
- foreman does not return these attributes on a normal "show" of a host and you need an additional API request to get them
- not all compute resources allow updating these for existing machines
- different CRs support different options here, so validating them is a tad hard (we'd have just to accept "everything")

evgeni:
So `interfaces` are even worse than `compute_attributes`…

There is an API for managing interfaces directly (without the Host) as you can see in #792, but doing so we would have the same problem we had with parameters "if we do it after the host creation, it will not propagate to the CR for initial deployment" (see https://github.com/theforeman/foreman-ansible-modules/pull/941).

And then the API supports both, managing the interface as the entity in Foreman but also managing the NIC in a virtual system.

So I think what we could do here is to take the spec @Fobhep started in #792 and use the "ansible" part of it (as returned by `_foreman_spec_helper`) to at least validate the input from the user (c.f. https://theforeman.org/api/2.1/apidoc/v2/hosts/create.html that lists all the possible entries in that list) and then feed that to foreman.

And idempotency is then left as an exercise to the reader

mdellweg:
What if we added the interfaced in the same call for host creation, but in the update case we do it properly, separate and idempotent.
(host update does not have the problem that a host might be started without network.)

evgeni:
discussion during triage:
- if we extend this PR to describe the various possible settings that go into an interface (I can provide pointers how to do that later), we are good to merge this and then continue to make it idempotent as a second step
- only thing to make sure: if we provide a list of interfaces at creation and then update the same host with the same interfaces, it should not duplicate them (I am kinda sure it won't but having tried that out would be good)

evgeni:
example how to do a `list` that contains `dicts` with more strict content checks:
https://github.com/theforeman/foreman-ansible-modules/blob/04a98bc7a96aefc7bbb52058b961171de90bc4eb/plugins/modules/repository_set.py#L252-L255

also needs similar updates to documentation:
https://github.com/theforeman/foreman-ansible-modules/blob/04a98bc7a96aefc7bbb52058b961171de90bc4eb/plugins/modules/repository_set.py#L46-L63

achevalet:
@evgeni thanks for the pointers, I will amend the PR according to the apidoc.

achevalet:
@evgeni I've updated the foreman_spec but those 3 points remain unclear to me:

- all parameters defined in the foreman_spec are passed even if they are not specified (with `None` value), so I have filtered them here:
https://github.com/theforeman/foreman-ansible-modules/pull/979/files#diff-c1ec80744bf25b0075ff44bfa0b77cd0R448-R451
Is it correct to do it that way? 

- I've tried to use `entity` type for `subnet`, `subnet6` and `domain` here:
https://github.com/theforeman/foreman-ansible-modules/pull/979/files#diff-c1ec80744bf25b0075ff44bfa0b77cd0R390-R392
But I get this error: `implementation error: unknown type entity requested for ...`

- The update of an existing host fails with:
`Failed to ensure entity state: 'NoneType' object is not iterable`

Any help would be appreciated.

evgeni:
Thanks @achevalet!

For the first question: yes, this looks correct. We do a similar kind of filtering for the top-level parameters automatically here:
https://github.com/theforeman/foreman-ansible-modules/blob/bc64c255b739351dd9c88c0a186afcd1be05e08a/plugins/module_utils/foreman_helper.py#L320
But for sub options like in your case, the individual modules have been doing the filtering themself:
https://github.com/theforeman/foreman-ansible-modules/blob/bc64c255b739351dd9c88c0a186afcd1be05e08a/plugins/modules/repository_set.py#L271

I think extending the automation to sub options might be a good idea (:eyes: @mdellweg), but sounds like out-of-scope for the current PR.

For the second question: I guess you mean `type='entity'`, not `type='element'`? At least that matches the error you posted. Yes, this is currently not implemented for sub options either. I'll think how this can be made better.

And third: oof, I'll give it a try myself, not sure from just looking at the code.

plugins/modules/host.py
```diff
@@ -225,6 +381,39 @@ def main():
             provision_method=dict(choices=['build', 'image', 'bootdisk']),
             image=dict(type='entity', scope=['compute_resource']),
             compute_attributes=dict(type='dict'),
+            interfaces_attributes=dict(type='list', elements='dict', options=dict(
+                mac=dict(),
+                ip=dict(),
+                ip6=dict(),
+                type=dict(default='interface', choices=['interface', 'bmc', 'bond', 'bridge']),
```
evgeni: if I read https://theforeman.org/api/2.1/apidoc/v2/hosts/create.html correctly, the `interface` default is present server-side and we've been usually not setting it *additionally* as a default for such cases in the modules.
```suggestion
                type=dict(choices=['interface', 'bmc', 'bond', 'bridge']),
```

achevalet:
Sorry, yes I mean `type='entity'` (comment updated). 
Thanks for your help!

plugins/modules/host.py
```diff
@@ -225,6 +381,39 @@ def main():
...
+                mode=dict(choices=[
+                    'balance-rr',
+                    'active-backup',
+                    'balance-xor',
+                    'broadcast',
+                    '802.3ad',
+                    'balance-tlb',
+                    'balance-alb',
+                ]),
+                attached_devices=dict(type='list'),
```
evgeni: the sanity tests complain:
` Argument 'attached_devices' in argument_spec found in interfaces_attributes defines type as list but elements is not defined`

I think adding `elements='str'` here is sufficient

plugins/modules/host.py
```diff
@@ -256,6 +445,10 @@ def main():
         elif 'owner_group' in module.foreman_params:
             module.foreman_params['owner_type'] = 'Usergroup'
 
+        if 'interfaces_attributes' in module.foreman_params:
+          filtered = [nic for nic in ({k: v for k, v in obj.items() if v} for obj in module.foreman_params['interfaces_attributes']) if nic]
+          module.foreman_params['interfaces_attributes'] = filtered
```
evgeni:
```suggestion
            filtered = [nic for nic in ({k: v for k, v in obj.items() if v} for obj in module.foreman_params['interfaces_attributes']) if nic]
            module.foreman_params['interfaces_attributes'] = filtered
```

your indentation is off ;)

plugins/modules/host.py
```diff
@@ -116,6 +116,126 @@
...
+      type:
+        description:
+          - Interface type, e.g. bmc. Default is interface
+        type: str
+        choices:
+          - 'interface'
+          - 'bmc'
+          - 'bond'
+          - 'bridge'
+        default: interface
```
evgeni:
```suggestion
```

;)

plugins/modules/host.py
```diff
@@ -116,6 +116,126 @@
...
+        description:
+          - IPv4 address of interface
+        type: str
+      ip6:
+        description:
+          - IPv6 address of interface
+        type: str
+      type:
+        description:
+          - Interface type, e.g. bmc. Default is interface
```
evgeni:
```suggestion
          - Interface type.
```

plugins/modules/host.py
achevalet: indeed :)

evgeni:
Thanks @achevalet! I've finally found some time to look at this.

Creating a host with `interface_attributes` seem to work (I only tested identifier/ip/mac, but there is no reason why it should differ for the other params), but when I try to run the same playbook a second time, I get an error: "Identifier has already been taken" -- which is because it's trying to create *another* interface, with the same parameters. I'll have a look what can be done to avoid this.

Another thing (but unrelated to the comment above): for subnet, subnet6 and domain you've used the "id", which is correct for the API, but confusing for the user -- they usually want to put the name in there. We should make the module do the lookup for us before merging this.

achevalet:
Thanks the update.
On my side after rebase, I still have this error when I run the playbook a second time:
`Failed to ensure entity state: TypeError: 'NoneType' object is not iterable`
I'm using the minimum attributes (others are inherited from the compute profile):
```
        interfaces_attributes:
        - compute_attributes:
            network: "<my_network>"
```
Tested on ovirt and vmware, same error.

For the second point, should we update `entity` type to manage sub options?

evgeni:
So I took your work here and applied the logic we discussed with @mdellweg above: if we *create* a host, we pass the interfaces directly, but if we *update* it, we use the dedicated interfaces api.

the code is here: https://github.com/evgeni/foreman-ansible-modules/tree/interfaces -- if you want I can just push it to your fork, which will update this PR and we can discuss further.

this also avoid the `NoneType` error you mention.

evgeni:
also ping @Fobhep who tried to implement interfaces in #792 -- I honestly wonder if we really need a dedicated interfaces module?

achevalet:
Awesome! Yes please push :)

evgeni:
There we go.

cc @cmeissner if you want to try it out? :)

cmeissner:
> There we go.
> 
> cc @cmeissner if you want to try it out? :)

We give it a try tomorrow I think

Fobhep:
> also ping @Fobhep who tried to implement interfaces in #792 -- I honestly wonder if we really need a dedicated interfaces module?

I think the idea back then was to use a dedicated module for a dedicated resource and avoid the more tough way using the host module :)
However - thanks to you and @achevalet  this is probably not really needed anymore

evgeni:
> > also ping @Fobhep who tried to implement interfaces in #792 -- I honestly wonder if we really need a dedicated interfaces module?
> 
> I think the idea back then was to use a dedicated module for a dedicated resource and avoid the more tough way using the host module :)
> However - thanks to you and @achevalet  this is probably not really needed anymore
> 

Cool, so you'd say merging this would close #757 ans #792?

Oh and note to myself, this needs a changelog entry

plugins/modules/host.py
```diff
@@ -197,11 +353,44 @@
...
+                unique_identifier = possible_identifier
+                break
+        else:
+            unique_identifier = None
+            # TODO: this shouldn't happen
+
+        if 'mac' in interface:
+            interface['mac'] = interface['mac'].lower()
+
+        existing_interface = next((x for x in current_interfaces if x.get(unique_identifier) == interface[unique_identifier]), None)
```
achevalet:
```suggestion
        existing_interface = next((x for x in current_interfaces if unique_identifier and x.get(unique_identifier) == interface[unique_identifier]), None)
```

achevalet:
Now the update tries to remove the interface :)
```
fatal: [localhost]: FAILED! => {"changed": false, "error": {"errors": {"primary": ["can't delete primary interface of managed host"], "provision": ["can't delete provision interface of managed host"]}
```

Fobhep:
> > > also ping @Fobhep who tried to implement interfaces in #792 -- I honestly wonder if we really need a dedicated interfaces module?
> > 
> > 
> > I think the idea back then was to use a dedicated module for a dedicated resource and avoid the more tough way using the host module :)
> > However - thanks to you and @achevalet  this is probably not really needed anymore
> 
> Cool, so you'd say merging this would close #757 ans #792?
> 
> Oh and note to myself, this needs a changelog entry

Given the fact that our inital plan was anyways to port all features from a standalone module to the host module I would say so, yes. I don't see a lot of people using an extra module to handle interfaces, when really I can also do it via the host-module. But if someone disagrees, I will gladly be persuaded otherwise

plugins/modules/host.py
achevalet: note: the parameter `name` actually defines (or is defined by) the host's FQDN and `identifier` is optional.
If those are not specified, it ends up with an empty interface list.

evgeni: Yeah, so I needed to "find" an existing interface somehow to know if we're updating it or creating a new one. My idea was to look if there is an interface defined with the same identifier, hostname or mac (and I didn't add mac in my code, ups). This behavior is not documented yet.

achevalet: I think it would make sense for PXE deployment (`mac` is probably the best option since it is required) but for image based provisioning the changes won't be propagated to the VM since most compute resources does not manage host update (afaik, same applies to `compute_attributes`). Shouldn't we just exclude `interfaces_attributes` in this case?

evgeni: Is there *anything* that we could ask the user to provide to make the entry unique? If you for example use identifier:fam01 (like I did in the tests), would the interface be properly deployed with that name inside the OS, even if we don't define ip/mac to match on?

evgeni: So I had a few more thoughts:
* I'll look what hammer does tomorrow
* If both the user and API both provide exactly one interface, that's the one we're updating, no need to match
* In the case we couldn't find an unique identifier, proceed to add a new interface and warn the user that this is probably not what they want

achevalet: Sounds good! We just need to be careful with Foreman setting `ignore_puppet_facts_for_provisioning`, disabled by default. Puppet facts may update some values and/or add new interfaces. If new interfaces are created by puppet facts, we need to match, and if the identifier has changed, we might create a new interface at each run, which is not what we want :)

evgeni: I've implemented the "only one" and "warning" proposals from above.

Now to hammer: 
* Given the non-declarative way hammer is designed, you have individual commands to "add", "update" and "delete" host interfaces.
* Let's ignore adding, that is trivial.
* Deleting an interface requires two inputs: host id (or name) and interface id (which you need to look up manually via `hammer host interfaces list` before)
* Updating also (technically) requires host id/name and interface id, but nothing *enforces* the entry of the ID, and if you don't you get the good old "Interfaces some interfaces are invalid" error because it tries to add an interface with an identifier that's already present.

So, all in all, I'm no way wiser now.

cmeissner:
Works for me. Nice will be the lookup for entities e.g. `subnet` and `domain`.

evgeni:
okay, so I've pushed some changes that make subnet/domain lookup work, but it's not as pretty as I'd like it to be

ping @mdellweg for ideas

quick recap of our problem: currently, our lookup code only looks at entity and entity_list elements of the "main" spec, but in this module (and in some other modules before that), we actually have a list of hashes, and some parts of those hashes are entities. :exploding_head: I've now implemented the lookup specifically for the `interfaces_attributes`, but that specific thing makes it even worse, as we need to (yes, really: https://projects.theforeman.org/issues/31311) to submit the interfaces differently on create and update.

evgeni:
Maybe we can also fix the apidoc to accept the interfaces on host update too...

Edit: given that's what hammer does, it probably should work with a small apidoc patch

evgeni:
But even if we use "same api request", the lookup is needed the same way, so let's not overcomplicate things.

Or rather: I needed to apply enough other workarounds to make update work using the "same api request" so that it's not worth it.

plugins/module_utils/foreman_helper.py
```diff
@@ -771,6 +771,11 @@ def lookup_entity(self, key, params=None):
             # Already looked up or not an entity(_list) so nothing to do
             return self.foreman_params[key]
 
+        result = self._lookup_entity(self.foreman_params[key], entity_spec, params)
+        self.set_entity(key, result)
+        return result
+
+    def _lookup_entity(self, param, entity_spec, params=None):
```
mdellweg: I know naming is hard, but having `param` and `params` in the same parameter list doesn't look sane.

How about `identifier`?

plugins/modules/host.py
```diff
@@ -256,13 +453,41 @@ def main():
...
+                for interface in module.foreman_params['interfaces_attributes']:
+                    for key, entity_spec in interfaces_foreman_spec.items():
+                        if key in interface and entity_spec.get('resolve', True) and entity_spec.get('type') in {'entity', 'entity_list'}:
+                            interface[key] = module._lookup_entity(interface[key], entity_spec)
+                    if not entity:
+                        interface = _flatten_entity(interface, interfaces_foreman_spec)
+                    interfaces.append(interface)
+                module.foreman_params['interfaces_attributes'] = interfaces
+
+        if entity and 'interfaces_attributes' in module.foreman_params:
```
mdellweg, line 488: this logic is determining whether we add the interfaces at create or manipulate them after update?

I think this is worth a comment.

plugins/modules/host.py
```diff
@@ -225,6 +421,7 @@ def main():
             provision_method=dict(choices=['build', 'image', 'bootdisk']),
             image=dict(type='entity', scope=['compute_resource']),
             compute_attributes=dict(type='dict'),
+            interfaces_attributes=dict(type='list', elements='dict', options=interfaces_ansible_spec),
```
mdellweg: `entity_list` does not work here, because this is not a list of entities, but a list of dicts that contain entities?
Do we need a new foreman_spec type? Why is `nested_list` not the right type?
Maybe we are "just" missing the lookup facilities inside a `nested_list`.

mdellweg (commented):
Added my thoughts.

plugins/modules/host.py
evgeni: Yeah, `nested_list` might work here.

evgeni: done

plugins/modules/role.py
```diff
@@ -96,7 +96,7 @@
 
 filter_foreman_spec = dict(
     id=dict(invisible=True),
-    permissions=dict(type='entity_list', required=True),
+    permissions=dict(type='entity_list', required=True, resolve=False),
```
mdellweg, line 99: Interesting side effect.

plugins/modules/host.py
```diff
@@ -116,6 +116,126 @@
       - When this parameter is set, the module will not be idempotent.
     type: dict
     required: false
+  interfaces_attributes:
+    description:
+      - Additional interfaces specific attributes.
+      - When this parameter is set, the module will not be idempotent.
```
mdellweg: Can you elaborate why? Will modules like this be added in every run?

plugins/module_utils/foreman_helper.py
```diff
@@ -818,17 +823,27 @@ def lookup_entity(self, key, params=None):
...
+    def auto_lookup_nested_entities(self):
+        for key, entity_spec in self.foreman_spec.items():
+            if entity_spec.get('type') in {'nested_list'}:
+                for nested_key, nested_spec in self.foreman_spec[key]['foreman_spec'].items():
+                    for item in self.foreman_params.get(key, []):
+                        if (nested_key in item and nested_spec.get('resolve', True) and not nested_spec.get('resolved', False)
+                                and nested_spec.get('type') in {'entity', 'entity_list'}):
+                            item[nested_key] = self._lookup_entity(item[nested_key], nested_spec)
+                            nested_spec['resolved'] = True
+
```
mdellweg, line 841: This is not yet fully recursive, right?
Maybe we can incorporate the nested lookup into the `_lookup_entity` eventually, in a followup pr...

plugins/modules/host.py
```diff
@@ -197,11 +353,52 @@
...
+
+    current_interfaces = module.list_resource('interfaces', params=scope)
+    current_interfaces_ids = {x['id'] for x in current_interfaces}
+    expected_interfaces_ids = set()
+
+    for interface in interfaces:
+        if 1 == len(current_interfaces) == len(interfaces):
+            existing_interface = current_interfaces[0]
+        else:
+            for possible_identifier in ['identifier', 'name']:
```
mdellweg: would the mac address serve as an identifier, too?

plugins/modules/host.py
```diff
@@ -197,11 +353,52 @@
...
+                if possible_identifier in interface:
+                    unique_identifier = possible_identifier
+                    break
+            else:
+                unique_identifier = None
+                warning_msg = "The provided interface definition has no unique identifier and thus cannot be matched against existing interfaces. " \
+                    "This will always create a new interface and might not be the desired behaviour."
```
mdellweg, line 395: This is the reason for not being idempotent?
Then the docs up there could be "...be not idempotent unless ... is specified".

mdellweg (approved):
I left some comments. Nothing that should stop us from merging necessarily.
But consider the doc improvement.

plugins/module_utils/foreman_helper.py
```diff
@@ -1633,3 +1650,39 @@ def ensure_puppetclasses(module, entity_type, entity, expected_puppetclasses=Non
...
+interfaces_spec = dict(
+    id=dict(invisible=True),
+    mac=dict(),
+    ip=dict(),
+    ip6=dict(),
+    type=dict(choices=['interface', 'bmc', 'bond', 'bridge']),
+    name=dict(),
+    subnet=dict(type='entity'),
+    subnet6=dict(type='entity'),
+    domain=dict(type='entity', resource_type='subnets'),
```
cmeissner: wrong `resource_type`. Have to be `domains` instead I would guess.

cmeissner (changes requested):

plugins/modules/host.py
evgeni: Because I didn't update the docs ;)

It's only not idempotent if you set compute attrs! :)

plugins/modules/host.py
evgeni: yes, sure

plugins/module_utils/foreman_helper.py
evgeni: good catch!

plugins/module_utils/foreman_helper.py
evgeni: Yeah. It's not. But I'd also defer that to another PR :)

mdellweg (approved):

PR #981, opened 2020-09-26
Add Smart Proxy module

### LINKED ISSUE #149: smart-proxy - module doesn't exist

mdellweg:
Co-authored-by: Matthias Dellweg <2500@gmx.de>

This is a replacement for #314


plugins/modules/smart_proxy.py
```diff
@@ -0,0 +1,110 @@
+#!/usr/bin/python
+# -*- coding: utf-8 -*-
+#
+# This file is part of Ansible
```
evgeni: No it's not ;)

plugins/modules/smart_proxy.py
```diff
@@ -0,0 +1,110 @@
...
+    default: on_demand
+    required: false
+extends_documentation_fragment:
+  - theforeman.foreman.foreman
+  - theforeman.foreman.foreman.entity_state
+  - theforeman.foreman.foreman.taxonomy
+'''
+
+EXAMPLES = '''
+# Create a local Smart Proxy
```
evgeni, line 63: With Katello you can't create proxies like that (they need certs and stuff), so the examples and docs should probably reflect that oddity.

plugins/modules/smart_proxy.py
```diff
@@ -0,0 +1,110 @@
...
+short_description: Manage Smart Proxies
+description:
+  - Create, update and delete Smart Proxies
+author:
+  - "James Stuart (@jstuart)"
+  - "Matthias M Dellweg (@mdellweg)"
+options:
+  name:
+    description:
+      - Name or title of the Smart Proxy
```
evgeni: I think for our current implementation it's name only

plugins/modules/smart_proxy.py
```diff
@@ -0,0 +1,110 @@
...
+    required: true
+  download_policy:
+    description:
+      - The download policy for the Smart Proxy
+      - Only used for Katello installations.
+    choices:
+      - background
+      - immediate
+      - on_demand
+    default: on_demand
```
evgeni: It shouldn't have a default, as otherwise you'll confuse Foreman only installations or always error out as you use required plugins below

plugins/modules/smart_proxy.py
```diff
@@ -0,0 +1,110 @@
...
+extends_documentation_fragment:
+  - theforeman.foreman.foreman
+  - theforeman.foreman.foreman.entity_state
+  - theforeman.foreman.foreman.taxonomy
+'''
+
+EXAMPLES = '''
+# Create a local Smart Proxy
+- name: "Create Smart Proxy"
+  foreman_smart_proxy:
```
evgeni: theforeman.foreman.smart_proxy

plugins/modules/smart_proxy.py
mdellweg: I added a katello version of the test. Nothing special was needed. Can you elaborate?

plugins/modules/smart_proxy.py
mdellweg: Did you implement a bot for reviewing that line? Feels like a dejavu.

plugins/modules/smart_proxy.py
evgeni: Yeah, in tests that'll work, but the proxy won't work in a real world scenario as the proxy won't be set up properly server side

mdellweg: Is that preparation, postparation or missing features of this very module?

In the former two cases, i'd rather leave that as an exercise to the interested role writer.

evgeni: This very module won't be able to create a working Katello proxy on it's own (but can be certainly used to mimic parts ot the foreman-proxy-content scenario of the foreman installer using Ansible).

I think this module should just gain a notes section in the docs pointing out that Katello proxies should usually be created using the installer.

We could also warn/fail if users try to, but I don't think that's needed.

plugins/modules/smart_proxy.py
evgeni: No, no bots were harmed during this review

plugins/modules/smart_proxy.py
```diff
@@ -0,0 +1,110 @@
+#!/usr/bin/python
+# -*- coding: utf-8 -*-
+#
+# Ansible is free software: you can redistribute it and/or modify
```
evgeni: This is still talking about ansible and not 'the program' (or whatever the original GPL says).

evgeni:
In #314 we had discussed that for tests, one could use an alternative spelling/capitalization of the hostname, which should avoid the whole localhost hackery you mention in the header of the test file?

mdellweg:
> In #314 we had discussed that for tests, one could use an alternative spelling/capitalization of the hostname, which should avoid the whole localhost hackery you mention in the header of the test file?

I tried the adding-a-slash hackery, and it was able to create the proxy reference, but saved the url without the slash. First of all this breaks idempotence, and also i consider this a bug in foreman and therefore it might get fixed eventually.
Secondly, using "localhost" saves us a lot of trouble with vcr rewriting the request, and in this case even the response.

plugins/modules/smart_proxy.py
mdellweg: I added a note about using `foreman-installer` to create an actual smartproxy.
And imho, it is not even katello specific.

evgeni: It is (there is no foreman-proxy scenario in the installer, only foreman-proxy-content ;))

Thanks!

tests/test_playbooks/katello_smart_proxy.yml
```diff
@@ -0,0 +1,61 @@
...
+        expected_change: true
+    - include_tasks: tasks/smart_proxy.yml
+      vars:
+        smart_proxy_state: present
+        expected_change: false
+    - include_tasks: tasks/smart_proxy.yml
+      vars:
+        smart_proxy_download_policy: background
+        smart_proxy_state: present
+        expected_change: true
```
evgeni, line 41: can we have another invocation with `expected_change: false` please?

evgeni (changes requested):
a test nit-pick, otherwise this can go  in!

plugins/modules/smart_proxy.py
```diff
@@ -0,0 +1,114 @@
...
+    type: str
+  url:
+    description:
+      - URL of the Smart Proxy
+    required: true
+    type: str
+  download_policy:
+    description:
+      - The download policy for the Smart Proxy
+      - Only used for Katello installations.
```
evgeni:
```suggestion
      - Only available for Katello installations.
```

(to match the other strings we have)

mdellweg: I should not have taken it from hostgroup, where "used" is proper.

evgeni (approved):

PR #982, opened 2020-09-28
build docs from foreman_helper strings

evgeni:


evgeni:
the output is rather ugly, because our docstrings are not in a format Sphinx likes, but I thought it'd be a start

mdellweg (approved):

PR #983, opened 2020-09-28
update docstrings so that sphinx don't throw warnings anymore

evgeni:
it's still not perfect, but much better

### DIFF
--- a/plugins/module_utils/foreman_helper.py
+++ b/plugins/module_utils/foreman_helper.py
@@ -302,7 +302,9 @@ class ForemanAnsibleModule(AnsibleModule):
     """ Baseclass for all foreman related Ansible modules.
         It handles connection parameters and adds the concept of the `foreman_spec`.
         This adds automatic entities resolution based on provided attributes/ sub entities options.
+
         It adds the following options to foreman_spec 'entity' and 'entity_list' types:
+
         * search_by (str): Field used to search the sub entity. Defaults to 'name' unless `parent` was set, in which case it defaults to `title`.
         * search_operator (str): Operator used to search the sub entity. Defaults to '='. For fuzzy search use '~'.
         * resource_type (str): Resource type used to build API resource PATH. Defaults to pluralized entity key.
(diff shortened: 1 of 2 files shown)


plugins/module_utils/foreman_helper.py
```diff
@@ -717,15 +730,20 @@ def _validate_supported_payload(self, resource, action, payload):
...
     def _create_entity(self, resource, desired_entity, params, foreman_spec):
-        """Create entity with given properties
-
-            Parameters:
-                resource (string): Plural name of the api resource to manipulate
-                desired_entity (dict): Desired properties of the entity
-                params (dict): Lookup parameters (i.e. parent_id for nested entities) (optional)
-                foreman_spec (dict): Description of the entity structure
-            Return value:
-                The new current state if the entity
```
evgeni, line 728: I took the liberty to fix that typo XD

plugins/module_utils/foreman_helper.py
```diff
@@ -809,29 +839,37 @@ def _copy_entity(self, resource, desired_entity, current_entity, params):
...
+
+        :param resource: Plural name of the api resource to manipulate
+        :type resource: str
+        :param current_entity: Current properties of the entity
+        :type current_entity: dict
+        :param params: Lookup parameters (i.e. parent_id for nested entities)
+        :type params: dict, optional
+
+        :return: The new current state of the entity
+        :rtype: dict
```
evgeni: this is IMHO wrong, the method returns `None`

mdellweg: With `None` being "the new state of the entity", so it is technically `NoneType`, but we always meant `Union(NoneType, dict)`.

mdellweg (approved):

plugins/module_utils/foreman_helper.py
evgeni: So that's also what https://github.com/sphinx-doc/sphinx/issues/1874 recommends and I've implemented it now ;)

PR #986, opened 2020-09-28
Update inventory to consume reports API

sjha4:
Opening this for feedback and testing. There are some differences I would like feedback on. The grouping seems to be incomplete in the original script. :confused: 
Host_api_inventory:
https://codebeautify.org/jsonviewer/cb3c2148
Report_api_inventory:
https://codebeautify.org/jsonviewer/cb2ec15f

### DIFF
--- a/changelogs/fragments/inventory_plugin_report_api.yml
+++ b/changelogs/fragments/inventory_plugin_report_api.yml
@@ -1,0 +1,2 @@
+minor_changes:
+   - inventory plugin - add support for the Report API present in Foreman 1.24 and later
--- a/plugins/inventory/foreman.py
+++ b/plugins/inventory/foreman.py
@@ -83,7 +83,7 @@ DOCUMENTATION = '''
         type: string
       batch_size:
         description: Number of hosts per batch that will be retrieved from the Foreman API per individual call
-        type: integer
+        type: int
         default: 250
       foreman:
         description: foreman server related configuration
@@ -97,7 +97,7 @@ DOCUMENTATION = '''
         type: dict
         poll_interval:
             description: The polling interval between 2 calls to the report_data endpoint while polling.
-            type: integer
+            type: int
             default: 10
         want_organization:
             description: Toggle, if true the inventory will fetch organization the host belongs to and create groupings for the same.
@@ -299,18 +299,23 @@ class InventoryModule(BaseInventoryPlugin, Cacheable, Constructable):
     def _fetch_params(self):
         options = ("no", "yes")
         params = dict()
-        self.want_location = self.get_option('report').get('want_location')
-        self.want_organization = self.get_option('report').get('want_organization')
-        self.want_IPv4 = self.get_option('report').get('want_ipv4')
-        self.want_IPv6 = self.get_option('report').get('want_ipv6')
-        self.want_facts = self.get_option('report').get('want_facts')
-        self.want_host_group = self.get_option('report').get('want_host_group')
-        self.want_hostcollections = self.get_option('report').get('want_hostcollections')
-        self.want_subnet = self.get_option('report').get('want_subnet')
-        self.want_subnet_v6 = self.get_option('report').get('want_subnet_v6')
-        self.want_smart_proxies = self.get_option('report').get('want_smart_proxies')
-        self.want_content_facet_attributes = self.get_option('report').get('want_content_facet_attributes')
-        self.want_host_params = self.get_option('report').get('want_host_params')
+        report_options = {'want_location': True, 'want_organization': True, 'want_ipv4': True, 'want_ipv6': True,
+                          'want_host_group': True, 'want_hostcollections': False, 'want_subnet': True, 'want_subnet_v6': True, 'want_smart_proxies': True,
+                          'want_content_facet_attributes': False, 'want_host_params': True}
+        report_options.update(self.get_option('report') or {})
+
+        self.want_location = report_options['want_location']
+        self.want_organization = report_options['want_organization']
+        self.want_IPv4 = report_options['want_ipv4']
+        self.want_IPv6 = report_options['want_ipv6']
+        self.want_host_group = report_options['want_host_group']
+        self.want_hostcollections = report_options['want_hostcollections']
+        self.want_subnet = report_options['want_subnet']
+        self.want_subnet_v6 = report_options['want_subnet_v6']
+        self.want_smart_proxies = report_options['want_smart_proxies']
+        self.want_content_facet_attributes = report_options['want_content_facet_attributes']
+        self.want_host_params = report_options['want_host_params']
+        self.want_facts = self.get_option('want_facts')
         self.host_filters = self.get_option('host_filters')
 
         params["Organization"] = options[self.want_organization]
@@ -330,7 +335,10 @@ class InventoryModule(BaseInventoryPlugin, Cacheable, Constructable):
         return params
 
     def _use_inventory_report(self):
-        use_inventory_report = self.get_option('foreman').get('use_reports_api')
+        try:
+            use_inventory_report = self.get_option('foreman').get('use_reports_api')
+        except Exception:
+            use_inventory_report = False
         if not use_inventory_report:
             return False
         status_url = "%s/api/v2/status" % self.foreman_url
@@ -342,7 +350,10 @@ class InventoryModule(BaseInventoryPlugin, Cacheable, Constructable):
         url = "%s/ansible/api/v2/ansible_inventories/schedule" % self.foreman_url
         session = self._get_session()
         params = {'input_values': self._fetch_params()}
-        self.poll_interval = self.get_option('report').get('poll_interval')
+        try:
+            self.poll_interval = self.get_option('report').get('poll_interval')
+        except Exception:
+            self.poll_interval = 10
         ret = session.post(url, json=params)
         if not ret:
             raise Exception("Error scheduling inventory report on foreman. Please check foreman logs!")
--- a/plugins/inventory/foreman.py
+++ b/plugins/inventory/foreman.py
@@ -99,6 +99,10 @@ DOCUMENTATION = '''
             description: The polling interval between 2 calls to the report_data endpoint while polling.
             type: int
             default: 10
+        max_timeout:
+            description: Timeout before falling back to old host API when using report_data endpoint while polling.
+            type: int
+            default: 600
         want_organization:
             description: Toggle, if true the inventory will fetch organization the host belongs to and create groupings for the same.
             type: boolean
@@ -133,10 +137,6 @@ DOCUMENTATION = '''
             default: True
         want_content_facet_attributes:
             description: Toggle, if true the inventory will fetch content view details that the host is tied to.
-            type: boolean
-            default: True
-        want_host_params:
-            description: Toggle, if true the inventory will fetch host parameters.
             type: boolean
             default: True
         want_hostcollections:
@@ -301,7 +301,7 @@ class InventoryModule(BaseInventoryPlugin, Cacheable, Constructable):
         params = dict()
         report_options = {'want_location': True, 'want_organization': True, 'want_ipv4': True, 'want_ipv6': True,
                           'want_host_group': True, 'want_hostcollections': False, 'want_subnet': True, 'want_subnet_v6': True, 'want_smart_proxies': True,
-                          'want_content_facet_attributes': False, 'want_host_params': True}
+                          'want_content_facet_attributes': False}
         report_options.update(self.get_option('report') or {})
 
         self.want_location = report_options['want_location']
@@ -314,7 +314,7 @@ class InventoryModule(BaseInventoryPlugin, Cacheable, Constructable):
         self.want_subnet_v6 = report_options['want_subnet_v6']
         self.want_smart_proxies = report_options['want_smart_proxies']
         self.want_content_facet_attributes = report_options['want_content_facet_attributes']
-        self.want_host_params = report_options['want_host_params']
+        self.want_params = self.get_option('want_params')
         self.want_facts = self.get_option('want_facts')
         self.host_filters = self.get_option('host_filters')
 
@@ -329,7 +329,7 @@ class InventoryModule(BaseInventoryPlugin, Cacheable, Constructable):
         params["Subnet v6"] = options[self.want_subnet_v6]
         params["Smart Proxies"] = options[self.want_smart_proxies]
         params["Content Attributes"] = options[self.want_content_facet_attributes]
-        params["Host Parameters"] = options[self.want_host_params]
+        params["Host Parameters"] = options[self.want_params]
         if self.host_filters:
             params["Hosts"] = self.host_filters
         return params
@@ -351,21 +351,28 @@ class InventoryModule(BaseInventoryPlugin, Cacheable, Constructable):
         session = self._get_session()
         params = {'input_values': self._fetch_params()}
         try:
-            self.poll_interval = self.get_option('report').get('poll_interval')
+            self.poll_interval = int(self.get_option('report').get('poll_interval'))
+            self.max_timeout = int(self.get_option('report').get('max_timeout'))
+            max_polls = int(self.max_timeout / self.poll_interval)
         except Exception:
             self.poll_interval = 10
+            max_polls = 60
         ret = session.post(url, json=params)
         if not ret:
             raise Exception("Error scheduling inventory report on foreman. Please check foreman logs!")
         url = "{0}/{1}".format(self.foreman_url, ret.json().get('data_url'))
+        polls = 0
         response = session.get(url)
         while response:
-            if response.status_code != 204:
+            if response.status_code != 204 or polls > max_polls:
                 break
             sleep(self.poll_interval)
+            polls += 1
             response = session.get(url)
         if not response:
             raise Exception("Error receiving inventory report from foreman. Please check foreman logs!")
+        elif (response.status_code == 204 and polls > max_polls):
+            raise Exception("Timeout receiving inventory report from foreman. Check foreman server and max_timeout in foreman.yml")
         else:
             return response.json()
 
@@ -380,7 +387,8 @@ class InventoryModule(BaseInventoryPlugin, Cacheable, Constructable):
         self.hosts = dict()
         try:
             inventory_report_response = self._post_request()
-        except Exception:
+        except Exception as exc:
+            self.display.warning("Failed to use Reports API, falling back to Hosts API: {0}".format(exc))
             self._populate_host_api()
             return
         self.group_prefix = self.get_option('group_prefix')
@@ -395,6 +403,9 @@ class InventoryModule(BaseInventoryPlugin, Cacheable, Constructable):
                 group_name = self.inventory.add_group(group_name)
                 self.inventory.add_child(group_name, host_name)
 
+            host_params = host.pop('host_parameters', {})
+            fact_list = host.pop('facts', {})
+
             if self.get_option('legacy_hostvars'):
                 hostvars = self._get_hostvars(host)
                 self.inventory.set_variable(host_name, 'foreman', hostvars)
@@ -407,8 +418,7 @@ class InventoryModule(BaseInventoryPlugin, Cacheable, Constructable):
                         self.inventory.set_variable(host_name, k, v)
                     except ValueError as e:
                         self.display.warning("Could not set host info hostvar for %s, skipping %s: %s" % (host, k, to_text(e)))
-            host_params = host.pop('host_parameters', {})
-            fact_list = host.pop('facts', {})
+
             content_facet_attributes = host.get('content_attributes', {}) or {}
             if self.get_option('want_facts'):
                 self.inventory.set_variable(host_name, 'foreman_facts', fact_list)
@@ -469,22 +479,16 @@ class InventoryModule(BaseInventoryPlugin, Cacheable, Constructable):
                             self.display.warning("Could not create groups for host collections for %s, skipping: %s" % (host_name, to_text(e)))
 
             # set host vars from params
-                if self.get_option('want_params'):
-                    filtered_params = {}
-                    for p in params:
-                        if 'name' in p and 'value' in p:
-                            filtered_params[p['name']] = p['value']
-
-                    if self.get_option('legacy_hostvars'):
-                        self.inventory.set_variable(host_name, 'foreman_params', filtered_params)
-                    else:
-                        for k, v in filtered_params.items():
-                            try:
-                                self.inventory.set_variable(host_name, k, v)
-                            except ValueError as e:
-                                self.display.warning("Could not set hostvar %s to '%s' for the '%s' host, skipping:  %s" %
-                                                     (k, to_native(v), host, to_native(e)))
-
+            if self.get_option('want_params'):
+                if self.get_option('legacy_hostvars'):
+                    self.inventory.set_variable(host_name, 'foreman_params', params)
+                else:
+                    for k, v in params.items():
+                        try:
+                            self.inventory.set_variable(host_name, k, v)
+                        except ValueError as e:
+                            self.display.warning("Could not set hostvar %s to '%s' for the '%s' host, skipping:  %s" %
+                                                 (k, to_native(v), host, to_native(e)))
             strict = self.get_option('strict')
             hostvars = self.inventory.get_host(host_name).get_vars()
             self._set_composite_vars(self.get_option('compose'), hostvars, host_name, strict)
--- a/plugins/inventory/foreman.py
+++ b/plugins/inventory/foreman.py
@@ -85,6 +85,64 @@ DOCUMENTATION = '''
         description: Number of hosts per batch that will be retrieved from the Foreman API per individual call
         type: integer
         default: 250
+      foreman:
+        description: foreman server related configuration
+        type: dict
+        use_reports_api:
+            description: Use Reporting API.
+            type: boolean
+            default: False
+      report:
+        description: report API specific configuration
+        type: dict
+        poll_interval:
+            description: The polling interval between 2 calls to the report_data endpoint while polling.
+            type: integer
+            default: 10
+        want_organization:
+            description: Toggle, if true the inventory will fetch organization the host belongs to and create groupings for the same.
+            type: boolean
+            default: True
+        want_location:
+            description: Toggle, if true the inventory will fetch location the host belongs to and create groupings for the same.
+            type: boolean
+            default: True
+        want_ipv4:
+            description: Toggle, if true the inventory will fetch ipv4 address of the host.
+            type: boolean
+            default: True
+        want_ipv6:
+            description: Toggle, if true the inventory will fetch ipv6 address of the host.
+            type: boolean
+            default: True
+        want_host_group:
+            description: Toggle, if true the inventory will fetch host_groups and create groupings for the same.
+            type: boolean
+            default: True
+        want_subnet:
+            description: Toggle, if true the inventory will fetch subnet.
+            type: boolean
+            default: True
+        want_subnet_v6:
+            description:  Toggle, if true the inventory will fetch ipv6 subnet.
+            type: boolean
+            default: True
+        want_smart_proxies:
+            description: Toggle, if true the inventory will fetch smart proxy that the host is registered to.
+            type: boolean
+            default: True
+        want_content_facet_attributes:
+            description: Toggle, if true the inventory will fetch content view details that the host is tied to.
+            type: boolean
+            default: True
+        want_host_params:
+            description: Toggle, if true the inventory will fetch host parameters.
+            type: boolean
+            default: True
+        want_hostcollections:
+            description: Toggle, if true the inventory will fetch host collections for a hosts and create groupings for the same.
+            type: boolean
+            default: True
 '''
 
 EXAMPLES = '''
@@ -96,9 +154,9 @@ password: secure
 validate_certs: False
 host_filters: 'organization="Web Engineering"'
 '''
-
+import json
 from distutils.version import LooseVersion
-
+from time import sleep
 from ansible.errors import AnsibleError
 from ansible.module_utils._text import to_bytes, to_native, to_text
 from ansible.module_utils.common._collections_compat import MutableMapping
@@ -123,7 +181,7 @@ class InventoryModule(BaseInventoryPlugin, Cacheable, Constructable):
     def __init__(self):
 
         super(InventoryModule, self).__init__()
-
+        self.MINIMUM_FOREMAN_VERSION_FOR_REPORTING_API = '1.24.0'
         # from config
         self.foreman_url = None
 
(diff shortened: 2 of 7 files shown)


plugins/inventory/foreman.py
```diff
@@ -85,6 +85,60 @@
...
         type: integer
         default: 250
+      foreman:
+        use_reports_api:
+            description: Use Reporting API
+            type: boolean
+            default: False
+      report:
+        poll_interval:
+            description: want_organization
```
evgeni: the descriptions of the `report` options might need a bit more work ;)

plugins/inventory/foreman.py
```diff
@@ -238,10 +293,214 @@ def _get_hostvars(self, host, vars_prefix='', omitted_vars=()):
...
+        params["Facts".encode('ascii')] = options[self.want_facts]
+        params["Host Group".encode('ascii')] = options[self.want_host_group]
+        params["Host Collections".encode('ascii')] = options[self.want_hostcollections]
+        params["Subnet".encode('ascii')] = options[self.want_subnet]
+        params["Subnet v6".encode('ascii')] = options[self.want_subnet_v6]
+        params["Smart Proxies".encode('ascii')] = options[self.want_smart_proxies]
+        params["Content Attributes".encode('ascii')] = options[self.want_content_facet_attributes]
+        params["Host Parameters".encode('ascii')] = options[self.want_host_params]
+        if self.host_filters:
+            params["Hosts".encode('ascii')] = self.host_filters
```
evgeni: why the `encode('ascii')` everywhere?

plugins/inventory/foreman.py
```diff
@@ -238,10 +293,214 @@ def _get_hostvars(self, host, vars_prefix='', omitted_vars=()):
...
+        return params
+
+    def _use_inventory_report(self):
+        use_inventory_report = self.get_option('foreman').get('use_reports_api')
+        if not use_inventory_report:
+            return False
+        status_url = "%s/api/v2/status" % self.foreman_url
+        result = self._get_json(status_url)
+        foreman_version = (LooseVersion(result.get('version')) >= LooseVersion('1.24.0'))
+        return foreman_version
```
evgeni: technically this returns "is foreman version at least 1.24", not the version itself, so I'd just do
```suggestion
        return (LooseVersion(result.get('version')) >= LooseVersion('1.24.0'))
```

and maybe put the `1.24.0` as a constant like `MINIMUM_FOREMAN_VERSION_FOR_REPORTING_API`?

evgeni: Also, could this check if the `ansible` plugin is installed and return `False` if not?

Probably also good idea to issue warnings if either the version is too old or the plugin not installed but the user requested the feature (then, obviously, my above comment becomes moot)?

evgeni: ah, it *tries* that later:

```python
        try:
            inventory_report_response = self._post_request()
        except Exception:
            self._populate_host_api()
            return
```

plugins/inventory/foreman.py
```diff
@@ -238,10 +293,214 @@ def _get_hostvars(self, host, vars_prefix='', omitted_vars=()):
...
+        self.hosts = dict()
+        try:
+            inventory_report_response = self._post_request()
+        except Exception:
+            self._populate_host_api()
+            return
+        self.group_prefix = self.get_option('group_prefix')
+        h_data = json.loads(inventory_report_response)
+        null = None  # noqa: F841
+        false = False  # noqa: F841
```
evgeni: what are these for?

plugins/inventory/foreman.py
```diff
@@ -238,10 +293,214 @@ def _get_hostvars(self, host, vars_prefix='', omitted_vars=()):
...
+        host_data = eval(json.dumps(h_data).encode('ascii'))
```
evgeni: `eval`? no.

tests/inventory/tests.foreman.yml
```diff
@@ -4,3 +4,20 @@ user: admin
...
+  want_ipv6: True
+  want_host_group: True
+  want_subnet: True
+  want_subnet_v6: True
+  want_smart_proxies: True
+  want_content_facet_attributes: True
+  want_host_params: True
+  want_facts: True
+  poll_interval: 10
+  want_hostcollections: True
```
evgeni: those are all default, why setting them?

plugins/inventory/foreman.py
```diff
@@ -85,6 +85,60 @@
         description: Number of hosts per batch that will be retrieved from the Foreman API per individual call
         type: integer
         default: 250
+      foreman:
+        use_reports_api:
```
evgeni, line 91: why is that a sub-entry of `foreman` instead of being to-level as all the others?

plugins/inventory/foreman.py
```diff
@@ -85,6 +85,60 @@
...
+            description: Use Reporting API
+            type: boolean
+            default: False
+      report:
```
evgeni, line 95:
```suggestion
      report:
        description: report API specific configuration
        type: dict
```

evgeni: (that's what the test failure "ERROR! Unable to retrieve documentation from 'theforeman.foreman.foreman' due to: All (sub-)options and return values must have a 'description' field" is about)

evgeni (changes requested):

plugins/inventory/foreman.py
sjha4: Pushed some changes to add better description. :)

plugins/inventory/foreman.py
sjha4: Meant to revert that. I was having some unrelated issues with encoding while working on this and was trying some fixes. Removed these.

plugins/inventory/foreman.py
sjha4: > eval? no.

These were to help with eval since json returned was ruby and interpreted in python. It wasn't working without these. However, I was trying to use eval since I was facing some issues parsing the returned json from report API that I don't seem to be seeing anymore. Removed these and eval.

plugins/inventory/foreman.py
sjha4: Removed.

tests/inventory/tests.foreman.yml
sjha4: Just added as an example and for testing the reading from config part. We can remove these if not needed.

plugins/inventory/foreman.py
sjha4: This is the way it was setup in the contrib script to read from config. Motivation here was that users upgrading wouldn't have to change their config files.

plugins/inventory/foreman.py
sjha4: I wasn't sure what was happening with the test failures. Thanks..Updated.

plugins/inventory/foreman.py
evgeni: Okay, yeah, I can see compat being a reason.

plugins/inventory/foreman.py
```diff
@@ -85,6 +85,64 @@
...
+        want_organization:
+            description: Toggle, if true the inventory will fetch organization the host belongs to and create groupings for the same.
+            type: boolean
+            default: True
+        want_location:
+            description: Toggle, if true the inventory will fetch location the host belongs to and create groupings for the same.
+            type: boolean
+            default: True
+        want_ipv4:
+            description: Toggle, if true the inventory will fetch ipv4.
```
evgeni:
```suggestion
            description: Toggle, if true the inventory will fetch IPv4 address of the host.
```

plugins/inventory/foreman.py
```diff
@@ -85,6 +85,64 @@
...
+            type: boolean
+            default: True
+        want_ipv6:
+            description: Toggle, if true the inventory will fetch ipv6.
```
evgeni:
```suggestion
            description: Toggle, if true the inventory will fetch IPv6 address of the host.
```

plugins/inventory/foreman.py
```diff
@@ -85,6 +85,64 @@
...
+            default: False
+      report:
+        description: report API specific configuration
+        type: dict
+        poll_interval:
+            description: The polling interval between 2 calls to the report_data endpoint while polling.
+            type: integer
+            default: 10
+        want_organization:
+            description: Toggle, if true the inventory will fetch organization the host belongs to and create groupings for the same.
```
evgeni, line 107: I wonder if we could change all these "Toggle, if true the inventory will fetch …" into "Include … in the inventory data" (or something like that?

plugins/inventory/foreman.py
```diff
@@ -238,10 +297,200 @@ def _get_hostvars(self, host, vars_prefix='', omitted_vars=()):
                 hostvars[vars_prefix + k] = v
         return hostvars
 
+    def to_safe(self, word):
```
evgeni: There is a `to_safe_group_name` that we import from `ansible.plugins.inventory` and use in the non-report version, should it be used here too, instead of implementing it ourself?

plugins/inventory/foreman.py
```diff
@@ -238,10 +297,200 @@ def _get_hostvars(self, host, vars_prefix='', omitted_vars=()):
...
+            self._populate_host_api()
+            return
+        self.group_prefix = self.get_option('group_prefix')
+        host_data = json.loads(inventory_report_response)
+        for host in host_data:
+            if not(host) or (host["name"] in self.cache.keys()):
+                continue
+            dns_name = self.inventory.add_host(host['name'])
+            host_name = dns_name
+            group_name = host.get('hostgroup_title', host.get('hostgroup_name'))
```
evgeni, line 400: @ezr-ondrej has recently fixed a bug where we had accidentally omitted empty parent groups from the import, is the same needed in this part too? see https://github.com/theforeman/foreman-ansible-modules/pull/931

tests/inventory/tests.foreman.yml
evgeni: I'd prefer to keep the test config minimal

evgeni:
Thanks @sjha4! I think the biggest thing missing here now are tests, as currently they run against an foreman-only container and thus the inventory silently falls back to the non-reporting mode.

I'll try to refactor the current tests into something that can be run twice with two different containers.

plugins/inventory/foreman.py
sjha4: Updated and tested with regular hostgroup with hosts. Could you explain the exact data setup I should have to test this.

tests/inventory/tests.foreman.yml
sjha4: Sounds good.. Will remove this file from the commit after testing. :+1:

plugins/inventory/foreman.py
evgeni: Have a look at the test playbook changes in #931. Essentially: create a groupA, create a groupB with parent=groupA, now put a host into groupB, the inventory should report it as being in ansbibke group groupA/groupB (or whatever safe does from it, I think groupA_groupB).

Technically it should also be in groupA, but I think that's not implemented in API mode either. Right @ezr-ondrej?

sjha4: Ok..I tested with parent/child hostgroup and I see the same behavior with reporting api and host api.
The data looks like this with both: (Empty parent hostgroup: "test_host_group", Child Hostgroup: "test_group_child")

"foreman_test_host_group": {
        "children": [
            "foreman_test_host_group_test_group_child"
        ]
    }, 
    "foreman_test_host_group_test_group_child": {
        "hosts": [
            "virt-who-0e93693d.7096.4346.afb6.1c95bd1153e9-1"
        ]
    },

evgeni: (in groupA of course works, that's what Ansible does for us)

evgeni: Thanks! I had the tests a bit refactored last week and will try to add them to this PR next week when back home, shouldn't be too hard 🤘

sjha4:
@evgeni: Does this look good to you?

evgeni:
rebased and added a basic set of tests.

evgeni:
So while the tests are executed, for some reason it falls back to hosts api for me. Will investigate further.

evgeni:
Okay, a few observations:
* `default` does not work for `dict` sub-options in inventories (I've opened an RFE: https://github.com/ansible/ansible/issues/72377), so that `_fetch_params()`, `self.get_option('foreman').get('use_reports_api')` and `self.poll_interval = self.get_option('report').get('poll_interval')` fail terribly if the user didn't provide any settings
* even after providing all the settings I run into 500 errors with Foreman, and error reporting for reports is broken: https://projects.theforeman.org/issues/31191, https://projects.theforeman.org/issues/31192
* I'll retry with a fixed container (currently building)

sjha4:
> default does not work for dict sub-options in inventories (I've opened an RFE: ansible/ansible#72377), so that _fetch_params(), self.get_option('foreman').get('use_reports_api') and self.poll_interval = self.get_option('report').get('poll_interval') fail terribly if the user didn't provide any settings

Good catch..I will try to work around this in the script locally..

evgeni:
Okay, so the reason it doesn't work in my container is that the template requires Katello:
https://github.com/theforeman/foreman/blob/f0f1d3e90852124c9fe3ece4ac833bac835f5796/app/views/unattended/report_templates/ansible_-_ansible_inventory.erb#L114-L115

I don't understand why, tho.

evgeni:
@ares do you know why the Ansible inventory report template says it needs Katello?

sjha4:
> default does not work for dict sub-options in inventories (I've opened an RFE: ansible/ansible#72377), so that _fetch_params(), self.get_option('foreman').get('use_reports_api') and self.poll_interval = self.get_option('report').get('poll_interval') fail terribly if the user didn't provide any settings

Pushed an update to handle the issue with defaults in nested options.

plugins/inventory/foreman.py
```diff
@@ -238,10 +296,235 @@ def _get_hostvars(self, host, vars_prefix='', omitted_vars=()):
                 hostvars[vars_prefix + k] = v
         return hostvars
 
+    def _fetch_params(self):
+        options = ("no", "yes")
+        params = dict()
+        try:
+            self.want_location = self.get_option('report').get('want_location')
+        except Exception:
+            self.want_location = True
```
evgeni: Can we, instead of the whole `try/except` jumping, do something like that?

```python
report_options = {'want_location': True, 'want_organization': True, … }
report_options.update(self.get_option('report') or {})
self.want_location = report_options.get('want_location')
```

`get_option()` returns `None` if the option is not set at all (so in this case we take the `or {}`) and a dict with the set options if any were set.

plugins/inventory/foreman.py
```diff
@@ -85,6 +85,64 @@
         description: Number of hosts per batch that will be retrieved from the Foreman API per individual call
         type: integer
         default: 250
+      foreman:
+        description: foreman server related configuration
+        type: dict
+        use_reports_api:
+            description: Use Reporting API.
+            type: boolean
```
evgeni, line 93: these should be `type: bool`, not that Ansible properly parses sub-options for inventories anyways.

plugins/inventory/foreman.py
```diff
@@ -85,6 +85,64 @@
...
+            default: False
+      report:
+        description: report API specific configuration
+        type: dict
+        poll_interval:
+            description: The polling interval between 2 calls to the report_data endpoint while polling.
+            type: integer
```
evgeni: these should be `type: int`, not that Ansible properly parses sub-options for inventories anyways.

plugins/inventory/foreman.py
```diff
@@ -238,10 +296,201 @@ def _get_hostvars(self, host, vars_prefix='', omitted_vars=()):
                 hostvars[vars_prefix + k] = v
         return hostvars
 
+    def _fetch_params(self):
+        options = ("no", "yes")
+        params = dict()
+        report_options = {'want_location': True, 'want_organization': True, 'want_IPv4': True, 'want_IPv6': True, 'want_host_group': True, 'want_hostcollections': False, 'want_subnet': True, 'want_subnet_v6': True, 'want_smart_proxies': True, 'want_content_facet_attributes': False, 'want_host_params': True}
```
evgeni:
```suggestion
        report_options = {'want_location': True, 'want_organization': True, 'want_ipv4': True, 'want_ipv6: True, 'want_host_group': True, 'want_hostcollections': False, 'want_subnet': True, 'want_subnet_v6': True, 'want_smart_proxies': True, 'want_content_facet_attributes': False, 'want_host_params': True}
```

it's lower case in the docs, so it should be lower case here

plugins/inventory/foreman.py
```diff
@@ -238,10 +296,201 @@ def _get_hostvars(self, host, vars_prefix='', omitted_vars=()):
...
+        report_options.update(self.get_option('report') or {})
+        
+        self.want_location = report_options['want_location']
+        self.want_organization = report_options['want_organization']
+        self.want_IPv4 = report_options['want_IPv4']
+        self.want_IPv6 = report_options['want_IPv6']
```
evgeni:
```suggestion
        self.want_IPv4 = report_options['want_ipv4']
        self.want_IPv6 = report_options['want_ipv6']
```

plugins/inventory/foreman.py
evgeni: and the linter complains: `E501 line too long (308 > 160 characters)` ;)

plugins/inventory/foreman.py
sjha4: I see type: boolean in other params in the file. Same with type: integer ..

sjha4:
@evgeni..How do you rerun the tests on this? I am not sure why it timed out..

evgeni:
Close/reopen is the simplest way

sjha4:
@evgeni: You have nay ideas about some tests are getting canceled?

evgeni:
@sjha4 I do not, but my guess would be some rate limiting or so, will look into it on monday

evgeni:
Ah,  now I remember. I fixed the container used to properly "work", but that resulted in it not generating the report as that needs a dynflow worker.

plugins/inventory/foreman.py
```diff
@@ -83,8 +83,66 @@
...
+            type: boolean
+            default: True
+        want_content_facet_attributes:
+            description: Toggle, if true the inventory will fetch content view details that the host is tied to.
+            type: boolean
+            default: True
+        want_host_params:
+            description: Toggle, if true the inventory will fetch host parameters.
+            type: boolean
+            default: True
```
evgeni: How does that compare to the `want_params` option we have outside the `report` section?

plugins/inventory/foreman.py
```diff
@@ -238,10 +296,203 @@ def _get_hostvars(self, host, vars_prefix='', omitted_vars=()):
...
+        self.want_organization = report_options['want_organization']
+        self.want_IPv4 = report_options['want_ipv4']
+        self.want_IPv6 = report_options['want_ipv6']
+        self.want_host_group = report_options['want_host_group']
+        self.want_hostcollections = report_options['want_hostcollections']
+        self.want_subnet = report_options['want_subnet']
+        self.want_subnet_v6 = report_options['want_subnet_v6']
+        self.want_smart_proxies = report_options['want_smart_proxies']
+        self.want_content_facet_attributes = report_options['want_content_facet_attributes']
+        self.want_host_params = report_options['want_host_params']
```
evgeni: Interestingly, I had to explicitly set that to `True` as otherwise it didn't fetch the params for me

plugins/inventory/foreman.py
sjha4: Serves the same purpose except this one is more report specific as in sent to the report API..

evgeni: Right, but isn't it weird (from a user PoV) to have to set it again, just because a different backend is used?

plugins/inventory/foreman.py
sjha4: I ran it and I see host params..
```
"virt-who-fb1b6f1e.c091.42f4.bd69.fc1bcdcd0dae-1": {
                "foreman_content_attributes": null, 
                "foreman_host_group": null, 
                "foreman_host_parameters": {
                    "remote_execution_connect_by_ip": false, 
                    "remote_execution_effective_user_method": "sudo", 
                    "remote_execution_ssh_user": "root"
                }, 
                "foreman_id": 11, 
                "foreman_organization": "Default Organization", 
                "foreman_smart_proxies": [], 
                "foreman_subnet": null, 
                "foreman_subnet6": null

```

evgeni: Hmm. Need to look further. I was missing the domain name.

evgeni: Right, so when I don't set it, I get:

```json
            "testhost1.example.com": {
                "foreman_host_group": "group_a",
                "foreman_host_parameters": {
                    "remote_execution_connect_by_ip": false,
                    "remote_execution_effective_user_method": "sudo",
                    "remote_execution_ssh_user": "root",
                    "testparam1": "testvalue1",
                    "testparam2": "testvalue2"
                },
                "foreman_id": 1,
                "foreman_ipv4": null,
                "foreman_ipv6": null,
                "foreman_location": "Default Location",
                "foreman_organization": "Default Organization",
                "foreman_smart_proxies": [],
                "foreman_subnet": null,
                "foreman_subnet6": null
            },
```

and when I set it, I get:

```json
            "testhost1.example.com": {
                "foreman_architecture_id": null,
                "foreman_architecture_name": null,
                "foreman_build": false,
                "foreman_capabilities": [
                    "build"
                ],
                "foreman_certname": "testhost1.example.com",
                "foreman_comment": null,
                "foreman_compute_profile_id": null,
                "foreman_compute_profile_name": null,
                "foreman_compute_resource_id": null,
                "foreman_compute_resource_name": null,
                "foreman_created_at": "2020-11-10 10:01:13 UTC",
                "foreman_disk": null,
                "foreman_domain_id": 1,
                "foreman_domain_name": "example.com",
                "foreman_enabled": true,
                "foreman_environment_id": 1,
                "foreman_environment_name": "production",
                "foreman_global_status": 0,
                "foreman_global_status_label": "OK",
                "foreman_hostgroup_id": 1,
                "foreman_id": 1,
                "foreman_image_file": "",
                "foreman_image_id": null,
                "foreman_image_name": null,
                "foreman_installed_at": null,
                "foreman_ip": null,
                "foreman_ip6": null,
                "foreman_last_compile": null,
                "foreman_last_report": null,
                "foreman_location_id": 2,
                "foreman_location_name": "Default Location",
                "foreman_mac": null,
                "foreman_managed": false,
                "foreman_medium_id": null,
                "foreman_medium_name": null,
                "foreman_model_id": null,
                "foreman_model_name": null,
                "foreman_operatingsystem_id": 1,
                "foreman_operatingsystem_name": "CentOS 7.7",
                "foreman_organization_id": 1,
                "foreman_organization_name": "Default Organization",
                "foreman_owner_id": 4,
                "foreman_owner_name": "Admin User",
                "foreman_owner_type": "User",
                "foreman_provision_method": "build",
                "foreman_ptable_id": null,
                "foreman_ptable_name": null,
                "foreman_puppet_ca_proxy": null,
                "foreman_puppet_ca_proxy_id": null,
                "foreman_puppet_ca_proxy_name": null,
                "foreman_puppet_proxy": null,
                "foreman_puppet_proxy_id": null,
                "foreman_puppet_proxy_name": null,
                "foreman_puppet_status": 0,
                "foreman_pxe_loader": null,
                "foreman_realm_id": null,
                "foreman_realm_name": null,
                "foreman_sp_ip": null,
                "foreman_sp_mac": null,
                "foreman_sp_name": null,
                "foreman_sp_subnet_id": null,
                "foreman_subnet6_id": null,
                "foreman_subnet6_name": null,
                "foreman_subnet_id": null,
                "foreman_subnet_name": null,
                "foreman_updated_at": "2020-11-10 10:01:15 UTC",
                "foreman_uptime_seconds": null,
                "foreman_use_image": null,
                "foreman_uuid": null,
                "testparam1": "testvalue1",
                "testparam2": "testvalue2"
            },
```

evgeni: Aha. The one with actually failed back to non-reports.

plugins/inventory/foreman.py
evgeni: For facts, you also do use the global one: `self.want_facts = self.get_option('want_facts')`

plugins/inventory/foreman.py
```diff
@@ -238,10 +296,203 @@ def _get_hostvars(self, host, vars_prefix='', omitted_vars=()):
...
+        result = self._get_json(status_url)
+        foreman_version = (LooseVersion(result.get('version')) >= LooseVersion(self.MINIMUM_FOREMAN_VERSION_FOR_REPORTING_API))
+        return foreman_version
+
+    def _post_request(self):
+        url = "%s/ansible/api/v2/ansible_inventories/schedule" % self.foreman_url
+        session = self._get_session()
+        params = {'input_values': self._fetch_params()}
+        try:
+            self.poll_interval = self.get_option('report').get('poll_interval')
```
evgeni:
```suggestion
            self.poll_interval = int(self.get_option('report').get('poll_interval'))
```

otherwise you get a string (as Ansible doesn't convert nested params for us) and `sleep(a_string)` doesn't work.

plugins/inventory/foreman.py
```diff
@@ -238,10 +296,203 @@ def _get_hostvars(self, host, vars_prefix='', omitted_vars=()):
...
+            raise Exception("Error scheduling inventory report on foreman. Please check foreman logs!")
+        url = "{0}/{1}".format(self.foreman_url, ret.json().get('data_url'))
+        response = session.get(url)
+        while response:
+            if response.status_code != 204:
+                break
+            sleep(self.poll_interval)
+            response = session.get(url)
+        if not response:
+            raise Exception("Error receiving inventory report from foreman. Please check foreman logs!")
```
evgeni: this loop needs some kind of timeout, as otherwise if the api does not work for some reason (= does not return anything but `204`), we'll end up in and endless loop.

plugins/inventory/foreman.py
```diff
@@ -238,10 +296,203 @@ def _get_hostvars(self, host, vars_prefix='', omitted_vars=()):
...
+            self._populate_host_api()
+
+    def _populate_report_api(self):
+        self.groups = dict()
+        self.hosts = dict()
+        try:
+            inventory_report_response = self._post_request()
+        except Exception:
+            self._populate_host_api()
+            return
```
evgeni:
```suggestion
        except Exception as exc:
            self.display.warning("Failed to use Reports API, falling back to Hosts API: {}".format(exc))
            self._populate_host_api()
            return
```

Otherwise the use will never know why/how it has failed back to the slow one.

plugins/inventory/foreman.py
```diff
@@ -238,10 +296,203 @@ def _get_hostvars(self, host, vars_prefix='', omitted_vars=()):
...
 
+                    if self.get_option('legacy_hostvars'):
+                        self.inventory.set_variable(host_name, 'foreman_params', filtered_params)
+                    else:
+                        for k, v in filtered_params.items():
+                            try:
+                                self.inventory.set_variable(host_name, k, v)
+                            except ValueError as e:
+                                self.display.warning("Could not set hostvar %s to '%s' for the '%s' host, skipping:  %s" %
+                                                     (k, to_native(v), host, to_native(e)))
```
evgeni: I think the code should be indented at the same level as `# set host vars from params`, as otherwise it's only executed when `self.want_hostcollections` is `True`, which seems not what it should do?

plugins/inventory/foreman.py
```diff
@@ -238,10 +296,203 @@ def _get_hostvars(self, host, vars_prefix='', omitted_vars=()):
...
+                        except ValueError as e:
+                            self.display.warning("Could not create groups for host collections for %s, skipping: %s" % (host_name, to_text(e)))
 
-        for host in self._get_hosts():
+            # set host vars from params
+                if self.get_option('want_params'):
+                    filtered_params = {}
+                    for p in params:
+                        if 'name' in p and 'value' in p:
+                            filtered_params[p['name']] = p['value']
```
evgeni: this doesn't work, as the `params` we get from the Reports API don't have name/value in there, but are just a dict: `{'testparam1': 'testvalue1', 'testparam2': 'testvalue2', 'remote_execution_ssh_user': 'root', 'remote_execution_effective_user_method': 'sudo', 'remote_execution_connect_by_ip': False}`

evgeni: the fix is to use `params` directly in the later code, not doing the `filtered_params` thing

plugins/inventory/foreman.py
```diff
@@ -238,10 +296,203 @@ def _get_hostvars(self, host, vars_prefix='', omitted_vars=()):
...
+                hostvars = self._get_hostvars(host, self.get_option('vars_prefix'), omitted_vars)
+
+                for k, v in hostvars.items():
+                    try:
+                        self.inventory.set_variable(host_name, k, v)
+                    except ValueError as e:
+                        self.display.warning("Could not set host info hostvar for %s, skipping %s: %s" % (host, k, to_text(e)))
+            host_params = host.pop('host_parameters', {})
+            fact_list = host.pop('facts', {})
+            content_facet_attributes = host.get('content_attributes', {}) or {}
```
evgeni: I think this needs to happen before

```python
                omitted_vars = ('name', 'hostgroup_title', 'hostgroup_name')
                hostvars = self._get_hostvars(host, self.get_option('vars_prefix'), omitted_vars)
```

(or be added to "omitted") as otherwise the above will unconditionally add these to the result as `foreman_host_params` etc, which is not what we want (see the `# set host vars from params` below)

plugins/inventory/foreman.py
```diff
@@ -238,10 +296,205 @@ def _get_hostvars(self, host, vars_prefix='', omitted_vars=()):
...
+                        except ValueError as e:
+                            self.display.warning("Could not create groups for host collections for %s, skipping: %s" % (host_name, to_text(e)))
 
-        for host in self._get_hosts():
+            # set host vars from params
+            if self.get_option('want_params'):
+                filtered_params = {}
+                for p in params:
+                    if 'name' in p and 'value' in p:
+                        filtered_params[p['name']] = p['value']
```
evgeni:
```suggestion
```

plugins/inventory/foreman.py
```diff
@@ -238,10 +296,205 @@ def _get_hostvars(self, host, vars_prefix='', omitted_vars=()):
...
 
+                if self.get_option('legacy_hostvars'):
+                    self.inventory.set_variable(host_name, 'foreman_params', filtered_params)
```
evgeni:
```suggestion
                    self.inventory.set_variable(host_name, 'foreman_params', params)
```

plugins/inventory/foreman.py
evgeni:
```suggestion
```

Can be removed, as you are now using the global `want_params` attr.

plugins/inventory/foreman.py
evgeni: yeah, not sure, maybe it's not strict

plugins/inventory/foreman.py
```diff
@@ -238,10 +296,205 @@ def _get_hostvars(self, host, vars_prefix='', omitted_vars=()):
...
+                omitted_vars = ('name', 'hostgroup_title', 'hostgroup_name')
+                hostvars = self._get_hostvars(host, self.get_option('vars_prefix'), omitted_vars)
+
+                for k, v in hostvars.items():
+                    try:
+                        self.inventory.set_variable(host_name, k, v)
+                    except ValueError as e:
+                        self.display.warning("Could not set host info hostvar for %s, skipping %s: %s" % (host, k, to_text(e)))
+
+            content_facet_attributes = host.get('content_attributes', {}) or {}
```
evgeni, line 422:
```suggestion
            content_facet_attributes = host.get('content_attributes', {})
```

the `or {}` is not necessary, `get` will do this for us nicely

plugins/inventory/foreman.py
sjha4: We discussed this briefly when migrating to reports with tower initially and the time to generate the report is dictated by the number of hosts to be captured and both the report generation in foreman and inventory sync in tower are async jobs. Given that, should we add a (user-specified?) timeout in the script itself?

evgeni: While this is true for Tower, it's not for Ansible itself, here the inventory fetch is synchronous.

The case where I hit the issue was: my dynflow workers died, so while the request to generate an inventory was accepted, it was never processed, the api kept returning 204 and nothing worked.

Is there a guesstimate for large deployments? Ist 5 minutes a good timeout?

Partially unrelated: this triggers a regeneration on each call - is there a way to use an older report?

sjha4: Vladimir had done some load testing of this on the tower side initially. His tests showed roughly 3 minutes for 10k hosts. I think 10 minutes should be a good conservative timeout for waiting on the report generation to finish if we have that as default timeout..Pushed a change to add that.

I am not sure about the second question. Presently, If the report API fails, we fall back to the old host API.

plugins/inventory/foreman.py
sjha4: I added this back. It wasn't handling the None case properly without this..

plugins/inventory/foreman.py
evgeni: I should not write review comments on mobile… Sorry :)

> Partially unrelated: this triggers a regeneration on each call - is there a way to use an older report?

What I meant is: is there a way (in the API) to say "give me the last rendered version of this report, without generating a new one"? I can imagine on bigger environments, people would like to just render it on a timer, and then use the pre-rendered version instead of triggering a fresh render every time ansible runs. But this is totally out of scope here, so let's get on with the real thing.

plugins/inventory/foreman.py
mdellweg: So get returns None if the key was already set to None. But also as a default if we dont specify anything.
It should be `content_facet_attributes = host.get('content_attributes') or {}`.
Downside (, but the original version has the same problem): False, [], "" and similar is all translated to {}.

evgeni:
Thank you @sjha4! I think this is as good as we can get from the old code base and we should merge this now. However, I have a few thoughts how we should improve in future:
* Lots of the logic around setting host params, groups, etc is now duplicated in both implementations, we should try to DRY that up.
* I still dislike the `dict` based config, so I thought we could move it (all the params we have in `reporting` and `foreman`) to the "main" level and then provide some code that will look into the `reporting` and `foreman` dicts to remain compatible with the old style.
* The tests need some cleanup, but that's on me.

plugins/inventory/foreman.py
evgeni: Yeah. That's also what I thought after Samir said it didn't work as expected.

I am not sure if we write `host.get('content_attributes') or {}` now and someone will refactor that in a month to `host.get('content_attributes', {})` without thinking further. The current version is more explicit?

sjha4:
@evgeni : Thanks for the extensive reviews here!

changelogs/fragments/inventory_plugin_report_api.yml
```diff
@@ -0,0 +1,2 @@
+minor_changes:
+   - inventory plugin - add support for the Report API present in Foreman 1.24 and later
```
sjha4, line 2: @evgeni: Is this good?

evgeni: Looks good, thanks!

evgeni (approved):

PR #989, opened 2020-09-29
fix a few typos in docstrings

evgeni:


plugins/module_utils/foreman_helper.py
```diff
@@ -1285,7 +1290,7 @@ def _foreman_spec_helper(spec):
     # We have to ensure that apypie is available before using it.
     # There is two cases where we can call _foreman_spec_helper() without apypie available:
     # * When the user calls the module but doesn't have the right Python libraries installed.
-    #   In this case nothing will works and the module will warn teh user to install the required library.
+    #   In this case nothing will works and the module will warn the user to install the required library.
```
mdellweg, line 1293: Awww

plugins/module_utils/foreman_helper.py
```diff
@@ -64,7 +64,7 @@
     hostgroups='title',
     locations='title',
     operatingsystems='title',
-    # TODO: Organizations should be search by title (as foreman allow nested orgs) but that's not the case ATM.
+    # TODO: Organizations should be search by title (as foreman allows nested orgs) but that's not the case ATM.
```
mdellweg, line 67: Wondering, how much :cake: and :beers: it would take to remove that TODO.

mdellweg (approved):

plugins/module_utils/foreman_helper.py
evgeni: Last time my proposed expense was denied, something about "you don't need a country of the size of Luxembourg" :woman_shrugging:

PR #991, opened 2020-09-29
move apidoc patching in an own method

evgeni:
this makes it cleaner to override it instead of connect()

### DIFF
--- a/plugins/module_utils/foreman_helper.py
+++ b/plugins/module_utils/foreman_helper.py
@@ -105,9 +105,8 @@ class KatelloMixin():
         required_plugins.append(('katello', ['*']))
         super(KatelloMixin, self).__init__(foreman_spec=foreman_spec, required_plugins=required_plugins, **kwargs)
 
-    @_exception2fail_json(msg="Failed to connect to Foreman server: {0}")
-    def connect(self):
-        super(KatelloMixin, self).connect()
+    def apply_apidoc_patches(self):
+        super(KatelloMixin, self).apply_apidoc_patches()
 
         self._patch_content_uploads_update_api()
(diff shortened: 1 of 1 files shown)


plugins/module_utils/foreman_helper.py
```diff
@@ -494,13 +493,14 @@ def connect(self):
         )
 
         self.status()
+        self.apply_apidoc_patches()
+        self.check_required_plugins()
 
+    def apply_apidoc_patches(self):
```
mdellweg, line 499: So we turn this hack into a facility now.
This would be the perfect time to add a docstring. ;)

How about:
"Calls some workarounds to fix the local representation of the api docs.
When adding another workaround, consider that the endpoint in question may depend on a plugin to be available.
Also a workaround should be specific to a range of server/plugin versions where possible."

evgeni: Yeah, sounds good.

I wanted to add some more metadata to the patches and make this method more robust (in the future)

mdellweg: Do we have an example where such a bug was actually fixed upstream.

evgeni: Yeah, for example https://projects.theforeman.org/issues/10359 was fixed in 1.23, so we won't need to call `_patch_location_api()` on modern installs.

mdellweg: Looking forward to see how that works. ;)

evgeni: Used a slightly reworded docstring.

mdellweg (approved):

PR #993, opened 2020-09-30
make apidoc patching more explicit in terms of fixed versions

evgeni:
Draft because I am still debating whether this should become a decorator or integrated in any other way.


plugins/module_utils/foreman_helper.py
```diff
@@ -511,7 +524,8 @@ def connect(self):
             verify_ssl=self._foremanapi_validate_certs,
         )
 
-        self.status()
+        _status = self.status()
+        self.foreman_version = _status.get('version', '0.0.0')
```
ekohl: Should this already be a `LooseVersion` object?

evgeni: I debated that and wasn't sure where else I'd might need the string.

On the other hand, there are no other users of that right now, and `LooseVersion` has a working `__str__`… Yeah, I'll change that.

plugins/module_utils/foreman_helper.py
```diff
@@ -90,6 +92,32 @@ def inner(self, *args, **kwargs):
     return decor
 
 
+def _check_patch_needed(fixed_version=None, plugins=None):
```
evgeni, line 95: I wonder if instead of *check if required* this decorator should add those patch functions to a list (like `self._apidoc_patches`) and then `apply_apidoc_patches` becomes:

```python
def apply_apidoc_patches(self):
    for apidoc_patch in self._apidoc_patches:
        apidoc_patch()
````

Comment (evgeni) [reply plugins/module_utils/foreman_helper.py:95]: But that probably needs more thought, as the decoration happens *before* we know the server version…

File: plugins/module_utils/foreman_helper.py [hunk 3]
```diff
@@ -90,6 +92,32 @@ def inner(self, *args, **kwargs):
...
+    :type plugins: list, optional
+    """
+
+    def decor(f):
+        @wraps(f)
+        def inner(self, *args, **kwargs):
+            if plugins is not None:
+                for plugin in plugins:
+                    if not self.has_plugin(plugin):
+                        return
```
mdellweg: If you think it's easier to read:
```suggestion
            if plugins is not None and not all(self.has_plugin(plugin) for plugin in plugins):
                return
```

plugins/module_utils/foreman_helper.py
mdellweg: yes, maybe that is over-engineered. But should we move the "katello-patches" to the base class anyway now?

Also we could call all members starting with "_patch_" (even if that sounds rather rubyesque).

mdellweg (approved):

plugins/module_utils/foreman_helper.py
evgeni: moving katello patches is a good idea I think, yeah.

PR #994, opened 2020-09-30
check remote_execution plugin is installed if rex proxies are set

evgeni:


mdellweg (approved):
It should have been obvious...

PR #995, opened 2020-09-30
document rex bug

evgeni:
See https://projects.theforeman.org/issues/30651

### DIFF
--- a/plugins/modules/subnet.py
+++ b/plugins/modules/subnet.py
@@ -129,6 +129,7 @@ options:
     description:
       - Remote execution Smart proxies for this subnet
       - This option is only available if the remote_execution plugin is installed.
+      - This will always report changed when used with remote_execution < 4.1.0, due to a bug in plugin.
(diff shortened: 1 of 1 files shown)


plugins/modules/subnet.py
```diff
@@ -129,6 +129,7 @@
     description:
       - Remote execution Smart proxies for this subnet
       - This option is only available if the remote_execution plugin is installed.
+      - This will always report changed when used with remote_execution < 4.1.0, due to a bug in plugin.
```
mdellweg:
```suggestion
      - This will always report I(changed=true) when used with I(remote_execution < 4.1.0), due to a bug in the plugin.
```

Not quite sure if "I" is the right one in the second place.

evgeni: It's as good as any other, as they don't fit anyways :see_no_evil:

mdellweg (approved):

PR #1000, opened 2020-10-01
don't make job invocations taxonomic

### LINKED ISSUE #999: make job_invocation module not taxonomic
If I'm reading https://github.com/theforeman/foreman_remote_execution/blob/master/app/controllers/api/v2/job_invocations_controller.rb correctly, then:
* Job invocations are not taxonomic
* There is no update or destroy method
* There is a cancel method (as you mentioned above) which will require special handling
* There is a rerun method which we probably don't need just now

What I was actually looking for: you set search_by description, which is rather unusual, but I could find anything better yet

_Originally posted by @evgeni in https://github.com/theforeman/foreman-ansible-modules/pull/864#issuecomment-677322713_

evgeni:
the api doesn't support that

Fixes: #999

### DIFF
--- a/plugins/modules/job_invocation.py
+++ b/plugins/modules/job_invocation.py
@@ -120,7 +120,6 @@ options:
         type: int
 extends_documentation_fragment:
   - theforeman.foreman.foreman
-  - theforeman.foreman.foreman.taxonomy
 '''
 
 EXAMPLES = '''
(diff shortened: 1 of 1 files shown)


evgeni:
@pondrejk that's one thing I forgot in your PR.

mdellweg (approved):

PR #1001, opened 2020-10-01
Add live test targets

mdellweg:


evgeni (approved):

evgeni:
@pondrejk this is interesting for you I think

Makefile
```diff
@@ -19,17 +19,19 @@ PYTEST = pytest -n 4 --boxed -v
...
+	@echo "  help             to show this message"
+	@echo "  info             to show infos about the collection"
+	@echo "  lint             to run code linting"
+	@echo "  test             to run unit tests"
+	@echo "  livetest         to run test playbooks live (without vcr)"
+	@echo "  sanity           to run santy tests"
+	@echo "  setup            to set up test, lint"
+	@echo "  test-setup       to install test dependencies"
+	@echo "  test_<test>      to run a specific unittest"
+	@echo "  livetest_<test>  to run a specific unittest"
```
pondrejk, line 31: maybe we should say here what is the difference between the two

Makefile
```diff
@@ -57,12 +59,18 @@ test-check-mode: $(MANIFEST) | tests/test_playbooks/vars/server.yml
...
 	$(PYTEST) -k 'not test_crud.py'
 
+livetest: $(MANIFEST) | tests/test_playbooks/vars/server.yml
+	pytest -v 'tests/test_crud.py::test_crud' --vcrmode live
+
 test_%: FORCE $(MANIFEST) | tests/test_playbooks/vars/server.yml
 	pytest -v 'tests/test_crud.py::test_crud[$*]' 'tests/test_crud.py::test_check_mode[$*]' $(FLAGS)
 
+livetest_%: FORCE $(MANIFEST) | tests/test_playbooks/vars/server.yml
+	pytest -v 'tests/test_crud.py::test_crud[$*]' --vcrmode live
```
pondrejk, line 69: could we have $(FLAGS) here as well?

evgeni:
@mdellweg do you want to incorporate the comments from Peter in a new PR?

Makefile
mdellweg: I guess so... PR will follow.

PR #1013, opened 2020-10-05
vendor apypie

evgeni:


ehelms:
:+1:

What version of the library are we vendoring at this time? What is the process for updates?

evgeni:
@ehelms right now it's 0.2.2 from pypi, and `make vendor` should do everything for you ;)

but this will be documented.

evgeni:
great... old ansible still doesn't like me :fire:

evgeni:
okay, that's better…

Makefile
```diff
@@ -125,6 +125,10 @@ doc: $(MANIFEST)
 	antsibull-docs collection --use-current --squash-hierarchy --dest-dir ./docs/plugins $(NAMESPACE).$(NAME)
 	make -C docs html
 
+vendor:
+	git clone --depth=1 https://github.com/Apipie/apypie/ build/apypie-git
```
ehelms: Do we want to lock this to a version that gets set or updated ? Or just always take master?

evgeni: I would lock versions, just *right now* I need master for it to not explode ;)

plugins/module_utils/foreman_helper.py
```diff
@@ -26,7 +26,10 @@
 from ansible.module_utils import six
 
 try:
-    import apypie
+    try:
+        from ansible_collections.theforeman.foreman.plugins.module_utils import _apypie as apypie
+    except ImportError:
```
evgeni, line 33: this is a terrible hack to make `_foreman_spec_helper` work *outside* the ansible context (as we do it in our tests). I'd much more prefer if the tests could pretend `ansible_collections` is an importable namespace, but I'm hittin' my Python hackery limits here.

vendor.py
```diff
@@ -0,0 +1,38 @@
+#!/usr/bin/env python3
```
evgeni, line 1: This file should probably not live in the top-level directory, but I didn't have a good idea where to put it.

plugins/module_utils/foreman_helper.py
mdellweg: would it work to import relative?
`like import ._apypie as apypie`

mdellweg: maybe it's breaking ansible machinery...

vendor.py
mdellweg: `dist_tools` ? Is there more we can move out of the main directory?

vendor.py
```diff
@@ -0,0 +1,50 @@
...
+                if not output_lines[-1].startswith('import '):
+                    if output_lines[-1] != '':
+                        output_lines.append('')
+                    output_lines.append('')
+            output_lines.append(line)
+    if buffer_lines:
+        output_lines.extend(buffer_lines)
+        buffer_lines.clear()
+
+print("\n".join(output_lines))
```
mdellweg: Would it be easier to maintain a stack of patches? i.e. with quilt?

evgeni: It's been a while I've seen "easier" and "quilt" in one sentence ;-)

I am not sure. The above should work rather reliably even if we do changes to apypie, a patch (or a series) might not. And something would still need to do the job of "concat all the files together" (well, `cat` exists, granted).

mdellweg: Let's keep it that way, and time will show.

plugins/module_utils/foreman_helper.py
evgeni: Ansible (pre 2.10) can't do relative imports :(

vendor.py
evgeni: well, the `Makefile`, but moving it would require more work than worth it IMHO

ehelms: This is just a build script right? I am ok with it being top level unless we can drop it in a scripts or build type directory.

evgeni: Yepp, it's just used for the "mangle apypie code to make it loadable as a module util of Ansible" step, which we only have to execute when we pull in a new apypie version.

mdellweg:
You need to adjust

https://github.com/theforeman/foreman-ansible-modules/blob/fe0f1d0d6f1ccd5cf2d00ff0794c7f5b940a6a32/plugins/module_utils/foreman_helper.py#L514

mdellweg (approved):
All issues i found have been addressed.
Thank you!

evgeni:
I'll hold off merging this until apypie 0.3.0 is released and I've updated the PR to vendor 0.3.0 ;)

Thulium-Drake:
Thanks for this, this makes deploying/using the collection on RHEL based systems a million times easier! :-)

No more funky tasks using SCL pip :-)

mdellweg:
Now let's ask ansible to vendor requests. :facepalm:

Thulium-Drake:
Not sure if that really is required, python-requests is a fairly well packaged library (at least under Debian/RHEL)

PR #1016, opened 2020-10-07
allow unsetting entities

### LINKED ISSUE #969: Unable to unset compute_profile in hostgroups via ansible
##### SUMMARY
It's not possible to unset `compute_profile` field via ansible. If we set `compute_profile` either via Ui or via Ansible we have only the possibility to unset the `compute_profile` field to `null` via UI not via ansible.

##### ISSUE TYPE
 - Bug Report

##### ANSIBLE VERSION

```
ansible 2.9.10
```

##### KATELLO/FOREMAN VERSION

```
2.1.2
```

##### APYPIE VERSION

```
Version: 0.2.2
```

##### STEPS TO REPRODUCE

1. Create a hostgroup with `compute_profile` set.
2. Set `compute_profile` in ansible module call to `None`

##### EXPECTED RESULTS

`Compute profile` in UI should show `Inherit ...` or is `empty`

##### ACTUAL RESULTS

Ansible runs into the following error:

```
fatal: [localhost]: FAILED! => {"changed": false, "msg": "Found no results while searching for compute_profiles with name=\"\""}
```

evgeni:
draft, needs apypie change

### DIFF
--- a/plugins/module_utils/foreman_helper.py
+++ b/plugins/module_utils/foreman_helper.py
@@ -78,6 +78,12 @@ ENTITY_KEYS = dict(
 )
 
 
+class NoEntity(object):
+    @staticmethod
+    def __bool__():
+        return False
+
+
 def _exception2fail_json(msg='Generic failure: {0}'):
     """
     Decorator to convert Python exceptions into Ansible errors that can be reported to the user.
@@ -662,7 +668,7 @@ class ForemanAnsibleModule(AnsibleModule):
 
     def find_resource_by(self, resource, search_field, value, **kwargs):
         if not value:
-            return None
+            return NoEntity()
         search = '{0}{1}"{2}"'.format(search_field, kwargs.pop('search_operator', '='), value)
         return self.find_resource(resource, search, **kwargs)
 
@@ -1442,7 +1448,7 @@ def _flatten_entity(entity, foreman_spec):
     if entity is None:
         entity = {}
     for key, value in entity.items():
-        if key in foreman_spec and foreman_spec[key].get('ensure', True) and (value is not None or foreman_spec[key].get('type', 'str') == 'entity'):
+        if key in foreman_spec and foreman_spec[key].get('ensure', True) and value is not None:
             spec = foreman_spec[key]
             flat_name = spec.get('flat_name', key)
             property_type = spec.get('type', 'str')
--- a/plugins/module_utils/foreman_helper.py
+++ b/plugins/module_utils/foreman_helper.py
@@ -661,6 +661,8 @@ class ForemanAnsibleModule(AnsibleModule):
         return result
 
     def find_resource_by(self, resource, search_field, value, **kwargs):
+        if not value:
+            return None
         search = '{0}{1}"{2}"'.format(search_field, kwargs.pop('search_operator', '='), value)
         return self.find_resource(resource, search, **kwargs)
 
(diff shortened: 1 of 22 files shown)


evgeni:
this is the apypie change needed: https://github.com/Apipie/apypie/pull/74

@cmeissner if you still want to try it out ;)

plugins/module_utils/foreman_helper.py
```diff
@@ -1437,12 +1439,15 @@ def _flatten_entity(entity, foreman_spec):
     if entity is None:
         entity = {}
     for key, value in entity.items():
-        if key in foreman_spec and foreman_spec[key].get('ensure', True) and value is not None:
+        if key in foreman_spec and foreman_spec[key].get('ensure', True) and (value is not None or foreman_spec[key].get('type', 'str') == 'entity'):
```
mdellweg: What other cases do we have, where a value can be `None`?
I think we filter all the `None`s we get from ansible...

tests/test_playbooks/domain.yml
```diff
@@ -27,25 +30,42 @@
...
+      vars:
+        domain_state: "present"
+        domain_dns_proxy: ""
+        expected_change: true
+    - name: unset dns_proxy again, no change
+      include_tasks: tasks/domain.yml
+      vars:
+        domain_state: "present"
+        domain_dns_proxy: ""
+        expected_change: false
```
mdellweg, line 66: :1st_place_medal:

plugins/module_utils/foreman_helper.py
evgeni: True, but we also (ab)use `_flatten_entity` to mangle what the API responds, and that's full of `None`.

mdellweg: We could (ab-)use `False` here. Or create our own `NoEntity` constant.

evgeni: Not sure I follow, how could we end up with `False` here? The arg spec says it's a string, so it will (or `None`).

I thought about `NoEntity` and ended up thinking it's too unobvious to the user (and then differs to the list behaviour, where unset is `set to []`)

mdellweg: I thought in `lookup_entity` if `""` we set it to `NoEntity`, which we define as a new class maybe, (or `False` or `""`).
Then it will not trigger the `None` test here and still work with the `if value` below.
I think the `None` behaving different if this is about entities is quite confusing.

mdellweg: The user will not see `NoEntity` but still specifying `""`.

```
class NoEntity:
    @staticmethod
    def __bool__():
        return False
```
Should do the trick.

evgeni: Ah, okay, yeah that might work. Lemme try it.

plugins/module_utils/foreman_helper.py
```diff
@@ -1445,7 +1453,10 @@ def _flatten_entity(entity, foreman_spec):
             flat_name = spec.get('flat_name', key)
             property_type = spec.get('type', 'str')
             if property_type == 'entity':
-                result[flat_name] = value['id']
+                if value:
```
evgeni: should this be kept, or become `if not isinstance(value, NoEntity)`? (then it also doesn't need the `__bool__`).

evgeni: `    "msg": "Failed to ensure entity state: TypeError: 'NoEntity' object has no attribute '__getitem__'"`

This answers my question, I guess.

evgeni: (that's Python2, if you wonder)

mdellweg: I thought about not even instantiating the NoEntity, but using the class (as a singleton). And would `if value is NoEntity:` work?

plugins/module_utils/foreman_helper.py
```diff
@@ -661,6 +665,8 @@ def find_resource(self, resource, search, params=None, failsafe=False, thin=None
         return result
 
     def find_resource_by(self, resource, search_field, value, **kwargs):
+        if not value:
+            return NoEntity()
```
```suggestion
            return NoEntity
```

plugins/module_utils/foreman_helper.py
```diff
@@ -1445,7 +1451,10 @@ def _flatten_entity(entity, foreman_spec):
             flat_name = spec.get('flat_name', key)
             property_type = spec.get('type', 'str')
             if property_type == 'entity':
-                result[flat_name] = value['id']
+                if not isinstance(value, NoEntity):
```
```suggestion
                if value is not NoEntity:
```
And you can switch branches to avoid the `not`.

mdellweg (commented):
That is what i thought of.

mdellweg (approved):

PR #1021, opened 2020-10-09
refactor inventory tests to allow testing different configurations

evgeni:
this is preparation for allowing tests in #986, essentially doing the following:
* make the inventory selection configurable (by using the playbook name as a prefix)
* refactoring the playbook to allow overrides of the common parameters
* splitting out the tasks into an own file, so they can be included in another playbook with other variables


mdellweg (approved):
Does not look too scary...

PR #1022, opened 2020-10-09
wrong module name in the example

markgraf:
I think "theforeman.foreman.foreman_smart_proxy" has one foreman too much...


evgeni (approved):
You can't have enough foreman!

PR #1023, opened 2020-10-09
subnet: externalipam_group support

evgeni:


evgeni:
@cmeissner based on https://github.com/theforeman/foreman/pull/8067 I think this should be enough to get the group feature working after all… could you please test in your env?

plugins/module_utils/foreman_helper.py
```diff
@@ -92,10 +92,12 @@ def inner(self, *args, **kwargs):
     return decor
 
 
-def _check_patch_needed(fixed_version=None, plugins=None):
+def _check_patch_needed(introduced_version=None, fixed_version=None, plugins=None):
     """
     Decorator to check whether a specific apidoc patch is required.
 
+    :param introduced_version: The version of Foreman the API bug was introduced.
+    :type introduced_version: str, optional
```
mdellweg, line 107: Somehow i was missing this, but didn't want to ask...

mdellweg (approved):
CI looks like a galaxy hickup.

plugins/module_utils/foreman_helper.py
evgeni: My very first version of that had it, but then I was like "fuck, we'll never need it". Well, never didn't last a week.

evgeni:
I'd like to wait with the merge until @cmeissner has a chance to verify the APIdoc patch is correct.

ehelms (approved):

cmeissner:
@evgeni unfortunately I ran still in this error:

```... Unsupported parameters for (theforeman.foreman.subnet) module: externalipam_group ...```

Are there any needs of updating components. I only installed the latest version (1.4.0) of the glory ansible modules.

mdellweg:
> @evgeni unfortunately I ran still in this error:
> 
> `... Unsupported parameters for (theforeman.foreman.subnet) module: externalipam_group ...`
> 
> Are there any needs of updating components. I only installed the latest version (1.4.0) of the glory ansible modules.

Well, it's not even merged to develop, so you would need to test with this feature branch.

I theory, this should work, https://docs.ansible.com/ansible/latest/user_guide/collections_using.html#installing-a-collection-from-a-git-repository

I think the best option is to get a local copy of this repository, check out evgeni's branch and run make dist. You can then install the collection from the generated tar.gz or point your ansible.cfg to the build/collections directory (see https://github.com/theforeman/foreman-ansible-modules/blob/develop/ansible.cfg#L2).

Whatever works for you should be a good addition to:
https://github.com/theforeman/foreman-ansible-modules/tree/develop/docs#installation

But you can also "just" apply the diff as a patch to your installed version, or copy the files there.

cmeissner (commented):
With ansible modules build from feature branch it works as expected. ﻿

evgeni:
Perfect, thanks for testing!

mdellweg:
> With ansible modules build from feature branch it works as expected.

/me `make dance`

Issue #1024, opened 2020-10-14
parameters: add option "hidden" to parameters to not show values in gui


fgoebel:
- [x] global_parameter
- [ ] smart_class_parameter
- [ ] nested parameter (org, loc)
- [ ] domain, subnet, host, hostgroup, ... ("direct" params, for the lack of a better name)

##### SUMMARY
When I set parameters with these modules (global_parameters, organizations, locations,...) it is not possible to toggle the "hidden" option like in the gui.

```
# task example:
- theforeman.foreman.location:
    username: "{{ foreman_web_user }}"
    password: "{{ foreman_web_password }}"
    server_url: "{{ foreman_server }}"
    name: "some location"
    organizations:
    - "some Org"
    state: present
    parameters:
      - name: password
        parameter_type: string
        value: verySecretPasswordIdoNotWantToShowInGui
        hidden: true
```

What do you think?

##### ISSUE TYPE
 - Feature Idea

mdellweg:
At least for `global_parameter` this should be fairly easy to implement.
Do you want to give it a try?

mdellweg:
Oh, i went ahead, and did some experiments. Turns out, that hidden values are also hidden from the api. So creating a hidden parameter will not be idempotent. Also `hidden_value` is reported as `hidden_value?` by the API. This needs a little bit more designing.

https://theforeman.org/api/2.1/apidoc/v2/common_parameters/show.html

fgoebel:
hey mdellweg,
thank you for your first test and your answer to my feature request.

I have never done any development yet, so i think i can barely help with this.
But if you would maybe assist me, I am happy to test your changes and maybe get some development started this way!

mdellweg:
I removed the "easy" label, because the thing is a bit tricky. We are in the process of pushing a breaking change through apypie to solve this. But, i invite you to test my PR #1025 .

evgeni:
#1025 is now merged.

while reviewing it, we noticed that `smart_class_parameter` has partial support for hidden parameters already, but is missing the idempotency bit that @mdellweg implemented for `global_parameter` in #1025 

and this still needs porting to the other parameter handlers we have (yay)

ExplGamma:
Linking for posterity: https://github.com/theforeman/foreman-ansible-modules/pull/1785 added support for creating hidden params. Unfortunately I've been unable to get them working with hostgroups when updating params (ie just trying to set previously non-hidden params as hidden)

PR #1025, opened 2020-10-16
Support hidden parameters in global_parameter

mdellweg:
affects #1024

### DIFF
--- a/changelogs/fragments/global_parameter-hidden_values.yaml
+++ b/changelogs/fragments/global_parameter-hidden_values.yaml
@@ -1,0 +1,2 @@
+minor_changes:
+  - allow to set hidden flag on ``global_parameter`` (https://github.com/theforeman/foreman-ansible-modules/issues/1024)
--- a/changelogs/fragments/global_parameter-hidden_values.yaml
+++ b/changelogs/fragments/global_parameter-hidden_values.yaml
@@ -1,2 +1,2 @@
 minor_changes:
-  - allow to set hidden flag on ``global_parameter`` (https://github.com/theforeman/foreman-ansible-modules/issues/1024)
+  - global_parameter - allow to set hidden flag (https://github.com/theforeman/foreman-ansible-modules/issues/1024)
--- a/plugins/module_utils/foreman_helper.py
+++ b/plugins/module_utils/foreman_helper.py
@@ -514,24 +514,6 @@ class ForemanAnsibleModule(AnsibleModule):
             if create_param is not None and update_param is None:
                 _content_view_filter_rule_update['params'].append(create_param)
 
-    @_check_patch_needed()
-    def _patch_common_parameter_index_show_api(self):
-        """
-        This is a workaround for wrong validation of the 'show_hidden' querystring in 'common_parameter' index and show.
-        """
-
-        _common_parameter_methods = self.foremanapi.apidoc['docs']['resources']['common_parameters']['methods']
-
-        _common_parameter_index = next(x for x in _common_parameter_methods if x['name'] == 'index')
-        _common_parameter_show = next(x for x in _common_parameter_methods if x['name'] == 'show')
-
-        _index_param = next((x for x in _common_parameter_index['params'] if x['name'] == 'show_hidden'), None)
-        _show_param = next((x for x in _common_parameter_show['params'] if x['name'] == 'show_hidden'), None)
-        if _index_param is not None:
-            _index_param['expected_type'] = 'string'
-        if _show_param is not None:
-            _show_param['expected_type'] = 'string'
-
     def check_requirements(self):
         if not HAS_APYPIE:
             self.fail_json(msg=missing_required_lib("requests"), exception=APYPIE_IMP_ERR)
@@ -570,7 +552,6 @@ class ForemanAnsibleModule(AnsibleModule):
         self._patch_templates_resource_name()
         self._patch_location_api()
         self._patch_subnet_rex_api()
-        self._patch_common_parameter_index_show_api()
 
         # Katello
         self._patch_content_uploads_update_api()
--- a/plugins/module_utils/foreman_helper.py
+++ b/plugins/module_utils/foreman_helper.py
@@ -514,6 +514,24 @@ class ForemanAnsibleModule(AnsibleModule):
             if create_param is not None and update_param is None:
                 _content_view_filter_rule_update['params'].append(create_param)
 
+    @_check_patch_needed()
+    def _patch_common_parameter_index_show_api(self):
+        """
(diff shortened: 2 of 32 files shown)


plugins/module_utils/foreman_helper.py
```diff
@@ -511,6 +511,24 @@ def _patch_cv_filter_rule_api(self):
             if create_param is not None and update_param is None:
                 _content_view_filter_rule_update['params'].append(create_param)
 
+    @_check_patch_needed()
+    def _patch_common_parameter_index_show_api(self):
+        """
+        This is a workaround for wrong validation of the 'show_missing' querystring in 'common_parameter' index and show.
+        """
```
mdellweg: Not sure if this is an api bug or an apidoc bug.

plugins/modules/global_parameter.py
```diff
@@ -111,14 +111,18 @@
 
 
 class ForemanCommonParameterModule(ForemanEntityAnsibleModule):
-    pass
+    def remove_sensitive_fields(self, entity):
+        if entity and 'hidden_value?' in entity:
+            entity['hidden_value'] = entity.pop('hidden_value?')
+        return super(ForemanCommonParameterModule, self).remove_sensitive_fields(entity)
```
mdellweg, line 126: Should we remove value if `hidden_value` is `True`?

plugins/module_utils/foreman_helper.py
evgeni: You mean show_hidden, not show_missing, right? Why is bool not the right validator here?

evgeni: You mean show_hidden, not show_missing, right? Why is bool not the right validator here?

evgeni: Or do we just fail to encode them properly when they are passed as GET params and not inside JSON?

plugins/modules/global_parameter.py
evgeni: I think that's reasonable, yes

evgeni (commented):
Comments

plugins/module_utils/foreman_helper.py
mdellweg: Yes, show_hidden.

apidoc:
https://theforeman.org/api/2.1/apidoc/v2/common_parameters/index.html
api:
https://github.com/theforeman/foreman/blob/ae96cb70130968bdcc8b7a9a9af7edc7c6e0d9a8/app/views/api/v2/common/show_hidden.json.rabl

It only works with string `"true"`.

evgeni: Right. GET params are always strings, as that's the only thing that you can pass (contrary to JSON which has types). The API doc defines it as bool, so the api could potentially convert it properly, but doesn't.

I think the right fix here is not to patch the apidoc but make our code (or rather, apypie) send booleans as true/false (lowercase strings) when passing them as GET params (right now it sends True/False (well, `str(value)`)).

evgeni: FWIW, the api should also accept the string "1", which it currently doesn't, but that's not our problem.

mdellweg: so it's an apypie bug?

evgeni: I think it is, yeah.

plugins/modules/global_parameter.py
```diff
@@ -132,7 +145,7 @@ def main():
     )
 
     with module.api_connection():
-        entity = module.lookup_entity('entity')
+        entity = module.lookup_entity('entity', params={'show_hidden': True})
```
evgeni, line 148: smart_class_parameter also has support for `hidden_value`, but doesn't pass `show_hidden`, is that a bug (in smart_class_parameter)?

plugins/modules/global_parameter.py
```diff
@@ -141,6 +154,8 @@ def main():
             module.foreman_params['value'] = parameter_value_to_str(module.foreman_params['value'], module.foreman_params['parameter_type'])
             if entity and 'value' in entity:
                 entity['value'] = parameter_value_to_str(entity['value'], entity.get('parameter_type', 'string'))
+            if entity and 'hidden_value?' in entity:
+                entity['hidden_value'] = entity.pop('hidden_value?')
```
evgeni, line 158: this smells like we'll have a HiddenValueEntity at some point, as the same code also exists in smart_class_parameter ;)

changelogs/fragments/global_parameter-hidden_values.yaml
```diff
@@ -0,0 +1,2 @@
+minor_changes:
+  - allow to set hidden flag on ``global_parameter`` (https://github.com/theforeman/foreman-ansible-modules/issues/1024)
```
evgeni:
```suggestion
  - global_parameter - allow to set hidden flag (https://github.com/theforeman/foreman-ansible-modules/issues/1024)
```

evgeni:
the host and activation_key fixtures look broken, but I don't think you had to re-record them anyways?

tests/test_playbooks/fixtures/activation_key-11.yml
```diff
@@ -314,12 +252,14 @@ interactions:
...
-      Cookie:
-      - _session_id=4981a4276e1accb5572d962fd7ebc583
       User-Agent:
       - apypie (https://github.com/Apipie/apypie)
     method: GET
+<<<<<<< HEAD
     uri: https://centos7-katello-3-14.yatsu.example.com/katello/api/activation_keys/3/product_content?content_access_mode_all=true
+=======
+    uri: https://foreman.example.org/katello/api/activation_keys/3/product_content?content_access_mode_all=true
+>>>>>>> 128f9666... fixup! Support hidden parameters in global_parameter
```
evgeni: yeah, broken

mdellweg:
> the host and activation_key fixtures look broken, but I don't think you had to re-record them anyways?

Technically i had to, before you "fixed" them in #1013

plugins/modules/global_parameter.py
mdellweg: That probably means that we do not yet support hidden values for smart_class_parameter.
We are also still missing that feature for nested parameter in host, domain, you-name-it.

plugins/modules/global_parameter.py
mdellweg: :nose: not today?

evgeni: absolutely, just wanted to raise awareness

plugins/modules/global_parameter.py
evgeni: not support them in an idempotent way? but yeah.

evgeni (approved):

PR #1026, opened 2020-10-16
WIP: Added logic to set lifecycle_environments for smart_proxies.

Thulium-Drake:
This commit uses the logic from sync_plans to attach lifecycle_environments to smart_proxies.

~~WIP: Does not fully work yet!~~ Works, needs tests


mdellweg:
Looks good on a first glance.
Things to be done at least:
- ~[ ] we need a changelog~
- [x] the docstring needs `elements: str`
- [x] I'd like to have tests for this feature (too much code to just believe it...)

Thulium-Drake:
It does not fully work yet, I'm having issues converting the lifecycle_environments to their ID numbers. And then requesting the current ids on the smart_proxy entity itself. I'm using ```module.lookup_entity('lifecycle_environments')``` now, but it throws the following error:

```
msg: 'Failed to list resource: KeyError: "Unknown action ''index''. Supported actions: add_lifecycle_environment, available_lifecycle_environments, cancel_sync, lifecycle_environments, remove_lifecycle_environment, sync, sync_status"'
```

I tried the adding/removing logic by hardcoding ```desired_lifecycle_environment_ids``` and ```current_lifecycle_environment_ids``` and it works fine.

"Just" need to get the IDs :-)

evgeni:
I think in this case you don't need to set the resource type. Or set it to lifecycle_environments, because that's the resource you're looking for. Caosule_content is just the endpoint you need for assignment later on.

evgeni:
Oh, and given the smart proxy module was in no released version yet, I don't think this *needs* a changelog entry.

Tests would be good tho.

Thulium-Drake:
@evgeni Thanks for the pointers! I got it working :-)

With regards to the tests, do I 'just' add stuff to the playbook in the tests directory?

EDIT: it works, the results are idempotent (I can run it multiple times against my server without it complaining), but it always returns as Changed. Not sure where that came from, anyone have a clue? :-)

mdellweg:
We have dedicated to writing (extending) the tests:

https://github.com/theforeman/foreman-ansible-modules/blob/develop/docs/testing.md#writing-tests

evgeni:
Essentially, yes. But also no ;)

Have a look at the testing docs: https://theforeman.github.io/foreman-ansible-modules/develop/testing.html

And my recent post about writing modules: 
https://theforeman.org/2020/10/how-to-write-a-new-foreman-ansible-module-in-20-lines-of-code.html

plugins/modules/smart_proxy.py
```diff
@@ -101,13 +110,53 @@ def main():
...
+        if handle_lifecycle_environments:
+            module.lookup_entity('lifecycle_environments')
+
+        lifecycle_environments = module.foreman_params.pop('lifecycle_environments', None)
+        smart_proxy = module.run()
+
+        payload = {
+            'id': smart_proxy['id'],
+        }
+        entity = module.resource_action('capsule_content', 'lifecycle_environments', payload)
```
mdellweg: This will not work in check_mode.
Also you need to add `record_change=False` as this is only a lookup call.

plugins/modules/smart_proxy.py
```diff
@@ -127,7 +127,7 @@ def main():
         payload = {
             'id': smart_proxy['id'],
         }
-        entity = module.resource_action('capsule_content', 'lifecycle_environments', payload)
+        entity = module.resource_action('capsule_content', 'lifecycle_environments', payload, record_change=False)
```
mdellweg: :+1:

mdellweg: But still in "check_mode" when you just faked to create a new smart_proxy, but didn't, this call will fail.
And i just realize, our testing strategy does not cover this.

Thulium-Drake: What would need to be done to fix that? :-)

mdellweg: You would hold the `fake_entity` from here:
https://github.com/theforeman/foreman-ansible-modules/blob/fe0f1d0d6f1ccd5cf2d00ff0794c7f5b940a6a32/plugins/module_utils/foreman_helper.py#L898

So you would not do that call but fake the result (as an empty list).

plugins/modules/smart_proxy.py
```diff
@@ -101,13 +110,53 @@ def main():
...
+        if handle_lifecycle_environments:
+            module.lookup_entity('lifecycle_environments')
+
+        lifecycle_environments = module.foreman_params.pop('lifecycle_environments', None)
+        smart_proxy = module.run()
+
+        payload = {
+            'id': smart_proxy['id'],
+        }
+        entity = module.resource_action('capsule_content', 'lifecycle_environments', payload, record_change=False)
```
evgeni: FWIW, I think we usually use `entity` for the *main* entity of the module (so in this case the `smart_proxy`) and store other things in differently named variables.

plugins/modules/smart_proxy.py
```diff
@@ -101,13 +110,53 @@ def main():
...
+
+        if handle_lifecycle_environments:
+            desired_environment_ids = set(lifecycle_environment['id'] for lifecycle_environment in lifecycle_environments)
+            current_environment_ids = set(lifecycle_environment['id'] for lifecycle_environment in entity['results']) if entity else set()
```
evgeni: e.g. here, `if entity` does it refer to the smart proxy or to the "set of LCEs attached to it"? (both work, but the later is not really an entity)

plugins/modules/smart_proxy.py
Thulium-Drake: I've renamed it smart_proxy_content (that sounded logical to me :) )

plugins/modules/smart_proxy.py
Thulium-Drake: gothcha, fixed it :-)

plugins/modules/smart_proxy.py
```diff
@@ -101,13 +110,53 @@ def main():
...
+
+            module.record_before('smart_proxy_content/lifecycle_environment_ids', current_environment_ids)
+            module.record_after('smart_proxy_content/lifecycle_environment_ids', desired_environment_ids)
+            module.record_after_full('smart_proxy_content/lifecycle_environment_ids', desired_environment_ids)
+
+            if desired_environment_ids != current_environment_ids:
+                if not module.check_mode:
```
evgeni: Is that needed? `resource_action` is noop if we're in check mode (and didn't pass `ignore_check_mode`) and will call `set_changed` for us.

Thulium-Drake: I removed it and it still works as I want, so I guess not. This would warrant a closer look at the code from the sync_plan module. As I copied that code to make this.

plugins/modules/smart_proxy.py
Thulium-Drake: I tried running the module with check_mode, it does what you describe:

```
--- before
+++ after
@@ -10,4 +10,4 @@
   - 3
   url: https://fm-proxy.dev.foreman.lab:9090
 smart_proxy_content/lifecycle_environment_ids:
-- []
+- - 4
```

plugins/modules/smart_proxy.py
```diff
@@ -101,13 +110,50 @@ def main():
...
+        smart_proxy = module.run()
+
+        payload = {
+            'id': smart_proxy['id'],
+        }
+        smart_proxy_content = module.resource_action('capsule_content', 'lifecycle_environments', payload, record_change=False)
+
+        if handle_lifecycle_environments:
+            desired_environment_ids = set(lifecycle_environment['id'] for lifecycle_environment in lifecycle_environments)
+            current_environment_ids = set(lifecycle_environment['id'] for lifecycle_environment in smart_proxy_content['results']) if smart_proxy_content else set()
```
evgeni: line too long :roll_eyes: 
sometimes I don't like linters.

plugins/modules/smart_proxy.py
```diff
@@ -101,13 +110,50 @@ def main():
...
+        if handle_lifecycle_environments:
+            module.lookup_entity('lifecycle_environments')
+
+        lifecycle_environments = module.foreman_params.pop('lifecycle_environments', None)
+        smart_proxy = module.run()
+
+        payload = {
+            'id': smart_proxy['id'],
+        }
+        smart_proxy_content = module.resource_action('capsule_content', 'lifecycle_environments', payload, record_change=False)
```
evgeni:
```suggestion
        current_lces = module.resource_action('capsule_content', 'lifecycle_environments', payload, record_change=False)
```

this would make the lines short enough and is IMHO more correct, as the action returns you a list of LCEs, not a smart proxy with content object.

plugins/modules/smart_proxy.py
mdellweg: That's OK. Linters never seem to like me.

plugins/modules/smart_proxy.py
Thulium-Drake: Yeah, that sounds better, I now used the name of the endpoint, sec, disregard the current push :-)

plugins/modules/smart_proxy.py
Thulium-Drake: fixed

plugins/modules/smart_proxy.py
```diff
@@ -101,13 +110,50 @@ def main():
...
+        if handle_lifecycle_environments:
+            module.lookup_entity('lifecycle_environments')
+
+        lifecycle_environments = module.foreman_params.pop('lifecycle_environments', None)
+        smart_proxy = module.run()
+
+        payload = {
+            'id': smart_proxy['id'],
+        }
+        current_lces = module.resource_action('capsule_content', 'lifecycle_environments', payload, record_change=False)
```
evgeni: This needs to be guarded by a `if handle_lifecycle_environments` as otherwise this is also executed on non-Katello installations and fails. (so just move the payload and current lines behind the `if` in the next line?)

plugins/modules/smart_proxy.py
```diff
@@ -101,13 +110,50 @@ def main():
...
+        required_plugins=[('katello', ['lifecycle_environments', 'download_policy'])],
     )
 
     with module.api_connection():
-        module.run()
+        handle_lifecycle_environments = not module.desired_absent and 'lifecycle_environments' in module.foreman_params
+        if handle_lifecycle_environments:
+            module.lookup_entity('lifecycle_environments')
+
+        lifecycle_environments = module.foreman_params.pop('lifecycle_environments', None)
```
evgeni:
```suggestion
        lifecycle_environments = module.foreman_params.pop('lifecycle_environments', [])
```

plugins/modules/smart_proxy.py
```diff
@@ -28,12 +28,19 @@
...
+  - "Jeffrey van Pelt (@Thulium-Drake)"
 options:
   name:
     description:
       - Name of the Smart Proxy
     required: true
     type: str
+  lifecycle_environments:
+    description:
+      - Lifecycle Environments synced to the Smart Proxy
```
evgeni:
```suggestion
      - Lifecycle Environments synced to the Smart Proxy.
      - Only available for Katello installations.
```

tests/test_playbooks/katello_smart_proxy.yml
```diff
@@ -34,6 +34,18 @@
       vars:
         smart_proxy_state: present
         expected_change: false
+    - include_tasks: tasks/smart_proxy.yml
+      vars:
+        smart_proxy_state: present
+        smart_proxy_lifecycle_environments:
+          - 'Dev'
```
evgeni: You'll have to create this env like in https://github.com/theforeman/foreman-ansible-modules/blob/6a4c479bec608c3d430bb94d9e86a0e09ac5173b/tests/test_playbooks/content_view_version.yml#L18-L33

Also probably cool to have a tests that replaces Dev with Test (or smth similar, so we also test the "remove" branch)

evgeni:
Do you know how to setup an environment to re-record the test fixtures? Or do you need help with that?

plugins/modules/smart_proxy.py
mdellweg: Hmm, beginning to think, that our tests actually capture this. ;)
Let's see whether the check_mode test will pass.

plugins/modules/smart_proxy.py
Thulium-Drake: Makes sense, fixed!

plugins/modules/smart_proxy.py
Thulium-Drake: fixed!

plugins/modules/smart_proxy.py
Thulium-Drake: Fixed!

tests/test_playbooks/katello_smart_proxy.yml
Thulium-Drake: I made some changed to the tests, is that what you meant? :-) It now created 2 LCE's and tries to bind them one at a time.

Thulium-Drake:
@evgeni Not exactly, do I need to mod an existing 'regular' installation of Foreman (with a Smart Proxy in this case)?

I read a few notes that the primary Smart Proxy should be modded a bit for the tests

tests/test_playbooks/katello_smart_proxy.yml
evgeni: Yes, that's exactly what I wanted! Thanks!

Thulium-Drake: Awesome, nice!

mdellweg:
I'd recommend to use a throwaway forklift box (centos7-katello-3.17 or similar) for this purpose.

Thulium-Drake:
Check, I'll need to prep a system with Vagrant, this will need some time (need to prepare a laptop with plain libvirt first, I only have Proxmox or Docker boxes atm).

evgeni:
We might also just be able to nicely ask @mdellweg to record this for us, as he has a setup that works ;)

mdellweg:
> Check, I'll need to prep a system with Vagrant, this will need some time (need to prepare a laptop with plain libvirt first, I only have Proxmox or Docker boxes atm).

That should also work... hard to say. Just don't use one of you production machines.

Thulium-Drake:
It's OK, I needed an excuse to reimage my old laptop anyway ;-)

I'll let you know!

evgeni:
given you added tests, I would expect the number of fixtures to increase, did you forget to `git add`? :)

Thulium-Drake:
Nope.. it only mentioned changes, here's what I did:

```
$ make record_smart_proxy
rm -f tests/test_playbooks/fixtures/smart_proxy-*.yml
pytest -v 'tests/test_crud.py::test_crud[smart_proxy]' --vcrmode record 
================================================ test session starts ================================================
platform linux2 -- Python 2.7.16, pytest-3.10.1, py-1.7.0, pluggy-0.8.0 -- /usr/bin/python2
cachedir: .pytest_cache
rootdir: /home/jeff/git/foreman-ansible-modules, inifile:
plugins: vcr-1.0.2, xdist-1.26.1, forked-1.0.1
collected 1 item                                                                                                    

tests/test_crud.py::test_crud[smart_proxy] PASSED                                                             [100%]

============================================= 1 passed in 12.64 seconds =============================================
$ git status
On branch smart_proxy_lifecycles
Your branch is up to date with 'origin/smart_proxy_lifecycles'.

Changes not staged for commit:
  (use "git add <file>..." to update what will be committed)
  (use "git checkout -- <file>..." to discard changes in working directory)

	modified:   tests/test_playbooks/fixtures/smart_proxy-0.yml
	modified:   tests/test_playbooks/fixtures/smart_proxy-1.yml
	modified:   tests/test_playbooks/fixtures/smart_proxy-2.yml
	modified:   tests/test_playbooks/fixtures/smart_proxy-3.yml
```

Also checked the results on my laptop:
```
$ make test_smart_proxy
install -m 644 -DT plugins/modules/smart_proxy.py build/src/plugins/modules/smart_proxy.py
ansible-galaxy collection build build/src --force
Created collection for theforeman.foreman at /home/jeff/git/foreman-ansible-modules/theforeman-foreman-1.3.0.tar.gz
ansible-galaxy collection install -p build/collections theforeman-foreman-1.3.0.tar.gz --force
Process install dependency map
Starting collection install process
Installing 'theforeman.foreman:1.3.0' to '/home/jeff/git/foreman-ansible-modules/build/collections/ansible_collections/theforeman/foreman'
pytest -v 'tests/test_crud.py::test_crud[smart_proxy]' 'tests/test_crud.py::test_check_mode[smart_proxy]' 
================================================ test session starts ================================================
platform linux2 -- Python 2.7.16, pytest-3.10.1, py-1.7.0, pluggy-0.8.0 -- /usr/bin/python2
cachedir: .pytest_cache
rootdir: /home/jeff/git/foreman-ansible-modules, inifile:
plugins: vcr-1.0.2, xdist-1.26.1, forked-1.0.1
collected 2 items                                                                                                   

tests/test_crud.py::test_crud[smart_proxy] PASSED                                                             [ 50%]
tests/test_crud.py::test_check_mode[smart_proxy] PASSED                                                       [100%]

============================================= 2 passed in 11.65 seconds =============================================
```

Some things don't seem to agree with each other ;-)

evgeni:
aaah
you need to `make record_katello_smart_proxy` instead, the ugly cases where the tests are split into multiple playbooks ;)

(and please drop the updated `smart_proxy` fixtures, as they are not really updated)

Thulium-Drake:
Aaaah, gotcha, hold my tea!

Thulium-Drake:
This one's a bit weird... It's a default installation made with Vagrant, but it doesn't have Pulp node? (whereas the Webinterface does show Pulp and Pulpcore features)

```
TASK [Ensure Smart Proxy 'Smart Proxy' is present] *****************************
task path: /home/jeff/git/foreman-ansible-modules/tests/test_playbooks/tasks/smart_proxy.yml:2
<tests> ESTABLISH LOCAL CONNECTION FOR USER: jeff
<tests> EXEC /bin/sh -c 'echo ~jeff && sleep 0'
<tests> EXEC /bin/sh -c '( umask 77 && mkdir -p "` echo /home/jeff/.ansible/tmp `"&& mkdir "` echo /home/jeff/.ansible/tmp/ansible-tmp-1603179344.23-27797-139558929942926 `" && echo ansible-tmp-1603179344.23-27797-139558929942926="` echo /home/jeff/.ansible/tmp/ansible-tmp-1603179344.23-27797-139558929942926 `" ) && sleep 0'
Using module file /home/jeff/git/foreman-ansible-modules/build/collections/ansible_collections/theforeman/foreman/plugins/modules/smart_proxy.py
<tests> PUT /home/jeff/.ansible/tmp/ansible-local-27430xhmJ3v/tmpRIYBSK TO /home/jeff/.ansible/tmp/ansible-tmp-1603179344.23-27797-139558929942926/AnsiballZ_smart_proxy.py
<tests> EXEC /bin/sh -c 'chmod u+x /home/jeff/.ansible/tmp/ansible-tmp-1603179344.23-27797-139558929942926/ /home/jeff/.ansible/tmp/ansible-tmp-1603179344.23-27797-139558929942926/AnsiballZ_smart_proxy.py && sleep 0'
<tests> EXEC /bin/sh -c '../vcr_python_wrapper.py /home/jeff/.ansible/tmp/ansible-tmp-1603179344.23-27797-139558929942926/AnsiballZ_smart_proxy.py && sleep 0'
<tests> EXEC /bin/sh -c 'rm -f -r /home/jeff/.ansible/tmp/ansible-tmp-1603179344.23-27797-139558929942926/ > /dev/null 2>&1 && sleep 0'
    raise HTTPError(http_error_msg, response=self)
fatal: [tests]: FAILED! => {
    "changed": false, 
    "error": {
        "displayMessage": "This request may only be performed on a Smart proxy that has the Pulp Node feature.", 
        "errors": [
            "This request may only be performed on a Smart proxy that has the Pulp Node feature."
        ]
    }, 
    "invocation": {
        "module_args": {
            "download_policy": null, 
            "lifecycle_environments": [
                "Dev"
            ], 
            "locations": null, 
            "name": "Smart Proxy", 
            "organizations": null, 
            "server_url": "https://foreman.example.com", 
            "state": "present", 
            "url": "http://localhost:8000", 
    The full traceback is:
  File "/tmp/ansible_smart_proxy_payload_cye86cgq/ansible_smart_proxy_payload.zip/ansible_collections/theforeman/foreman/plugins/module_utils/foreman_helper.py", line 1033, in resource_action
    result = self._resource_call(resource, action, resource_payload, options=options, data=data, files=files)
  File "/tmp/ansible_smart_proxy_payload_cye86cgq/ansible_smart_proxy_payload.zip/ansible_collections/theforeman/foreman/plugins/module_utils/foreman_helper.py", line 577, in _resource_call
    return self._resource(resource).call(*args, **kwargs)
  File "/home/jeff/.local/lib/python3.7/site-packages/apypie/resource.py", line 74, in call
    return self.api.call(self.name, action, params, headers, options, data, files)
  File "/home/jeff/.local/lib/python3.7/site-packages/apypie/api.py", line 243, in call
    return self._call_action(action, params, headers, data, files)
  File "/home/jeff/.local/lib/python3.7/site-packages/apypie/api.py", line 255, in _call_action
    headers, data, files)
  File "/home/jeff/.local/lib/python3.7/site-packages/apypie/api.py", line 293, in http_call
    request.raise_for_status()
  File "/usr/lib/python3/dist-packages/requests/models.py", line 940, in raise_for_status
    raise HTTPError(http_error_msg, response=self)
fatal: [tests]: FAILED! => {
    "changed": false, 
    "error": {
        "displayMessage": "This request may only be performed on a Smart proxy that has the Pulp Node feature.", 
        "errors": [
            "This request may only be performed on a Smart proxy that has the Pulp Node feature."
        ]
    }, 
    "invocation": {
        "module_args": {
            "download_policy": null, 
            "lifecycle_environments": [
                "Dev"
            ], 
            "locations": null, 
            "name": "Smart Proxy", 
            "organizations": null, 
            "server_url": "https://foreman.example.com", 
            "state": "present", 
            "url": "http://localhost:8000", 
            "username": "admin", 
            "validate_certs": false
        }
    }, 
    "msg": "Error while performing lifecycle_environments on capsule_content: 500 Server Error: Internal Server Error for url: https://foreman.example.com/katello/api/capsules/7/content/lifecycle_environments"
```

evgeni:
Heh. I think that's because the pulpnode is associated with the FQDN name, not the `localhost` name. But not exactly sure.

Thulium-Drake:
Hmm, when I open 'Smart Proxy' in the proxies list I see a lot of 404 errors, see below)

```
2020-10-20T07:46:34 [I|app|bcdc4162] Backtrace for 'ProxyAPI operation FAILED' error (ProxyAPI::ProxyException): ERF12-3113 [ProxyAPI::ProxyException]: Unable to detect puppet path ([RestClient::NotFound]: 404 Not Found) for proxy http://localhost:8000/pulp/status/puppet
2020-10-20T07:46:37 [I|app|13fcb158] Backtrace for 'Failed to index' error (Foreman::WrappedException): ERF50-5345 [Foreman::WrappedException]: Unable to connect ([ProxyAPI::ProxyException]: ERF12-5356 [ProxyAPI::ProxyException]: Unable to get PuppetCA certificates ([RestClient::NotFound]: 404 Not Found) for proxy http://localhost:8000/puppet/ca)
```

Thulium-Drake:
I've tried a few URLs, but I can't get the Pulp node to talk to me.

Which triggered me to look in my other lab, Pulp Node is actually a different feature.. So I'm not sure if this test can be done with a single node Foreman (which would also explain why capsule_content is a different API endpoint, as the primary smart proxy doesn't do selective LCE, all of them are on there by default)

evgeni:
Yeah, I think you're right, this requires a "real" proxy to be tested.

@mdellweg thoughts?

mdellweg:
Oh, yeah, the proxy on the same machine is probably not even able to install another instance of pulp (resource conflicts) and therefore cannot have pulp_node feature.
Isn't there a katello-proxy box configured in forklift?

Thulium-Drake:
I don't see a specific katello-proxy in this file https://github.com/theforeman/forklift/blob/master/vagrant/boxes.d/99-local.yaml.example

It does mention a REX proxy, but I'm not sure if it will install what we need

mdellweg:
Hmm, Yeah.
There are `*-foreman-proxy-*` boxes. Don't know if they include `pulp-node` or if it can be installed into them.

evgeni:
the foreman-proxy boxes are really katello proxies ;)

so yeah, you can use them

Thulium-Drake:
spinning them up now!

plugins/modules/smart_proxy.py
```diff
@@ -101,13 +111,49 @@ def main():
...
+        required_plugins=[('katello', ['lifecycle_environments', 'download_policy'])],
     )
 
     with module.api_connection():
-        module.run()
+        handle_lifecycle_environments = not module.desired_absent and 'lifecycle_environments' in module.foreman_params
+        if handle_lifecycle_environments:
+            module.lookup_entity('lifecycle_environments')
+
+        lifecycle_environments = module.foreman_params.pop('lifecycle_environments', [])
```
mdellweg:
```suggestion
            module.lookup_entity('lifecycle_environments')
            lifecycle_environments = module.foreman_params.pop('lifecycle_environments', [])

```

plugins/modules/smart_proxy.py
```diff
@@ -101,13 +111,49 @@ def main():
...
+        smart_proxy = module.run()
+
+        if handle_lifecycle_environments:
+            payload = {
+                'id': smart_proxy['id'],
+            }
+            current_lces = module.resource_action('capsule_content', 'lifecycle_environments', payload, record_change=False)
```
mdellweg: There's still some thought needed around this in check_mode.

plugins/modules/smart_proxy.py
Thulium-Drake: fixed!

tests/test_playbooks/katello_smart_proxy.yml
```diff
@@ -34,6 +44,50 @@
       vars:
         smart_proxy_state: present
         expected_change: false
+    - include_tasks: tasks/location.yml
```
mdellweg, line 31: All those tasks should be part of the localhost play (and therefore not be recorded).

tests/test_playbooks/katello_smart_proxy.yml
```diff
@@ -34,6 +44,50 @@
...
+          label: dev
+          prior: Library
+        - name: Test
+          label: test
+          prior: Dev
+    - include_tasks: tasks/smart_proxy.yml
+      vars:
+        smart_proxy_state: present
+        smart_proxy_lifecycle_environments:
+          - 'Dev'
```
mdellweg, line 64: :+1: that we add an env in the first creation call.
This is really an important edge case for the check_mode logic.

plugins/modules/smart_proxy.py
Thulium-Drake: So basically what we need is to give check mode the list of what _should_ happen, right?

I've been trying to make the code do that, but it doesn't seem to work, it still barfs with

```
TASK [assert] ******************************************************************
task path: /home/jeff/git/foreman-ansible-modules/tests/test_playbooks/tasks/smart_proxy.yml:17
fatal: [tests]: FAILED! => {
    "assertion": "result.changed == expected_change", 
    "changed": false, 
    "evaluated_to": false, 
    "msg": "Ensuring Smart Proxy is present failed! (expected_change: False)"

```

mdellweg: So let's try to get `check_mode` work. I think it should be:
```suggestion
            if smart_proxy['id'] == -1:
                # just created a fake entity, so fake the environments too
                current_lces = {'results': []}
            else:
                current_lces = module.resource_action('capsule_content', 'lifecycle_environments', payload, ignore_check_mode=True, record_change=False)
```

mdellweg (changes requested):
Impressive, you are almost there.

plugins/modules/smart_proxy.py
Thulium-Drake: Thanks a lot! That was the fix!

I was trying with ```module.check_mode``` and without the ```ignore_check_mode``` in the resource action :-)

tests/test_playbooks/katello_smart_proxy.yml
Thulium-Drake: fixed!

tests/test_playbooks/katello_smart_proxy.yml
mdellweg: Now i was wrong, sorry. Can you make this the first task in the test play?

mdellweg: I think, we technically don't need the two tasks that precede this.

Thulium-Drake: That depends, I had to create them in order to run these the smart_proxy test specifically on a 'clean' slate. What is the current preferred workflow for testing specific modules? Should they be self-containing where possible?

Thulium-Drake: Should be good now :-)

If it's OK, I'll update the fixtures

tests/test_playbooks/katello_smart_proxy.yml
```diff
@@ -26,6 +44,19 @@
...
+        lifecycle_environment_name: "{{ item.name }}"
+        lifecycle_environment_label: "{{ item.label }}"
+        lifecycle_environment_prior: "{{ item.prior }}"
+      loop:
+        - name: Dev
+          label: dev
+          prior: Library
+        - name: Test
+          label: test
+          prior: Dev
```
evgeni, line 51: I'd think this part should be on `localhost` (and thus not recorded), as this is essentially "set-up" which is only needed when we run against a live instance, in the recorded case the searches for the LCEs will work as those are still recorded as part of the `smart_proxy` module invocations.

tests/test_playbooks/katello_smart_proxy.yml
evgeni: See my comment in https://github.com/theforeman/foreman-ansible-modules/pull/1026/files?file-filters%5B%5D=.yml#r509003611 -- I think @mdellweg is wrong here and it should be part of the `localhost` play. As we need those LCEs *present* when we talk to a Katello box, but *assume* they are present when talking against recordings.

plugins/modules/smart_proxy.py
```diff
@@ -101,13 +111,53 @@ def main():
...
-        module.run()
+        handle_lifecycle_environments = not module.desired_absent and 'lifecycle_environments' in module.foreman_params
+        if handle_lifecycle_environments:
+            module.lookup_entity('lifecycle_environments')
+            lifecycle_environments = module.foreman_params.pop('lifecycle_environments', [])
+
+        smart_proxy = module.run()
+
+        if handle_lifecycle_environments:
+            if smart_proxy['id'] == -1:
```
evgeni: the `id = -1` this is something that we shouldn't expose in the modules. in other modules we lookup the entity twice and decide on the before/after state essentially:

https://github.com/theforeman/foreman-ansible-modules/blob/a5a7be69678eb7992882f95b12e5de26b7f37027/plugins/modules/role.py#L118-L128

we could introduce a "newly created" property, or something, but the `-1` should remain an implementation detail.

tests/test_playbooks/katello_smart_proxy.yml
Thulium-Drake: That sounds right, and the code to create them is already tested elsewhere.

Do you agree @mdellweg ? :-)

mdellweg: I think, i was misunderstood. I wanted that the very first task in the test section is a create smart proxy _with_ a lce.
And i was asking whether we are confident with deleting the two tasks that test the smart proxy without specifying any lce.

tests/test_playbooks/katello_smart_proxy.yml
mdellweg: Agreed. My request was to remove https://github.com/theforeman/foreman-ansible-modules/pull/1026/files#diff-6a0fba6bb0a7bef6d299c3dbccf84992c22c44143566227aa3ab2e983db0b993L29-R67

tests/test_playbooks/katello_smart_proxy.yml
Thulium-Drake: Oh, right, my bad, fixing it now :)

tests/test_playbooks/katello_smart_proxy.yml
Thulium-Drake: Fixed!

tests/test_playbooks/katello_smart_proxy.yml
Thulium-Drake: With regards to the first 2 tasks, I think we can, it still has to create one, except with some more options right now

plugins/modules/smart_proxy.py
Thulium-Drake: I made a fix, is this what you meant?

plugins/modules/smart_proxy.py
```diff
@@ -118,21 +118,21 @@ def main():
     )
 
     with module.api_connection():
+        smart_proxy = module.lookup_entity('entity')
+        new_smart_proxy = module.run()
```
mdellweg: Don't you want to have this call in the old place? So the `lifecycle_environments = module.foreman_params.pop('lifecycle_environments', [])` gets this parameter out of the way for the `run()`?

Thulium-Drake: Done! :-)

Thulium-Drake: I seem to have broken it..

plugins/modules/smart_proxy.py
```diff
@@ -120,8 +120,8 @@ def main():
     with module.api_connection():
         handle_lifecycle_environments = not module.desired_absent and 'lifecycle_environments' in module.foreman_params
         if handle_lifecycle_environments:
-            lifecycle_environments = module.foreman_params.pop('lifecycle_environments', [])
             module.lookup_entity('lifecycle_environments')
+            lifecycle_environments = module.foreman_params.pop('lifecycle_environments', [])
 
         smart_proxy = module.lookup_entity('entity')
         new_smart_proxy = module.run()
```
mdellweg: Yeah, that should also work.

plugins/modules/smart_proxy.py
Thulium-Drake: Fixed!

plugins/modules/smart_proxy.py
```diff
@@ -101,13 +111,55 @@ def main():
...
+
+        if handle_lifecycle_environments:
+
+            if smart_proxy:
+                payload = {
+                    'id': new_smart_proxy['id'],
+                }
+                current_lces = module.resource_action('capsule_content', 'lifecycle_environments', payload, ignore_check_mode=True, record_change=False)
+            else:
+                current_lces = {'results': []}
```
mdellweg, line 137: I just realized, this is not only for check_mode now, but also skipping the extra call if we just created a new entity and, therefore, know the result will be empty.
:1st_place_medal:

mdellweg (approved):
Nice piece of work!

evgeni (approved):

evgeni:
Thanks a ton @Thulium-Drake!

Thulium-Drake:
You're welcome! :-)

Is the module in time for the next update on the Ansible Collection? (So I can incorporate it in my foreman role ;-) )

@mdellweg Thanks for your time reviewing this!

evgeni:
Yes, it will be part of 1.4.0 release which will go out either today or tomorrow.

Issue #646, opened 2019-12-09
define rules for roles


evgeni:
##### SUMMARY
as per discussion in https://github.com/theforeman/foreman-ansible-modules/issues/535, we need a set of rules (besides naming) how roles should behave to be included in this collection.

copying from the original issue, we thought about:
* roles should be topic specific / atomic / small / however you want to name it (= do one job and do it well)
* roles need to be generic for their topic (= no special cases for "that one weird repository")
* related roles should be driven by similar input, to ease combining them
##### ISSUE TYPE
 - Feature Idea

evgeni:
cc @nesanton for the input from your customer

evgeni:
input from @nesanton:
* make roles "atomic" as in "only one family of modules", a foreman-related role should not use katello modules, and vice versa. any further glueing should be done via variables.
* roles need either have sane defaults that just work or a sane way to bail out and tell the user that there are missing prerequisites.

@evgeni:
* avoid Satellite naming in variables/tasks/role names etc
* copy https://github.com/RedHatSatellite/satellite-ansible-roles/blob/master/README.md

evgeni:
During triage, @sthirugn asked for an example of "satellite specific" workflow that we'd not want 1:1 here. My example was "content acquisition and management". The below bleeds into #644, but I'm gona write it down here and y'all can read it ;)

Satellite users usually get their content from Red Hat, so they need to get a manifest from the portal, upload that manifest to Katello/Satellite, enable Red Hat repositories, sync them, create content views.
Katello users (while they can do the same), will often use CentOS, Debian, etc, where the first steps are different: create a product and repository, point these at the upstream mirror servers. Then you can go and sync/create CVs.

So ideally, our roles are so "atomic" (I hate that word) that they can be used in either scenario. Which in return means, that if someone has a role for a full Satellite workflow (from manifest to CV), I'd love to see them splitting that role in at least two roles: one for "acquire content" (manifest to enable repo) and one for "manage content" (sync, publish). Then, if someone wants the same for CentOS/Debian, they need a new acquire_content role, but can re-use the manage content one.

bagasse:
As I will start to work to moving our internal 'bare' configuration playbooks to roles/collection, maybe if our internal workflows are applicable here I will contribute them here instead of keeping it internal:

**Foreman:**
* As a user, I want to configure taxonomies (Orgs, Locs)
[...]

PR #1027, opened 2020-10-18
add a role to manage products, repositories, and repository_sets

wbclark:


mdellweg:
Where do you want to go with this? ATM it's a role that calls a single module once without even mangling any parameter, and i do not see any benefit.
What about a role that accepts a list of products that contain lists of repositories with URLs. It then `state=present`s them and syncs all (maybe in parallel)?

ehelms:
I agree @mdellweg and this presents the first opportunity to discuss encouraging a best practices structure for how a user might define and manage their data. A format we've used to do this using RHEL as an example:

```
rhel_product:
  name: Red Hat Enterprise Linux Server
  repository_sets:
    - name: Red Hat Enterprise Linux 7 Server (RPMs)
      basearch: x86_64
      releasever: 7Server
    - name: Red Hat Enterprise Linux 6 Server (RPMs)
      basearch: x86_64
      releasever: 6Server
    - name: Red Hat Enterprise Linux 5 Server (RPMs)
      basearch: x86_64
      releasever: 5Server
    - name: Red Hat Enterprise Linux 7 Server - Extras (RPMs)
      basearch: x86_64
      releasever: ''
    - name: Red Hat Enterprise Linux 7 Server - Optional (RPMs)
      basearch: x86_64
      releasever: ''
  repositories:
    "7Server":
      name: Red Hat Enterprise Linux 7 Server RPMs x86_64 7Server
    "6Server":
      name: Red Hat Enterprise Linux 6 Server RPMs x86_64 6Server
    "5Server":
      name: Red Hat Enterprise Linux 5 Server RPMs x86_64 5Server
    "7Server Extras":
      name: Red Hat Enterprise Linux 7 Server - Extras RPMs x86_64
    "7Server Optional":
      name: Red Hat Enterprise Linux 7 Server - Optional RPMs x86_64 7Server
```

wbclark:
I've updated this role to enable and synchronize repositories, using the below structure for the data:

```yaml
- hosts: localhost
  roles:
    - role: theforeman.foreman.reposync
      vars:
        username: "admin"
        password: "changeme"
        products:
          - name: Red Hat Enterprise Linux Server
            repository_sets:
              - name: Red Hat Enterprise Linux 7 Server (RPMs)
                basearch: x86_64
                releasever: 7Server
              - name: Red Hat Enterprise Linux 6 Server (RPMs)
                basearch: x86_64
                releasever: 6Server
              - name: Red Hat Enterprise Linux 7 Server - Extras (RPMs)
                basearch: x86_64
                releasever: ''
              - name: Red Hat Enterprise Linux 7 Server - Optional (RPMs)
                basearch: x86_64
                releasever: ''
          - name: Red Hat Enterprise Linux for x86_64
            repository_sets:
              - name: Red Hat Enterprise Linux 8 for x86_64 - BaseOS (RPMs)
                basearch: x86_64
                releasever: 8Server
              - name: Red Hat Enterprise Linux 8 for x86_64 - AppStream (RPMs)
                basearch: x86_64
                releasever: 8server
```

Note that it was sufficient to define `repository_sets` per `product` and not yet necessary to define the actual `repositories`. This is because the `repository_sync` module synchronizes all repositories in a product by default when specific repositories are not passed.

TODO: Write and record tests, although I expect I'll face a similar issue to what I'm seeing with #978 , which is `"msg": "Error while performing upload on subscriptions: Can't overwrite existing cassette ('fixtures/manifest_role-0.yml') in your current record mode ('none')`

wbclark:
At this time the role is working to enable repositories and trigger sync tasks.

However I'm getting errors while syncing on my `centos7-katello-nightly` box so I'm going to provision a more stable release to record the tests.

roles/repositories/defaults/main.yml
```diff
@@ -0,0 +1,2 @@
+---
+validate_certs: true
```
evgeni: this is the default
```suggestion
```

roles/repositories/tasks/main.yml
```diff
@@ -0,0 +1,28 @@
+---
+- name: 'Enable Repositories'
+  theforeman.foreman.repository_set:
+    username: "{{ username }}"
+    password: "{{ password }}"
+    server_url: "{{ server_url }}"
+    validate_certs: "{{ validate_certs }}"
```
evgeni:
```suggestion
    validate_certs: "{{ validate_certs | default(omit) }}"
```

roles/repositories/tasks/main.yml
```diff
@@ -0,0 +1,28 @@
...
+  with_subelements:
+    - "{{ products }}"
+    - repository_sets
+
+- name: 'Synchronize Repositories'
+  theforeman.foreman.repository_sync:
+    username: "{{ username }}"
+    password: "{{ password }}"
+    server_url: "{{ server_url }}"
+    validate_certs: "{{ validate_certs }}"
```
evgeni:
```suggestion
    validate_certs: "{{ validate_certs | default(omit) }}"
```

tests/test_playbooks/repositories_role.yml
```diff
@@ -0,0 +1,35 @@
...
+  vars_files:
+    - vars/server.yml
+  roles:
+    - role: repositories
+      vars:
+        username: "{{ foreman_username }}"
+        password: "{{ foreman_password }}"
+        server_url: "{{  foreman_server_url }}"
+        validate_certs: false
+        organization: "Test Organization"
```
evgeni, line 28: this needs to be created first against `localhost`

tests/test_playbooks/repositories_role.yml
```diff
@@ -0,0 +1,35 @@
...
+  gather_facts: false
+  vars_files:
+    - vars/server.yml
+  roles:
+    - role: repositories
+      vars:
+        username: "{{ foreman_username }}"
+        password: "{{ foreman_password }}"
+        server_url: "{{  foreman_server_url }}"
+        validate_certs: false
```
evgeni:
```suggestion
        validate_certs: "{{ foreman_validate_certs }}"
```

roles/repositories/README.md
```diff
@@ -0,0 +1,60 @@
...
+
+Example Playbooks
+-----------------
+
+Enable and synchronize several Red Hat Repositories. There are a few important points to note about the structure of the data in this example. The first is that RHEL 8 repos have a different product name than previous RHEL versions. Additionally the RHEL 8 product already contains the `basearch` so it should not be specified on the RHEL 8 `repository_sets`, and the naming convention for `releasever` changed with RHEL 8 since system purpose removes the need for separate distributions like `Server` and `Workstation`. Finally, the optional and extras repositories do not have point releases so `releasever` should be omitted.
+
+```yaml
+- hosts: localhost
+  roles:
+    - role: theforeman.foreman.reposync
```
evgeni: this role is called `repositories` ;)

roles/repositories/tasks/main.yml
```diff
@@ -0,0 +1,28 @@
...
+  theforeman.foreman.repository_set:
+    username: "{{ username }}"
+    password: "{{ password }}"
+    server_url: "{{ server_url }}"
+    validate_certs: "{{ validate_certs }}"
+    organization: "{{ organization }}"
+    product: "{{ item.0.name }}"
+    name: "{{ item.1.name }}"
+    state: enabled
+    repositories:
```
evgeni, line 13: What if I want to enable multiple repo-sets of the same product? I think currently this will fail (or rather the last repo-set will win)

evgeni:
I have one major design issue with this role: It's called `repositories`, while in reality it's only `redhat_(cdn_)repositories`. Nothing wrong with providing a Red Hat specific role here, but I think we should *try* to also accommodate other users as good as possible.

My vision would be that the user has a data structure (call it `products`) that encapsulates all required knowledge to obtain these products (and can later be used to provide them to the clients). Based on the example playbook in your readme, it could be something like this:

```yaml
products:
  - name: Red Hat Enterprise Linux Server
    type: redhat
    repository_sets:
      - name: Red Hat Enterprise Linux 7 Server (RPMs)
        basearch: x86_64
        releasever: 7Server
  - name: CentOS 8
    type: yum
    repositories:
      - name: BaseOS x86_64
        url: http://mirror.centos.org/centos/8/BaseOS/x86_64/os/
  - name: Debian 10
    type: deb
    repositories:
      - name: main
        url: http://deb.debian.org/debian
        deb_components: main
        deb_architectures: amd64
        deb_releases: buster
```

The user feeds that list to a `repositories` role, which invokes a `redhat_repositories` (what you have written), `yum_repositories` and `deb_repositories` (both to be written, they create products and repositories the "normal way") roles, based on the `type` and finishes off with a `repository_sync` for all products.

(the alternative is not to have `type` and make the user invoke the individual roles, but seems less friendly)

mdellweg:
I think, the type is really something on the repository level *. Also why split it into different roles. We can use that structure and just one `repositories` role to set it all up.

*) IMHO it can be beneficial to have the same (software-)product packaged for different OSes in the same (katello-)product.

ehelms:
> I think, the type is really something on the repository level *.

Agreed, I think it might create overhead for users to have to define type at the product level when it only applies to the repository itself (and a product can have multiple types in it).

evgeni:
The problem of doing that on the repository level is that in the Red Hat case, you don't ever create products (or repositories), Katello does all that for you.

ehelms:
This is where I would love to see either `vars.yaml.example` or `vars/products.yaml.example` or something equivalent that lays out an example layout of the products structure at the top level for users to reference or copy.

ehelms:
> The problem of doing that on the repository level is that in the Red Hat case, you don't ever create products (or repositories), Katello does all that for you.

In that case, I would think the types, if at the product level, would be `redhat` and `custom`. But I wonder if we can omit that and let the existence of `repository_sets` and `repositories` do the switching for us given that is the power of roles?

evgeni:
Ah, yeah, that'd be possible. Great idea!

mdellweg:
> given that is the power of roles?

I couldn't have said it better.

evgeni:
So, to reiterate, the new structure would be:

```yaml
products:
  - name: Red Hat Enterprise Linux Server
    repository_sets:
      - name: Red Hat Enterprise Linux 7 Server (RPMs)
        basearch: x86_64
        releasever: 7Server
  - name: CentOS 8
    repositories:
      - name: BaseOS x86_64
        content_type: yum
        url: http://mirror.centos.org/centos/8/BaseOS/x86_64/os/
  - name: Debian 10
    repositories:
      - name: main
        content_type: deb
        url: http://deb.debian.org/debian
        deb_components: main
        deb_architectures: amd64
        deb_releases: buster
  - name: Foreman Client
    repositories:
      - name: Foreman Client Debian
        url: http://deb.atix.de/debian
        content_type: deb
        deb_components: main
        deb_architectures: amd64
        deb_releases: buster
      - name: Foreman Client CentOS
        url: http://yum.theforeman.org/client/latest
        content_type: yum
```

Or is content_type not required either?

ehelms:
> Or is content_type not required either?

Definitely required on a repository.

ehelms:
I am torn if I want a role that enables/ensures products and repositories exist *and* sync's the repositories at the same time. I definitely want a role that I can use to sync repositories which would then create some duplication with this role. Further, if while managing my repositories I want to make a change to a repository I may not want to sync it (e.g. if I were to enable HTTP on the repository). However, if I changed the URL of the repository, in that case I may want to sync? But perhaps we should give user more control over that via two separate roles:

 - repositories
 - repository_sync

bagasse:
Sorry to jump in this conversation :smile:. If we manage also custom products/repos, can we manage also some other "content management" things that products/repos depends on, like content_credential, sync_plans ? We worked on some playbooks originally, that we are moving to collection/roles that follow some usecases we have (like I described in [this comment](https://github.com/theforeman/foreman-ansible-modules/issues/646#issuecomment-608480355)). Has stated here our primary goal is to define some data structures that the users follow and just provide it as variables to role(s) that enforce it in our katello instances.
Maybe we can only sync repos only on change by default (and control this behavior via a variable) ?

evgeni:
@ehelms I'd prefer not to have logic that decides whether a sync is needed or not based on what has changed (url vs http publish in your example), this feels too magic. A tristate "sync: always/change/never" might be a fine compromise. Or just have two roles that accept the same data (structure) for management and syncing.

@bagasse heey, welcome back! Ideally, yes, we should also allow management of the supplementary entities, the question is always how much in detail can we go without making it too complicated to use. Take the sync plans for example, we could add a sync_plan key (with a cron like value) to each product or repo in the example above and autogenerate sync  plan entries from that, or we could define plans at the same level as the products and the plans refer to products. As a user I'd probably prefer the first? But then it again becomes magic in a way...

bagasse:
> As a user I'd probably prefer the first? But then it again becomes magic in a way...

I don't know for others, but we don't use too much sync plans, we just have one for internal repo sync (hourly) and two other (daily) for external repos (centos, epel, ovirt, foreman/katello...), so defining sync_plans at the same level as products and refer to it in products is fine for us.

roles/repositories/tasks/main.yml
```diff
@@ -0,0 +1,28 @@
...
+      - releasever: "{{ item.1.releasever | default(omit) }}"
+        basearch: "{{ item.1.basearch | default(omit) }}"
+  with_subelements:
+    - "{{ products }}"
+    - repository_sets
+
+- name: 'Synchronize Repositories'
```
ehelms: I'm still not sold that enabling repositories should imply a sync of the repositories.

roles/repositories/tasks/main.yml
```diff
@@ -0,0 +1,74 @@
...
+    deb_architectures: "{{ item.1.deb_architectures | default(omit) }}"
+    deb_releases: "{{ item.1.deb_releases | default(omit) }}"
+    download_policy: "{{ item.1.download_policy | default('on demand') }}"
+    mirror_on_sync: "{{ item.1.mirror_on_sync | default(true) }}"
+    state: present
+  with_subelements:
+    - "{{ products | selectattr('repositories', 'defined') | list }}"
+    - repositories
+
+- name: 'Create Sync Plans'
```
ehelms: Is there a benefit to putting this inside the repositories role instead of split out into it's own role? I would think the latter provides more composibility for users.

roles/repositories/README.md
```diff
@@ -0,0 +1,131 @@
+theforeman.foreman.repositories
+===============================
+
+This role enabled Red Hat Repositories, creates Custom Products and Repositories, and defines Sync Plans.
```
ehelms:
```suggestion
This role enables Red Hat Repositories, creates Custom Products and Repositories, and defines Sync Plans.
```

roles/repositories/tasks/main.yml
wbclark: The benefit is that defining it in the same role allows us to specify that a repository should be synced only when changed.

I'm still debating whether that's worth doing, as well as researching the data structures involved when registering the results of the with_subelements loops above.

wbclark:
I want to extend a big thank you to @ehelms @mdellweg @evgeni @bagasse for all the reviews. I have greatly benefited from the effort you've put towards sharing your thoughts and requirements for this feature.

I've pushed an update and want to summarize what has changed:

I. The role now supports creating Products and custom Repositories.
II. The role now allows to define Sync Plans and assign products to them.
III. All pieces of the data structure are optional. You may or may not enable Red Hat Repositories. You may or may not create custom Products and Repositories. You may or may not define Sync Plans. If any of these are empty, e.g. `Products` is passed as an empty list, the role should not fail.

I still haven't touched the synchronize repository task and am leaning towards defining a tri-state initial sync strategy at the Product level that allows syncs to be performed always, never, or only when the Product is changed.

I need to update the tests as well and will get to work on that shortly.

I'm still interested in any further feedback, and I want to solicit feedback in particular about the clarity of the README. Any further feedback is highly valued.

roles/repositories/tasks/main.yml
wbclark: Ah, I thought this comment was relevent to Syncs, not Sync Plans. My apologies. Please disregard my prior comment.

wbclark: The benefit here is that the user can define data for their Products alongside data for their Sync Plans.

Regarding composibility, I've taken care to ensure that each piece should be optional and tasks should be skipped when the required data structure is not defined.

@ehelms do you think there would be value in defining this role to be re-usable so that it can be passed multiple times with different variables for any users who desire that approach?

ehelms: I'm just having a hard time seeing the value provided by putting this all in a single role versus providing users more flexibility by having roles for configuring repositories, sync plans and taking the action to sync a repository. I think `sync` is a great example of this. Not all users will want to sync their repositories when they create or enable them. Users will want to sync repositories out of band of enable and creation. Therefore, we will end up with a stand-alone sync role anyway. So I perceive it as better to just write the stand-alone sync role (it could always be included by the repositories role).

evgeni: I think at some point I expressed the desire of having a "meta" data structure that defines "all things content" and all content related roles feed from the same structure, to avoid duplication of information in similar but not identical structures. This might be influenced by that idea?

wbclark: Yes. One goal I have is that the user should never have to define the same data in multiple places.

That's still possible with when composing role, but what it looks like is a single composite role, which receives the data once, and passes it to the child roles.

IMO at that point it's easier and more maintainable to just have a single role

ehelms: > That's still possible with when composing role, but what it looks like is a single composite role, which receives the data once, and passes it to the child roles.

I do not really understand this nuance. But I do not feel my point was addressed either. A dedicated sync role gives more power to users and allows use in this case.

A user will definitely have a common data structure spanning these roles. The repositories structure might be used by a user when defining their content views for example.

I think we are in that weird space between providing "workflows" and providing "components" for use by a user. My concern is if a user does not want to initiate a sync everytime they change something (add a repository, or product, change a property of a repository) they have to incur a sync. And maybe they prefer to let sync plans drive that, maybe their CI/CD but the control is taken away from them. Whereas a stand alone sync role allows the user to construct playbooks that define their workflow.

roles/repositories/README.md
```diff
@@ -0,0 +1,131 @@
...
+===============================
+
+This role enables Red Hat Repositories, creates Custom Products and Repositories, and defines Sync Plans.
+
+Role Variables
+--------------
+
+This role supports the [Common Role Variables](https://github.com/theforeman/foreman-ansible-modules/blob/develop/README.md#common-role-variables).
+
+- `products`: List of products to enable and synchronize. Each product is represented as a dictionary and can include `repository_sets` which represent Red Hat Repositories and should be used when the product name matches an existing Red Hat Product. Each repository set must have a `name` and should specify the `basearch` and/or `releasever` only when multiple versions are available for that product. Alternatively a custom product can have associated `repositories` which represent custom repositories. Custom repositories are required to have a `url` and `content_type` and may require additional fields. A variety of examples are demonstrated in the data structure below:
```
mdellweg: Custom repositories require a name, too, right?

Personal preference: Include a line break after each sentence.

roles/repositories/README.md
```diff
@@ -0,0 +1,131 @@
...
+    cron_expression: 0 6 8 * *
+    sync_date: 2020-11-08 00:06:00 UTC
+    products:
+      - Foreman Client
+```
+
+Example Playbooks
+-----------------
+
+Enable several Red Hat Repositories and create a weekly sync plan. There are a few important points to note about the structure of the data in this example. The first is that RHEL 8 repos have a different product name than previous RHEL versions. Additionally the RHEL 8 product already contains the `basearch` so it should not be specified on the RHEL 8 `repository_sets`, and the naming convention for `releasever` changed with RHEL 8 since system purpose removes the need for separate distributions like `Server` and `Workstation`. Finally, the optional and extras repositories do not have point releases so `releasever` should be omitted.
```
mdellweg: Maybe make the "points" a bullet list.

roles/repositories/README.md
```diff
@@ -0,0 +1,131 @@
...
+        sync_plans:
+          - name: Weekly Sync
+            interval: weekly
+            sync_date: 2020-11-07 00:00:00 UTC
+            products:
+              - Red Hat Enterprise Linux Server
+              - Red Hat Software Collections (for RHEL Server)
+              - Red Hat Enterprise Linux for x86_64
+              - CentOS 8
+              - Debian 10
```
mdellweg:
```suggestion
              - Red Hat Enterprise Linux Server
              - Red Hat Enterprise Linux for x86_64
```

roles/repositories/tasks/main.yml
```diff
@@ -0,0 +1,74 @@
...
+    validate_certs: "{{ validate_certs | default(omit) }}"
+    organization: "{{ organization }}"
+    name: "{{ item.1.name }}"
+    url: "{{ item.1.url }}"
+    product: "{{ item.0.name }}"
+    content_type: "{{ item.1.content_type }}"
+    deb_components: "{{ item.1.deb_components | default(omit) }}"
+    deb_architectures: "{{ item.1.deb_architectures | default(omit) }}"
+    deb_releases: "{{ item.1.deb_releases | default(omit) }}"
+    download_policy: "{{ item.1.download_policy | default('on demand') }}"
```
mdellweg: Any reason, we are overriding katello's default here?

roles/repositories/tasks/main.yml
evgeni: After thinking more on this, I agree, sync plans should go into an own role.

roles/repositories/tasks/main.yml
evgeni: I think, at some point, we mentioned a flag that will result in a sync "always", "never", "if changed".

Honestly, this could also be done with an own role (and give the user better flexibility what they sync). The only caveat is that `import_role` doesn't register a result, so you can't automatically see if something changed by running the role. The role would need to register that itself and we'd have to document that for the user.

roles/repositories/tasks/main.yml
wbclark: This has not been the case in my testing. For example the test playbook here has 2 repository sets under the same product, and both are successfully enabled.

I admit I am very impressed by Ansible's `with_subelements`

tests/test_playbooks/repositories_role.yml
wbclark: Done.

roles/repositories/tasks/main.yml
wbclark: No and indeed it fails for Debian repositories unless I use `default(omit)` so I have made the change.

roles/repositories/README.md
wbclark: I've made a first pass at cleaning that up

roles/repositories/README.md
wbclark: Done, but I admit I'm still not fully satisfied with the overall flow and formatting of the doc

wbclark:
I have split all functionality related to sync plans off with https://github.com/theforeman/foreman-ansible-modules/pull/1065

roles/repositories/README.md
```diff
@@ -0,0 +1,108 @@
+theforeman.foreman.repositories
+===============================
+
+This role defines Products and enabled Red Hat Repositories and Custom Repositories.
```
ehelms: Nit: `defines Products, enabled Red Hat Repositories and Custom Repositories.`

roles/repositories/README.md
```diff
@@ -0,0 +1,108 @@
...
+    repository_sets:
+      - name: Red Hat Enterprise Linux 8 for x86_64 - BaseOS (RPMs)
+        releasever: 8
+      - name: Red Hat Enterprise Linux 8 for x86_64 - AppStream (RPMs)
+        releasever: 8
+  - name: CentOS 8
+    repositories:
+      - name: BaseOS x86_64
+        content_type: yum
+        url: http://mirror.centos.org/centos/8/BaseOS/x86_64/os/
```
ehelms, line 54: Mind adding a second repostitory to this example for AppStream to show why you would want to do a `CentOS 8` product instead of a `CentOS` product with 7 and 8 in it?

ehelms:
I just had a few thoughts around the README since I know users will look to those examples as a best practice. Looks good @wbclark !

roles/repositories/tasks/main.yml
```diff
@@ -0,0 +1,50 @@
...
+  theforeman.foreman.product:
+    username: "{{ username }}"
+    password: "{{ password }}"
+    server_url: "{{ server_url }}"
+    validate_certs: "{{ validate_certs | default(omit) }}"
+    organization: "{{ organization }}"
+    name: "{{ item.name }}"
+    state: present
+  with_items:
+    - "{{ products | selectattr('repositories', 'defined') | list }}"
```
wbclark, line 44: TODO: Should technically be able to define a Product without defining Repositories on that Product. I'm not sure if there is a real use for that, but Katello allows it and I noticed it would be nice to have when developing test playbooks for other roles.

roles/repositories/README.md
evgeni: Should this rather be "This role defines Products and Custom Repositories and enables Red Hat Repositories."
As it does not *enable* custom repos?

roles/repositories/README.md
```diff
@@ -0,0 +1,108 @@
...
+===============================
+
+This role defines Products and enabled Red Hat Repositories and Custom Repositories.
+
+Role Variables
+--------------
+
+This role supports the [Common Role Variables](https://github.com/theforeman/foreman-ansible-modules/blob/develop/README.md#common-role-variables).
+
+- `products`: List of products to enable and synchronize.
```
evgeni:
```suggestion
- `products`: List of products to manage.
```

roles/repositories/README.md
```diff
@@ -0,0 +1,108 @@
...
+  - name: Foreman Client
+    repositories:
+      - name: Foreman Client Debian
+        url: http://deb.atix.de/debian
+        content_type: deb
+        deb_components: main
+        deb_architectures: amd64
+        deb_releases: buster
+      - name: Foreman Client CentOS
+        url: http://yum.theforeman.org/client/latest
```
evgeni, line 68:
```suggestion
        url: https://yum.theforeman.org/client/latest/
```

roles/repositories/README.md
wbclark: Yes, I think so. I'm also leaning to renaming the role `Products` with the shape it is taking

roles/repositories/README.md
```diff
@@ -0,0 +1,108 @@
...
+        deb_architectures: amd64
+        deb_releases: buster
+  - name: Foreman Client
+    repositories:
+      - name: Foreman Client Debian
+        url: http://deb.atix.de/debian
+        content_type: deb
+        deb_components: main
+        deb_architectures: amd64
+        deb_releases: buster
```
evgeni:
```suggestion
      - name: Foreman Client Debian 10
        url: https://apt.atix.de/Debian10/
        content_type: deb
        deb_components: main
        deb_architectures: amd64
        deb_releases: stable
```

roles/repositories/README.md
evgeni:
```suggestion
      - name: Foreman Client CentOS 7
        url: https://yum.theforeman.org/client/latest/el7/x86_64/
```

roles/repositories/README.md
```diff
@@ -0,0 +1,108 @@
...
+      - name: Red Hat Enterprise Linux 8 for x86_64 - AppStream (RPMs)
+        releasever: 8
+  - name: CentOS 8
+    repositories:
+      - name: BaseOS x86_64
+        content_type: yum
+        url: http://mirror.centos.org/centos/8/BaseOS/x86_64/os/
+  - name: Debian 10
+    repositories:
+      - name: main
```
evgeni:
```suggestion
      - name: Debian 10 main
```

roles/repositories/tasks/main.yml
evgeni: Cool,thanks!

roles/repositories/tasks/main.yml
```diff
@@ -0,0 +1,50 @@
...
+  theforeman.foreman.repository_set:
+    username: "{{ username }}"
+    password: "{{ password }}"
+    server_url: "{{ server_url }}"
+    validate_certs: "{{ validate_certs | default(omit) }}"
+    organization: "{{ organization }}"
+    product: "{{ item.0.name }}"
+    name: "{{ item.1.name }}"
+    state: enabled
+    repositories:
```
evgeni, line 13: the repo set module has a `all_repositories` parameter: https://theforeman.github.io/foreman-ansible-modules/develop/plugins/repository_set_module.html#parameter-all_repositories

Should we expose that here? There is also an example how to combine that with `repositories` so that the user can have a single data structure but some sets have `repositories` and the others end up being `all`.

roles/repositories/README.md
```diff
@@ -0,0 +1,108 @@
+theforeman.foreman.repositories
```
evgeni, line 1: I wonder if `theforeman.foreman.content` is not a more appropriate name? but wouldn't block on this

roles/repositories/tasks/main.yml
```diff
@@ -0,0 +1,50 @@
...
+    server_url: "{{ server_url }}"
+    validate_certs: "{{ validate_certs | default(omit) }}"
+    organization: "{{ organization }}"
+    name: "{{ item.name }}"
+    state: present
+  with_items:
+    - "{{ products | selectattr('repositories', 'defined') | list }}"
+
+- name: 'Create Repositories'
+  theforeman.foreman.repository:
```
evgeni, line 47: would you mind exposing the "other" params of the module here too, with `|default(omit)`, see https://theforeman.github.io/foreman-ansible-modules/develop/plugins/repository_module.html

this would allow people also manage container/ansible/puppet repos with that

and we could document something like "the `repositrory` takes any parameter the `repository` module supports"

roles/repositories/README.md
ehelms: Content seems so broad spectrum that given this is focused on repositories, I would stick to that personally.

evgeni: wfm, was just a thought :)

roles/repositories/tasks/main.yml
wbclark: Done.

roles/repositories/tasks/main.yml
wbclark: I added a new task for Products which are specified with `all_repositories: True` and documented it appropriately. I added data in the test playbook to test for this task as well.

After using the module in this way, my general feeling is that this is something we should discourage users from doing unless they fully understand the consequences of enabling a dozen point releases that they are likely not going to use.

roles/repositories/README.md
wbclark: Done.

roles/repositories/tasks/main.yml
wbclark: A caveat is that `all_repositories` requires either `name` + `product`, OR `label` to be specified. I added support for the labels but not for product + name. I could add that still if there is value in it (the `name`s are more specific I think and avoid the problem of enabling a dozen unused point releases?) but it would require a 3rd task or a bit of rethinking my approach to the looping.

EDIT: the reason for the awkwardness is that you would need to specify the `repository_sets` in order to give them `name`s, which breaks assumptions in the prior `with_subelements` loop over Products with repository sets. We could instead expect it to be called something other than repository_sets in this case but that makes the data structure itself awkward.

wbclark: I also tried to test multiple labels but using the `rhel-server-rhscl-6-rpms` label resulted in working Ansible but "The path /content/dist/rhel/server/6/6.9/i386/rhscl/1/os does not seem to be a valid repository. If you think this is an error, please try refreshing your manifest." from Katello

ehelms (approved):
Everything looks good to me, I'm excited to get this particular role out there for use. Will let @evgeni have final say in review.

roles/repositories/README.md
```diff
@@ -0,0 +1,115 @@
...
+===============================
+
+This role defines Products and enabled Red Hat Repositories and Custom Repositories.
```
evgeni:
```suggestion
This role defines Products and Custom Repositories and enables Red Hat Repositories.
```

roles/repositories/README.md
```diff
@@ -0,0 +1,115 @@
...
+
+Role Variables
+--------------
+
+This role supports the [Common Role Variables](https://github.com/theforeman/foreman-ansible-modules/blob/develop/README.md#common-role-variables).
+
+- `products`: List of products to manage.
+  Each product is represented as a dictionary and can include `repository_sets` which represent Red Hat Repositories and should be used when the product name matches an existing Red Hat Product.
+  Each element of `repository_sets` must have a `name` and should specify the `basearch` and/or `releasever` only when multiple versions are available for that Product.
+  All repository sets for a Red Hat Product can be enabled by omitting `repository_sets` and instead specifying that the Product has `all_repositories: True`. When using this option it is also necessary to specify a list of repository `label`s for the Product (e.g. rhel-7-server-rpms). Be wary that this option can result in enabling a large number of unused repositories that, if added to sync plans, can greatly increase sync times and rapidly fill disk space.
```
evgeni, line 14: Why the `label` limitation? Can't it be enabled with `product` and `name` too?

evgeni: Ah, I see in https://github.com/theforeman/foreman-ansible-modules/pull/1027/files#r541297185

Okay for now I guess.

evgeni (changes requested):
one doc nitpick ;)

roles/repositories/README.md
wbclark: Yeah, this feels a lot more natural. :+1:

evgeni (approved):

evgeni:
@wbclark last thing, really the last (:sweat_smile:) can you please add a changelog entry? new modules get that automatically, but new roles do not.

evgeni:
I don't think `roles` is a valid section for changelogs, can you please use `minor_changes`?

evgeni (approved):

PR #1031, opened 2020-10-19
Verify only safe methods are used in check_mode

mdellweg:
Adds a method matcher in check_mode that disallowes POST, PUT, PATCH and
DELETE.


evgeni (approved):

evgeni:
super safe!

PR #1032, opened 2020-10-21
use felix' ansible sphinx plugin

evgeni:
this contains better CSS for the plugin docs and still provides the
lexer


mdellweg (approved):
sounds good to me

If we could pin it to a tag, that would be even better.

evgeni:
there are none (yet), the repo might move inside ansible-community org and maybe even into the antsibull repo, I'll update us to use that as soon the fate is decided.

PR #1033, opened 2020-10-21
cache python dependencies for quicker setup

evgeni:


evgeni:
mhh, no, this can be simplified to use less caches
- the ansible version is irrelevant (we won't cache that anyways)
- the checkmode vs build jobname is irrelevant

mdellweg:
What is being cached here? The package list of pip and the downloaded packages? Can't we just reuse the same cache for all jobs?

evgeni:
This caches `~/.cache/pip` which contains whatever `pip` downloaded from pypi.org (so wheels and stuff). But really just the downloads, the dependency resolution and installation still happens "live".

My reasoning behind using different caches was that different steps use a different set of packages, and thus re-using a cache from the docs step is "bad" in the normal test step, as that would mean to re-download all the non-docs deps (as the docs step didn't cache them).

evgeni:
but at least the python version should be irrelevant, we have few deps that *differ* based on python version.

mdellweg:
Wouldn't that just give a cache containing all the wheels for tests and docs? And therefore be good for everyone?

I still don't see, who saves the cache, and whether it's augmented by each job. Is it cached across runs?

evgeni:
`actions/cache` has a "post" step (so you don't see it in the workflow definition) that collects the cache

and yeah, you're right, the "worst" that'd happen is we seed the cache too much and then only partially update it.

mdellweg (approved):
Not that i fully understand the mechanics, but this looks reasonable.

PR #1034, opened 2020-10-21
document possible template kind choices in os_default_template

evgeni:


plugins/modules/os_default_template.py
```diff
@@ -36,6 +36,24 @@
       - name of the template kind
     required: true
     type: str
+    choices:
```
evgeni, line 39: this is not a doc fragment, because the two modules use different names for that property :roll_eyes:

evgeni: also, the lists differ, the one module accepts "snippet", the other does not.

mdellweg (approved):

PR #1040, opened 2020-10-22
rename katello_manifest test to subscription_manifest

evgeni:
the module was renamed, but we forgot to adjust the test name


mdellweg (approved):
Let's avoid confusion!

evgeni:
rebased.

PR #1048, opened 2020-10-30
document cvmanager → FAM migration path

evgeni:


evgeni:
@pcfe I'd love to hear your input given you've been a cvmanager user for a while :)

pcfe:
all in all I like it very much!
- clear
- concise
- good examples
- clear statement of what's different

Thanks Evgeni.

The ony section that might benefit from expanding is auto-update. Consider the following scenario
- one _el8 base_ CV (BaseOS, AppStream assorted ks trees and Sat tools)
- one _el7 base_ CV (base, ks tree and Sat tools)
- one app _foo_ CV (el release agnostic)
- one app _bar-el7_ CV
- one app _bar-el8_ CV

CCVs
- CCV _einz_ with el8 base + foo + bar-el8
- CCV _zwei_ with el7 base + foo
- CCV _drei_ with el7 base + foo + bar-el7

if a user opts for auto publish, then AFAICT publishing a new version of _el8 base_ then _foo_ and finally _bar-el8_ will result in 3 updates of _einz_. So ain't nobody got time for that and users tend to not use auto publish (mind you I have not tried auto publish in ages, I could be wrong)

long story short; IMNSHO the PR is definitely very good to use as is, but if you were to finesse further, then an example where a CCV uses multiple CVs with more than one component CVs set to latest and an explanation on how many CCV updates that entails.

evgeni:
> if a user opts for auto publish, then AFAICT publishing a new version of _el8 base_ then _foo_ and finally _bar-el8_ will result in 3 updates of _einz_. So ain't nobody got time for that and users tend to not use auto publish (mind you I have not tried auto publish in ages, I could be wrong)

I'd disagree. Not because the described scenario won't trigger 3 publishes of `einz` (it will), but because I don't think users should trigger publishes of base, foo and bar in a row -- they are in separate CVs because they have different lifecycles/cadences, so updating them at the same time seems weird?

> long story short; IMNSHO the PR is definitely very good to use as is, but if you were to finesse further, then an example where a CCV uses multiple CVs with more than one component CVs set to latest and an explanation on how many CCV updates that entails.

I think I can still come up with a nice example for that, showing what's possible, even if *usually* not required.

evgeni:
@pcfe updated with an example how to use `resource_info` to find the latest version of a CV

docs/cvmanager.md
```diff
@@ -0,0 +1,142 @@
...
+
+[cvmanager](https://github.com/RedHatSatellite/katello-cvmanager) is a tool to automate Content View management workflows. It hasn't been updated since a long time and users are encouraged to migrate to Ansible and the Foreman Ansible Modules.
+
+This document documents how the various workflows can be translated into Ansible playbooks.
+
+Please note that `cvmanager` configuration refers to Content Views by their labels, while Ansible uses the names of the Content Views for that.
+
+## Cleanup of old Content Views
+
+[TBD](https://github.com/theforeman/foreman-ansible-modules/pull/958)
```
evgeni, line 11: I don't think this should block this PR.

ehelms: > I don't think this should block this PR.

Agreed.

docs/cvmanager.md
```diff
@@ -0,0 +1,142 @@
+# Migrating from `cvmanager` to Foreman Ansible Modules
+
+[cvmanager](https://github.com/RedHatSatellite/katello-cvmanager) is a tool to automate Content View management workflows. It hasn't been updated since a long time and users are encouraged to migrate to Ansible and the Foreman Ansible Modules.
```
ehelms: Consider `been updated in a long time`

docs/cvmanager.md
```diff
@@ -0,0 +1,142 @@
...
+If, for example, the old `cvmanager` configuration contained the following entries:
+
+```yaml
+:ccv:
+  ccv-RHEL7-automated:
+    cv-RHEL7: 23.0
+    cv-tools: latest
+```
+
+You can achieve the same using the `content_view` modules:
```
ehelms: Should this be `module` since the example shows only one module being used?

ehelms (approved):
Could of nits on the wording

evgeni:
Thanks! Updated.

Let's see if Patrick has any last words, and then this goes in!

pcfe:
LGTM!
(while I would like 958 in too I agree that that should not block this PR.)

example with `resource_info` much appreciated, thanks Evgeni

Apologies for delay, saw update yesterday after lunch and then had a _oh look a :chipmunk: _moment

evgeni:
Thanks!

PR #1049, opened 2020-11-06
mark rex subnet api patch only for < 2.2

evgeni:
it was fixed as part of https://projects.theforeman.org/issues/30651


mdellweg (approved):
fixed_version is referring to the foreman version, right?

evgeni:
> fixed_version is referring to the foreman version, right?

Correct, we don't (yet) have a good way to check plugin versions.

PR #1054, opened 2020-11-10
debian related tests moved to separate files

### LINKED ISSUE #897: tag to mark upstream-only tasks in test playbook
##### SUMMARY
Basically I'd welcome a way to exclude tasks that won't work downstream (debian repo tests, etc.). I suppose the solution would be to have an 'upstream_only'  tag + a way to tell 'make record_' to pass down the --skip-tags option.

Different suggestions are welcome :)

##### ISSUE TYPE
 - Feature Idea

pondrejk:
so that they can be easily left out when testing downstream, fixes #897


evgeni:
Thanks @pondrejk!

Could you please rebase on the rpm fix in #1058? This is good to go then.

evgeni (approved):
ack pending rebase

PR #1058, opened 2020-11-11
only decode RPM headers if they were returned as bytes

evgeni:
RPM 4.15 and newer returns strings:
  * https://github.com/rpm-software-management/rpm/commit/84920f898315d09a57a3f1067433eaeb7de5e830
  * https://bugzilla.redhat.com/show_bug.cgi?id=1631292

### DIFF
--- a/plugins/modules/content_upload.py
+++ b/plugins/modules/content_upload.py
@@ -74,7 +74,7 @@ RETURN = ''' # '''
 import os
 import traceback
 
-from ansible.module_utils._text import to_bytes
+from ansible.module_utils._text import to_bytes, to_native
 from ansible_collections.theforeman.foreman.plugins.module_utils.foreman_helper import KatelloAnsibleModule, missing_required_lib
 
 try:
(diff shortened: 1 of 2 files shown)


plugins/modules/content_upload.py
```diff
@@ -113,11 +113,17 @@ def get_rpm_info(path):
...
-    name = rpmhdr[rpm.RPMTAG_NAME].decode('ascii')
     epoch = rpmhdr[rpm.RPMTAG_EPOCHNUM]
-    version = rpmhdr[rpm.RPMTAG_VERSION].decode('ascii')
-    release = rpmhdr[rpm.RPMTAG_RELEASE].decode('ascii')
-    arch = rpmhdr[rpm.RPMTAG_ARCH].decode('ascii')
+    if isinstance(rpmhdr[rpm.RPMTAG_NAME], bytes):
+        name = rpmhdr[rpm.RPMTAG_NAME].decode('ascii')
+        version = rpmhdr[rpm.RPMTAG_VERSION].decode('ascii')
+        release = rpmhdr[rpm.RPMTAG_RELEASE].decode('ascii')
+        arch = rpmhdr[rpm.RPMTAG_ARCH].decode('ascii')
```
mdellweg: Isn't there this strange `to_native` thing in ansible?

evgeni: There is, and I'm every time confused I when I have to use it…

mdellweg (approved):

PR #1061, opened 2020-11-16
flag password_hash explicitly as no_log=False

evgeni:
`password_hash` is just the hashing mechanism, not the actual hash, but
Ansible doesn't know that and yields a warning:

    [WARNING]: Module did not set no_log for password_hash

Also see https://docs.ansible.com/ansible/latest/dev_guide/developing_modules_documenting.html#documentation-block


mdellweg (approved):
Probably ok without a changelog. ;)

PR #1062, opened 2020-11-16
assert we didn't yield any warnings

### LINKED ISSUE #1055: [WARNING]: The following parameters are not supported by your server when performing create on repositories: {'organization_id'}. They were ignored.
##### SUMMARY
trying to create a repo using the below playbook

##### ISSUE TYPE
 - Bug Report

##### ANSIBLE VERSION

```
2.10
```

##### COLLECTION VERSION

```
develop
```

##### KATELLO/FOREMAN VERSION

```
3.16
```

##### STEPS TO REPRODUCE

```yaml
    - name: "Create files repository 'docker-compose-linux-x86_64' for product 'docker-compose'"
      theforeman.foreman.repository:
        server_url: "{{ foreman_url }}"
        organization: "{{ foreman_initial_organization }}"
        name: "docker-compose-linux-x86_64"
        state: present
        content_type: "file"
        product: "docker-compose"

```

##### EXPECTED RESULTS

no warning

##### ACTUAL RESULTS

```

```

evgeni:


evgeni:
Hm, seems not to work, doesn't break without #1061

tests/test_crud.py
```diff
@@ -88,6 +88,9 @@ def test_crud(tmpdir, module, vcrmode):
         run = run_playbook_vcr(tmpdir, module, record=record)
     assert run.rc == 0
 
+    for event in run.events:
+        assert not event.get('event_data', {}).get('warning', False)
```
mdellweg:
```suggestion
        assert not event.get('event_data', {}).get('warning', False), str(event.event_data.warning)
```
Let's see the warning.

evgeni: It's a bool, so there is nothing to see, but it's also the wrong bool. It'll only show you play warnings, I want task warnings, and that's what I've implemented now.

evgeni:
Well, the new errors are mostly valid.

* ~~activation_key - the warning is intentional here, I'll need to see how we can ignore it for that one case~~
* job_invocation - needs investigation
* repository, repository_deb: #1055

mdellweg (approved):
I like!

evgeni:
repository re-record is failing due to https://projects.theforeman.org/issues/31322, yay

plugins/modules/repository.py
```diff
@@ -294,6 +294,10 @@ def main():
         entity_opts={'scope': ['product']},
     )
 
+    # KatelloEntityAnsibleModule automatically adds organization to the enity scope
```
```suggestion
    # KatelloEntityAnsibleModule automatically adds organization to the entity scope
```

evgeni:
okay, only job_invocation left

evgeni:
great, that seems to be a bug in `prepare_params` :roll_eyes:

evgeni:
fix in https://github.com/Apipie/apypie/pull/87

evgeni:
~~and note to myself, when fixing job_invocation tapes, also clean out the `None`s from the params~~ this will be done in https://github.com/theforeman/foreman-ansible-modules/pull/1067

evgeni:
@mdellweg mind re-reviewing, as this got quite a few changes after your ack?

mdellweg (approved):
Still, looks good!

evgeni:
Thanks, latest push just adds a changelog fragment.

PR #1064, opened 2020-11-17
clarify documentation which module should be used for Red Hat Reposit…

### LINKED ISSUE #1059: The documentation should better describe the difference between a repository and repository_set
##### SUMMARY

It took me a while and question on IRC to figure, repository_set stand for RH repository. I'd suggest to modify the docs in a way

1) `repository_set – Enable/disable Repositories in Repository Sets` => `repository_set – Enable/disable Red Hat Repositories`
2) `repository – Manage Repositories` => `repository – Manage non RH Repositories` or even `repository – Manage Custom Repositories`
3) the repository detail should mention in synopsis, that for managing RH repositories, please use repository_set module.

##### ISSUE TYPE
 - Documentation

Not sure what the correct nomenclature should be, but some if the suggestion could simplify the first timer experience.

evgeni:
…ories

Fixes: #1059

### DIFF
--- a/plugins/modules/repository.py
+++ b/plugins/modules/repository.py
@@ -27,6 +27,11 @@ short_description: Manage Repositories
 description:
   - Crate and manage repositories
 author: "Eric D Helms (@ehelms)"
+notes:
+  - You can configure certain aspects of existing Red Hat Repositories (like I(download_policy)) using this module,
+    but you can't create (enable) or delete (disable) them.
+  - If you want to enable or disable Red Hat Repositories available through your subscription,
+    please use the M(theforeman.foreman.repository_set) module instead.
 options:
   name:
     description:
--- a/plugins/modules/repository_set.py
+++ b/plugins/modules/repository_set.py
@@ -23,9 +23,9 @@ DOCUMENTATION = '''
 ---
 module: repository_set
 version_added: 1.0.0
-short_description: Enable/disable Repositories in Repository Sets
+short_description: Enable/disable Red Hat Repositories available through subscriptions
 description:
-  - Enable/disable repositories in repository sets
+  - Enable/disable Red Hat Repositories that are available through subscriptions
 author: "Andrew Kofink (@akofink)"
 options:
   name:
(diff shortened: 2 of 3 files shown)


evgeni:
cc @ares

plugins/modules/repository.py
```diff
@@ -27,6 +27,10 @@
 description:
   - Crate and manage repositories
 author: "Eric D Helms (@ehelms)"
+notes:
+  - You can configure certain aspects of existing Red Hat Repositories (like I(download_policy)) using this module,
+    but you can't create (enable) or delete (disable) them.
+  - If you want to enable or disable Red Hat Repositories, please use the M(theforeman.foreman.repository_set) module instead.
```
```suggestion
  - If you want to enable or disable Red Hat Repositories available in your subscription manifest, please use the M(theforeman.foreman.repository_set) module instead.
```
Or should we even mention the `manifest` module here?

plugins/modules/repository_set.py
```diff
@@ -23,9 +23,9 @@
 ---
 module: repository_set
 version_added: 1.0.0
-short_description: Enable/disable Repositories in Repository Sets
+short_description: Enable/disable Red Hat Repositories
 description:
-  - Enable/disable repositories in repository sets
+  - Enable/disable Red Hat Repositories
```
mdellweg: Maybe this should refer to the subscription manifest mechanics.

plugins/modules/repository.py
evgeni: I wouldn't mention the manifest here, no. I'd word it:

```
If you want to enable or disable Red Hat Repositories available through your subscription,
please use the M(theforeman.foreman.repository_set) module instead.
```

plugins/modules/repository_set.py
evgeni: the manifest is a file, I'd not mention it here. "subscription" is a good keyword tho, so maybe:

```
Enable/disable Red Hat Repositories available through subscriptions
```

ares (approved):
Thanks, that solves my confusion over this module :-)

mdellweg (approved):

PR #1065, opened 2020-11-17
add a role to manage sync plans

wbclark:


wbclark:
This is currently based on #1027 for the reason that in the test playbook, I want to ensure products exist before trying to add them to sync plans and it was convenient to use the role for that.

roles/sync_plans/README.md
```diff
@@ -0,0 +1,58 @@
...
+sync_plans:
+  - name: Weekly Sync
+    interval: weekly
+    sync_date: 2020-11-07 00:00:00 UTC
+    products:
+      - Red Hat Enterprise Linux Server
+      - Red Hat Software Collectiions (for RHEL Server)
+      - Red Hat Enterprise Linux for x86_64
+      - CentOS 8
+      - Debian 10
```
evgeni, line 23: how much work would it be to make this accept both, a list of strings, and a list of dicts? why you ask? then users can have one `products` definition and feed that to both `repositories` and `sync_plans` modules. at least for the simple case.

roles/sync_plans/README.md
```diff
@@ -0,0 +1,58 @@
...
+=============================
+
+This role defines Sync Plans.
+
+Role Variables
+--------------
+
+This role supports the [Common Role Variables](https://github.com/theforeman/foreman-ansible-modules/blob/develop/README.md#common-role-variables).
+
+- `sync_plans`: List of sync plans to create. Each sync plan is represented as a dictionary which specifies the `name` of the sync plan and the `products` assigned to the sync plan. It also specifies the `interval` which can be 'hourly', 'daily', 'weekly', or 'custom cron'. In case the 'custom cron' `interval` is used, it should also specify the `cron_expression`. Finally the sync plan should have a `sync_date` which specifies the first time that the sync plan will run.
```
evgeni, line 11: I wonder how much this should duplicate the `sync_plan` module docs.

roles/sync_plans/README.md
evgeni: (and your test playbook shows nicely how things get duplicated in simple cases)

wbclark: I decided against that approach for the reason that a product can be assigned to no sync plans, one sync plan, or multiple sync plans, and because the sync plan only needs to know the name of the assigned product and not the full details of its structure.

Specifying them together in a single structure only really makes sense imo in the case where there is a single sync plan with all products assigned. I think that is most likely to occur on a smaller deployment with a relatively fewer count of enabled products, and in that case, we really aren't saving that much work by preventing the user from needing to list the product names a second time under the sync plan.

wbclark: The other approach I considered was allowing the sync plan <-> product assignments to exist on the products data structure, but the benefits are again limited for the same reasons. Sync plans would still need to have their full attributes defined elsewhere, so we aren't really saving much work.

Ultimately I settled on the current approach because I think it is the easiest one to think about, and that to me is more important than a slight reduction in the amount of typing we are asking the user to do.

ehelms: Given there is really only that all products use case, I would lean towards wait for future enhancement if someone requests it to add that and leave this as a list. The notion that you have to supply the product name twice, and correctly twice is more a fault of the current design. And, separately, we could explore some documentation that shows how to define your data such that you can re-use names via yaml methods, e.g.

```
product_names:
  rhel_6: Red Hat Enterprise Linux 6
  rhel_7: Red Hat Enterprise Linux 7

products:
  - name: "{{ product_names['rhel_6'] }}"
  - name: "{{ product_names['rhel_7'] }}"
 
sync_plans:
   - name: Weekly Sync
     interval: weekly
     sync_date: 2020-11-07 00:00:00 UTC
     products:
       - "{{ product_names['rhel_6'] }}"
```

evgeni: Yeah, documentation is what I ended up in my head too. And it's even easier than what you have:

```
sync_plans:
   - name: Weekly Sync
     interval: weekly
     sync_date: 2020-11-07 00:00:00 UTC
     products: "{{ products | map(attribute='name') | list }}"
```

William, if you can incorporate that in the examples, I'm happy!

wbclark: Updated with additional documentation. Thanks for the suggestion on using yaml methods to avoid duplicating data.

roles/sync_plans/README.md
wbclark: My general sense is that the role docs should be complete enough so that the average user could avoid using the modules or their documentation at all in cases where a role exists.

roles/sync_plans/README.md
```diff
@@ -0,0 +1,78 @@
...
+        password: "changeme"
+        organization: "Default Organization"
+        sync_plans:
+          - name: Weekly Sync
+            interval: weekly
+            sync_date: 2020-11-07 00:00:00 UTC
+            products: "{{ products | map(attribute='name') | list }}"
+```
+
+The above example assumes that a yaml dictionary `products` is already defined in Ansible variables. It uses yaml methods to select the name of each product from that dictionary, convert them all to a list, and pass that list to the definition of the sync plan.
```
ehelms, line 78: If possible, add a link here to https://github.com/theforeman/foreman-ansible-modules/tree/develop/roles/repositories#role-variables as that shows the yaml dictionary to which you are referring them to and brings things full circle.

ehelms (approved):
Small documentation thought, but I am happy to see it go in as it stands.

roles/sync_plans/README.md
evgeni: I always struggle to find the right balance between "this is required" and "a copy that gets out of date soon"

So lets go with what we have

evgeni (approved):
ACK

PR #1066, opened 2020-11-18
don't manually include route params when preparing

evgeni:
this has been ported to apypie in 0.3.0 (https://github.com/Apipie/apypie/pull/79) and we vendor a recent enough version


mdellweg (approved):

mdellweg:
Deleting code without breaking anything == improvement

PR #1067, opened 2020-11-18
always clean None params from dics and lists

evgeni:


evgeni:
this has the potential for breaking more things where we don't have tests for, but I prefer to live by "if you like it, then you should have put a test on it" :)

mdellweg:
Yep! What we don't test we cannot support.
Also I think we have been sufficiently reluctant to merge untested stuff.

plugins/module_utils/foreman_helper.py
```diff
@@ -329,7 +329,14 @@ def __init__(self, **kwargs):
...
         aliases = {alias for arg in argument_spec.values() for alias in arg.get('aliases', [])}
-        self.foreman_params = {k: v for (k, v) in self.params.items() if v is not None and k not in aliases}
+        self.foreman_params = {}
+        for (k, v) in self.params.items():
+            if v is not None and k not in aliases:
+                if isinstance(v, dict):
+                    v = {subk: subv for (subk, subv) in v.items() if subv is not None}
+                elif isinstance(v, list) and v and isinstance(v[0], dict):
+                    v = [{subk: subv for (subk, subv) in element.items() if subv is not None} for element in v]
+                self.foreman_params[k] = v
```
mdellweg: Should this be fully recursive again?

evgeni: I honestly don't expect further nesting, but we can.

mdellweg: I bet the apipie specs allow it. And maybe it will spare you an hour of debugging once `foreman-weird-plugin` uses it.

plugins/module_utils/foreman_helper.py
```diff
@@ -1510,6 +1503,27 @@ def _recursive_dict_keys(a_dict):
     return keys
 
 
+def _recursive_dict_without_none(a_dict, exclude=None):
+    """
+    Remove all entries with `None` value from a dict, recursively.
+    Also drops all entries with keys in `exclude`
```
```suggestion
    Also drops all entries with keys in `exclude` in the top level.
```
I agree that we should not drop the aliases from the nested structures.

mdellweg (approved):
Left a small idea for the docstring.

PR #1068, opened 2020-11-19
Activation key role

wbclark:


roles/activation_keys/README.md
```diff
@@ -0,0 +1,79 @@
...
+
+The following fields are required for an activation key but have defaults which make them optional for this role:
+
+- `lifecycle_environment`: Lifecycle Environment to assign to hosts registered with this activation key. Defaults to "Library".
+- `content_view`: Content View to assign to hosts registered with this activation key. Defaults to "Default Organization View".
+
+The following fields are not required but nevertheless have defaults in this role which make it easier to use:
+
+- `auto_attach`: Auto Attach behavior for the activation key. When true, it will attempt to attach a minimum of subscriptions (from the subset of assigned subscriptions on the activation key; selects from all subscriptions in the organization if none are assigned) to cover any present products on the host. When false, it will attempt to attach all subscriptions assigned on the activation key to the host at registration time.
+- `unlimited_hosts`: Allow an unlimited number of hosts to register with the activation key when true. When false, the `max_hosts` parameter which sets a numerical limit on the nnumber of hosts that can be registered becomes required. Defaults to true.
```
evgeni: I *think* these are the API defaults, so you should not need to override them.

Of course https://theforeman.org/plugins/katello/3.17/api/apidoc/v2/activation_keys/create.html doesn't tell if these are default…

But there is a migration doing this: https://github.com/Katello/katello/blob/abbff485b07e94f8dd1dc1caeff4fc23f4e1b356/db/migrate/20141204203609_add_default_value_to_auto_attach.rb

ehelms: I also think these are by default true and true so we can set them to omit rather than specifying them.

ehelms (approved):

roles/activation_keys/README.md
```diff
@@ -0,0 +1,79 @@
...
+The main data structure for this role is the list of `activation_keys`. Each `activation_key` requires the following fields:
+
+- `name`: The name of the activation key.
+
+The following fields are required for an activation key but have defaults which make them optional for this role:
+
+- `lifecycle_environment`: Lifecycle Environment to assign to hosts registered with this activation key. Defaults to "Library".
+- `content_view`: Content View to assign to hosts registered with this activation key. Defaults to "Default Organization View".
+
+The following fields are optional in the sense that the API will use default values when they are omitted:
```
evgeni: I wonder if we should write "API" here, it's an implementation detail the user won't care about. How about "Foreman" or "server"?

evgeni (approved):
One wording nit in the docs, but not blocking

roles/activation_keys/README.md
wbclark: Good point. I think "server" is the clearest.

evgeni: There are a few more api strings in the following sections, mind tackling those too?

roles/activation_keys/README.md
```diff
@@ -0,0 +1,79 @@
...
+The following fields are required for an activation key but have defaults which make them optional for this role:
+
+- `lifecycle_environment`: Lifecycle Environment to assign to hosts registered with this activation key. Defaults to "Library".
+- `content_view`: Content View to assign to hosts registered with this activation key. Defaults to "Default Organization View".
+
+The following fields are optional in the sense that the server will use default values when they are omitted:
+
+- `auto_attach`: Auto Attach behavior for the activation key. When true, it will attempt to attach a minimum of subscriptions (from the subset of assigned subscriptions on the activation key; selects from all subscriptions in the organization if none are assigned) to cover any present products on the host. When false, it will attempt to attach all subscriptions assigned on the activation key to the host at registration time. API defaults to true.
+- `unlimited_hosts`: Allow an unlimited number of hosts to register with the activation key when true. When false, the `max_hosts` parameter which sets a numerical limit on the nnumber of hosts that can be registered becomes required. API defaults to true.
```
ehelms: I think this is a candidate to convert from API to Server

roles/activation_keys/README.md
```diff
@@ -0,0 +1,79 @@
...
+- `name`: The name of the activation key.
+
+The following fields are required for an activation key but have defaults which make them optional for this role:
+
+- `lifecycle_environment`: Lifecycle Environment to assign to hosts registered with this activation key. Defaults to "Library".
+- `content_view`: Content View to assign to hosts registered with this activation key. Defaults to "Default Organization View".
+
+The following fields are optional in the sense that the server will use default values when they are omitted:
+
+- `auto_attach`: Auto Attach behavior for the activation key. When true, it will attempt to attach a minimum of subscriptions (from the subset of assigned subscriptions on the activation key; selects from all subscriptions in the organization if none are assigned) to cover any present products on the host. When false, it will attempt to attach all subscriptions assigned on the activation key to the host at registration time. API defaults to true.
```
ehelms: I think this is a candidate to convert from API to Server

Issue #541, opened 2019-10-21
Improve subscription search by name when adding subs to activation keys


wbclark:
##### SUMMARY
When a user wants to add subscriptions to an activation key, currently they can use the `pool_id` to get an exact match, but using the subscription `name` will fail in the common case that the `name` is one to many. Searching by `name` should be able to handle the one to many case and provide additional search filters such as `service_level` and `expiration_date` to give the user better control over what subscriptions are added to the activation key without depending on the `pool_id` which won't be available until manifest has been imported and/or custom products have been created.

##### ISSUE TYPE
 - Feature Idea

wbclark:
@sean797 : I have opened https://github.com/Katello/katello/pull/8396 to allow Katello API users to search subscriptions using `upstream_pool_id` so that we can use the functionality here

PR #1069, opened 2020-11-19
Add Subscriptions to Activation Keys by Upstream Pool ID

wbclark:


plugins/modules/activation_key.py
```diff
@@ -275,10 +283,14 @@ def main():
             if subscriptions is not None:
                 desired_subscriptions = []
                 for subscription in subscriptions:
-                    if subscription.get('name') is not None and subscription.get('pool_id') is None:
+                    if subscription.get('name') is not None and subscription.get('pool_id') is None and subscription.get('upstream_pool_id') is None:
```
mdellweg: If the required_one_of and mutually_exclusive work properly, this logic can be simplified a lot.

evgeni: that's true, I think this check pre-dates the `required_one_of` and `mutually_exclusive` above, worth checking if we can drop it

plugins/modules/activation_key.py
```diff
@@ -47,21 +47,28 @@
...
           - Pool ID of the Subscription to be added.
-          - Mutually exclusive with I(name).
+          - Mutually exclusive with I(name) and I(upstream_pool_id).
+        type: str
+        required: false
+      upstream_pool_id:
+        description:
+          - Upstream Pool ID of the Subscription to be added.
+          - Mutually exclusive with I(name) and I(pool_id).
+          - Supported since Katello 3.15.0
```
evgeni: I must admit I prefer not to have version numbers in doc strings, for two reasons:
* upstream moves quickly (heck, 3.15 is not supported anymore ;)) and this makes the comment kinda out of date quickly
* downstream the users don't necessarily know which component versions they have.

I wonder if there is a trivial way to deduct support for that param from the apidoc and warn the user in the case it's not supported.

either way, shouldn't block this pr

evgeni:
can you please add a changelog fragment in `changelogs/fragments`? something like:

```yaml
minor_changes:
  - activation_key - add support for selecting subscriptions by ``upstream_pool_id``
```

wbclark:
#541 is worded in a way that is pretty general but we could consider closing it with this

wbclark:
@mdellweg @evgeni Thanks for the reviews. I've pushed an update based on your feedback.

plugins/modules/activation_key.py
```diff
@@ -275,10 +282,14 @@ def main():
...
                 for subscription in subscriptions:
-                    if subscription.get('name') is not None and subscription.get('pool_id') is None:
+                    if subscription.get('name') is not None:
                         desired_subscriptions.append(module.find_resource_by_name('subscriptions', subscription['name'], params=scope, thin=True))
                     if subscription.get('pool_id') is not None:
                         desired_subscriptions.append(module.find_resource_by_id('subscriptions', subscription['pool_id'], params=scope, thin=True))
+                    if subscription.get('upstream_pool_id') is not None:
+                        desired_subscriptions.append(
+                            module.find_resource_by('subscriptions', 'upstream_pool_id', subscription['upstream_pool_id'], params=scope, thin=True)
+                        )
```
mdellweg, line 292: :+1:

mdellweg (approved):

evgeni:
Is upstream pool id unique? Or do we have the same issue as with the name that we get multiple subs that have the same name but different expiration dates or support levels?

I'd not close #541 here

evgeni (approved):

mdellweg:
> Is upstream pool id unique? Or do we have the same issue as with the name that we get multiple subs that have the same name but different expiration dates or support levels?

The module Docs (in this PR) state that they are not unique.

evgeni:
Then let's not close #541, thanks!

wbclark:
@evgeni @mdellweg 

To answer the previous question, Upstream Pool IDs are "more unique" than Subscription Names, but not guaranteed to be fully unique.

It used to be the case that entitlement count for an existing pool on a manifest could not be modified on the RH portal, so an user wishing to add additional entitlements from the same upstream pool could only create a new downstream pool with the additional entitlements.

Now the 'subscription allocations' interface on the RH portal is updated and allows modifying the account of existing downstream pools and does not allow creating a separate downstream pool from the same upstream pool; however, nothing prevents that from existing on a manifest and users could still use an older manifest that was created in that fashion.

What is worse is that additionally there is an interface in Katello itself for adding subscriptions to the manifest; this creates a 'bind entitlements to allocation' task that, when additional entitlements are added from an upstream pool which is already used on the manifest, does in fact create a separate downstream pool instead of updating the quantity on the existing downstream pool.

Interestingly, for subscriptions that create a derived entitlement that is for 'temporarily unmapped guests', the Candlepin database stores the upstream pool IDs for these meaning they are not unique for Candlepin itself, but the upstream_pool_id field in the katello_pools table of Foreman database is actually blank for those derived entitlements. (And there are edge cases where an user would want to attach such derived entitlements to an activation key. To my knowledge that would happen with a 'guests of virt-who-*' entitlement and not with the 'temporarily unmapped guest' entitlement, but that still presents a barrier to being able to use upstream pool IDs in all cases).

After researching and confirming these issues, I agree with the call to leave #541 open.

evgeni:
Whoa, thanks for that elaborate explanation! Much appreciated!

PR #1080, opened 2020-11-27
compose inventory_hostname

### LINKED ISSUE #1070: Option to import hosts with hostname/shortname only
##### SUMMARY
A option which we could use in the plugin to remove the domain part from "host.domain" when importing a Foreman inventory would be very useful to have a inventory only with hostname/shortname of hosts imported from Foreman.

##### ISSUE TYPE
 - Feature Idea

evgeni:
Fixes: #1070

### DIFF
--- a/plugins/inventory/foreman.py
+++ b/plugins/inventory/foreman.py
@@ -144,16 +144,29 @@ DOCUMENTATION = '''
         description: Toggle, if true the inventory will fetch content view details that the host is tied to.
         type: boolean
         default: True
+      hostnames:
+        description:
+          - A list of templates in order of precedence to compose inventory_hostname.
+          - If the template results in an empty string or None value it is ignored.
+        type: list
+        default: ['name']
 '''
 
 EXAMPLES = '''
 # my.foreman.yml
 plugin: theforeman.foreman.foreman
-url: http://localhost:2222
-user: ansible-tester
-password: secure
-validate_certs: False
+url: https://foreman.example.com
+user: ansibleinventory
+password: changeme
 host_filters: 'organization="Web Engineering"'
+
+# shortname.foreman.yml
+plugin: theforeman.foreman.foreman
+url: https://foreman.example.com
+user: ansibleinventory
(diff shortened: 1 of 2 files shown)


plugins/inventory/foreman.py
```diff
@@ -492,87 +539,93 @@ def _populate_report_api(self):
...
             hostvars = self.inventory.get_host(host_name).get_vars()
             self._set_composite_vars(self.get_option('compose'), hostvars, host_name, strict)
             self._add_host_to_composed_groups(self.get_option('groups'), hostvars, host_name, strict)
             self._add_host_to_keyed_groups(self.get_option('keyed_groups'), hostvars, host_name, strict)
 
     def _populate_host_api(self):
+        hostnames = self.get_option('hostnames')
+        strict = self.get_option('strict')
         for host in self._get_hosts():
-            if host.get('name'):
```
evgeni, line 503: while this looks like a big change in the diff, it's actually just mostly re-indenting the code

evgeni:
this is inspired by the [vmware](https://github.com/ansible-collections/community.vmware/blob/main/plugins/inventory/vmware_vm_inventory.py) inventory, which supports similar composition

plugins/inventory/foreman.py
nesanton: took me a while to comprehend that ;)

nesanton:
fwiw lgtm

plugins/inventory/foreman.py
```diff
@@ -144,16 +144,29 @@
...
+password: changeme
 host_filters: 'organization="Web Engineering"'
+
+# shortname.foreman.yml
+plugin: theforeman.foreman.foreman
+url: https://foreman.example.com
+user: ansibleinventory
+password: changeme
+hostnames:
+  - name.split('.')[0]
```
nesanton, line 169: Never seen companies putting dots in shortnames. Oh wait ...

nesanton (approved):

PR #1083, opened 2020-12-01
add user management role

IncredibleRichie:
Hi everybody, we want to share/add this role.
As i have seen on other pr's i need a testplaybook and a test, could someone explain me how to record the test?


mdellweg:
Thank you @IncredibleRichie for sharing this role. In triage we discussed, that it brings good value, but it needs to be polished to be included.
First of all it's name should probably be `theforeman.foreman.users` so it can be vendored easily.
Writing tests is almost the same as for modules:
  * adding a test playbook `users_role.yml`
  * linking the apidoc json in `tests/fixtures/apidoc` accordingly
  * generating fixtures with `make record_users_role`
The test playbook can assume that the whole collection is installed as `theforeman.foreman`.

IncredibleRichie:
I hope you had a good start to the year.
I don't know if I understood correctly but renaming it to theforeman.foreman.users doesn't work because ansible can't find the role. I renamed it to foreman_users. 

I have now followed the steps but I think something is still not working for me, the test tells me that the foreman server cannot be reached. What am I doing wrong?

mdellweg:
Happy new year, @IncredibleRichie ! :fireworks: 

There is an ongoing discussion about namespacing variables here https://github.com/theforeman/foreman-ansible-modules/issues/1118 . Maybe keep a look on this.

Renaming it to `users` should work without a name conflict, and than it will be accessible via `theforeman.foreman.users`.
Also i think there are still some `or_` prefixes left.

roles/users/defaults/main.yml
```diff
@@ -1,44 +1,44 @@
...
 map_external_usergroups: false
-
+hide_password_in_log: true
 # users
 # to create multiple users copy
 # from  - username: to organizations: - "ORG" and paste it below.
 # The indentation must be correct
-or_users:
+users:
   - username: "Gandalf"
```
mdellweg: Maybe you want to provide an example here, but the role default would be better off to be empty.

roles/users/defaults/main.yml
```diff
@@ -1,44 +1,44 @@
...
+      - "{{ location }}"
     organizations:
-      - "{{ or_organization }}"
+      - "{{ organization }}"
 
 # Usergroups
 # For multiple usergroups it is to proceed as before with the previous example
-or_usergroups:
+usergroups:
   - usergroupname: "Hobbits"
```
mdellweg: same here

IncredibleRichie:
Thank you, ok i will follow the discussion.
I renamed the role to users and removed the or_ prefix. 
My error still occurs, can´t remember if i´ve to set up more but there is no foreman server reachable for the test. Maybe I can ask someone next week to have a look at it.

roles/users/tasks/main.yml
```diff
@@ -65,6 +66,7 @@
     attr_photo: "{{ attr_photo | default(omit) }}"
     state: present
   when: enable_ldap_auth |bool
+  no_log: "{{ hide_password_in_log | default (true) }}"
```
mdellweg: `hide_password_in_log` is in the roles default. So no use to provide another default here.

IncredibleRichie:
You're absolutely right I wanted to make it as simple as possible, but then it's no longer clean.

IncredibleRichie:
I have now got the tests up and running, but before I upload them I wonder if I can prevent the passwords from getting into the "fixtures" files. 
Or do I have to delete them manually?

roles/users/defaults/main.yml
```diff
@@ -6,59 +6,3 @@ create_usergroups: true
 enable_ldap_auth: false
 map_external_usergroups: false
 hide_password_in_log: true
-# users
```
mdellweg: You can still provide an empty array:
`users: []`

mdellweg:
Passwords are always a delicate thing. Thanks that you are asking. Are the passwords you are talking about the ones to access the foreman or are they used as the users passwords?
In the latter case, we could just use dumb passwords, right? The other solution might be to find a way to filter the passwords in the vcr_python_wrapper. I'd like to avoid that.

We do already filter authentication info and the server name:
https://github.com/theforeman/foreman-ansible-modules/blob/develop/tests/vcr_python_wrapper.py#L12
https://github.com/theforeman/foreman-ansible-modules/blob/develop/tests/vcr_python_wrapper.py#L113

IncredibleRichie:
For me it is about the ldap password, the other values are already "dumb" passwords.
I understand, what if i delete them manually?

mdellweg:
> For me it is about the ldap password, the other values are already "dumb" passwords.
> I understand, what if i delete them manually?

Then they will reappear if someone needs to rerecord.

IncredibleRichie:
Right but not when i delete it in the test-playbook too. Doesn´t the rerecorder need his own password for his ldap?

mdellweg:
> Right but not when i delete it in the test-playbook too. Doesn´t the rerecorder need his own password for his ldap?

In that case it should not be in the playbook verbatim, but provided via a variable, like this https://github.com/theforeman/foreman-ansible-modules/blob/develop/tests/test_playbooks/vars/server.yml.example#L14 . At the same time, the matcher should ignore it's value and the persister should filter it in the vcr_python_wrapper.
Also if recording that test needs some intricate setup, please leave a note in the playbook like here: https://github.com/theforeman/foreman-ansible-modules/blob/15322e91b61618074908ad02e4b2dfbb9f058113/tests/test_playbooks/smart_proxy.yml#L2

tests/test_playbooks/users_role.yml
```diff
@@ -25,6 +31,7 @@
     server_url: "{{ foreman_server_url }}"
     validate_certs: "{{ foreman_validate_certs }}"
     organization: "Test Organization"
+    location: Minga
```
mdellweg: ;)
I believe that location is where your server is located. But you should add `tasks/location.yml` to the included tasks and use, i guess, "Test Location".

IncredibleRichie:
I have moved the variables into the server.yml and made a note in the playbook. Unfortunately, nothing is filtered, so I have replaced the sensitive data. Any idea?

tests/test_playbooks/users_role.yml
```diff
@@ -0,0 +1,73 @@
+---
+# The following sensitive variables can be specified in the server.yml
+# ldap_host:
+# ldap_account:
+# ldap_password:
+# ldap_base_dn:
+# ldap_groups_base:
```
mdellweg, line 7: If those are sensitive, you should provide dumb examples in the `server.yml.example´ and, more importantly, filter them in `vcr_python_wrapper` from being written to the `tests/test_playbooks/fixtures/users_role-*.yml` files.

tests/test_playbooks/users_role.yml
IncredibleRichie: :-)

mdellweg (commented):
Can you make `tests/fixtures/apidoc/users_role.json` a symlink?

This looks like a big step in the right direction. :+1:

tests/test_playbooks/users_role.yml
IncredibleRichie: I have added examples and extended the filter in the "wrapper.py". The result/output is now good for me.

IncredibleRichie: Can you explain to me why the tests here are not successful? Mine were ok. 
I can only find a message like "can not open to record" or similar.

mdellweg: Sorry for the delay. I will try to find time to dig into this.
Looking at the logs it says it found 1 similar request with no failed matcher. This does not make sense to me.

IncredibleRichie: N.p. . Me neither ^^

mdellweg:
Hey, i finally got around it. My observation is, that the `re{quest,sponse}.body` is of type `bytes` and `json.dumps` makes it `str`.
So the short answer is "add `.encode()` to `json.dumps()`. But as i did some cleanup on the way to understanding, you may just want to cherrypick this comit:

https://github.com/mdellweg/foreman-ansible-modules/commit/692f72e908851b08fa2dac96dbfda24bf9af4f00

However i still want to ask if it's safe to filter "host" from _all_ responses in _all_ tests.

IncredibleRichie:
Thank you. i finally got around to it too, even though there was only a small part to do. 

I picked your changes in the vcr_wrapper on the fly. 

As for your question, I don't think everyone needs to know the hostname/ip

tests/vcr_python_wrapper.py
```diff
@@ -108,45 +108,44 @@ def filter_response(response):
...
-            json_sbody['results'][0]['base_dn'] = 'FILTERED'
-        if "groups_base" in json_sbody['results'][0]:
-            json_sbody['results'][0]['groups_base'] = 'FILTERED'
-            response['body']['string'] = json.dumps(json_sbody)
-    except Exception:
+        json_body = json.loads(response['body']['string'])
+    except json.JSONDecodeError:
         pass
+    else:
+        saved_str = response['body']['string']
```
mdellweg: This is used nowhere, right?

IncredibleRichie: I think so too

IncredibleRichie: Removed it

IncredibleRichie: Unfortunately, I cannot identify the errors that the linter outputs.

mdellweg:
Ahh, it took me some time, too.
Seems like there was another PR merged that added filtering for post data in the meantime.
So the code that is executed is the new develop and your changes merged. Somehow this did not produce a visible merge conflict, but still the merge affects the functionality. Can you please rebase, and look at the new `filter_request` function that (in turn) calls the old `filter_request_uri` plus another filter? I'd suggest to add your filter in that manner, too.

IncredibleRichie:
After some problems i got it rebased. Now lint is green but there are still some issues with the filter i guess

tests/vcr_python_wrapper.py
```diff
@@ -107,11 +107,44 @@ def filter_response(response):
...
+                json_body['results'][0]['base_dn'] = 'FILTERED'
+            if "groups_base" in json_body['results'][0]:
+                json_body['results'][0]['groups_base'] = 'FILTERED'
+        response['body']['string'] = json.dumps(json_body).encode()
     return response
 
 
 def filter_request_uri(request):
     request.uri = urlunparse(urlparse(request.uri)._replace(netloc="foreman.example.org"))
+    if request.body is not None:
```
mdellweg, line 150: Here i think, you should create a new filter function (like `filter_request_manifest`) and call it in the `filter_request` function:
https://github.com/theforeman/foreman-ansible-modules/pull/1083/files#diff-98ccd5edeea7871fb84b2db28d5b369f17bbdfa862faf5ab73618886be95e0a7R157

tests/vcr_python_wrapper.py
```diff
@@ -107,11 +107,44 @@ def filter_response(response):
...
+            json_body['host'] = 'FILTERED'
+        if "account" in json_body:
+            json_body['account'] = 'FILTERED'
+        if "base_dn" in json_body:
+            json_body['base_dn'] = 'FILTERED'
+        if "groups_base" in json_body:
+            json_body['groups_base'] = 'FILTERED'
+        # another for the search result
+        if 'results' in json_body:
+            if "host" in json_body['results'][0]:
```
mdellweg: It seems this line is failing if the results array is empty.
Maybe you should filter this only if `length(json_body["results"]) ==1` or you should filter every item in results.

@evgeni thoughts?

IncredibleRichie:
I´ve added your proposed changes

tests/vcr_python_wrapper.py
```diff
@@ -121,9 +146,24 @@ def filter_request_manifest(request):
...
+            if 'auth_source_ldap' in json_body:
+                json_body['auth_source_ldap'] = 'FILTERED'
+                request.body = json.dumps(json_body).encode()
+    return request
+
+
 def filter_request(request):
     request = filter_request_uri(request)
     request = filter_request_manifest(request)
+    request = filter_request_ldap(request)
```
mdellweg, line 165: :+1:

tests/vcr_python_wrapper.py
```diff
@@ -121,9 +146,24 @@ def filter_request_manifest(request):
     return request
 
 
+def filter_request_ldap(request):
+    request.uri = urlunparse(urlparse(request.uri)._replace(netloc="foreman.example.org"))
+    if request.body is not None:
+        try:
+            json_body = json.loads(request.body)
+        except json.JSONDecodeError:
```
mdellweg: https://stackoverflow.com/a/35214768
Seems like we should check for `ValueError` instead, to make py27 happy.

tests/vcr_python_wrapper.py
```diff
@@ -121,9 +146,24 @@ def filter_request_manifest(request):
     return request
 
 
+def filter_request_ldap(request):
+    request.uri = urlunparse(urlparse(request.uri)._replace(netloc="foreman.example.org"))
```
mdellweg: This replacement is done in `filter_request_uri`.

tests/vcr_python_wrapper.py
```diff
@@ -107,6 +107,31 @@ def filter_response(response):
         # headers should be case insensitive, but for some reason they weren't for me
         response['headers'].pop(header.lower(), None)
         response['headers'].pop(header, None)
+    try:
+        json_body = json.loads(response['body']['string'])
+    except json.JSONDecodeError:
```
mdellweg, line 112:
```suggestion
    except ValueError:
```

tests/vcr_python_wrapper.py
```diff
@@ -107,6 +107,31 @@ def filter_response(response):
...
+        if "account" in json_body:
+            json_body['account'] = 'FILTERED'
+        if "base_dn" in json_body:
+            json_body['base_dn'] = 'FILTERED'
+        if "groups_base" in json_body:
+            json_body['groups_base'] = 'FILTERED'
+        # another for the search result
+        if 'results' in json_body:
+            if len(json_body["results"]) == 1:
+                if "host" in json_body['results'][0]:
```
mdellweg, line 126: Still seems to fail on this line, but i don't see why.

tests/vcr_python_wrapper.py
mdellweg: This function seems to break a lot of existing tests. Maybe we need to restrict it to the user role test?

IncredibleRichie: Not sure how to do it, there are still some areas i don´t know/understand.

tests/vcr_python_wrapper.py
IncredibleRichie: Maybe there is always a result, but empty? Then it would make sence to filter every item

tests/vcr_python_wrapper.py
mdellweg: we have some test specific filters like:
https://github.com/theforeman/foreman-ansible-modules/blob/develop/tests/vcr_python_wrapper.py#L161

But that variable is not available in the scope of `filter_request`.

IncredibleRichie:
I close this pull request and open a new one. I have set up a contaier ldap so there is no sensitive data

PR #1086, opened 2020-12-03
add module to import puppet classes from proxies

evgeni:


ehelms (approved):

evgeni:
the job_template sanity failures are fixed in https://github.com/theforeman/foreman-ansible-modules/pull/1088

evgeni:
Merging with failing checks, as those are not from this PR and a fix is open.

PR #1092, opened 2020-12-09
don't scope create/update requests by Organization/Location

evgeni:
Foreman/Apipie allows the same short parameter name to be used multiple
times in an API definition and we fill them identically, assuming they
mean the same.
However, this is not true. For example, when creating a Host, the "outer"
organization_id means to "scope" the request by that ID, while the
"inner" says that the Host is to be created in said Organization.
While this does not negatively affect Host *creation*, you cannot
*update* a Host to be moved into a different Organization.

### DIFF
--- a/changelogs/fragments/bz1901716-dont_scope_requests.yml
+++ b/changelogs/fragments/bz1901716-dont_scope_requests.yml
@@ -1,0 +1,2 @@
+bugfixes:
+  - host - allow moving hosts between Organizations and Locations (https://bugzilla.redhat.com/show_bug.cgi?id=1901716)
--- a/plugins/module_utils/foreman_helper.py
+++ b/plugins/module_utils/foreman_helper.py
@@ -363,6 +363,14 @@ class ForemanAnsibleModule(AnsibleModule):
     def set_changed(self):
         self._changed = True
 
+    def _patch_host_update(self):
+        _host_methods = self.foremanapi.apidoc['docs']['resources']['hosts']['methods']
+
+        _host_update = next(x for x in _host_methods if x['name'] == 'update')
+        for param in ['location_id', 'organization_id']:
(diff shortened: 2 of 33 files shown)


evgeni:
Draft because I didn't update the tests and would like some feedback on the (rather hacky) approach before continuing.

There is sadly no information in the APIdoc what the *use* of a specific param is, so I refrained from putting this into `apypie`.

mdellweg:
Is the main diffference here, that one is a url parameter and the other is supposed to go into the body payload?
Do we even need to care for the scoping of queries here?

evgeni:
```
FAILED tests/test_crud.py::test_crud[host_interface_attributes]
FAILED tests/test_crud.py::test_crud[host]
FAILED tests/test_crud.py::test_crud[scc_account]
```

I have expected to break more tests with this change :rofl:

evgeni:
> Is the main diffference here, that one is a url parameter and the other is supposed to go into the body payload?
> Do we even need to care for the scoping of queries here?

No, both go into the JSON body:

bad (old):
```json
{"location_id": 16, "organization_id": 15, "host": {"name": "test-host.example.net", "location_id": 16, "organization_id": 15, "build": false, "managed": false}}
```

good (new):
```json
{"host": {"name": "test-host.example.net", "location_id": 16, "organization_id": 15, "build": false, "managed": false}}
```

mdellweg:
So we could "hot patch" the docs "fam style" to never use the outer appearance?

evgeni:
We probably could, both sounds like a hack :(

ehelms:
Conceptually this makes sense to me to do.

evgeni:
Updated to use doc patching, seems more straight forward and scoped.

ehelms (approved):

mdellweg (approved):

PR #1093, opened 2020-12-09
Remove scc_product search by friendly_name as its no longer unique

Manisha15:
Earlier scc_product name was not unique but friendly_name was, therefore, it was used to find the product. But with SCC manager v1.8.6, the name is unique but not friendly_name, thus, friendly_name is not needed anymore.


plugins/modules/scc_product.py
```diff
@@ -63,7 +63,9 @@
 def main():
     module = KatelloAnsibleModule(
         foreman_spec=dict(
-            scc_product=dict(required=True, type='entity', aliases=['friendly_name'], scope=['scc_account'], thin=False),
+            # scc_product search by friendly_name alias is deprecated in version v1.8.6
+            # scc_product=dict(required=True, type='entity', aliases=['friendly_name'], scope=['scc_account'], thin=False),
+            scc_product=dict(required=True, type='entity', scope=['scc_account'], thin=False),
```
quba42: If I understand correctly, I don't think we actually need to remove this alias. (If we do remove it we also need to change the example in the module documentation). This alias just means users can use either `scc_product` or `friendly_name`, to specify the search string for the module. Whether that search string then looks over the `name` field, or the `friendly_name` field from the API replies is controlled exclusively via the `ENTITY_KEYS` dict, no?

Of course it might be a bit misleading to leave the alias `friendly_name`, and not actually have it search over the `friendly_name` fields from the API replies, but my intuition is that this is less bad than breaking peoples plays.

plugins/module_utils/foreman_helper.py
```diff
@@ -73,7 +73,9 @@
     #       Applying this will need to record a lot of tests that is out of scope for the moment.
     # organizations='title',
     scap_contents='title',
-    scc_products='friendly_name',
+    # NOTE: scc_product is searched by friendly_name before version v1.8.6.
+    # It is deprecated in version v1.8.6 and should not be used to search scc_product from version v1.8.6
+    # scc_products='friendly_name',
```
quba42: I think a more accurate version of the comment would be something like:
```
# For SCC manager versions prior to v1.8.6 the 'name' field was not unique and 'friendly_name' was used instead.
# with version v1.8.6 this has been reversed (the 'name' field is now unique, but 'friendlly_name' is not).
```
We aren't really deprecating anything, we are simply dropping the `ENTITY_KEYS` redirect. As a result the newer version of the module won't work right for older versions of the SCC manager. I do not know how one could build this change in a backwards compatible way, and it is probably not worth the effort. Users will simply need to update their ansible modules and the SCC manager plugin in tandem...

quba42 (changes requested):

plugins/modules/scc_product.py
mdellweg: If the new "unique name" looks different than what was the "unique friendly_name", users playbooks will break anyway. In that case removing the alias might make the need and the knowlege where to adjust more obvious.

quba42: I guess this would become easier to evaluate once some new test fixtures are added to this PR (I presume we will need new test fixtures for this change :wink:).

quba42:
Note: Here is the Foreman packaging PR for the relevant SCC manager release: https://github.com/theforeman/foreman-packaging/pull/6112

plugins/modules/scc_product.py
quba42: I think I am in favor of chucking out the `friendly_name` Alias. After all, `name` was never the same thing as `friendly_name`, so in some sense it never made sense as an Alias.

evgeni:
I must admit that I think that `scc_manager` with that behaviour change should not be 1.8.6, but 2.0.0, but I am not a release manager for that plugin, so :woman_shrugging:

Now to the changes to the module, which I more care about ;-)

The changes as they are now are backwards incompatible, so we'd have to release FAM 2.0 if we merge this as-is and I think that is not ideal. Can we instead *detect* the version of the plugin and act accordingly?

From the top of my head, I see two options for this:

1. poll [`/api/statuses`](https://theforeman.org/api/2.3/apidoc/v2/ping/statuses.html) and see what version of the plugin is installed:
    ```python
    try:
        statuses = module.foremanapi.resource('ping').call('statuses')
        plugins = statuses['results']['plugins']
        scc_version = '1.0.0'  # fallback
        for plugin in plugins:
            if 'foreman_scc_manager' in plugin:
                scc_version = plugin.split(',')[1]  # yes, we get a string, eww
    except:
        scc_version = '1.8.6'
    if if LooseVersion(scc_version) >= LooseVersion('1.8.6'):
        scc_search_by = 'name'
    else:
        scc_search_by = 'friendly_name'
    …
    ```

2. looking at the [diff for 1.8.6](https://github.com/ATIX-AG/foreman_scc_manager/compare/v1.8.5...v1.8.6), there was a new param (`katello_gpg_key_id`) added to the `scc_account` API endpoint, so we could match on this being present

Manisha15:
@evgeni Can you please tell me where exactly this version checking should be done, in foreman_helper or in scc_product module plugin? Also, Is it possible to record fixtures for both the cases of versions, i.e, have fixtures for version < 1.8.6 as well as version >=1.8.6 in the repository?

evgeni:
I'd do it here in the scc module, and yes, recording both should be possible by having two test playbooks.

If you're working tomorrow, we could go over the changes together?

Manisha15:
Yes we can discuss the changes tomorrow. Thank you.

Manisha15:
@evgeni I recorded fixtures for two versions of SCC Manager but for old version of scc_product it is checking fixtures of new scc_product version. How can this be fixed?

plugins/modules/scc_product.py
```diff
@@ -72,6 +71,22 @@ def main():
...
     with module.api_connection():
+        try:
+            statuses = module.foremanapi.resource('ping').call('statuses')
+            plugins = statuses['results']['foreman']['plugins']
+            scc_version = '1.0.0'  # fallback
+            for plugin in plugins:
+                if 'foreman_scc_manager' in plugin:
+                    scc_version = plugin.split(',')[1]
+        except Exception:
+            scc_version = '1.8.6'
```
evgeni: I think I got the logic here a tad wrong: the `try/except` is to support Foreman versions that didn't have the `statuses` API yet, but then we should not fall-back to using 1.8.6 as the version.

```suggestion
        scc_version = '1.0.0'  # fallback
        try:
            statuses = module.foremanapi.resource('ping').call('statuses')
            plugins = statuses['results']['foreman']['plugins']
            for plugin in plugins:
                if 'foreman_scc_manager' in plugin:
                    scc_version = plugin.split(',')[1]
        except Exception:
            pass
```

That should be cleaner

evgeni (changes requested):
The failure you see is because the `scc_manager.json` apidoc fixture you use is from a Foreman version that has no `statuses` API and it always ends up in the "fallback" case in CI (recording works as that fixture is not used there) and thus miss-identifies the running plugin versions.

If you update the JSON to a current version it should work (regardless of my change suggestion above, but that would be nice to have)

plugins/modules/scc_product.py
```diff
@@ -16,6 +16,7 @@
 # along with this program.  If not, see <http://www.gnu.org/licenses/>.
 
 from __future__ import absolute_import, division, print_function
+from distutils.version import LooseVersion
```
evgeni: Oh, could you please move this import to line 60 where the other Ansible-related imports are? Thanks!

plugins/modules/scc_product.py
```diff
@@ -58,12 +56,12 @@
 
 
 from ansible_collections.theforeman.foreman.plugins.module_utils.foreman_helper import KatelloAnsibleModule
-
+from distutils.version import LooseVersion
```
evgeni, line 64: `plugins/modules/scc_product.py:61:1: E302 expected 2 blank lines, found 1`

the linter is picky, you need another empty line after the import.

evgeni:
now the APIdocs don't have SCC Manager at all?

plugins/modules/scc_product.py
Manisha15: Fixed!

Manisha15:
> now the APIdocs don't have SCC Manager at all?

Changed it and now it should work.

evgeni:
Thanks @Manisha15! This looks great!

I think the only concern I have here is the removal of the `friendly_name` alias. For users on 1.8.5 and below, `friendly_name` is still "valid" (well, as valid as it was before), so we'd break their playbooks with this change, which means it's an API break and thus would require a 2.0 release (which I'd love to avoid).

How about we keep the `friendly_name` alias, but document it like this:

```
options:
  scc_product:
    description:
      - Full name of the product of suse customer center account.
      - The I(friendly_name) alias is deprecated as it refers to an attribute that does not uniquely identify a product and not used for product lookups since SCC Manager 1.8.6.
    required: true
    type: str
    aliases:
      - friendly_name
```

We can then remove it in 2.0 whenever that is released.

Manisha15:
> - Full name of the product of suse customer center account.
>       - The I(friendly_name) alias is deprecated as it refers to an attribute that does not uniquely identify a product and not used for product lookups since SCC Manager 1.8.6.

This change is throwing sanity check error:
` Argument 'friendly_name' is listed in DOCUMENTATION.options, but not accepted by the module argument_spec`

Manisha15:
> > * Full name of the product of suse customer center account.
> >   - The I(friendly_name) alias is deprecated as it refers to an attribute that does not uniquely identify a product and not used for product lookups since SCC Manager 1.8.6.
> 
> This change is throwing sanity check error:
> ` Argument 'friendly_name' is listed in DOCUMENTATION.options, but not accepted by the module argument_spec`

Is the error because we don't use alias option in argument_spec at all and instead search for scc_product (see example, it uses `scc_product: "Product1"` and not `friendly_name: "product 1"`) and later use search_by argument to search for friendly name?

plugins/modules/scc_product.py
```diff
@@ -58,12 +61,13 @@
...
 
 from ansible_collections.theforeman.foreman.plugins.module_utils.foreman_helper import KatelloAnsibleModule
+from distutils.version import LooseVersion
 
 
 def main():
     module = KatelloAnsibleModule(
         foreman_spec=dict(
-            scc_product=dict(required=True, type='entity', aliases=['friendly_name'], scope=['scc_account'], thin=False),
+            scc_product=dict(required=True, type='entity', scope=['scc_account'], thin=False),
```
evgeni: CI complains about this now, as you allow the `friendly_name` alias in the documentation, but not in the spec.

Please revert this line to read
```python
scc_product=dict(required=True, type='entity', aliases=['friendly_name'], scope=['scc_account'], thin=False),
```
and we should be good to go!

Manisha15: Fixed!

evgeni:
@quba42 @mdellweg could you have a look whether the solution Manisha and I came up here here is sane?

Otherwise this is an ACK and I'll merge it ASAP :)

evgeni (approved):

quba42 (approved):
The code changes LGTM. Thanks @evgeni for all your help on this one. I think I learned a thing or two about Ansible modules. (And that we should get a better release process for the Foreman SCC manager plugin :stuck_out_tongue_winking_eye:)

I don't think I fully understand how recording two sets of test fixtures works, but I trust your expertise (and the green test result) on that.

quba42 (approved):

plugins/modules/scc_product.py
```diff
@@ -72,6 +76,22 @@ def main():
     module.task_timeout = 4 * 60
 
     with module.api_connection():
+        scc_version = '1.0.0'  # fallback
+        try:
+            statuses = module.foremanapi.resource('ping').call('statuses')
```
mdellweg, line 81: Didn't we call ping in the `api_connect` process already? Maybe we can "Just" cache the result there?

This would be fine in a subsequent PR, however.

plugins/modules/scc_product.py
```diff
@@ -72,6 +76,22 @@ def main():
...
+                    scc_version = plugin.split(',')[1]
+        except Exception:
+            pass
+
+        if LooseVersion(scc_version.strip()) < LooseVersion('1.8.6'):
+            scc_search_by = 'friendly_name'
+        else:
+            scc_search_by = 'name'
+
+        module.foreman_spec['scc_product']['search_by'] = scc_search_by
```
mdellweg, line 94: This is a first, to my knowlege. We patch the foreman_spec on the fly that is completely in our controll.
:+1: anyway.

mdellweg (approved):
@quba42 as far as I understand, the existence of the scc_product_old.yml playbook introduces a second test for that module using the corresponding old api-fixtures.
There is really no technical relationship between the name of the test playbook and the model. It's just a convenient convention.

plugins/modules/scc_product.py
evgeni: it's not `ping`, it's `statuses`, different (more expensive) API

plugins/modules/scc_product.py
evgeni: It is, and it isn't, because when we write the spec, we never talked to the server ;)

Manisha15:
Thanks @evgeni for your help and merging the PR. @quba42 and @mdellweg Thanks for reviewing the PR.

PR #1098, opened 2020-12-14
refactor resolved tracking into an own function

evgeni:


plugins/module_utils/foreman_helper.py
```diff
@@ -1541,6 +1540,10 @@ def _recursive_dict_without_none(a_dict, exclude=None):
     return result
 
 
+def _is_resolved(what):
+     return (what is None or isinstance(what, (dict, list)))
```
mdellweg: Not sure if it's really better:
```suggestion
     return isinstance(what, (dict, list, NoneType))
```

evgeni: You can't `isinstance` on `NoneType` :(

mdellweg:
If you pass `{entity,nested}_spec` to `_is_resolved` you can add the type check there too.
I'd consider non entity types to be inherently "resolved".

mdellweg:
If you pass `{entity,nested}_spec` to `_is_resolved` you can add the type check there too.
I'd consider non entity types to be inherently "resolved".

plugins/module_utils/foreman_helper.py
```diff
@@ -1541,6 +1540,10 @@ def _recursive_dict_without_none(a_dict, exclude=None):
     return result
 
 
+def _is_resolved(what):
+    return (what is None or isinstance(what, (dict, list)))
```
evgeni: Right, you can't check if it's a list, when it's already a list of strings. If it's a list, the elements need to be dicts.

mdellweg: That means the test in the nested thing is not entirely correct, but it didn't trigger because we didn't have entity_lists there?
If you passed the spec, you could make the test more specific.

evgeni: Exactly.

evgeni: How about that.

plugins/module_utils/foreman_helper.py
```diff
@@ -1541,6 +1539,19 @@ def _recursive_dict_without_none(a_dict, exclude=None):
...
 
 
+def _is_resolved(spec, what):
+    if spec.get('type') not in ('entity', 'entity_list'):
+        return True
+
+    if spec.get('type') == 'entity' and (what is None or isinstance(what, dict)):
+        return True
+
+    if spec.get('type') == 'entity_list' and isinstance(what, list) and what and isinstance(what[0], dict):
```
evgeni: It can also be a list of `None`, right?

mdellweg: Having a list lookup with `failsafe=True` is probably odd, but i think it can be mixed.

evgeni: We do have code for it ;)

https://github.com/theforeman/foreman-ansible-modules/blob/023c9624efa5cbb69fae1571e35db69a152e3985/plugins/module_utils/foreman_helper.py#L790

mdellweg (approved):
Looks good!

PR #1101, opened 2020-12-16
enable python 3.9 tests

evgeni:
this includes #1100 because I am lazy

3.9 is not excluded from sanity, because it should skip it automatically


evgeni:
```
WARNING: Skipping sanity test 'pylint' on unsupported Python 3.9.
```

Works as advertized :)

mdellweg (approved):

PR #1105, opened 2020-12-18
add activation_keys parameter to ease configuring activation keys

evgeni:


evgeni:
this is a result of a discussion with @ericzolf who wanted to config AKs but didn't know they are just a param internally (and he's right, a user shouldn't have to know that)

plugins/module_utils/foreman_helper.py
```diff
@@ -1378,7 +1387,7 @@ def run(self, **kwargs):
                 if entity['digest'] in [digest, digest_stripped]:
                     self.foreman_params.pop('scap_file')
 
-        super(ForemanScapDataStreamModule, self).run(**kwargs)
+        return super(ForemanScapDataStreamModule, self).run(**kwargs)
```
evgeni, line 1398: accidental find while introducing the same bug above XD

plugins/doc_fragments/foreman.py
```diff
@@ -270,6 +270,13 @@ class ModuleDocFragment(object):
       - Only available for Katello installations.
     required: false
     type: str
+  activation_keys:
+    description:
+      - Activation Keys used for deployment.
+      - Coma separated list.
```
mdellweg: Let's put a little consciousness in that list.
```suggestion
      - Comma separated list.
```

plugins/doc_fragments/foreman.py
```diff
@@ -270,6 +270,13 @@ class ModuleDocFragment(object):
...
+      - Only available for Katello installations.
+    required: false
+    type: str
```
mdellweg, line 279: What happens if one *also* provides the param? Should we document who wins?

plugins/module_utils/foreman_helper.py
```diff
@@ -282,15 +282,24 @@ def __init__(self, **kwargs):
...
             ('openscap', ['openscap_proxy']),
         ]
         mutually_exclusive = kwargs.pop('mutually_exclusive', []) + [['medium', 'kickstart_repository']]
         super(HostMixin, self).__init__(foreman_spec=foreman_spec, required_plugins=required_plugins, mutually_exclusive=mutually_exclusive, **kwargs)
 
+    def run(self, **kwargs):
+        if 'activation_keys' in self.foreman_params:
+            if 'parameters' not in self.foreman_params:
+                self.foreman_params['parameters'] = []
+            self.foreman_params['parameters'].append({'name': 'kt_activation_keys', 'value': self.foreman_params.pop('activation_keys')})
```
mdellweg: This looks like the `kt_activation_keys` could end up twice in this list. :exploding_head:

evgeni: Same as when the user provides the same param twice: 🤯

mdellweg: So, not a new problem here. :heavy_check_mark:

evgeni: given the code, and #1106 - I don't think the check from #1106 will catch that :(

mdellweg: Because the check will run as the last part of the `super().__init__()`.
Too bad.
I put the check in the `__init__` to check before we perform the ping. But maybe we need to move it into `run`.

ehelms (approved):

plugins/module_utils/foreman_helper.py
evgeni: followup in https://github.com/theforeman/foreman-ansible-modules/pull/1114

PR #1107, opened 2020-12-18
start dedicated info modules

evgeni:
starting off with `host` as that's rather simple.

next on my list would be `repository` which is a bit more involved, but I hope to be able to abstract most of it into the helper

### DIFF
--- a/plugins/doc_fragments/foreman.py
+++ b/plugins/doc_fragments/foreman.py
@@ -325,3 +325,29 @@ options:
     type: str
     required: False
 '''
+
+    INFOMODULE = '''
+options:
+  name:
+    description:
+      - Name of the resource to fetch information for.
+      - Mutually exclusive with I(search).
+    required: false
+    type: str
+  location:
+    description:
+      - Label of the Location to scope the search for.
+    required: false
+    type: str
+  organization:
+    description:
+      - Name of the Organization to scope the search for.
+    required: false
+    type: str
+  search:
+    description:
+      - Search query to use
+      - If None, and I(name) is not set, all resources are returned.
+      - Mutually exclusive with I(name).
+    type: str
+'''
--- a/plugins/doc_fragments/foreman.py
+++ b/plugins/doc_fragments/foreman.py
@@ -351,3 +351,9 @@ options:
       - Mutually exclusive with I(name).
     type: str
 '''
+
+    KATELLOINFOMODULE = '''
+options:
+  organization:
+    required: true
+'''
--- a/plugins/module_utils/foreman_helper.py
+++ b/plugins/module_utils/foreman_helper.py
@@ -1287,6 +1287,38 @@ class ForemanStatelessEntityAnsibleModule(ForemanAnsibleModule):
         return '_'.join(class_name.split('_')[1:-1])
 
 
+class ForemanInfoAnsibleModule(ForemanStatelessEntityAnsibleModule):
+    """
+    Base class for Foreman info modules that fetch information about entities
+    """
+    def __init__(self, **kwargs):
+        self._resources = []
+        foreman_spec = dict(
+            name=dict(),
+            search=dict(),
+            organization=dict(type='entity'),
+            location=dict(type='entity'),
+        )
+        foreman_spec.update(kwargs.pop('foreman_spec', {}))
+        super(ForemanInfoAnsibleModule, self).__init__(foreman_spec=foreman_spec, **kwargs)
+
+    def run(self, **kwargs):
+        """
+        lookup entities
+        """
+        self.auto_lookup_entities()
+
+        resource = self.foreman_spec['entity']['resource_type']
+
+        if 'name' in self.foreman_params:
+            self._resources = self.lookup_entity('entity')
+        else:
+            self._resources = self.list_resource(resource, self.foreman_params['search'], _flatten_entity(self.foreman_params, self.foreman_spec))
+
+    def exit_json(self, **kwargs):
+        super(ForemanInfoAnsibleModule, self).exit_json(resources=self._resources, **kwargs)
+
+
 class ForemanEntityAnsibleModule(ForemanStatelessEntityAnsibleModule):
     """ Base class for Foreman entities. To use it, subclass it with the following convention:
         To manage my_entity entity, create the following sub class::
--- a/plugins/module_utils/foreman_helper.py
+++ b/plugins/module_utils/foreman_helper.py
@@ -1311,12 +1311,14 @@ class ForemanInfoAnsibleModule(ForemanStatelessEntityAnsibleModule):
         resource = self.foreman_spec['entity']['resource_type']
 
         if 'name' in self.foreman_params:
-            self._resources = self.lookup_entity('entity')
+            self._info_result = {self.entity_name: self.lookup_entity('entity')}
         else:
-            self._resources = self.list_resource(resource, self.foreman_params['search'], _flatten_entity(self.foreman_params, self.foreman_spec))
+            _flat_entity = _flatten_entity(self.foreman_params, self.foreman_spec)
+            self._info_result = {resource: self.list_resource(resource, self.foreman_params['search'], _flat_entity)}
 
     def exit_json(self, **kwargs):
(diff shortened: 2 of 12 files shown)


evgeni:
@ehelms as discussed the other day :)

plugins/modules/host_info.py
```diff
@@ -0,0 +1,77 @@
...
+
+def main():
+    module = ForemanHostInfo()
+
+    with module.api_connection():
+        module.run()
+
+
+if __name__ == '__main__':
+    main()
```
evgeni, line 81: so much real code… ;)

mdellweg: mostly blank lines anyway. ;)

plugins/modules/host_info.py
```diff
@@ -0,0 +1,77 @@
...
+  type: list
+  elements: dict
+'''
+
+from ansible_collections.theforeman.foreman.plugins.module_utils.foreman_helper import (
+    ForemanInfoAnsibleModule,
+)
+
+
+class ForemanHostInfo(ForemanInfoAnsibleModule):
```
mdellweg, line 69: If you called this a `ForemanHostInfoModule` the resource lookup was broken, right?

evgeni: Yeah, then I'd need to adjust `entity_name_from_class` ;)

We could, tho, just /me being lazy.

plugins/modules/host_info.py
evgeni: I wondered about folding `with connection: run` into an `autorun` method?

evgeni: so then the whole main would become:
```
def main():
    ForemanHostInfo().autorun()
```

plugins/module_utils/foreman_helper.py
```diff
@@ -1399,11 +1431,9 @@ class KatelloAnsibleModule(KatelloMixin, ForemanAnsibleModule):
     pass
 
 
-class KatelloEntityAnsibleModule(KatelloMixin, ForemanEntityAnsibleModule):
+class KatelloScopedMixin(KatelloMixin):
```
evgeni, line 1452: I wonder if we really need this, or if the scope enforcement should happen in `KatelloMixin` directly.

plugins/modules/host_info.py
mdellweg: I don't think it's too much boilerplate, and it keeps the simple modules in the same stucture as the ones where we need some manipulation between `api_connection` and `run`.
I think it's a good thing, that you can see this shared inheritance.

plugins/modules/host_info.py
mdellweg: `ForemanHostInfomodule` could work, but maybe that introduces even more confusion.

plugins/module_utils/foreman_helper.py
mdellweg: Maybe yes, i guess, you need to look into an non entity katello module to decide.

tests/test_playbooks/host_info.yml
```diff
@@ -0,0 +1,5 @@
+---
+# TBD
```
ehelms: Is this TBD in this PR or a follow up?

evgeni: This PR, I waa just lazy ;)

evgeni:
So I've added tests, and they led me to some design-y questions…

Currently, the modules return their data in the `resources` attribute (so if you register the host info as `host`, to access it you need to do `host.resources…`), which is an inheritance from the `resource_info` module. But the data structure returned will differ, based on `name` (one item) vs `search` (array of items). Which leads me to the following questions:
* Should the attribute be called `<whatever_resource>` instead? We know the resource name after all
* Should the attribute vary and be `<whatever_resource>` for a single name result, vs `<whatever_resource>s` for the search result? We know both singular and plural names too
* Should the returned data structure be always an array?

tests/test_playbooks/host_info.yml
```diff
@@ -0,0 +1,101 @@
...
+        validate_certs: "{{ foreman_validate_certs }}"
+        name: "test-host.{{ host.domain }}"
+        organization: "{{ host.organization }}"
+        location: "{{ host.location }}"
+      register: host_info
+    - name: check host details
+      assert:
+        that:
+          - host_info['resources']['name'] == "test-host.{{ host.domain }}"
+          - host_info['resources']['domain_name'] == host.domain
```
evgeni: this is the part I mostly wonder about

tests/test_playbooks/host_info.yml
```diff
@@ -0,0 +1,101 @@
...
+        validate_certs: "{{ foreman_validate_certs }}"
+        search: "name = test-host.{{ host.domain }}"
+        organization: "{{ host.organization }}"
+        location: "{{ host.location }}"
+      register: host_info
+    - name: check host details
+      assert:
+        that:
+          - host_info['resources'][0]['name'] == "test-host.{{ host.domain }}"
+          - host_info['resources'][0]['domain_name'] == host.domain
```
evgeni: vs this

mdellweg:
I think i prefer to be specific and use `<whatever_resource>(s)` accordingly. So we can expect
```
result => {
"state": "OK",
"host": {
  "name": "whoever.whereev.er",
  ...
}
}
```
Am i getting this right?

evgeni:
for `name: host.example.com` you'd get:

```
ok: [tests] => {
    "changed": false,
    "host": {
        "id": 12,
        "name": "host.example.com",
        …
     }
}
```

and for for `search: name = host.example.com` you'd get:

```
ok: [tests] => {
    "changed": false,
    "hosts": [
        {
          "id": 12,
          "name": "host.example.com",
          …
        }
     ]
}
```

evgeni:
Updated to reflect the latest discussion results.

evgeni:
~~Should add docs in `RETURNS`…~~ done

plugins/module_utils/foreman_helper.py
```diff
@@ -1287,6 +1287,40 @@ def entity_name_from_class(self):
...
+
+    def run(self, **kwargs):
+        """
+        lookup entities
+        """
+        self.auto_lookup_entities()
+
+        resource = self.foreman_spec['entity']['resource_type']
+
+        if 'name' in self.foreman_params:
```
ehelms, line 1313: Checking on my pythonic learning, should this be a try-rescue block for a KeyError ?

ehelms (approved):
Overall looks good to me, I just had the one question to help with my python learning.

plugins/module_utils/foreman_helper.py
mdellweg: This looks like politely asking if a KeyError would be expected.

evgeni: a try/except block? no-ish.
the code doesn't access `foreman_params['name']`, so there is nothing that would raise the `KeyError`.
(well, `lookup_entity` might at some point, but that's weird)

plugins/doc_fragments/foreman.py
```diff
@@ -325,3 +325,35 @@ class ModuleDocFragment(object):
...
+
+    INFOMODULE = '''
+options:
+  name:
+    description:
+      - Name of the resource to fetch information for.
+      - Mutually exclusive with I(search).
+    required: false
+    type: str
+  location:
```
mdellweg, line 337: Do all resourced have taxonomies?
Well probably all that we have infomodules for at this point. Which is good enough for now.

plugins/doc_fragments/foreman.py
```diff
@@ -325,3 +325,35 @@ class ModuleDocFragment(object):
...
+      - Search query to use
+      - If None, and I(name) is not set, all resources are returned.
+      - Mutually exclusive with I(name).
+    type: str
+'''
+
+    KATELLOINFOMODULE = '''
+options:
+  organization:
+    required: true
```
mdellweg, line 358: :+1:

mdellweg (approved):

plugins/doc_fragments/foreman.py
evgeni: Yeah, `setting` is not taxonomied, but I've left that to future-me to fix :)

PR #1113, opened 2020-12-22
Add lifecycle environments role

### LINKED ISSUE #1110: lifecycle environments role
##### SUMMARY
role for managing lifecycle environments with variables

##### ISSUE TYPE
 - Feature Idea

ehelms:


evgeni:
* `prior` is an optional field in the module (and then Katello defaults to "Library"), should the role also allow omitting it?
* the module has `label` as a field that can be set, do we want to expose that?

evgeni:
It looks like you added ` tests/fixtures/apidoc/lifecycle_environments_role.json` as a file, not as a symlink to `katello.json`?

ehelms:
> * `prior` is an optional field in the module (and then Katello defaults to "Library"), should the role also allow omitting it?

That is tricky. That makes sense as a default and it could also lead users to defining their environment heirarchy incorrectly. For example, they might assume giving a list in order will create them in order and they all end up attached to Library. Right now I am leaning towards being explicit rather than implicit.

evgeni (approved):

roles/lifecycle_environments/README.md
```diff
@@ -0,0 +1,63 @@
...
+This role supports the [Common Role Variables](https://github.com/theforeman/foreman-ansible-modules/blob/develop/README.md#common-role-variables).
+
+The main data structure for this role is the list of `lifecycle_environments`. Each `lifecycle_environment` requires the following fields:
+
+- `name`: The name of the lifecycle environment.
+- `prior`: The name of the previous lifecycle environment to attach to in sequence
+
+The following fields are optional and will be omitted by default:
+
+- `description`: Description of the lifecycle environment
```
evgeni, line 18: we could add `label` here, or ignore it, the ACK stands :)

roles/lifecycle_environments/README.md
```diff
@@ -0,0 +1,64 @@
...
+
+Role Variables
+--------------
+
+This role supports the [Common Role Variables](https://github.com/theforeman/foreman-ansible-modules/blob/develop/README.md#common-role-variables).
+
+The main data structure for this role is the list of `lifecycle_environments`. Each `lifecycle_environment` requires the following fields:
+
+- `name`: The name of the lifecycle environment.
+- `prior`: The name of the previous lifecycle environment to attach to in sequence
```
wbclark:
```suggestion
- `prior`: The name of the previous lifecycle environment to attach to in sequence. For the first lifecycle environment in a new path, set the prior lifecycle environment to Library.
```

This will likely be helpful to somebody.

roles/lifecycle_environments/tasks/main.yml
```diff
@@ -0,0 +1,15 @@
+---
+- name: 'Create Lifecycle Environments'
+  theforeman.foreman.lifecycle_environment:
+    username: "{{ username }}"
```
willtome: It would be good to namespace these like `foreman_username` to avoid conflict with other variables. Also, should default to omit so that env vars could be used.

evgeni: We decided not to prefix the common ones (user, pass, url, org), but omit is a good point!

willtome: I think that leaves a lot of room for variable conflicts. Ansible has a flat namespace so without a prefix that leaves very generic variable names tied specifically to this role. Here is an example of how I intend to use these roles for reference. https://github.com/redhat-cop/agnosticd/blob/development/ansible/configs/smart-management/sample_vars_patching.yml

ehelms: I filed https://github.com/theforeman/foreman-ansible-modules/issues/1117 since the lack of omit is present across various roles.

IPvSean: I am assuming we are talking about the variable `username` versus the parameter `username` here, which I agree with willtome.  If you are already using username (e.g. for Ansible Tower, or another box) there can be massive conflicts.  The extra_vars will always win, so if someone is using Ansible Tower survey tool and happened to choose survey they could mess up this role and not understand why.

willtome: Opened #1118 to continue this discussion

mdellweg:
> > * `prior` is an optional field in the module (and then Katello defaults to "Library"), should the role also allow omitting it?
> 
> That is tricky. That makes sense as a default and it could also lead users to defining their environment heirarchy incorrectly. For example, they might assume giving a list in order will create them in order and they all end up attached to Library. Right now I am leaning towards being explicit rather than implicit.

How about making this a role about livecycle-environment-paths instead of single lce?
You would provide a list of list's of lce-names and the role would care about "prior" internally.

```yaml
lce-paths:
  - ["dev", "test", "prod"]
  - ["apple", "pear", "orange"]
```

ehelms:
> > > * `prior` is an optional field in the module (and then Katello defaults to "Library"), should the role also allow omitting it?
> > 
> > 
> > That is tricky. That makes sense as a default and it could also lead users to defining their environment heirarchy incorrectly. For example, they might assume giving a list in order will create them in order and they all end up attached to Library. Right now I am leaning towards being explicit rather than implicit.
> 
> How about making this a role about livecycle-environment-paths instead of single lce?
> You would provide a list of list's of lce-names and the role would care about "prior" internally.
> 
> ```yaml
> lce-paths:
>   - ["dev", "test", "prod"]
>   - ["apple", "pear", "orange"]
> ```

I like this idea. How would you see supporting label and description parameters? Should I modify it to:

```yaml
lifecycle_environment_paths:
  - - name: "Dev"
      description: Environment for development machines
    - name: "Test"
    - name: "Prod" 
      description: Production environment for RHEL hosts
  - - name: Apple
    - name: Pear
    - name: Orange
```

evgeni:
Lists of lists of dicts are horrible to read, so hard no.

evgeni:
I think the current "API" is perfectly fine.

@ehelms can you add `default(omit)`?

mdellweg:
Good question. That would work for me, although i kind of dislike the double dash syntax for nested lists. I don't know of a better way.

ehelms:
Based on y'alls responses, I find myself asking:

 1) Is it better to have this role slightly more opinionated and focused on creating lifecycle environment paths for users (and they can use the modules directly if they need to do more advanced things like descriptions) ? e.g. ['Dev', 'Test', 'Prod']
 2) Or to have it allow full control with a harder to visually parse syntax that it has today?

I think as a 90% use case user, I'd want to just define the paths. And then be really annoyed the moment I want to add a description.

evgeni:
3. leave it as it is? ;)

I must say that the origina pathl example *is* readable

```yaml
lce-paths:
  - ["dev", "test", "prod"]
  - ["apple", "pear", "orange"]
```

until someone says "bah, inline lists" (or worse, a yaml linter does that) and goes

```yaml
lce-paths:
  - - "dev"
    - "test"
    - "prod"
  - - "apple"
    - "pear"
    - "orange"
```

which is horrible again. so I think I'd personally would just stick with the original 

```yaml
          - name: "Dev"
            prior: "Library"
          - name: "Test"
            prior: "Dev"
          - name: "Prod"
            prior: "Test"
```

ehelms:
@willtome since you've been crafting variable files in your tower usage, do you have an opinion to help sway us?

willtome:
Here is an example of how I’ve done it with another role. I’m partial to this format.

https://github.com/redhat-cop/agnosticd/blob/c417a78ab87e7811089286615b3fff5d5fe9c730/ansible/configs/smart-management/sample_vars_patching.yml#L408

ehelms:
Based on the feedback, I'd like to merge it as is then. @mdellweg ok with that?

evgeni:
I am OK with that -- but can you add the `| default(omit)` to the server_url etc params?

evgeni:
Was the readme update to the "paths" style intentional? This doesn't match the role code anymore :(

ehelms:
Noooo, my reset-branch script failed me from when I was playing around with the suggestion. Let me put it back.

roles/lifecycle_environments/README.md
```diff
@@ -0,0 +1,68 @@
...
+```yaml
+- hosts: localhost
+  roles:
+    - role: theforeman.foreman.lifecycle_environments
+      vars:
+        server_url: https://foreman.example.com
+        username: "admin"
+        password: "changeme"
+        organization: "Default Organization"
+        lifecycle_environment_paths:
```
evgeni:
```suggestion
        lifecycle_environments:
```

roles/lifecycle_environments/README.md
```diff
@@ -0,0 +1,68 @@
...
```
evgeni:
```suggestion
        lifecycle_environment:
```

roles/lifecycle_environments/README.md
```diff
@@ -0,0 +1,68 @@
...
+          - name: "Dev"
+            prior: "Library"
+          - name: "Test"
+            prior: "Dev"
+          - name: "Prod"
+            prior: "Dev"
+```
+
+Create two lifecycle environment paths:
```
evgeni: Should we call out the paths here in the description, so the users know them before looking at the YAML and building paths in their head?

roles/lifecycle_environments/README.md
```diff
@@ -0,0 +1,68 @@
...
+
+The main data structure for this role is the list of `lifecycle_environments`. Each `lifecycle_environment` requires the following fields:
+
+- `name`: The name of the lifecycle environment.
+- `prior`: The name of the previous lifecycle environment to attach to in sequence. For the first lifecycle environment in a new path, set the prior lifecycle environment to Library.
+
+The following fields are optional and will be omitted by default:
+
+- `description`: Description of the lifecycle environment
+- `label`: A permanent label for identifying the lifecycle environment to tools such as subscription-manager. This is created by the server if omitted.
```
evgeni:
```suggestion
- `label`: A permanent label for identifying the lifecycle environment to tools such as subscription-manager. This is created by the server if omitted. It can't be changed after the lifecycle environment has been created.
```

evgeni (changes requested):

evgeni (approved):

roles/lifecycle_environments/README.md
```diff
@@ -0,0 +1,68 @@
...
+
+Role Variables
+--------------
+
+This role supports the [Common Role Variables](https://github.com/theforeman/foreman-ansible-modules/blob/develop/README.md#common-role-variables).
+
+The main data structure for this role is the list of `lifecycle_environments`. Each `lifecycle_environment` requires the following fields:
+
+- `name`: The name of the lifecycle environment.
+- `prior`: The name of the previous lifecycle environment to attach to in sequence. For the first lifecycle environment in a new path, set the prior lifecycle environment to Library.
```
mdellweg: Should we add a note that the order of creation matters?

mdellweg (approved):
Maybe, the lcep role is not completely off the table?

My other thought: Can this role handle the case when you add/remove an lce to an existing path?

Approving anyway, as this surely allows for proper initial setup.

roles/lifecycle_environments/README.md
ehelms: Yea, that is a good point otherwise it will blow up.

ehelms:
I think this can handle add to an existing path... assuming the ordering in the list of data provided is right. Remove is a tricky one with lifecycle environments due to the amount of data and relationships they can contain. But I see your point on the ordering. I suppose it will throw an error and user will need to change their ordering. Not ideal, not terrible.

PR #1120, opened 2020-12-22
Omit username, password, and server_url in roles by default

### LINKED ISSUE #1117: Set common role variables in existing roles to `default(omit)` to allow use of environment variables
##### SUMMARY
The initial set of roles added often did not include omitting things like username by default to allow the use of environment variables. This issue is to systematically fix that among the existing roles.

##### ISSUE TYPE
 - Bug Report

wbclark:
Fixes #1117

### DIFF
--- a/roles/activation_keys/tasks/main.yml
+++ b/roles/activation_keys/tasks/main.yml
@@ -1,9 +1,9 @@
 ---
 - name: 'Create Activation Keys'
   theforeman.foreman.activation_key:
-    username: "{{ username }}"
-    password: "{{ password }}"
-    server_url: "{{ server_url }}"
+    username: "{{ username | default(omit) }}"
+    password: "{{ password | default(omit) }}"
+    server_url: "{{ server_url | default(omit) }}"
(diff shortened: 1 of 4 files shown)


mdellweg (approved):
Looks exactly like what was described in the issue.

mdellweg:
Do you want to add a little bit of text about this environment injection in the "common role variables" section of the readme?

wbclark:
> Do you want to add a little bit of text about this environment injection in the "common role variables" section of the readme?

Thanks for the suggestion. Though I'm not very experienced with AWX/Tower and my search engine skills are failing me here, so I'm not sure what exactly to write.

Can you recommend some resource about it?

mdellweg:
Oh, I wasn't even going for tower here, just state that if certain variables are not provided, there is a fallback to lookup environment vars. A bit like we do in the module docs.
https://github.com/theforeman/foreman-ansible-modules/blob/develop/plugins/doc_fragments/foreman.py#L31

evgeni:
created an own issue for the docs, and merging now.

Issue #838, opened 2020-06-17
katello_sync example leads to bad performance


evgeni:
##### SUMMARY
the example in `katello_sync` tells the user to search for all available repos and then trigger a sync on all of them in parallel. this can lead to a very high load on the system, especially on memory, sometimes killing MongoDB and friends, thus actually breaking the system.

I think we should not give such examples.

@sean797 you added that example in #211 -- care to weight in? :)

##### ISSUE TYPE
 - Bug Report

##### ANSIBLE VERSION

```

```

##### KATELLO/FOREMAN VERSION

```

```

##### APYPIE VERSION

```

```

##### STEPS TO REPRODUCE

```yaml

```

##### ACTUAL RESULTS

```

```

sean797:
Seems like a fair use case, shame we do not handle it well.

Removing it as a short term solution makes sense to me

PR #1127, opened 2021-01-11
Add repo sync option

### LINKED ISSUE #1119: Repository Sync Role
##### SUMMARY
Role to conduct a repository sync. Currently I see roles for enabling repositories and creating a sync plan but this issue is for functionality to do an on-demand sync. Example implementation here (https://github.com/redhat-cop/agnosticd/blob/development/ansible/roles/satellite-manage-repositories/tasks/version_6.x.yml)

I don't know if this would be best to add to existing role or create new one.

##### ISSUE TYPE
 - Feature Idea

willtome:
This PR adds the option `sync` to the repositories data structure to allow for an on-demand sync of a product when it is enabled. I believe this should be added to the current repositories role because it uses the same variable. The default behavior is not to sync. If `sync: true` the repos for that product will be synced. This is achieved using async operations so that multiple repos can be synced simultaneously. 

closes #1119


ehelms:
There was a discussion, at length, in the original role PR (https://github.com/theforeman/foreman-ansible-modules/pull/1027) around this. In that PR, we avoided this as we landed on introducing a role for syncing rather than baking it into the repositories role. That would avoid duplicated logic (as a role for syncing is certainly useful) and would allow users to compose functionality to their desire. And the repositories role could potentially import the role if we went that route DRYing things up.

willtome:
@ehelms updated PR to introduce a new role rather than an extension to the existing role.

evgeni:
AI: @evgeni wants to link the old issue where the `async` style broke things

evgeni:
Hey @willtome,

we had the `async: 9999, poll: 0` code as an example in the `repository_sync` module before, and have removed it (#838, #845) because we had reports that doing this with more than a few repositories would overload the system.

Is there a possibility to batch those syncs better? Like "only do X" at a time?

roles/repositories_sync/tasks/main.yml
```diff
@@ -0,0 +1,29 @@
+---
+- name: Sync repository
+  theforeman.foreman.repository_sync:
+    username: "{{ username | default(omit) }}"
+    password: "{{ password | default(omit) }}"
+    server_url: "{{ server_url | default(omit) }}"
+    validate_certs: "{{ validate_certs | default(omit) }}"
+    organization: "{{ organization }}"
+    product: "{{ item.0.name }}"
+    repository: "{{ item.1.name }}"
```
evgeni: If you omit `repository` the module will sync the whole product, which in most cases is what the user wants to do anyways.

roles/repositories_sync/tasks/main.yml
```diff
@@ -0,0 +1,29 @@
...
+  with_subelements:
+    - "{{ products | selectattr('repository_sets', 'defined') | list }}"
```
evgeni, line 11: this means it won't sync any custom repositories that the `repositories` role can create.

roles/repositories_sync/tasks/main.yml
mdellweg: If we go for the batching, we might want to iterate over the repositories to even out the load better.
If we do one product at a time, this might just be the easiest approach to batching.

Just my two thoughts here...

evgeni: Yeah, agree, "by product" is probably the cheapest batching we can get

willtome: OK. Will update to just sync by product and not async. Can an issue be opened to better queue requests in foreman? At the end of the day, it really shouldn't be the automation's responsibility to batch requests and manage load for the system.

roles/repositories_sync/tasks/main.yml
```diff
@@ -0,0 +1,13 @@
...
+    username: "{{ username | default(omit) }}"
+    password: "{{ password | default(omit) }}"
+    server_url: "{{ server_url | default(omit) }}"
+    validate_certs: "{{ validate_certs | default(omit) }}"
+    organization: "{{ organization }}"
+    product: "{{ item.0.name }}"
+  with_subelements:
+    - "{{ products | selectattr('repository_sets', 'defined') | list }}"
+    - repository_sets
+  when: item.0.sync | default(false) | bool
```
evgeni, line 13:
```suggestion
    product: "{{ item.name }}"
  loop: "{{ products }}"
  when: item.sync | default(false) | bool
```

There is no need to iterate over the repositories anymore, as we're syncing by-product now.
Also loop over all products, regardless of `repository_sets` (RH repos) or `repositories` (custom repos) attributes.

roles/repositories/tasks/main.yml
```diff
@@ -81,4 +81,4 @@
     verify_ssl_on_sync: "{{ item.1.verify_ssl_on_sync | default(omit) }}"
   with_subelements:
     - "{{ products | selectattr('repositories', 'defined') | list }}"
-    - repositories
+    - repositories
```
evgeni, line 84: this looks unrelated and unneeded?

evgeni (changes requested):
this needs tests -- do you know how to record those, or do you need help with them?

wbclark (commented):
Hi @willtome , thanks for the PR!

could you add some tests to this role? The documentation at https://theforeman.github.io/foreman-ansible-modules/develop/testing.html is really helpful for this

In short, what you need is a test playbook `tests/test_playbooks/repositories_sync_role.yml` and an apidoc fixture `tests/fixtures/apidoc/repositories_sync_role.json` which should just be a symlink to `katello.json` in the same dir

To record the test fixtures, there is a little first time setup: `make test-setup` and then edit `tests/test_playbooks/vars/server.yml` to point to your katello server. Once that setup is complete `make record_repositories_sync_role` would then record the fixtures.

You can look at the PRs for other roles to get an idea how to do this, or please feel free to reach out any time if you have any questions

PR #1128, opened 2021-01-13
Add some hints on how to find the pool IDs

ericzolf:
It's not obvious which ID is coming from where, hence extend the docs to explain it.

### DIFF
--- a/plugins/modules/activation_key.py
+++ b/plugins/modules/activation_key.py
@@ -62,8 +62,8 @@ options:
         description:
           - Pool ID of the Subscription to be added.
           - Mutually exclusive with I(name) and I(upstream_pool_id).
-          - Also named I(Candlepin Id) in the CSV export of the subscriptions,
-          - it is as well the UUID as output by I(hammer subscription list).
+          - Also named C(Candlepin Id) in the CSV export of the subscriptions,
+          - it is as well the C(UUID) as output by C(hammer subscription list).
         type: str
         required: false
       upstream_pool_id:
--- a/plugins/modules/activation_key.py
+++ b/plugins/modules/activation_key.py
@@ -62,12 +62,15 @@ options:
         description:
           - Pool ID of the Subscription to be added.
           - Mutually exclusive with I(name) and I(upstream_pool_id).
+          - Also named I(Candlepin Id) in the CSV export of the subscriptions,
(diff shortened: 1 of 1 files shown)


plugins/modules/activation_key.py
```diff
@@ -62,12 +62,15 @@
...
           - Mutually exclusive with I(name) and I(upstream_pool_id).
+          - Also named I(Candlepin Id) in the CSV export of the subscriptions,
+          - it is as well the UUID as output by I(hammer subscription list).
         type: str
         required: false
       upstream_pool_id:
         description:
           - Upstream Pool ID of the Subscription to be added.
           - Mutually exclusive with I(name) and I(pool_id).
+          - Also named I(Master Pools) at access.redhat.com.
```
```suggestion
          - Also named C(Master Pools) in the Red Hat Portal.
```

`I()` is for option names, so I'd avoid that here. See https://docs.ansible.com/ansible/latest/dev_guide/developing_modules_documenting.html#linking-and-other-format-macros-within-module-documentation

And I think we should call the portal by name, not url (and if url, then `U(https://access.redhat.com)`)

plugins/modules/activation_key.py
```diff
@@ -62,12 +62,15 @@
         description:
           - Pool ID of the Subscription to be added.
           - Mutually exclusive with I(name) and I(upstream_pool_id).
+          - Also named I(Candlepin Id) in the CSV export of the subscriptions,
+          - it is as well the UUID as output by I(hammer subscription list).
```
```suggestion
          - Also named C(Candlepin Id) in the CSV export of the subscriptions,
          - it is as well the C(UUID) as output by C(hammer subscription list).
```

Is the "CSV" export still a thing in modern Satellite/Katellos?

evgeni (changes requested):

plugins/modules/activation_key.py
ericzolf: And, yes, in Satellite 6.8 / Katello 3.16, CSV export is still a thing, even if I'm not sure it's modern :-)

image

ericzolf:
All changes accepted, thanks @evgeni !

plugins/modules/activation_key.py
evgeni: Oh, cool. I thought about `hammer csv` which is gone. :+1:

evgeni (approved):

PR #1112, opened 2020-12-21
Organization Role

hindman-redhat:
Draft pull request

Issue #1118, opened 2020-12-22
Role Variable Prefix


willtome:
##### SUMMARY
This issue is to open the discussion around variable naming convention for roles. Currently, "generic" variable names are used (eg. `server_url`, `username`, `manifest`, etc). This could lead to conflicting variables during runtime between other roles, facts, or extra_vars. I propose roles use variables with the prefix `foreman_` to "namespace" vars used by these roles.

Current state example:
```
server_url: https://satellite.example.com
username: admin
password: password
organization: "Default Organization"
manifest_path: /tmp/manifest.zip
activation_keys:
          - name: "Basic Activation Key"
```

Future State Example:
```
foreman_server_url: https://satellite.example.com
foreman_username: admin
foreman_password: password
foreman_organization: "Default Organization"

foreman_manifest_path: /tmp/manifest.zip

foreman_activation_keys:
          - name: "Basic Activation Key"
```

##### ISSUE TYPE
 - Feature Idea

evgeni:
Hey @willtome!

Thanks for raising this. I must admit I never liked how Ansible does (not) allow easy namespacing of variables by default. But oh well, here we are.

May I ask for a bit more background of your environment? Do you have one single `vars.yml` that is loaded by all playbooks? Or does Tower inject them somehow "too globaly"?

The reason I am asking is: when we initially designed this, we did talk about prefixes to namespace between *our* roles, but never about completely others. And not having a `foreman_` prefix seemed natural, as this reduces confusion when used  in Satelilte environments.

The trail of thought was: well, the user will be having one (or multiple) playbook(s) that are managing their Foreman/Satellite, and the common things they will need is the "url", "username" etc, so those should be named the same for all roles (granted, this would also be the case when we prefix them with `foreman_`).

But we might need to bite the bullet and add that prefix, confusing Satellite users (or making playbooks not work across both systems, which also sucks)

willtome:
@evgeni many of the use cases I currently have are for provisioning. This leads to single or multiple variable files that define all of the configurations for a deployment and seek to reduce the number of bespoke playbooks required by using these and other roles. It is hard to guarantee that a set of variables will only exist in a play being run against a foreman. For example, the same variables used to configure foreman may also be used to register hosts correctly and enable the proper repositories.

[...]

PR #1131, opened 2021-01-18
Organization Role 2

willtome:
This PR a copy of PR #1112 that meets current role naming conventions.


evgeni:
Given the outcome of #1118, so I am *inclined* to close this in favor of #1112, but your last commit actually has also changes to the README that I'd consider worth to have.

Thus, @willtome and @hindman-redhat can you two play rock-paper-scissors who will pick up the work here? :)
If neither of you has time, we can pick up the work from here -- just say a word.

willtome:
I'm swamped right now @evgeni. If you don't mind combining the 2 and closing these out that would be awesome!

evgeni:
taken over in #1204

PR #1132, opened 2021-01-25
scope content_view and kickstart tighter if data was provided

### LINKED ISSUE #1090: kickstart_repository (field "synced content") is not correctly set when creating a host via the ansibel module redhat.satellite.host
##### SUMMARY
Using the ansible module redhat.satellite.host to create a host on our Satellite server it seems the selected kickstart_repository is not correctly set in field "synced content" (although it is correctly set in the playbook). The field on the satellite host remains gray.
When trying to kickstart the host anyway the URL refers to a library URL instead to a lifecycle environment URL.
When doing this through the Satellite GUI, this it correctly set upon selecting the operating system (dynamically updated). 

##### ISSUE TYPE
 - Bug Report

##### ANSIBLE VERSION
ansible 2.8.16

##### KATELLO/FOREMAN VERSION
tfm-rubygem-katello-3.14.0.32-1.el7sat.noarch
foreman-1.24.1.32-1.el7sat.noarch

##### STEPS TO REPRODUCE
run playbook

```yam
    - name: "Create host {{ host.name }}"
      redhat.satellite.host:
        name: "{{ host.name }}.{{ domain_name }}"
        organization: "{{ organization }}"
        location: "{{ datacenter }}"
        hostgroup: "hg-{{ organization }}-{{ datacenter }}"
        lifecycle_environment: "{{ host.lifecycle_environment }}"
        content_view: "{{ host.content_view }}"
        architecture: "{{ host.architecture }}"
        operatingsystem: "{{ host.operatingsystem }}"
        kickstart_repository: "{{ host.kickstart_repository }}"
        ptable: "Kickstart default {{ organization }}"
        pxe_loader: "{{ host.pxe_loader }}"
        root_pass: "{{ host.root_pass }}"
        mac: "{{ network.mac }}"
        subnet: "{{ network.name }}/{{ network.bits }}"
        ip: "{{ network.ip }}"
        build: true
        comment : "{{ host.comment }}"
        state: present
        server_url: "{{ server_url }}"
        username: "{{ username }}"
        password: "{{ password }}"
```

##### EXPECTED RESULTS
Host is created but the field "Synced Content" remains empty where it should mention the kickstart repo.
When starting the server the URL to the kickstart image is incorrect.

The used URL refers to a Library path instead of a lifecycle path (should be test in our case)
http://capsule_server/pulp/repos/organisation/Library/content/dist/rhel/server/7/7.6/x86_64/kickstart/images/pxeboot/vmlinuz... No such file or directory

evgeni:
when the user provides a lifecycle environment (and content view),
we can look up the content view and kickstart repository with more
specific queries, thus actually finding what the user was looking for.

### DIFF
--- a/plugins/module_utils/foreman_helper.py
+++ b/plugins/module_utils/foreman_helper.py
@@ -297,8 +297,9 @@ class HostMixin(ParametersMixin):
             openscap_proxy=dict(type='entity', resource_type='smart_proxies'),
             content_source=dict(type='entity', scope=['organization'], resource_type='smart_proxies'),
             lifecycle_environment=dict(type='entity', scope=['organization']),
-            kickstart_repository=dict(type='entity', scope=['organization'], resource_type='repositories'),
-            content_view=dict(type='entity', scope=['organization']),
+            kickstart_repository=dict(type='entity', scope=['organization'], optional_scope=['lifecycle_environment', 'content_view'],
+                                      resource_type='repositories'),
+            content_view=dict(type='entity', scope=['organization'], optional_scope=['lifecycle_environment']),
             activation_keys=dict(),
         )
         foreman_spec.update(kwargs.pop('foreman_spec', {}))
@@ -786,8 +787,13 @@ class ForemanAnsibleModule(AnsibleModule):
     def find_puppetclasses(self, names, **kwargs):
         return [self.find_puppetclass(name, **kwargs) for name in names]
 
-    def scope_for(self, key):
-        return {'{0}_id'.format(key): self.lookup_entity(key)['id']}
+    def scope_for(self, key, scoped_resource=None):
+        # workaround for https://projects.theforeman.org/issues/31714
+        if scoped_resource in ['content_views', 'repositories'] and key == 'lifecycle_environment':
+            scope_key = 'environment'
+        else:
+            scope_key = key
+        return {'{0}_id'.format(scope_key): self.lookup_entity(key)['id']}
 
     def set_entity(self, key, entity):
         self.foreman_params[key] = entity
@@ -814,9 +820,12 @@ class ForemanAnsibleModule(AnsibleModule):
         else:
             params = {}
         try:
-            if 'scope' in entity_spec:
-                for scope in entity_spec['scope']:
-                    params.update(self.scope_for(scope))
+            for scope in entity_spec.get('scope', []):
+                params.update(self.scope_for(scope, resource_type))
+            for optional_scope in entity_spec.get('optional_scope', []):
+                if optional_scope in self.foreman_params:
+                    params.update(self.scope_for(optional_scope, resource_type))
+
         except TypeError:
             if failsafe:
                 if entity_spec.get('type') == 'entity':
@@ -1371,10 +1380,11 @@ class ForemanEntityAnsibleModule(ForemanStatelessEntityAnsibleModule):
                 self.foreman_params[self.entity_key] = self.foreman_params.pop(updated_key)
 
         params = kwargs.get('params', {})
-        entity_scope = self.foreman_spec['entity'].get('scope')
-        if entity_scope:
-            for scope in entity_scope:
-                params.update(self.scope_for(scope))
+        for scope in self.foreman_spec['entity'].get('scope', []):
+            params.update(self.scope_for(scope))
+        for optional_scope in self.foreman_spec['entity'].get('optional_scope', []):
+            if optional_scope in self.foreman_params:
+                params.update(self.scope_for(optional_scope))
         new_entity = self.ensure_entity(self.foreman_spec['entity']['resource_type'], self.foreman_params, entity, params=params)
         new_entity = self.remove_sensitive_fields(new_entity)
 
@@ -1492,6 +1502,7 @@ def _foreman_spec_helper(spec):
         'flat_name',
         'foreman_spec',
         'invisible',
+        'optional_scope',
         'resolve',
         'resource_type',
         'scope',
@@ -1506,6 +1517,7 @@ def _foreman_spec_helper(spec):
     }
     _ENTITY_SPEC_KEYS = {
         'failsafe',
+        'optional_scope',
         'resolve',
         'resource_type',
         'scope',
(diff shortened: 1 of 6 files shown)


evgeni:
`katello_hostgroup` tests are failing, because I changed how it looks up things :)

plugins/module_utils/foreman_helper.py
```diff
@@ -816,7 +827,11 @@ def _lookup_entity(self, identifier, entity_spec, params=None):
         try:
             if 'scope' in entity_spec:
                 for scope in entity_spec['scope']:
-                    params.update(self.scope_for(scope))
+                    if resource_type in ['content_views', 'repositories'] and scope == 'lifecycle_environment':
```
evgeni: this is horrible, but Katello keeps on calling LCEs "environments" in so many places…

plugins/module_utils/foreman_helper.py
```diff
@@ -315,6 +315,15 @@ def __init__(self, **kwargs):
             ak_param = {'name': 'kt_activation_keys', 'parameter_type': 'string', 'value': self.foreman_params.pop('activation_keys')}
             self.foreman_params['parameters'].append(ak_param)
 
+        if 'content_view' in self.foreman_params:
+            if 'lifecycle_environment' in self.foreman_params:
+                self.foreman_spec['content_view']['scope'].append('lifecycle_environment')
+
+        if 'kickstart_repository' in self.foreman_params:
+            for optional_scope in ['lifecycle_environment', 'content_view']:
```
evgeni: should this become an entry in the `foreman_spec`, so the spec becomes:

```python
kickstart_repository=dict(type='entity', scope=['organization'], optional_scope=['lifecycle_environment', 'content_view'], resource_type='repositories')
```

plugins/module_utils/foreman_helper.py
evgeni: https://projects.theforeman.org/issues/31714

plugins/module_utils/foreman_helper.py
```diff
@@ -786,8 +795,10 @@ def find_puppetclass(self, name, environment=None, params=None, failsafe=False,
     def find_puppetclasses(self, names, **kwargs):
         return [self.find_puppetclass(name, **kwargs) for name in names]
 
-    def scope_for(self, key):
-        return {'{0}_id'.format(key): self.lookup_entity(key)['id']}
+    def scope_for(self, key, scope_key=None):
+        if scope_key is None:
+            scope_key = key
+        return {'{0}_id'.format(scope_key): self.lookup_entity(key)['id']}
```
```suggestion
        return {'{0}_id'.format(scope_key or key): self.lookup_entity(key)['id']}
```

plugins/module_utils/foreman_helper.py
```diff
@@ -816,7 +827,11 @@ def _lookup_entity(self, identifier, entity_spec, params=None):
         try:
             if 'scope' in entity_spec:
                 for scope in entity_spec['scope']:
-                    params.update(self.scope_for(scope))
+                    if resource_type in ['content_views', 'repositories'] and scope == 'lifecycle_environment':
+                        scope_key = 'environment'
+                    else:
+                        scope_key = None
+                    params.update(self.scope_for(scope, scope_key))
```
mdellweg: I like the `optional_scope` idea, but i would traverse it as a second loop here instead of fiddling with the `foreman_spec` depending on user input.

plugins/module_utils/foreman_helper.py
```diff
@@ -816,7 +816,12 @@ def _lookup_entity(self, identifier, entity_spec, params=None):
         try:
             if 'scope' in entity_spec:
                 for scope in entity_spec['scope']:
-                    params.update(self.scope_for(scope))
+                    params.update(self.scope_for(scope, _scope_key(resource_type, scope)))
+            if 'optional_scope' in entity_spec:
```
evgeni: this now works for everything but the main `entity`, should that also get optional scopes? I think so!

mdellweg: Good call.

plugins/module_utils/foreman_helper.py
```diff
@@ -816,7 +817,12 @@ def _lookup_entity(self, identifier, entity_spec, params=None):
...
+                for optional_scope in entity_spec['optional_scope']:
```
mdellweg: With `entity_spec.get('optional_scope', [])` you didn't need the surrounding if.

plugins/module_utils/foreman_helper.py
```diff
@@ -1588,6 +1596,14 @@ def _is_resolved(spec, what):
     return False
 
 
+def _scope_key(resource_type, scope):
+    # workaround for https://projects.theforeman.org/issues/31714
+    key = None
+    if resource_type in ['content_views', 'repositories'] and scope == 'lifecycle_environment':
+        key = 'environment'
+    return key
```
mdellweg: Is there a way to embed this in `scope` i guess you want it always to be used, right?
I know the logic around this workaround is just :exploding_head: .

plugins/module_utils/foreman_helper.py
evgeni: I was copying the code above, refactored both.

plugins/module_utils/foreman_helper.py
evgeni: you mean directly in the `scope_for` func?

evgeni: I guess, if we (optionally) pass it the `res_type` so that it can do the same :exploding_head: logic.

mdellweg: Yes, that's where i see it, anyway. :+1:

But having the `res_type` only optionally passed can still produce incorrect results if one omits it. :disappointed:

mdellweg: So it seems the scope is not only dependent on what is scoped, but also for what it is scoped.
Maybe we should add the `resource_type` as a mandatory parameter to `scope_for`.

evgeni: Well, it's only really special for LCEs as they don't use their right name in params. I've not seen something like this with any other resource.

evgeni: I added it as an optional param for now. but made the automatic code always pass it.

so now only manual calls have it optional?

mdellweg: Yes, it's really a strange edge case. The code looks good for me now.

evgeni:
as :green_apple:  as it gets.

mdellweg (approved):
:green_apple:

evgeni:
(rebase to nicer history)

PR #1133, opened 2021-01-28
fix rstcheck requirement

evgeni:
it's gone in ansible-develop, but we still use it


ehelms (approved):

mdellweg (approved):
That looks even better.

PR #1135, opened 2021-02-02
Add a role to perform basic setup to register and patch RHEL clients

wbclark:
This PR is a WIP. It adds a minimal and somewhat opinionated role that performs all of the basic steps required to be ready to register and patch RHEL7 clients.

Comments and suggestions are welcome


wbclark:
Tests should be working now.

Thanks to @evgeni for the suggestion to use a real manifest and skip recording it in fixtures by modifying the vcr python wrapper.

wbclark:
ansible lint is now happy

tests/vcr_python_wrapper.py
```diff
@@ -159,7 +159,7 @@ def filter_request_uri(request):
     if test_params['test_name'] == 'host':
         fam_vcr.register_matcher('host_body', host_body_matcher)
         body_matcher = 'host_body'
-    elif test_params['test_name'] in ['subscription_manifest', 'manifest_role']:
+    elif test_params['test_name'] in ['subscription_manifest', 'manifest_role', 'basic_setup_rhel7_role']:
```
mdellweg: :+1: By reading your comment, i expected a much more involved change here. LOL

mdellweg: But if i read it correctly, the manifest will still be recorded. It's just not considered, when matching the requests again.
So be sure not to use a valuable manifest.

evgeni: Damn, you're right. I somehow thought it's not recorded then, but no.

We'd need to adjust https://github.com/theforeman/foreman-ansible-modules/blob/a071f171f6aa178545ec884657bc1e0511005505/tests/vcr_python_wrapper.py#L113-L115 to make it really unrecorded.

@wbclark I can look at this later this week in more detail if you want to?

evgeni: And twitter never fails to amaze me with timing, so: https://twitter.com/IanColdwater/status/1358805476616544258

roles/basic_setup_rhel7/README.md
```diff
@@ -0,0 +1,78 @@
+theforeman.foreman.basic_setup_rhel7
+====================================
+
+This role is an opinionated reuse of other roles in the collection, which creates a basic configuration for everything Katello needs to register and patch existing RHEL7 clients.
+
+That includes uploading a subscription manifest an organization, enabling the rhel-7-server-rpms repository and creating a sync plan for it, optionally syncing it immediately, and creating an activation key <name> to use when registering RHEL7 clients.
```
ehelms:
```suggestion
That includes uploading a subscription manifest to an organization, enabling the rhel-7-server-rpms repository and creating a sync plan for it, optionally syncing it immediately, and creating an activation key <name> to use when registering RHEL7 clients.
```

roles/basic_setup_rhel7/README.md
```diff
@@ -0,0 +1,78 @@
...
+
+The subscription manifest will be retrieved from the specified path on the ansible target host; optionally, it can be fetched first from the RHSM portal using the provided login credentials and manifest UUID. It will be uploaded to the specified organization.
```
ehelms:
```suggestion
The subscription manifest will be retrieved from the specified path on the Ansible target host; optionally, it can be fetched first from the RHSM portal using the provided login credentials and manifest UUID. It will be uploaded to the specified organization.
```

roles/basic_setup_rhel7/tasks/main.yml
```diff
@@ -0,0 +1,35 @@
...
+        sync_date: "{{ sync_plan_sync_date | default('2020-01-01 00:00:00 UTC') }}"
+        products:
+          - Red Hat Enterprise Linux Server
+
+- name: "Create Activation Key"
+  include_role:
+    name: theforeman.foreman.activation_keys
+  vars:
+    activation_keys:
+      - name: "{{ activation_key_name }}"
```
ehelms: I think this should have a default and reduce the amount that a user needs to specify when using this role.

roles/basic_setup_rhel7/README.md
```diff
@@ -0,0 +1,78 @@
...
+
+`sync_plan_interval`: 'hourly', 'daily', 'weekly', or 'custom cron'. See the [Sync Plan Role Documentation](https://github.com/theforeman/foreman-ansible-modules/blob/develop/roles/manifest/README.md#role-variables) for more information.
+
+`sync_plan_cron_expression`: Required when using the 'custom cron' `sync_plan_interval`.
+
+`sync_plan_sync_date`: Initial sync date for the sync plan, formatted as 'YYYY-MM-DD HH:MM:SS UTC'.
+
+`activation_key_name`: Name of the activation key to create.
+
+Example Playbooks
```
ehelms: Could we add an example that has only the bare minimum required set of information that needs to be supplied to the role? For example, not specifying sync plan information and not specifying activation key name with a downloaded manifest.

ehelms:
This role is tied tightly to RHEL 7. What is your vision of how RHEL 8 would be handled?

wbclark:
> This role is tied tightly to RHEL 7. What is your vision of how RHEL 8 would be handled?

I think generalizing and expanding on this role makes sense to do. This role could enable and optionally sync a few different repositories, with booleans to turn that on or off.

ehelms:
> I think generalizing and expanding on this role makes sense to do. This role could enable and optionally sync a few different repositories, with booleans to turn that on or off.

I take that to imply the role should get renamed to just be `rhel` instead of `rhel7` - which sounds great.

roles/basic_setup_rhel7/tasks/main.yml
wbclark: I agree, although I'm struggling to come up with a good name. I'll go with something like `base_activation_key` for now since it's a default org view / library key with auto-attach and no subs.... which is about as basic as you can get for an activation key.

wbclark:
I updated the role with the following changes:

1. Fixed variables for compatibility with https://github.com/theforeman/foreman-ansible-modules/pull/1145
2. Updated README.md based on feedback from @ehelms (additional example playbook, etc)
3. Re-named the role and removed some specific references to RHEL7 that should be more general to RHEL

TO-DOs:

1. Add support for RHEL8 and possibly other repos (app-stream, extras, optional, rhscl ?)
2. Add support for syncing repositories immediately

roles/basic_setup_rhel/README.md
```diff
@@ -0,0 +1,93 @@
+theforeman.foreman.basic_setup_rhel
+===================================
+
+This role is an opinionated reuse of other roles in the collection, which creates a basic configuration for everything Katello needs to register and patch existing RHEL clients.
+
+That includes uploading a subscription manifest to an organization, enabling the rhel-7-server-rpms repository and creating a sync plan for it, optionally syncing it immediately, and creating an activation key `Base_RHEL_Key` to use when registering RHEL clients.
```
ehelms: Thoughts on `base_rhel_key` to simplify for users? Having to remember what is capitalized and what isn't or fat-fingering the shift key is always a source of annoyance as a user.

ehelms:
Thoughts on dropping the `basic` from the role name? Tends to imply there is an advanced setup that exists or is coming.

roles/basic_setup_rhel/README.md
```diff
@@ -0,0 +1,93 @@
...
+`sync_plan_sync_date`: Initial sync date for the sync plan, formatted as 'YYYY-MM-DD HH:MM:SS UTC'.
+
+`activation_key_name`: Name of the activation key to create.
+
+Example Playbooks
+-----------------
+
+This minimal example assumes the manifest has already been downloaded to ~/manifest.zip on localhost (the Ansible control node) and uploads that manifest to the ACME organization. It enables the rhel-7-server-rpms repository and uses the role default sync plan with it. It creates an activation key with the role default name `Base_Activation_Key`.
+
+```yaml
```
ehelms: It'd be cool to have this same, bare minimum example, but with what is needed to perform the download to make it clearer to users how to differentiate making that choice.

ehelms:
> * Add support for RHEL8 and possibly other repos (app-stream, extras, optional, rhscl ?)

I'd say RHEL 8 + AppStream. Extras if others feel this is a common enough repository to get enabled by users. I think RHSCL is likely used but not as common.

mccun934:
+1 to just RHEL8 base + appstream and nothing else. 

I was thinking SCL because it is needed for Capsules but I think at the bare minimum case there won't be Capsules deploy so we can skip SCL as a default.

wbclark:
The role now enables and syncs RHEL7 and RHEL8 (base + appstream) by default, with options to skip enabling or syncing either.

Some issues remain:

1. Syncing repositories is slow as these are large repositories and we can't presently take advantage of parallelism when syncing across multiple products (it won't attempt to sync RHEL8 until the RHEL7 sync has completed). I've opened https://github.com/theforeman/foreman-ansible-modules/issues/1158 for this issue

2. Recording test fixtures is failing in my lab due to the below issue occuring when syncing RHEL7:

```
 Error:

PulpRpmClient::ApiError

Error message: the server returns an error HTTP status code: 502 Response headers: {"date"=>"Thu, 04 Mar 2021 02:52:23 GMT", "server"=>"Apache", "content-length"=>"445", "connection"=>"close", "content-type"=>"text/html; charset=iso-8859-1"} Response body: <!DOCTYPE HTML PUBLIC "-//IETF//DTD HTML 2.0//EN"> <html><head> <title>502 Proxy Error</title> </head><body> <h1>Proxy Error</h1> <p>The proxy server received an invalid response from an upstream server.<br /> The proxy server could not handle the request <em><a href="/pulp/api/v3/content/rpm/packages/">GET&nbsp;/pulp/api/v3/content/rpm/packages/</a></em>.<p> Reason: <strong>Error reading from remote server</strong></p></p> </body></html> 
```

I'll update Katello first before performing any additional troubleshooting on that sync failure.

3. The README requires updates to document the new behavior around RHEL8 repositories and syncing.

wbclark:
UPDATE: The above items are now addressed as this is now based on https://github.com/theforeman/foreman-ansible-modules/pull/1159

Therefore this is now ready for review

roles/basic_setup_rhel/tasks/main.yml
```diff
@@ -0,0 +1,72 @@
...
+
+- name: "Sync RHEL7 repository"
+  theforeman.foreman.repository_sync:
+    username: "{{ username }}"
+    password: "{{ password }}"
+    organization: "{{ foreman_organization }}"
+    server_url: "{{ server_url }}"
+    validate_certs: "{{ validate_certs }}"
+    product: Red Hat Enterprise Linux Server
+    wait_for_completion: false
```
ehelms: If we go this route we cannot say to users "Run this role and when it completes you are ready to go" and that goes against my understanding of the original intent.

wbclark: The alternative is that the role will take a very long time to run, as it first starts syncing `rhel-7-server-rpms` and waits for the task to complete, and only then starts to sync `rhel-8-for-x86_64-{baseos,appstream}-rpms`. In my lab (8cpu cores, 20gb mem) that takes about an hour.

My thinking is that documentation can inform the user that repositories have begun to sync by the time the role has finished running, and they can monitor progress at `https://katello.example.com/foreman_tasks/tasks` which is a page we want them to be familiar with regardless.

ehelms: My concern is that the repositories are unusable, registering a client is unusable until those repositories have synced. So we let users run a command designed to populate and put them in a state of ready to go when it's done but then they end up not in that state. So instead of monitoring this single command they kicked off for it to be "done". They now have to go hunt through the UI looking for tasks to monitor to let them know things are done, or user hammer to monitor the task or the API if they are not wanting to use the UI. I'd be all for this being an option to disable, but by default I think it should be enabled or, for me, it defeats the point and makes extra work for the user. 

I look at the goal of this role as:

 1) Run a single Ansible command
 2) When command completes, I can register a host and receive content

Is there a way, for this case, to parallelize the syncs? I know that for large sets of repositories we do want to be smart about it, for RHEL 7 + RHEL 8 I think Katello can handle managing syncing those optimally.

wbclark:
After discussing this today, we settled on implementing a variable `wait_for_repo_syncs` for this role which should have the fellowing behavior:

I. When true, the role should should sync RHEL7 and RHEL8 repos in parallel (this will likely take a bit of work in the repository_sync module) and should not complete until the sync tasks are completed. This will be be useful in cases where this role is expected to be the single source of truth about the completed configuration (for example if a playbook including this role is run from the REX interface, an user would expect to track the progress of the configuration from that single REX job).

II. When false, the role should create the sync tasks and it's up to the user to track the progress of those tasks. We expect this will be useful for users who are using this role in their own playbooks and are already familiar with Foreman Tasks and may want to move forward with working on other configuration which isn't blocked on syncs, while syncs run in the background.

roles/basic_setup_rhel/defaults/main.yml
```diff
@@ -0,0 +1,6 @@
+---
+enable_rhel7: true
+enable_rhel8: true
+sync_rhel7_now: true
+sync_rhel8_now: true
+wait_for_syncs: true
```
ehelms: Based on our new rules, I think this will, unfortunately, need to be:

```
basic_setup_rhel_enable_rhel7: true
basic_setup_rhel_enable_rhel8: true
basic_setup_rhel_sync_rhel7_now: true
basic_setup_rhel_sync_rhel8_now: true
basic_setup_rhel_wait_for_syncs: true
```

CC @evgeni

ehelms:
Not a blocker, more an opinion question for @evgeni and @mdellweg based on @wbclark and I chatting around this PR. Do y'all think it would make sense to allow the `repository_sync` module to handle lists as input similar to the `package` module does?

e.g.
```
product: A
repository_sync:
 -  RHEL 7
 -  RHEL Extras
```

I suppose this might be limited due to structure to only a sub-set of repositories per product given you can't specify a combination.

wbclark:
> Not a blocker, more an opinion question for @evgeni and @mdellweg based on @wbclark and I chatting around this PR. Do y'all think it would make sense to allow the `repository_sync` module to handle lists as input similar to the `package` module does?
> 
> e.g.
> 
> ```
> product: A
> repository_sync:
>  -  RHEL 7
>  -  RHEL Extras
> ```
> 
> I suppose this might be limited due to structure to only a sub-set of repositories per product given you can't specify a combination.

Based on conversation with @mdellweg and a removed example in the module docs, I think we can do instead:

1. Multiple sync tasks with `async: $some_reasonable_number` and `poll: 0` so that they can run in parallel and in the background

2. An `async_status` task that only runs when `basic_setup_rhel_wait_for_syncs: true`, so that the play waits for (Foreman) task completion.

I'm planning to try this implementation with the next update that I push.

roles/basic_setup_rhel/defaults/main.yml
evgeni: it'd be `foreman_…` tho ;)

roles/basic_setup_rhel/tasks/main.yml
```diff
@@ -0,0 +1,74 @@
...
+        products:
+          - Red Hat Enterprise Linux Server
+          - Red Hat Enterprise Linux for x86_64
+
+- name: "Create Activation Key"
+  include_role:
+    name: theforeman.foreman.activation_keys
+  vars:
+    foreman_activation_keys:
+      - name: "{{ activation_key_name | default('base_rhel_key') }}"
```
evgeni: this doesn't add any subs, thus assuming SCA is enabled?

roles/basic_setup_rhel/tasks/main.yml
```diff
@@ -0,0 +1,74 @@
...
+
+- name: "Enable RHEL7 repository"
+  include_role:
+    name: theforeman.foreman.repositories
+  vars:
+    foreman_products:
+      - name: Red Hat Enterprise Linux Server
+        repository_sets:
+          - name: Red Hat Enterprise Linux 7 Server (RPMs)
+            basearch: x86_64
```
evgeni: THe readme should point out that this does x86_64 only.

roles/basic_setup_rhel/tasks/main.yml
```diff
@@ -0,0 +1,74 @@
...
+  poll: "{{ 0 if not wait_for_syncs else omit }}"
+  when: enable_rhel8 and sync_rhel8_now
+
+- name: "Create Sync Plan"
+  include_role:
+    name: theforeman.foreman.sync_plans
+  vars:
+    foreman_sync_plans:
+      - name: "{{ sync_plan_name | default('RHEL Sync Plan') }}"
+        interval: "{{ sync_plan_interval | default(daily) }}"
```
evgeni:
```suggestion
        interval: "{{ sync_plan_interval | default('daily') }}"
```

evgeni:
All variables need to gain an `foreman_` prefix ;)

One thing that I wondered while reading the role is: how much more work would it be to have the role first generate apropriate `foreman_products`, `foreman_sync_plans` etc variables and then call the roles/modules once? If not too much, I'd love to see that, followed by a `basic_setup_centos` role that just does the vars differently and then does the same tasks.
But that's totally scope creep, it works fine w/o :)

ehelms:
> One thing that I wondered while reading the role is: how much more work would it be to have the role first generate apropriate `foreman_products`, `foreman_sync_plans` etc variables and then call the roles/modules once? If not too much, I'd love to see that, followed by a `basic_setup_centos` role that just does the vars differently and then does the same tasks.
> But that's totally scope creep, it works fine w/o :)

Scope creep I do agree. Valid ask? Very much so. Take a look at this content quickstart guide the Katello team is building out: https://github.com/theforeman/foreman-documentation/pull/439/files

Does that look familiar and like it could be automated? I'd love to see that included as a follow up.

roles/basic_setup_rhel/tasks/main.yml
wbclark: Auto-attach is enabled on activation keys by default, so it's not a strictly necessary assumption that SCA is enabled I think. The potential pitfall would be a subscription like 'Satellite Infrastructure Subscription' being consumed by RHEL machines because it also covers the RHEL product ID. But I believe auto-attach is supposed to prefer the minimal available subscription that covers all `/etc/pki/product/*.pem` product certificates present on the registering content host.

roles/basic_setup_rhel/tasks/main.yml
wbclark: We can always add the option to enable other architectures later if it's something that users want.

roles/basic_setup_rhel/tasks/main.yml
```diff
@@ -0,0 +1,94 @@
...
+
+- name: "Sync RHEL8 repositories"
+  theforeman.foreman.repository_sync:
+    username: "{{ username }}"
+    password: "{{ password }}"
+    organization: "{{ foreman_organization }}"
+    server_url: "{{ server_url }}"
+    validate_certs: "{{ validate_certs }}"
+    product: Red Hat Enterprise Linux for x86_64
+  async: "{{ ansible_check_mode | ternary(0, 14400) }}"
```
wbclark: This is a workaround for the checkmode test on CI, however there may be a better workaround such as exempting this role from the test altogether

wbclark: Hmmm, at least it is failing differently now :)

ehelms:
> Thoughts on dropping the `basic` from the role name? Tends to imply there is an advanced setup that exists or is coming.

Was still curious your thoughts on this.

wbclark:
I'm not attached to the `basic_setup_rhel` name @ehelms , in fact I don't like it very much. I don't like `setup_rhel` very much either.

How about `basic_rhel_content` or even `rhel_content` ? I think `theforeman.foreman.rhel_content` accurately conveys what this role does.

Regarding the basic part, I would like to later implement a role that does everything here + content lifecycle management (i.e. lifecycle environments and content views). I suppose that could simply be an expansion upon this role, off by default, in which case there is no reason to keep `basic` in the name.

wbclark:
> All variables need to gain an `foreman_` prefix ;)
> 
> One thing that I wondered while reading the role is: how much more work would it be to have the role first generate apropriate `foreman_products`, `foreman_sync_plans` etc variables and then call the roles/modules once? If not too much, I'd love to see that, followed by a `basic_setup_centos` role that just does the vars differently and then does the same tasks.
> But that's totally scope creep, it works fine w/o :)

+1 to similar role for centos.

Regarding having this role generate the exact data structures expected by the component roles its comprised of, I'm not opposed to the idea, but I'd like to better understand what exactly do we gain from implementing it that way?

mdellweg:
The sanity related failures should be fixed by rebasing.

mdellweg:
I believe, by skipping tasks in check_mode, your vcr tapes get out of sync with the calls to the vcr_python_wrapper.

roles/content_rhel/tasks/main.yml
```diff
@@ -0,0 +1,94 @@
...
+
+- name: "Sync RHEL8 repositories"
+  theforeman.foreman.repository_sync:
+    username: "{{ username }}"
+    password: "{{ password }}"
+    organization: "{{ foreman_organization }}"
+    server_url: "{{ server_url }}"
+    validate_certs: "{{ validate_certs }}"
+    product: Red Hat Enterprise Linux for x86_64
+  async: "{{ ansible_check_mode | ternary(0, 14400) }}"
```
mdellweg: 14400 sounds like the baudrate of my first modem. ;)

wbclark:
Fixed issues with check_mode by skipping that test for this role since check_mode does not support async tasks

I will examine remaining test failures tomorrow

wbclark:
By the way, https://github.com/theforeman/puppet-pulpcore/pull/178 was helpful to get all of the repositories to finally sync successfully

wbclark:
Hmmmm,

```
fatal: [tests]: FAILED! => {
    "msg": "The conditional check 'rhel7_job_result.finished' failed. The error was: error while evaluating conditional (rhel7_job_result.finished): 'dict object' has no attribute 'finished'"
}
```

But only on CI / build (3.7, stable-2.8) (pull_request)

wbclark:
`Using module file /opt/hostedtoolcache/Python/3.7.10/x64/lib/python3.7/site-packages/ansible/modules/windows/async_status.ps1`

`    "module_stderr": "/bin/sh: 1: powershell: not found\n",`

What in the hell...?

wbclark:
The issue is https://github.com/ansible/ansible/pull/65776

roles/content_rhel/README.md
```diff
@@ -0,0 +1,123 @@
+theforeman.foreman.content_rhel
+===================================
+
+This role is an opinionated reuse of other roles in the collection, which creates a basic configuration for everything Katello needs to register and patch existing RHEL clients.
```
evgeni:
```suggestion
This role is an opinionated reuse of other roles in the collection, which creates a basic configuration for everything needed to register and patch existing RHEL clients.
```

we try to avoid Foreman/Katello names to make branding easier :)

roles/content_rhel/README.md
```diff
@@ -0,0 +1,123 @@
...
+
+Repository behavior is controlled via the variables:
+
+`foreman_content_rhel_enable_rhel7`: Enable rhel-7-server-rpms repository (x86 architecture and 7Server release). Default true.
+
+`foreman_content_rhel_enable_rhel8`: Enable rhel-8-for-x86_64-baseos-rpms and rhel-8-for-x86_64-appstream-rpms (x86 architecture). Default true.
+
+`foreman_content_rhel_sync_now`: Sync repositories immediately after enabling. Default true.
+
+`foreman_content_rhel_wait_for_syncs`: Monitor status of sync tasks. When false, the sync tasks will continue running in the background after the playbook has finished running. This option is most useful when other automation (for example, registering and patching a client) requires the repository syncs to have completed.
```
evgeni: "Default true"? :)

roles/content_rhel/README.md
```diff
@@ -0,0 +1,123 @@
...
+- hosts: localhost
+  roles:
+    - role: theforeman.foreman.content_rhel
+      vars:
+        server_url: https://foreman.example.com
+        username: "admin"
+        password: "changeme"
+        organization: "ACME"
+        manifest_download: False
+        foreman_manifest_path: "~/manifest.zip"
```
evgeni: prefix plz

roles/content_rhel/README.md
```diff
@@ -0,0 +1,123 @@
...
+  roles:
+    - role: theforeman.foreman.content_rhel
+      vars:
+        server_url: https://foreman.example.com
+        username: "admin"
+        password: "changeme"
+        organization: "ACME"
+        manifest_download: True
+        rhsm_username: "happycustomer"
+        rhsm_password: "$ecur3p4$$w0rd"
```
evgeni: prefix

roles/content_rhel/README.md
```diff
@@ -0,0 +1,123 @@
...
+  roles:
+    - role: theforeman.foreman.content_rhel
+      vars:
+        server_url: https://foreman.example.com
+        username: "admin"
+        password: "changeme"
+        organization: "Default Organization"
+        manifest_download: True
+        rhsm_username: "happycustomer"
+        rhsm_password: "$ecur3p4$$w0rd"
```
evgeni: prefix :)

roles/content_rhel/README.md
```diff
@@ -0,0 +1,123 @@
...
+```yaml
+- hosts: localhost
+  roles:
+    - role: theforeman.foreman.content_rhel
+      vars:
+        server_url: https://foreman.example.com
+        username: "admin"
+        password: "changeme"
+        organization: "ACME"
+        manifest_download: False
```
evgeni: :)

tests/test_playbooks/content_rhel_role.yml
```diff
@@ -0,0 +1,35 @@
...
+        password: "{{ foreman_password }}"
+        server_url: "{{ foreman_server_url }}"
+        validate_certs: false
+        manifest_download: false
+        foreman_manifest_path: "{{ subscription_manifest_path }}"
+        foreman_sync_plan_name: "RHEL Sync Plan"
+        foreman_sync_plan_interval: daily
+        foreman_sync_plan_sync_date: 2021-02-02 00:00:00 UTC
+        foreman_activation_key_name: "basic_rhel_key"
+        foreman_content_rhel_wait_for_syncs: true
```
evgeni:
```suggestion
        foreman_content_rhel_wait_for_syncs: "{{ ansible_version['full'] is version('2.9', '>=') }}"
```

that should allow you to test unsync on 2.8, as discussed.

see https://docs.ansible.com/ansible/latest/user_guide/playbooks_tests.html#comparing-versions for details about `version` test and https://docs.ansible.com/ansible/latest/reference_appendices/special_variables.html for the `ansible_version` magic var

roles/content_rhel/tasks/main.yml
wbclark: It's a perfectly cromulent number..!

tests/test_playbooks/content_rhel_role.yml
wbclark: Ah, this is a much nicer approach than checking in tests/test_crud.py

Thanks!

wbclark:
Thanks @evgeni . All of the valuable feedback is addressed in the latest changes I have pushed. I'm expecting all tests will be green, if indeed it works to skip `async_status` tasks on Ansible 2.8 while the responses have been recorded in the API fixtures, but let's see :)

wbclark:
I have reduced the diff from over 62,000 lines to under 4,000 lines by eliminating requests/responses from the fixtures which return a foreman task in state: running, result: pending, after the first such response.

If my theory and implementation is correct, this should speed up testing rather significantly in the case where we are using async_status to check on the status of long running asynchronous tasks such as repository syncs.

wbclark:
Alas, it failed on CI / build (3.7, stable-2.9) with:

`    "msg": "Failed to connect to Foreman server: CannotOverwriteExistingCassetteException: Can't overwrite existing cassette ('fixtures/content_rhel_role-6.yml') in your current record mode ('none').\nNo match for the request (<The full traceback is:`

What's interesting to me is... `fixtures/content_rhel_role-6.yml` is not included in any commit here, does not exist on my local branch, and indeed was skipped when recording fixtures in the first place (along with `fixtures/content_rhel_role-4.yml`) which is a behavior that I found interesting but didn't question much further until now.

Confounding me further is the fact that this only caused a single check to fail, while I would have expected it to fail on all checks except for Ansible 2.8 where I am skipping async_status tasks due to https://github.com/ansible/ansible/pull/65776

evgeni:
@wbclark re-running it fixed it?!

changelogs/fragments/1097-content-rhel-role.yml
```diff
@@ -0,0 +1,2 @@
+minor_changes:
+  - Add a role `content_rhel` to perform basic setup for registering and syncing RHEL7 content hosts
```
ehelms: Can drop the `7` here

roles/content_rhel/README.md
```diff
@@ -0,0 +1,123 @@
+theforeman.foreman.content_rhel
+===================================
```
ehelms:
```suggestion
theforeman.foreman.content_rhel
===============================
```

evgeni (approved):

evgeni:
@wbclark thanks a ton for your work on this!

@mdellweg William has manually dropped a big chunk of "poll pending task" requests to speed up things and reduce the size of the fixtures. I wonder if that's something we should do by default? I.e. `GET /api/foreman_tasks/:id` and `state: pending` in reply → don't record.

mdellweg:
> @wbclark thanks a ton for your work on this!
> 
> @mdellweg William has manually dropped a big chunk of "poll pending task" requests to speed up things and reduce the size of the fixtures. I wonder if that's something we should do by default? I.e. `GET /api/foreman_tasks/:id` and `state: pending` in reply → don't record.

Sounds like a good idea. Would it be possible to record pending just once? There are codepaths for this we probably want to test at least once (or maybe we keep the old behavior in a dedicated test?).

evgeni:
I didn't fully explore it yet. Probably.

wbclark:
> > @wbclark thanks a ton for your work on this!
> > @mdellweg William has manually dropped a big chunk of "poll pending task" requests to speed up things and reduce the size of the fixtures. I wonder if that's something we should do by default? I.e. `GET /api/foreman_tasks/:id` and `state: pending` in reply → don't record.
> 
> Sounds like a good idea. Would it be possible to record pending just once? There are codepaths for this we probably want to test at least once (or maybe we keep the old behavior in a dedicated test?).

Indeed, the approach I have taken here is

1. Keep the first request, which returns a task with `state: planned, result: pending`

2. Keep the second request, which returns the task with `state: running, result: pending`

3. Drop all subsequent requests until the final one which returns the task with `state: stopped, result: success`

wbclark:
Thanks to all of those who gave their efforts to testing this module, reviewing the code, and providing design feedback !! I'm glad to see this soon in the hands of users

PR #1136, opened 2021-02-03
lookup the correct network_id for a given network

### LINKED ISSUE #1104: Automatically match network_name to network_id for the deployement of new VM´s on VMWare
##### SUMMARY

If I deploy a new Host over Foreman on my VMWare Cluster, with the host FAM, I need to manually check the correct network_id for my network or otherwise Foreman will always show a wrong network in the VM´s Interface tab.

Should be nearly the same like: https://github.com/theforeman/foreman-ansible-modules/issues/1041
##### ISSUE TYPE
 - Feature Idea

Automatically lookup the correct network_id for a network_name. I coul use as workaround the Ansible URI Module to do this, but, this is probably not the best solution, it would be better to implement it into the host FAM.

Here is an example of my workaround:

```
- name: Get network ids
  uri:
    url: "{{ foreman_server_url }}/api/compute_resources/{{ compute_ressource_id }}/available_networks"
    user: "{{ foreman_admin_user }}"
    password: "{{ foreman_admin_password }}"
    method: GET
    return_content: true
    force_basic_auth: true
    validate_certs: "{{ validate_foreman_certs }}"
  register: available_networks
  no_log: true

- set_fact:
    vsphere_network_id1: "{{ available_networks | json_query(query) | first }}"
  vars:
    query: "json.results[?name=='{{ foreman_host_vm.networkinterfaces[0].name }}'].id"
```

evgeni:
Fixes: #1104

draft because:
- [x] doesn't work on oVirt/RHV
- [x] only tested on VMware
- [x] no idea how to properly test that in CI

### DIFF
--- a/plugins/modules/host.py
+++ b/plugins/modules/host.py
@@ -491,7 +491,7 @@ def main():
             compute_resource = module.foreman_params['compute_resource']
             if 'compute_attributes' in module.foreman_params and 'cluster' in module.foreman_params['compute_attributes']:
                 cluster = module.find_cluster(module.foreman_params['compute_attributes']['cluster'], compute_resource)
-                module.foreman_params['compute_attributes']['cluster'] = cluster['id']
+                module.foreman_params['compute_attributes']['cluster'] = cluster['_api_identifier']
             else:
                 cluster = None
 
--- a/plugins/modules/host.py
+++ b/plugins/modules/host.py
@@ -115,6 +115,7 @@ options:
     description:
       - Additional compute resource specific attributes.
       - When this parameter is set, the module will not be idempotent.
+      - When you provide a I(cluster) here and I(compute_resource) is set, the cluster id will be automatically looked up.
     type: dict
     required: false
   interfaces_attributes:
@@ -251,6 +252,8 @@ options:
         description:
           - Additional compute resource specific attributes for the interface.
           - When this parameter is set, the module will not be idempotent.
+          - When you provide a I(network) here and I(compute_resource) is set, the network id will be automatically looked up.
+          - On oVirt/RHV I(cluster) is required in the hosts I(compute_attributes) for the lookup to work.
         type: dict
 extends_documentation_fragment:
   - theforeman.foreman.foreman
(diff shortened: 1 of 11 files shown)


evgeni:
cc @SimonLorentz @fgoebel @cmeissner if y'all want to play around with this.

plugins/modules/host.py
```diff
@@ -484,6 +484,17 @@ def main():
...
+        if 'compute_resource' in module.foreman_params and 'interfaces_attributes' in module.foreman_params:
+            avn_params = {'id': module.foreman_params['compute_resource']['id']}
+            available_networks = module.resource_action('compute_resources', 'available_networks', params=avn_params,
+                                                        ignore_check_mode=True, record_change=False)['results']
+
+            for interface in module.foreman_params['interfaces_attributes']:
+                if 'compute_attributes' in interface and 'network' in interface['compute_attributes']:
+                    network_name = interface['compute_attributes']['network']
+                    network_id = next((network['id'] for network in available_networks if network['name'] == network_name), network_name)
+                    interface['compute_attributes']['network'] = network_id
```
evgeni: what should happen if the network was not found in the available ones? currently, this silently takes the value from the playbook, and will thus fail, I think

SimonLorentz: I think, the best would be to print an error message, such as "network not found" or something

SimonLorentz:
I tried it against foreman 1.24.3 and 2.1.4 and it works fine with both, thank you very much :D

evgeni:
Thanks @SimonLorentz for testing! I guess you're also on VMware?

SimonLorentz:
Hi @evgeni, yes I tested it against VMware, shell I test it against proxmox or something else?

evgeni:
That would be greatly appreciated, yes. Proxmox is something I don't have easily available here.

SimonLorentz:
Okay I tried it against proxmox, it works, but in proxmox we have as id and name the same value, so that also works without the modification

evgeni:
Okay, updated to error out when the network can't be found. Also made it work on oVirt/RHV.

evgeni:
the devel failures are  unrelated and fixed in https://github.com/theforeman/foreman-ansible-modules/pull/1140

evgeni:
okay, maybe that's a bit more complicated for compute_profiles…

evgeni:
~~(compute_profile takes cluster name, not cluster id, …)~~ only on vmware, bleh

evgeni:
whoa, even better, when listing the networks of a cluster, vmware takes cluster name, ovirt cluster id, what is this…

evgeni:
@SimonLorentz would you mind retesting again in your envs? I did a few changes, which *shouldn't* affect things that already worked, but you never know.

evgeni:
@m-bucher you might be interested in this too!

tests/test_playbooks/vars/compute_profile.yml
```diff
@@ -41,22 +41,22 @@ ovirt:
   compute_profile:
     name: app-small
     attrs:
-      cluster: c95c55f3-705e-4182-92a6-f66754aadf02
+      cluster: Devel-Infra
```
evgeni, line 44: there, a working change! :)

tests/test_playbooks/vars/compute_profile.yml
```diff
@@ -41,22 +41,22 @@ ovirt:
...
       cores: 2
       sockets: 1
       memory: 1073741824
       ha: 0
       interfaces_attributes:
         0:
           name: ""
-          network: 390666e1-dab3-4c99-9f96-006b2e2fd801
+          network: ovirtmgmt
```
evgeni, line 52: there, a working change too! :)

tests/test_playbooks/vars/compute_profile.yml
```diff
@@ -41,22 +41,22 @@ ovirt:
...
           interface: virtio
       volumes_attributes:
         0:
-          size_gb: 16
-          storage_domain: 19c50090-1ab4-4023-a63f-75ee1018ed5e
+          size_gb: '16'
```
evgeni, line 56: this is "unrelated", but Foreman wants all params to be strings here, whyever.

ehelms (approved):

PR #1138, opened 2021-02-12
don't record subscription manifest content

evgeni:


wbclark (approved):
It looks good to me. The logic checks out and I can see in the updated fixtures where the manifest was filtered from the request body and replaced with FAKE_MANIFEST. Thanks @evgeni !

wbclark (approved):
ACK

PR #1140, opened 2021-02-15
properly guard the inventory plugin if requests is not installed

evgeni:
this wasn't tested by sanity before and thus slipped


mdellweg (approved):

mdellweg:
> this wasn't tested by sanity before and thus slipped

And now it is? Did upstream sanity change?

evgeni:
Yeah, recent change in Ansible devel: https://github.com/ansible/ansible/commit/1f3a90270b5b718ad1456e6bb11097724cd3d8ab

PR #1141, opened 2021-02-15
add a hint on user permissions when no entities are found

evgeni:


plugins/module_utils/foreman_helper.py
```diff
@@ -726,9 +726,11 @@ def find_resource(self, resource, search, params=None, failsafe=False, thin=None
         else:
             if len(results) > 1:
                 error_msg = "too many ({0})".format(len(results))
+                user_msg = ""
             else:
                 error_msg = "no"
-            self.fail_json(msg="Found {0} results while searching for {1} with {2}".format(error_msg, resource, search))
+                user_msg = "Does the '{0}' user have sufficient permissions to see those?".format(self._foremanapi_username)
```
ehelms, line 732: Is the API returning a 404?

evgeni: No, an empty result set.

ehelms: So we are sharing more information than the server does here by suggesting that this may be hidden behind permissions rather than not exist. I tend to shy away from that.

evgeni: This is rather common user error. Care to elaborate more why you don't like that?

We have similar hints in other places (like https://github.com/Apipie/apypie/blob/master/apypie/api.py#L201-L203 that also exists in the Ruby version and is displayed by hammer).

ehelms: Those hints are about the state of the server, this is about the state of the user and the conditions for existence. As I understand things, security concerns around APis are moving more towards entities either existing or not without trying to indicate that it may be due to permissions which would give an attacker a hint.

evgeni: I always thought the idea of the security aspect is not to differentiate 403 (forbidden) and 404 (not found) errors when *replying* to the user.

In this case the server replies "200 - OK, found 0" in either case, and we're just enriching the error we show to the user (in either case, because we can't know why the server found 0), without leaking any more information they had before.

ehelms: We know that, but does a user know that if they only see this output? If I saw this, I would think I got a permission denied of some sort.

PR #1142, opened 2021-02-17
use names for datacenter and quota in oVirt examples

evgeni:
the API accepts those since 1.22 for DC and 2.0 for quota

draft because it makes the execution non-idempotent, and I need to figure out how to map that…


evgeni:
nobody has forbidden evgeni to implement the translation the same way as we translate networks/datastores.

PR #1149, opened 2021-02-23
add bindep definition for RPM bindings

evgeni:
ansible-builder can use bindep to create a working Execution Environment
based on system packages, which we can benefit from by including RPM
bindings there, as those are hard to install from PyPI.

[1] https://ansible-builder.readthedocs.io/en/latest/definition.html

### DIFF
--- a/bindep.txt
+++ b/bindep.txt
@@ -1,0 +1,2 @@
+python3-rpm [(platform:redhat platform:base-py3)]
+rpm-python [(platform:redhat platform:base-py2)]
--- a/Makefile
+++ b/Makefile
@@ -5,7 +5,7 @@ MANIFEST := build/collections/ansible_collections/$(NAMESPACE)/$(NAME)/MANIFEST.
 
 ROLES := $(wildcard roles/*)
 PLUGIN_TYPES := $(filter-out __%,$(notdir $(wildcard plugins/*)))
-METADATA := galaxy.yml LICENSE README.md meta/runtime.yml requirements.txt changelogs/changelog.yaml CHANGELOG.rst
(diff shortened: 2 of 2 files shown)


bindep.txt
```diff
@@ -0,0 +1,2 @@
+python3-lxml [(platform:redhat platform:base-py3)]
+python-lxml [(platform:redhat platform:base-py2)]
```
mdellweg: ansible-builder will consult requirements.txt for python depenencies, and bindep.txt for system packages, right?
But the weird `rpm-py-installer` is living in requirements-dev.txt.

evgeni: Correct. `requirements-dev.txt` is not what `ansible-builder` will look at.

mdellweg: Who else is looking at `requirements.txt` should we move all the "optional" requirements there?

evgeni: besides `ansible-builder` (for pure python deps) and our CI? nobody.

we could move python-debian there, yeah

ehelms:
> by including RPM
> bindings there,

What exactly does this mean in our context? Or perhaps where do we use RPM bindings?

evgeni:
the `content_upload` module uses it to fetch information about the RPM if you upload to an RPM repo:

https://github.com/theforeman/foreman-ansible-modules/blob/a39adb3ff0979a46c953f3555874f31669cf1950/plugins/modules/content_upload.py#L102-L120

ehelms (approved):

evgeni:
funny how nobody complained about the wrong RPM names in the file :see_no_evil:

PR #1154, opened 2021-02-24
lookup storage domains and pods

evgeni:


evgeni:
cc @SimonLorentz @fgoebel @cmeissner @m-bucher

plugins/module_utils/foreman_helper.py
```diff
@@ -820,6 +820,36 @@ def find_network(self, name, compute_resource, cluster=None):
             self.fail_json(msg=err_msg)
         return network
 
+    def find_storage_domain(self, name, compute_resource, cluster=None):
```
evgeni, line 823: at this point, the four find methods look really similar, so probably should be dried up.

ehelms (approved):
Tests look good, if you want to DRY it up go for it

evgeni:
> Tests look good, if you want to DRY it up go for it

I'll do that in a follow up PR, thanks

PR #1159, opened 2021-03-04
Add wait_for_completion option to repository_sync module

### LINKED ISSUE #1158: Sync multiple products simultaneously
Currently when syncing repositories using the module `theforeman.foreman.repository_sync` I have to define either one ansible task per repository or one ansible task per product. Furthermore, each ansible task runs until the foreman task is completed, which under some circumstances takes considerable time.

It would be a significant improvement to define an option like `wait_for_completion: false` so that the play can continue while large repository syncs have been initiated. This way several repos across multiple products can be synced in parallel, as is supported by Katello, which would greatly reduce the execution time for plays which sync multiple large repositories.

wbclark:
Fixes #1158

### DIFF
--- a/plugins/module_utils/foreman_helper.py
+++ b/plugins/module_utils/foreman_helper.py
@@ -1158,7 +1158,7 @@ class ForemanAnsibleModule(AnsibleModule):
         return None
 
     def resource_action(self, resource, action, params, options=None, data=None, files=None,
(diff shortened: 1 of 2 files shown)


evgeni:
triage: ideally this is not needed and we'll try to figure out what really fails the tests in Williams content PR.

evgeni:
we got #1135 working without this, so closing for now

PR #1161, opened 2021-03-05
Allow disabling callback plugin even if it gets loaded

adamruzicka:
Foreman_ansible decides whether the callback should be used to prevent facts
from being uploaded in case ansible is used "just" as a remote execution
provider. It was implemented by setting ANSIBLE_CALLBACK_WHITELIST environment
variable, but that stomps over any user-configured callback plugins[2]. When we
moved to ansible-runner the whitelist env trick was dropped, but also led to
facts being uploaded even if we don't want them[1].

While we could read the configured callback plugins from ansible's configuration
and change it on the fly, it would be a bit fragile. The configuration can be in
different places and noone guarantees that the location or its contents won't
change in the foreseeable future.

This change allows us to selectively disable the callback plugin if we need to,
without any undesired side-effects.

[1] - https://projects.theforeman.org/issues/32020
[2] - https://projects.theforeman.org/issues/30373

### DIFF
--- a/plugins/callback/foreman.py
+++ b/plugins/callback/foreman.py
@@ -66,6 +66,13 @@ DOCUMENTATION = '''
         ini:
           - section: callback_foreman
             key: verify_certs
+      disable:
+        description:
+          - Toggle to make the callback plugin disable itself even if it is loaded.
+          - It can be set to '1' to prevent the plugin from being used even if it gets loaded.
+        env:
+          - name: FOREMAN_DISABLE
+        default: 0
 '''
 
 import os
@@ -104,6 +111,9 @@ class CallbackModule(CallbackBase):
     def set_options(self, task_keys=None, var_options=None, direct=None):
 
         super(CallbackModule, self).set_options(task_keys=task_keys, var_options=var_options, direct=direct)
+
+        if self.get_option('disable'):
+            self._disable_plugin()
 
         self.FOREMAN_URL = self.get_option('url')
         self.FOREMAN_SSL_CERT = (self.get_option('client_cert'), self.get_option('client_key'))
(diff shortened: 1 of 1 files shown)


mdellweg (approved):
LGTM

ezr-ondrej (approved):
:heart: it :+1:

plugins/callback/foreman.py
```diff
@@ -66,6 +66,13 @@
         ini:
           - section: callback_foreman
             key: verify_certs
+      disable:
+        description:
+          - Toggle to make the callback plugin disable itself even if it is loaded.
+          - It can be set to '1' to prevent the plugin from being used even if it gets loaded.
+        env:
+          - name: FOREMAN_DISABLE
```
ezr-ondrej: A thought:
Should this be more specific like `FOREMAN_CALLBACK_DISABLE`?
I'd expect this to work on all plugins not just callback if called so general, but that's probably not what we want here :thinking:

plugins/callback/foreman.py
```diff
@@ -105,6 +112,9 @@ def set_options(self, task_keys=None, var_options=None, direct=None):
 
         super(CallbackModule, self).set_options(task_keys=task_keys, var_options=var_options, direct=direct)
 
+        if self.get_option('disable'):
+            self._disable_plugin()
```
ezr-ondrej: [WARNING]: Skipping callback 'theforeman.foreman.foreman', unable to load due to: _disable_plugin() missing 1 required positional argument: 'msg'

ezr-ondrej (changes requested):

plugins/callback/foreman.py
adamruzicka: Oh, right, not really a graceful disable, will fix

adamruzicka: Done

ezr-ondrej (approved):
Thanks @adamruzicka! works smoothly now :)

PR #1166, opened 2021-03-10
support Foreman's structured plugin details

evgeni:
Foreman 2.4 [1] started to return plugin information in a structured
way, let's be nice and support that too :-)

[1] https://github.com/theforeman/foreman/pull/8207/

### DIFF
--- a/plugins/modules/scc_product.py
+++ b/plugins/modules/scc_product.py
@@ -81,8 +81,12 @@ def main():
             statuses = module.foremanapi.resource('ping').call('statuses')
             plugins = statuses['results']['foreman']['plugins']
             for plugin in plugins:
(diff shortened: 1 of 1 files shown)


evgeni:
@Manisha15 can you have a look and see if that works with 2.4 and current SCC plugin? I don't have that installed here.

Manisha15:
> @Manisha15 can you have a look and see if that works with 2.4 and current SCC plugin? I don't have that installed here.

Hey @evgeni, I tested it on Foreman 2.4 and with latest SCC plugin and it's working fine.

mdellweg (approved):

evgeni:
Thanks a ton!

PR #1167, opened 2021-03-10
Add a content_view_info module

ehelms:


ehelms:
There was heavy copy and paste involved in the creation, so highly possible I missed a subnet here or there.

plugins/modules/content_view_info.py
```diff
@@ -0,0 +1,81 @@
...
+
+EXAMPLES = '''
+- name: "Show a content_view"
+  theforeman.foreman.content_view_info:
+    username: "admin"
+    password: "changeme"
+    server_url: "https://foreman.example.com"
+    name: "CentOS 8"
+
+- name: "Show all content_views with domain example.com"
```
mdellweg: Not a subnet as such... But a domain. ;)

mdellweg (approved):
Only the title of the one example looks odd to me.

plugins/modules/content_view_info.py
```diff
@@ -0,0 +1,81 @@
...
+  theforeman.foreman.content_view_info:
+    username: "admin"
+    password: "changeme"
+    server_url: "https://foreman.example.com"
+    search: "name = 'CentOS 8'"
```
evgeni: This is not a valid search ;)
Our API *needs* double quotes.

```suggestion
    search: 'name = "CentOS 8"'
```

tests/test_playbooks/content_view_info.yml
```diff
@@ -0,0 +1,93 @@
...
+          - content_view_info['content_view']['name'] == "Test Content View"
+          - content_view_info['content_view']['repositories'][0]['name'] == "Test Repository"
+
+    - name: search content_view info
+      content_view_info:
+        username: "{{ foreman_username }}"
+        password: "{{ foreman_password }}"
+        server_url: "{{ foreman_server_url }}"
+        validate_certs: "{{ foreman_validate_certs }}"
+        search: "name = \"Test Content View\""
```
evgeni:
```suggestion
        search: 'name = "Test Content View"'
```

plugins/modules/content_view_info.py
```diff
@@ -0,0 +1,81 @@
...
+  - theforeman.foreman.foreman.infomodule
+'''
+
+EXAMPLES = '''
+- name: "Show a content_view"
+  theforeman.foreman.content_view_info:
+    username: "admin"
+    password: "changeme"
+    server_url: "https://foreman.example.com"
+    name: "CentOS 8"
```
evgeni, line 42: Given this is Katello, I'd expect you *need* to pass `organization`, otherwise you won't get anything nice back?

Your tests say something different, though

evgeni (changes requested):

plugins/modules/content_view_info.py
ehelms: Not for index calls, they have been treated as non-org scoped allowing you to get data across organizations.

evgeni: TIL :)

evgeni (approved):

PR #1169, opened 2021-03-11
Add content_view_version_info module

ehelms:
Opening as a draft, I keep running into:

```
fatal: [tests]: FAILED! => {
    "changed": false,
    "invocation": {
        "module_args": {
            "content_view": "Test Content View",
            "name": "blah",
            "search": "version = \"1.0\"",
            "server_url": "https://pipe-katello-server-nightly-centos7.wareagle.example.com",
            "username": "admin",
            "validate_certs": false
        }
    },
    "msg": "Unsupported parameters for (content_view_info) module: content_view Supported parameters include: location, name, organization, password, search, server_url, username, validate_certs"
}
```

I was following the model of `repository_info`.

### DIFF
--- a/plugins/doc_fragments/foreman.py
+++ b/plugins/doc_fragments/foreman.py
@@ -352,6 +352,25 @@ options:
     type: str
 '''
 
+    INFOMODULEWITHOUTNAME = '''
+options:
+  location:
+    description:
+      - Label of the Location to scope the search for.
+    required: false
+    type: str
+  organization:
+    description:
+      - Name of the Organization to scope the search for.
+    required: false
+    type: str
+  search:
+    description:
+      - Search query to use
+      - If None, all resources are returned.
+    type: str
+'''
+
     KATELLOINFOMODULE = '''
 options:
   organization:
--- a/plugins/modules/content_view_version_info.py
+++ b/plugins/modules/content_view_version_info.py
@@ -1,0 +1,90 @@
+#!/usr/bin/python
+# -*- coding: utf-8 -*-
+# (c) 2021 Eric Helms
+#
+# This program is free software: you can redistribute it and/or modify
+# it under the terms of the GNU General Public License as published by
+# the Free Software Foundation, either version 3 of the License, or
+# (at your option) any later version.
+#
+# This program is distributed in the hope that it will be useful,
+# but WITHOUT ANY WARRANTY; without even the implied warranty of
+# MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+# GNU General Public License for more details.
+#
+# You should have received a copy of the GNU General Public License
+# along with this program.  If not, see <http://www.gnu.org/licenses/>.
+
+from __future__ import absolute_import, division, print_function
+__metaclass__ = type
+
+
+DOCUMENTATION = '''
+---
+module: content_view_version_info
+version_added: 2.1.0
+short_description: Fetch information about Content Views
+description:
+  - Fetch information about Content Views
+author:
+  - "Eric Helms (@ehelms)"
+options:
+  content_view:
+    description:
+      - Content View to which the Version belongs
+    required: true
+    type: str
+extends_documentation_fragment:
+  - theforeman.foreman.foreman
+  - theforeman.foreman.foreman.katelloinfomodule
+  - theforeman.foreman.foreman.infomodulewithoutname
+'''
+
+EXAMPLES = '''
+- name: "Show a content view version"
+  theforeman.foreman.content_view_version_info:
+    username: "admin"
+    password: "changeme"
+    server_url: "https://foreman.example.com"
+    content_view: "CentOS 8 View"
+    search: 'version = "4.0"'
+
+- name: "Show all content view_versions for a content view"
+  theforeman.foreman.content_view_version_info:
+    username: "admin"
+    password: "changeme"
+    server_url: "https://foreman.example.com"
+    content_view: "CentOS 8 View"
+'''
+
+RETURN = '''
+content_view_versions:
+  description: List of all found content_view_versions and their details
+  returned: success and I(search) was passed
+  type: list
+  elements: dict
+'''
+
+from ansible_collections.theforeman.foreman.plugins.module_utils.foreman_helper import (
+    KatelloInfoAnsibleModule,
+)
+
+
+class KatelloContentViewVersionInfo(KatelloInfoAnsibleModule):
+    pass
+
+
+def main():
+    module = KatelloContentViewVersionInfo(
+        foreman_spec=dict(
+            content_view=dict(type='entity', scope=['organization'], required=True),
+            name=dict(invisible=True),
+        ),
+    )
+
+    with module.api_connection():
+        module.run()
+
+
+if __name__ == '__main__':
+    main()
--- a/plugins/module_utils/foreman_helper.py
+++ b/plugins/module_utils/foreman_helper.py
@@ -1348,7 +1348,9 @@ class ForemanInfoAnsibleModule(ForemanStatelessEntityAnsibleModule):
             location=dict(type='entity'),
         )
         foreman_spec.update(kwargs.pop('foreman_spec', {}))
-        mutually_exclusive = kwargs.pop('mutually_exclusive', []) + [['name', 'search']]
+        mutually_exclusive = kwargs.pop('mutually_exclusive', [])
+        if not foreman_spec['name'].get('invisible', False):
+            mutually_exclusive.extend([['name', 'search']])
         super(ForemanInfoAnsibleModule, self).__init__(foreman_spec=foreman_spec, mutually_exclusive=mutually_exclusive, **kwargs)
 
     def run(self, **kwargs):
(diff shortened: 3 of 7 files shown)


plugins/modules/content_view_version_info.py
```diff
@@ -0,0 +1,86 @@
...
+module: content_view_version_info
+version_added: 2.1.0
+short_description: Fetch information about Content Views
+description:
+  - Fetch information about Content Views
+author:
+  - "Eric Helms (@ehelms)"
+extends_documentation_fragment:
+  - theforeman.foreman.foreman
+  - theforeman.foreman.foreman.infomodule
```
evgeni:
```suggestion
  - theforeman.foreman.foreman.katelloinfomodule
  - theforeman.foreman.foreman.infomodule
```

(this will load the right organization definition)

plugins/modules/content_view_version_info.py
```diff
@@ -0,0 +1,86 @@
...
+DOCUMENTATION = '''
+---
+module: content_view_version_info
+version_added: 2.1.0
+short_description: Fetch information about Content Views
+description:
+  - Fetch information about Content Views
+author:
+  - "Eric Helms (@ehelms)"
+extends_documentation_fragment:
```
evgeni, line 37:
```suggestion
options:
  content_view:
    description:
      - Content View to which the Version belongs
    required: true
    type: str
extends_documentation_fragment:
```

evgeni (changes requested):
two comments about the docs, which will also make the tests pass when fixed

ACK otherwise

plugins/modules/content_view_version_info.py
```diff
@@ -0,0 +1,93 @@
...
+
+
+class KatelloContentViewVersionInfo(KatelloInfoAnsibleModule):
+    pass
+
+
+def main():
+    module = KatelloContentViewVersionInfo(
+        foreman_spec=dict(
+            content_view=dict(type='entity', scope=['organization'], required=False),
```
evgeni:
```suggestion
            content_view=dict(type='entity', scope=['organization'], required=True),
```

Or is it not required for search to work? If so, my doc string suggestion was wrong

ehelms: It might as well be in this format. I had tried to allow getting all versions for an environment but translating lifecycle_environment and environment was a pain I kicked down the road.

evgeni:
the other sanity failure is fixed in https://github.com/theforeman/foreman-ansible-modules/pull/1170

evgeni:
One thing that I'd love to do with this PR, before merging, is getting rid of the `name` parameter that currently gets injected from the base InfoModule class, but I have not thought how I'd do that nicely.

ehelms:
That being present doesn't keep this module from functioning right? Just provides a "false" parameter if someone savvy enough were to try to use it?

evgeni:
Correct, and removing it is an major change, *technically* ;)
(You could argue it didn't ever work, tho)

ehelms:
> (You could argue it didn't ever work, tho)

Argued. :) I'm OK waiting if you want to get that resolved first.

evgeni:
If you're not too blocked, I'd look into this Tue/Wed.

evgeni:
So I promised you a few thoughts, and here there are:

I think the easiest way to remove `name` from the docs is to introduce a new doc fragment `INFOMODULEWITHOUTNAME` similar to https://github.com/theforeman/foreman-ansible-modules/blob/dbd8ab270cac9d69f8b1cd1bd1a03e7ffb078260/plugins/doc_fragments/foreman.py#L329-L353 and omit the `name` from the options and make the search option more like:
```yaml
  search:
    description:
      - Search query to use
      - If None, all resources are returned.
    type: str
```

This will duplicate content from `INFOMODULE`, but there is no way to hide things from a fragment, and making it a composite like we did in other places would mean to add the `NAME` composite to almost every module, which seems a lot of noise/boilerplate for a corner case. Obviously you can also not create a new dog fragment, but define the correct params here in the module, given it's the only one.

You will also have to add `name=dict(invisitble=True)` to the `foreman_spec` to hide it from the accepted params (or sanity will yell at you). Again, creating a separate class for that is probably too much.

Ping me if you want me to just do the changes to your branch.

Another thing that came to my mind while looking at that: you didn't add a `version` param to the module, but relied on the search feature. I *guess* that's OK, but we might want to add a `version` param as a future improvement as it makes the module calling easier?

mdellweg:
If we go for the subclass thing, it would be `INFOMODULE` and `NAMEDINFOMODULE`? Or the name is added as its own snippet?
How often do we lookup items not by their `name`?

evgeni:
yeah, or `INFOMODULEWITHNAME`, pick a fork.

I think very few models have non-searchable names, so I'd expect this being really the exception until proven otherwise.

(the only other non-name thing I can think off are host_errata, which are searchable by hostname, not name…)

ehelms:
I implemented the withoutname idea... now I hit:

```
ERROR: plugins/modules/content_view_version_info.py:0:0: mutually_exclusive-unknown: mutually_exclusive contains terms which are not part of argument_spec: name
```

I tried to think of a solution but none of them felt clean. So figure I leave that up to the experts to suggest a path forward.

mdellweg:
You are hitting:
https://github.com/theforeman/foreman-ansible-modules/blob/dbd8ab270cac9d69f8b1cd1bd1a03e7ffb078260/plugins/module_utils/foreman_helper.py#L1351

And the `ForemanInfoAnsibleModule` in the class hierarchy assumes the `name` to be the lookup field in one more place:
https://github.com/theforeman/foreman-ansible-modules/blob/dbd8ab270cac9d69f8b1cd1bd1a03e7ffb078260/plugins/module_utils/foreman_helper.py#L1362

evgeni:
the following probably works?

```patch
diff --git plugins/module_utils/foreman_helper.py plugins/module_utils/foreman_helper.py
index 130dcd4d..5f5f448f 100644
--- plugins/module_utils/foreman_helper.py
+++ plugins/module_utils/foreman_helper.py
@@ -1348,7 +1348,9 @@ class ForemanInfoAnsibleModule(ForemanStatelessEntityAnsibleModule):
             location=dict(type='entity'),
         )
         foreman_spec.update(kwargs.pop('foreman_spec', {}))
-        mutually_exclusive = kwargs.pop('mutually_exclusive', []) + [['name', 'search']]
+        mutually_exclusive = kwargs.pop('mutually_exclusive', [])
+        if not foreman_spec['name'].get('invisible', False):
+            mutually_exclusive.extend([['name', 'search']])
         super(ForemanInfoAnsibleModule, self).__init__(foreman_spec=foreman_spec, mutually_exclusive=mutually_exclusive, **kwargs)
 
     def run(self, **kwargs):
```

The alternative would be to define a `ForemanNamelessInfoAnsibleModule` or so…

evgeni:
> And the `ForemanInfoAnsibleModule` in the class hierarchy assumes the `name` to be the lookup field in one more place:
> 
> https://github.com/theforeman/foreman-ansible-modules/blob/dbd8ab270cac9d69f8b1cd1bd1a03e7ffb078260/plugins/module_utils/foreman_helper.py#L1362

Nah, if it's not in there, it will just go straight to `else`, which is the intended way here

mdellweg:
> > And the `ForemanInfoAnsibleModule` in the class hierarchy assumes the `name` to be the lookup field in one more place:
> > https://github.com/theforeman/foreman-ansible-modules/blob/dbd8ab270cac9d69f8b1cd1bd1a03e7ffb078260/plugins/module_utils/foreman_helper.py#L1362
> 
> Nah, if it's not in there, it will just go straight to `else`, which is the intended way here

I thought that you would replace "name" with whatever specifies a resource here (e.g. "login" or "username" for users).
Going straight to the else means that "search" will be used. But maybe i'm missing something.

evgeni:
> > Nah, if it's not in there, it will just go straight to `else`, which is the intended way here
> 
> I thought that you would replace "name" with whatever specifies a resource here (e.g. "login" or "username" for users).
> Going straight to the else means that "search" will be used. But maybe i'm missing something.

The module currently doesn't add a dedicated resource identifier, so search needs to be used to get a specific version.

ehelms:
Green now, thanks @evgeni

plugins/doc_fragments/foreman.py
```diff
@@ -352,6 +352,26 @@ class ModuleDocFragment(object):
...
+  organization:
+    description:
+      - Name of the Organization to scope the search for.
+    required: false
+    type: str
+  search:
+    description:
+      - Search query to use
+      - If None, and I(name) is not set, all resources are returned.
+      - Mutually exclusive with I(name).
```
evgeni:
```suggestion
      - If None, all resources are returned.
```

plugins/modules/content_view_version_info.py
```diff
@@ -0,0 +1,94 @@
...
+    password: "changeme"
+    server_url: "https://foreman.example.com"
+    content_view: "CentOS 8 View"
+'''
+
+RETURN = '''
+content_view_version:
+  description: Details about the found content_view_version
+  returned: success and I(name) was passed
+  type: dict
```
evgeni:
```suggestion
```

There is currently no way to get a singular result, all results are searches.

evgeni (approved):
two documentation nitpicks, then good to go

thanks!

PR #1170, opened 2021-03-12
explicitly set key_path to no_log False

evgeni:
sanity tests now check that and it matches the general /key/ rule.
no_log=False is the explicit declaration of "this is not secret"


ehelms (approved):

mdellweg (approved):
Is this because it contains "key"?

evgeni:
> Is this because it contains "key"?

yes

PR #1155, opened 2021-02-25
Foreman facts uploading correctly merges facts

adamruzicka:
Adapted from https://github.com/ansible/ansible/pull/51546

PR #1172, opened 2021-03-13
Specify the timezone in timestamps

### LINKED ISSUE #1171: The reported facts and reports are ignoring machine timezone
##### SUMMARY
A fresh report causes out of sync configuration status. This is caused by the timestamp in the report being without timezone information. The same problem is present for facts upload. We should unify the way how we fetch the current time information and include the TZ information. At the same time, the format can be a bit tweaked so it's more native to the Ruby that parses it.

##### ISSUE TYPE
 - Bug Report

##### ANSIBLE VERSION
N/A
```

```

##### COLLECTION VERSION
all
```

```

##### KATELLO/FOREMAN VERSION
N/A
```

```

##### STEPS TO REPRODUCE
1. have a machine with non +00:00 TZ
2. create a report and upload it to Foreman
3. visit the report page, you can see the reported at information being wrong

##### EXPECTED RESULTS
The time zone should be correctly parsed

##### ACTUAL RESULTS

```
The time zone is ignored so the time is considered as UTC
```

ares:
fixes #1171

### DIFF
--- a/plugins/callback/foreman.py
+++ b/plugins/callback/foreman.py
@@ -132,8 +132,10 @@ def get_time():
 def get_now():
     """
     Return the current timestamp as a string to be sent over the network.
+    The time is always in UTC timezone and printed in a format Ruby
+    DateTime can easily parse.
     """
-    return datetime.utcnow().isoformat()
+    return datetime.datetime.now(datetime.timezone.utc).isoformat(sep=" ", timespec="seconds")
 
 
 class CallbackModule(CallbackBase):


ares:
Is aligned with changes in https://github.com/theforeman/foreman-ansible-modules/pull/1155, merging one of this will cause the need for rebase, I'm fine with the other one being merged first as this one is easier to rebase.

ares:
rebased, should be ready to get in

plugins/callback/foreman.py
```diff
@@ -132,8 +132,10 @@ def get_time():
 def get_now():
     """
     Return the current timestamp as a string to be sent over the network.
+    The time is always in UTC timezone and printed in a format Ruby
+    DateTime can easily parse.
     """
-    return datetime.utcnow().isoformat()
+    return datetime.datetime.now(datetime.timezone.utc).isoformat(sep=" ", timespec="seconds")
```
evgeni, line 138:
```
>>> datetime.datetime.now(datetime.timezone.utc).isoformat(sep=" ", timespec="seconds")
  File "<input>", line 1, in <module>
    datetime.datetime.now(datetime.timezone.utc).isoformat(sep=" ", timespec="seconds")
AttributeError: type object 'datetime.datetime' has no attribute 'datetime'
```

The `datetime` here is `datetime.datetime` already. You want to change the import if you need both `datetime.datetime` and `datetime.timezone`.

evgeni: The correct, version agnostic way to get the UTC time is `datetime.datetime.utcnow()`

Now I just need to understand why you set `sep` and `timespec` (as the later is not understood by Python 2)

evgeni:
#1333 was merged.

PR #1173, opened 2021-03-15
add user management role #2

IncredibleRichie:
#1083

### DIFF
--- a/roles/users/README.md
+++ b/roles/users/README.md
@@ -28,10 +28,10 @@ validate_certs: bool defualt (true)
 To switch on/off tasks use:
 
 ``` yaml
-create_users: bool defualt (true)
-create_usergroups: bool defualt (true)
-enable_ldap_auth: bool defualt (true)
-map_external_usergroups: bool defualt (true)
+create_users: bool default (true)
+create_usergroups: bool default (true)
+enable_ldap_auth: bool default (true)
+map_external_usergroups: bool default (true)
 ```
 
 For user creation use:
--- a/roles/users/README.md
+++ b/roles/users/README.md
@@ -6,38 +6,36 @@
 
 #### Tasks to switch on/off
 
-+ create_users
-+ create_usergroups
-+ enable_ldap_auth
-+ map_external_usergroups
++ foreman_create_users
++ foreman_create_usergroups
++ foreman_enable_ldap_auth
++ foreman_map_external_usergroups
 
 ### Role Variables
-
-#### Example values can be found in defaults/main.yaml
 
 Default:
 
 ``` yaml
-server_url: string(url) **required**
-username: string **required**
-password: string
-password: string
-validate_certs: bool defualt (true)
+foreman_server_url: string(url) **required**
+foreman_username: string **required**
+foreman_password: string
+foreman_password: string
+foreman_validate_certs: bool default (true)
 ```
 
 To switch on/off tasks use:
 
 ``` yaml
-create_users: bool default (true)
-create_usergroups: bool default (true)
-enable_ldap_auth: bool default (true)
-map_external_usergroups: bool default (true)
+foreman_create_users: bool default (true)
+foreman_create_usergroups: bool default (true)
+foreman_enable_ldap_auth: bool default (true)
+foreman_map_external_usergroups: bool default (true)
 ```
 
 For user creation use:
 
 ``` yaml
-users:
+foreman_users:
   - username: string **required**
     firstname: string
     lastname: string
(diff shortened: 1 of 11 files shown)


tests/test_playbooks/users_role.yml
```diff
@@ -0,0 +1,76 @@
+---
+# The following sensitive variables can be specified in the server.yml
+# ldap_host:
+# ldap_account:
+# ldap_password:
+# ldap_base_dn:
+# ldap_groups_base:
```
mdellweg: Not true anymore?

Can you add information about the steps that are necessary to run / record this test?

roles/users/README.md
```diff
@@ -0,0 +1,149 @@
...
+  - username: string **required**
+    firstname: string
+    lastname: string
+    email: string(email)
+    description: string
+    admin_rights: bool
+    user_password: string **required when internal**
+    default_location: string
+    default_organization: string
+    auth_source: sting **required**
```
mdellweg:
```suggestion
    auth_source: string **required**
```

roles/users/README.md
```diff
@@ -0,0 +1,149 @@
...
+validate_certs: bool defualt (true)
+```
+
+To switch on/off tasks use:
+
+``` yaml
+create_users: bool defualt (true)
+create_usergroups: bool defualt (true)
+enable_ldap_auth: bool defualt (true)
+map_external_usergroups: bool defualt (true)
```
mdellweg:
```suggestion
create_users: bool default (true)
create_usergroups: bool default (true)
enable_ldap_auth: bool default (true)
map_external_usergroups: bool default (true)
```

mdellweg: What is the usecase here? Would it make sense to just perform the actions if the corresponding data structures are not empty?

roles/users/README.md
```diff
@@ -0,0 +1,149 @@
...
+#### Example values can be found in defaults/main.yaml
+
+Default:
+
+``` yaml
+server_url: string(url) **required**
+username: string **required**
+password: string
+password: string
+validate_certs: bool defualt (true)
```
mdellweg: At least for this set of variables, there is a common text snippet used in the other roles, i think.

https://github.com/theforeman/foreman-ansible-modules/tree/develop/roles/manifest#role-variables

roles/users/README.md
```diff
@@ -0,0 +1,149 @@
...
+```
+
+To switch on/off tasks use:
+
+``` yaml
+create_users: bool defualt (true)
```
mdellweg: All top level vars need proper namespacing.
...even if it sounds silly.

tests/test_playbooks/users_role.yml
IncredibleRichie: I´ve added it to the server.example also. 
What do you mean exactly?

roles/users/README.md
IncredibleRichie: thx

roles/users/README.md
IncredibleRichie: I have a full playbook (vars) what describes my instance or in this case my user space, with the above vars in a playbook i can choose what to run without del/paste data.

tests/test_playbooks/users_role.yml
mdellweg: I think you mentioned you had a ldap container.
If that's something the developer needs to supply in order to record this test, it should be mentioned here.

Example:
https://github.com/theforeman/foreman-ansible-modules/blob/develop/tests/test_playbooks/realm.yml#L1

roles/users/README.md
mdellweg: I'm not sold on this. Do we have a similar thing in other roles?

mdellweg: To be a bit more verbose: I thought, the deal of this role is to define your instance and the role will do every thing needed. I don't understand, why i should specify everything to then not use part of it. I could still be convinced though.

Maybe as a guideline: Those roles fall into the 80/20 rule. If you need something really special, you can write your own. We want to serve the common use case.

tests/test_playbooks/users_role.yml
IncredibleRichie: Ah ok added some information.

roles/users/README.md
IncredibleRichie: Applied

roles/users/README.md
IncredibleRichie: Ok, i´ve changed them.

roles/users/README.md
IncredibleRichie: You dont have to specify anything, only the data for enabled tasks.

I think this happens very often as an admin. Example: 
If an admin already has the playbook written and wants to add an internal user without going to the whole process with ldap etc. he can only activate the create user part and run it. 
On the other side you can only create a auth source and link groups without create/update plenty users.

PR #1116, opened 2020-12-22
Add role for managing hostgroups

ehelms:
Putting into draft initially as I want to work on a few better examples. Ideas welcome on what we can show here.

PR #1174, opened 2021-03-16
Add ansible roles parameter to hostgroup module

### LINKED ISSUE #1123: [foreman-ansible-plugin] assign ansible roles to hostgroups
##### SUMMARY
I am about to create our hostgroups via these ansible-modules, but I can not find any way to assign ansible-roles to hostgroups/hosts. 
Searching around in community.foreman.com i found this question https://community.theforeman.org/t/no-api-available-for-attaching-ansible-roles-with-a-hostgroup-in-foreman-v2-1/21550
Unfortunately I have no experience with ruby-programming at all..

Maybe there are more plugin-features which would be nice to control via ansible, but this is what caught my eye.

Edit: In the first implementation there will be support for hostgroups. A new issue for hosts can be found at #1181 .

##### ISSUE TYPE
 - Feature Idea
 - Question

mdellweg:


mdellweg:
Some cool stuff is missing here i think.

- [x] guarding with installed ansible module
- [ ] looking into whether this works with hosts too (maybe not this PR)

plugins/modules/hostgroup.py
```diff
@@ -182,9 +190,25 @@ def main():
...
+            desired_ansible_role_ids = [item['id'] for item in ansible_roles]
+            current_ansible_role_ids = [
+                item['id'] for item in module.resource_action(
+                    'hostgroups', 'ansible_roles', {'id': entity['id']},
+                    ignore_check_mode=True, record_change=False,
+                )
+            ] if old_entity else []
+            if set(current_ansible_role_ids) != set(desired_ansible_role_ids):
+                module.resource_action(
+                    'hostgroups', 'assign_ansible_roles',
```
evgeni, line 211: will that also unassign roles that are not on that list?

mdellweg: There are only those two endpoints available (listing and setting).
But you are right (of course) and i should not just assume...

evgeni:
#1116 will need an update when this is merged

plugins/modules/hostgroup.py
```diff
@@ -55,6 +55,13 @@
   parameters:
     description:
       - Hostgroup specific host parameters
+  ansible_roles:
```
evgeni, line 58: Can you please add a "version_added: 2.1.0" to this section and also a changelog snippet? ;)

evgeni (changes requested):
ACK pending version_added/changelog

mdellweg:
The original request was written for host/hostgroup.
Should we just close the issue, rescope or split it when merging this?

changelogs/fragments/1123-ansible-roles-for-hostgroups.yml
```diff
@@ -0,0 +1,2 @@
+minor_changes:
+  - hostgroup - add a ``ansible_roles parameter`` (https://github.com/theforeman/foreman-ansible-modules/issues/1123)
```
evgeni:
```suggestion
  - hostgroup - add a ``ansible_roles`` parameter (https://github.com/theforeman/foreman-ansible-modules/issues/1123)
```

evgeni (approved):
tiny nitpick in the changelog

I'd probably just not close the original issue until the host support wasn't added too.

(or open a follow up, if you wish)

PR #1134, opened 2021-02-01
Content View Role

willtome:
Role for creating Content Views

Closes #1111

PR #1177, opened 2021-03-16
add general content-view management role

bmarlow:
add role for promoting, publishing and rolling back content-views


ehelms:
We also have this [content view role](https://github.com/theforeman/foreman-ansible-modules/pull/1134) under development. I see a lot of overlap here for arriving at a role that allows users to manage content views.

bmarlow:
> We also have this [content view role](https://github.com/theforeman/foreman-ansible-modules/pull/1134) under development. I see a lot of overlap here for arriving at a role that allows users to manage content views.

That role looks like it is for creating content-views and related.

This one is specifically around managing them and their lifecycles, whether that be promotion, publishing a new one, or rolling back a version.

One key difference is that this role is more around operationalization of the content-views versus creating them.  This view allows somebody to just set this on a schedule so that they can promote/publish (more than once) in an automated fashion, without having to keep editing their playbooks.

ehelms:
So, this is more of a content_view_version role? That could make sense to manage separately from the content view itself.

I'm not as clear on what rollback here means. Since all content view operations are simply promotions of some version to some environment. If for example I had version 10 in Test, and I wanted to version 8 instead. How would I go about that? 

I would think, treating entities as code, I would update my yaml from 10 to 8 and let Ansible run and adjust things for me. 

With the rollback, I assume I would have to run the playbook twice to go back two versions?

ehelms:
There are some variable naming updates that will be needed to ensure the role conforms to https://github.com/theforeman/foreman-ansible-modules#common-role-variables and attempts, where possible, to make use of existing variables defined for other roles such as lifecycle-environments (https://github.com/theforeman/foreman-ansible-modules/tree/develop/roles/lifecycle_environments).

bmarlow:
> So, this is more of a content_view_version role? That could make sense to manage separately from the content view itself.
> 
> I'm not as clear on what rollback here means. Since all content view operations are simply promotions of some version to some environment. If for example I had version 10 in Test, and I wanted to version 8 instead. How would I go about that?
> 
> I would think, treating entities as code, I would update my yaml from 10 to 8 and let Ansible run and adjust things for me.
> 
> With the rollback, I assume I would have to run the playbook twice to go back two versions?

content_view_version might be a more apt description of the role, truth be told I was struggling with what to name it.

The idea here is that the user doesn't need to know the content-view version, we will just loop over the dictionary and promote each of the lifecycle environments they want to N+1 of the content-view.  If N+1 means that they need to publish a new version of the content-view first, it will do that automatically.

Rollback will give N-1 content-view version for each of the defined lifecycle environments.  So yes, to roll back 2 versions you'd run it twice.

The number of versions to roll back could pretty easily get wrapped into a var in the main dictionary so that the roll back would do N-Y versions.

The main idea here is from customer experiences with patch management.  They otherwise don't want to affect the content-view other than publishing new versions, or promoting lifecycle environments (or the occasional rollback).  

The end state would be that the role gets called by AWX tower as a scheduled task that would align with a customers patching cadence/requirements (eg publish a new version every Tuesday, promote the dev LCE to the new version on Wednesday, so on and so forth).

This gives the customer the ability to do something of a 'set it and forget it' style, but at the same time getting the process to align with their business requirements.

roles/content_view_promotion_rollback_publish/tasks/promote.yml
```diff
@@ -0,0 +1,82 @@
...
+- name: "Gather Data For Current Content-View From Foreman"
+  theforeman.foreman.resource_info:
+    username: "{{ foreman.user }}"
+    password: "{{ foreman.password }}"
+    server_url: "{{ foreman.url }}"
+    organization: "{{ organization.key }}"
+    resource: content_views
+    search: name = "{{ content_view }}"
+    validate_certs: false
+  register: content_view_data
```
ehelms, line 12: You can use the dedicated info module now: https://github.com/theforeman/foreman-ansible-modules/blob/develop/plugins/modules/content_view_info.py

roles/content_view_promotion_rollback_publish/tasks/promote.yml
```diff
@@ -0,0 +1,82 @@
...
+- name: "Gather Data For Current Content-View Versions From Foreman"
+  theforman.foreman.resource_info:
+    username: "{{ foreman.user }}"
+    password: "{{ foreman.password }}"
+    server_url: "{{ foreman.url }}"
+    organization: "{{ organization.key }}"
+    resource: content_view_versions
+    params:
+      content_view_id: "{{ content_view_data.resources[0].id }}"
+  register: version_information
```
ehelms, line 25: There should be a dedicated module for this as well in the next few days: https://github.com/theforeman/foreman-ansible-modules/pull/1169

Could be good feedback before merging it whether it handles your use case.

roles/content_view_promotion_rollback_publish/tasks/promote.yml
```diff
@@ -0,0 +1,82 @@
...
+
+# get the highest version number for the content-view
+- name: "Get Highest Content-View Version That Isn't Library"
+  set_fact:
+    highest_version: "{{ item.value|int }}"
+  when: highest_version | default(0) | int < item.value|int and item.key != 'Library'
+  with_dict: "{{ environments }}"
+
+
+# only publish new view if necessary (when the current view is at or above library)
```
ehelms, line 57: I do not fully understand this logic. Or the circumstances under which it may occur.

roles/content_view_promotion_rollback_publish/vars/main.yml
```diff
@@ -0,0 +1,22 @@
+---
```
ehelms, line 1: Generally, I'd expect this file to be dropped from the final version given its specificity.

roles/content_view_promotion_rollback_publish/README.md
```diff
@@ -0,0 +1,86 @@
...
+```
+This is in the vars/main.yml for illistrative purposes only!  Please use a vault (or custom credential-type if using Tower).  STORING PASSWORDS IN PLAINTEXT IS BAD, MMM-KAY?
+
+
+Dependencies
+------------
+
+You need a Foreman user with admin access to the Organizations, Lifecycle_Environments, and Content_Views you wish to interact with.
+
+By default, the role will require a valid SSL certificate installed on your Foreman server that the ansible client can trace trust to.  To disable that update the 'FOREMAN_VALIDATE_CERTS variable in defaults/main.yml.'
```
ehelms: This is true of any of our roles. I'd promote this up to the README if its not already covered there.

roles/content_view_promotion_rollback_publish/README.md
```diff
@@ -0,0 +1,86 @@
...
+
+
+Example Playbook
```
ehelms, line 64: It would be helpful to see examples of the scenarios you outlined here -- https://github.com/theforeman/foreman-ansible-modules/pull/1177#issuecomment-800539612

roles/content_view_promotion_rollback_publish/README.md
```diff
@@ -0,0 +1,86 @@
...
+      - "content-view1"
+```
+Items not described in the inventory will not be affected (the exception being Library when a new content_view_version needs to be published.
+
+If you only want to use this role for staging new content_view_versions, simply set the variable **publish_only** to **True** (this can be done in vars/main.yml or elsewhere).  This action will only publish a new content_view_version to the content_views described in the dictionary.
+
+If you want to rollback (reverse promote) lifecycle environments simply set the variable **rollback** to **True** (this can be done in vars/main.yml or elsewhere).  This action will only roll back lifecycle_environments in the content views described in main dictionary.
+
+
+Foreman information is accessed by way of the dictionary 'foreman', formatted as such:
```
ehelms: I think this whole block of text down to line 62 can be dropped. This should be covered by the collections README.

roles/content_view_promotion_rollback_publish/README.md
```diff
@@ -0,0 +1,86 @@
...
+    lifecycle_environments:
+      - "Prod"
+    content_views:
+      - "content-view1"
+```
+Items not described in the inventory will not be affected (the exception being Library when a new content_view_version needs to be published.
+
+If you only want to use this role for staging new content_view_versions, simply set the variable **publish_only** to **True** (this can be done in vars/main.yml or elsewhere).  This action will only publish a new content_view_version to the content_views described in the dictionary.
+
+If you want to rollback (reverse promote) lifecycle environments simply set the variable **rollback** to **True** (this can be done in vars/main.yml or elsewhere).  This action will only roll back lifecycle_environments in the content views described in main dictionary.
```
ehelms:
```suggestion
If you want to rollback (reverse promote) lifecycle environments simply set the variable **rollback** to **True**.  This action will only roll back lifecycle_environments in the content views described in main dictionary.
```

I think this line could be dropped as users should not be editing the vars file inside the role. Same in the line above.

roles/content_view_promotion_rollback_publish/README.md
```diff
@@ -0,0 +1,86 @@
...
+For example, if the previously mentioned dictionary describes *ALL* of my Foreman environment, but I only want to promote the 'Prod' lifecycle_environment in the content-view 'content-view1' in organization 'org', my dictionary would look like this:
+```
+organizations:
+  org1:
+    lifecycle_environments:
+      - "Prod"
+    content_views:
+      - "content-view1"
+```
+Items not described in the inventory will not be affected (the exception being Library when a new content_view_version needs to be published.
```
ehelms:
```suggestion
Items not described in the inventory will not be affected (the exception being Library when a new content_view_version needs to be published).
```

roles/content_view_promotion_rollback_publish/README.md
```diff
@@ -0,0 +1,86 @@
...
+
+Requirements
+------------
+
+This role requires the theforeman.foreman module collection.
+
+Role Variables
+--------------
+
+The main amount of work is vars/main.yml.
```
ehelms: I think this should be dropped as it appears to encourage users to edit this file.

roles/content_view_promotion_rollback_publish/tasks/promote.yml
bmarlow: If your content-view's highest version number is 22, and that is also where your life-cycle environment is then we need to publish a new version prior to promoting- since you cannot promote to version 22+1 if version 23 doesn't exist.

roles/content_view_promotion_rollback_publish/README.md
bmarlow: removed it.

roles/content_view_promotion_rollback_publish/README.md
bmarlow: Done

roles/content_view_promotion_rollback_publish/README.md
bmarlow: Great idea, I have added some in the latest commit.

roles/content_view_promotion_rollback_publish/README.md
bmarlow: Gotcha, I have dropped those.

roles/content_view_promotion_rollback_publish/tasks/promote.yml
bmarlow: Does it make sense to have roles that utilize modules that aren't even in a main version yet?  By using the existing format, this role will be backwards compatible with older versions of the collection.

roles/content_view_promotion_rollback_publish/tasks/promote.yml
bmarlow: Please see comment above.

roles/content_view_promotion_rollback_publish/tasks/promote.yml
wbclark: Chiming in here, I think it does make sense. The primary way we expect users would consume this role would be to install the collection.

If they install the development branch from git they would have access to all of the latest modules. Of course it is more likely that they install one of the released versions from ansible galaxy, or a downstream RPM packaged version.... in each case still by the time this role is included in any of those releases, so too would be all modules currently present at development time.

Is this contrary to your expectation, @bmarlow ?

bmarlow: @wbclark I think its fair to assume people are going to get this by way of the collection so no need to worry there.

I guess my statement is more that I don't think it makes sense to build a role against a module that is in development as changes that happen to that module could potentially break the role.

As it sits (and based on the way that roles are consumed/delivered), I would think we wouldn't care which module got used as long as it worked (unless there was a significant performance benefit or depreciation issue) due to the fact that the user shouldn't really be modifying the role, but rather changing the variables that the role digests.

I haven't had a chance to use the new module yet, however my biggest concern would be that the data returned isn't structured the same way (maybe it is?) and it would require more significant refactoring (and lets face it, the data returned isn't the most easily parsable.

ehelms:
> content_view_version might be a more apt description of the role, truth be told I was struggling with what to name it.
> 
> The idea here is that the user doesn't need to know the content-view version, we will just loop over the dictionary and promote each of the lifecycle environments they want to N+1 of the content-view. If N+1 means that they need to publish a new version of the content-view first, it will do that automatically.

We discussed this PR during our triage this week (anybody is welcome to join btw) and arrived at the following thoughts on how to proceed.

There seems to be three primary functions contained in this that could be approached as independent roles:

 * a role to publish content views
 * a role to promote latest content views (perhaps a generic promote role with 'latest' as an option?)
 * a role to rollback content view versions (perhaps this could be a parameter to the promote)

We realize this creates additional work for additional roles but building re-usable roles would be beneficial to users and then also be consumed by this specific workflow. For example, https://github.com/theforeman/foreman-ansible-modules/issues/1209

I can also dedicate some time to help get these available.
> 
> Rollback will give N-1 content-view version for each of the defined lifecycle environments. So yes, to roll back 2 versions you'd run it twice.
> 
> The number of versions to roll back could pretty easily get wrapped into a var in the main dictionary so that the roll back would do N-Y versions.
> 
> The main idea here is from customer experiences with patch management. They otherwise don't want to affect the content-view other than publishing new versions, or promoting lifecycle environments (or the occasional rollback).
> 
> The end state would be that the role gets called by AWX tower as a scheduled task that would align with a customers patching cadence/requirements (eg publish a new version every Tuesday, promote the dev LCE to the new version on Wednesday, so on and so forth).
> 
> This gives the customer the ability to do something of a 'set it and forget it' style, but at the same time getting the process to align with their business requirements.

bmarlow:
@ehelms 

Just for clarification sake- is the intent here that a role _should_ only have one purpose and be as singular as possible?

For the record, the currently proposed role can do all, or just one of the things listed in the three proposed new roles.

I can refactor if that is the desire of the team here, just looking for a little clarification with regards to role structures.

bmarlow:
per maintainers request this has been replaced by PRs #1214 #1216 #1217

PR #1179, opened 2021-03-18
Fixes #31841 - fix bug in assigning error log level in callback report

yifatmakias:


plugins/callback/foreman.py
```diff
@@ -179,12 +179,15 @@ def _build_log(self, data):
         logs = []
         for entry in data:
             source, msg = entry
-            if 'failed' in msg:
+            if 'failed' in msg and msg['failed']:
                 level = 'err'
             elif 'changed' in msg and msg['changed']:
```
mdellweg:
```suggestion
            if msg.get('failed'):
                level = 'err'
            elif msg.get('changed'):
```

mdellweg (approved):
Looks good to me.

xprazak2:
Works, the report now shows correct count of failed tasks:
failed-count

There is a correct count in graphs as well, only in table the task is labeled as 'notice':

failed-task

So I am wondering whether any more updates are needed here or if it is something we should handle in foreman_ansible.

yifatmakias:
> Works, the report now shows correct count of failed tasks:
> failed-count
> 
> There is a correct count in graphs as well, only in table the task is labeled as 'notice':
> 
> failed-task
> 
> So I am wondering whether any more updates are needed here or if it is something we should handle in foreman_ansible.

@xprazak2 should be fixed now, Let me know if you find more issues.

plugins/callback/foreman.py
```diff
@@ -223,6 +223,32 @@ def send_facts(self):
                 self._display.warning(u'Sending facts to Foreman at {url} failed for {host}: {err}'.format(
                     host=to_text(host), err=to_text(err), url=to_text(self.foreman_url)))
 
+    def _build_log(self, data):
```
xprazak2: I think the method changed and it is now [build_log](https://github.com/theforeman/foreman-ansible-modules/blob/develop/plugins/callback/foreman.py#L95)

xprazak2:
It now works for 'command' as expected, unfortunately [failure](https://github.com/xprazak2/fail_role/blob/master/tasks/main.yml) in 'file' module is still only info. This makes me thing that maybe we need something more robust than scanning messages for failures as different modules do not have consistent output for errors.

file-module

yifatmakias:
> It now works for 'command' as expected, unfortunately [failure](https://github.com/xprazak2/fail_role/blob/master/tasks/main.yml) in 'file' module is still only info. This makes me thing that maybe we need something more robust than scanning messages for failures as different modules do not have consistent output for errors.
> 
> file-module

@xprazak2 I agree, Do you maybe have an idea how to deal with this? Maybe it is worth discussion int the next ansible meeting

plugins/callback/foreman.py
```diff
@@ -98,12 +98,15 @@ def build_log(data):
     config_report API.
     """
     for source, msg in data:
-        if 'failed' in msg:
+        if msg.get('failed') or msg.get('stderr_lines') or (msg.get('msg') and 'Err' in msg.get('msg')):
```
ezr-ondrej: - the stderr_lines can exist and not mean failure IMO :thinking: take wget, it writes in stderr just because stdout is for the download result.
- the last bit seems quite arbitrary and could be possibly find even in success.
- could we extract it in a method? It feels already quite complicated

ezr-ondrej (commented):
Just ideas, I'm not sure if it can be true for any modules, so feel free to ignore me :)

plugins/callback/foreman.py
ares: I assume ansible must somehow keep the status in the task object represented, it's stdout also shows green/red based on the status. Could it be found somewhere in the `source`? Or do the logs from `data` have more information?

yifatmakias: @ares 
I think there is no details about the task in the source or data but I will check again to make sure.
@ezr-ondrej 
I agree but there are errors that I managed to catch only this way

evgeni:
@ezr-ondrej @xprazak2 @ares any last comments or is that good to merge?

xprazak2 (approved):
Works as expected now, well done @yifatmakias!

PR #1183, opened 2021-03-23
Do not add certs for http requests

xprazak2:
When we want to do http requests,
we still add cert paths into requests.
With the following config:

[callback_foreman]
url='http://centos-devel.example.com'
verify_certs=false

cert paths will contain default values
and requests will fail if there are no certs
in default paths. But we do not care about
certs since we are doing http.

### DIFF
--- a/plugins/callback/foreman.py
+++ b/plugins/callback/foreman.py
@@ -166,6 +166,8 @@ class CallbackModule(CallbackBase):
         else:
             self._disable_plugin(u'The `requests` python module is not installed')
 
+        self.session = requests.Session()
+
         if self.foreman_url.startswith('https://'):
             if not os.path.exists(ssl_cert):
                 self._disable_plugin(u'FOREMAN_SSL_CERT %s not found.' % ssl_cert)
@@ -173,9 +175,8 @@ class CallbackModule(CallbackBase):
             if not os.path.exists(ssl_key):
                 self._disable_plugin(u'FOREMAN_SSL_KEY %s not found.' % ssl_key)
 
-        self.session = requests.Session()
-        self.session.verify = self._ssl_verify(str(self.get_option('verify_certs')))
-        self.session.cert = (ssl_cert, ssl_key)
+            self.session.verify = self._ssl_verify(str(self.get_option('verify_certs')))
+            self.session.cert = (ssl_cert, ssl_key)
 
     def _disable_plugin(self, msg):
         self.disabled = True


wbclark:
Re-running tests here as I expect the issue with galaxy.ansible.com was transient

yifatmakias (approved):
Thanks, @xprazak2. Looks good and works for me.

plugins/callback/foreman.py
```diff
@@ -173,9 +173,10 @@ def set_options(self, task_keys=None, var_options=None, direct=None):
             if not os.path.exists(ssl_key):
                 self._disable_plugin(u'FOREMAN_SSL_KEY %s not found.' % ssl_key)
 
+            self.session.cert = (ssl_cert, ssl_key)
```
evgeni, line 179: self.session does not exist at this point.

evgeni (changes requested):

wbclark:
Could we also update the commit message to be more clear that this is about the callback plugin specifically?

xprazak2:
Updated, I also modified the commit message.

plugins/callback/foreman.py
```diff
@@ -166,16 +166,18 @@ def set_options(self, task_keys=None, var_options=None, direct=None):
         else:
             self._disable_plugin(u'The `requests` python module is not installed')
 
+        self.session = requests.Session()
+        self.session.verify = self._ssl_verify(str(self.get_option('verify_certs')))
```
evgeni: should this also be in the `if https` part?

evgeni: (I have no idea what the default value of verify is…)

xprazak2: Default is 1, can be easily changed via config, moved as verify does not make much sense when using http anyway.

evgeni (approved):

plugins/callback/foreman.py
```diff
@@ -166,16 +166,18 @@ def set_options(self, task_keys=None, var_options=None, direct=None):
...
 
             if not os.path.exists(ssl_key):
                 self._disable_plugin(u'FOREMAN_SSL_KEY %s not found.' % ssl_key)
 
-        self.session = requests.Session()
-        self.session.verify = self._ssl_verify(str(self.get_option('verify_certs')))
-        self.session.cert = (ssl_cert, ssl_key)
+            self.session.verify = self._ssl_verify(str(self.get_option('verify_certs')))
+            self.session.cert = (ssl_cert, ssl_key)
+
```
evgeni, line 180: ERROR: plugins/callback/foreman.py:182:5: E303: too many blank lines (2)

;)

evgeni (approved):

PR #1184, opened 2021-03-24
 Set validate_certs to API call to avoid effects from environment vars

saito-hideki:
# Summary
If `REQUESTS_CA_BUNDLE` or `CURL_CA_BUNDLE` has been set as an environment variables, `requests` will use those variables with API call even if `validate_certs` set `False`.

This PR addresses the above issue.

### DIFF
--- a/plugins/inventory/foreman.py
+++ b/plugins/inventory/foreman.py
@@ -237,7 +237,11 @@ class InventoryModule(BaseInventoryPlugin, Cacheable, Constructable):
             params['page'] = 1
             params['per_page'] = self.get_option('batch_size')
             while True:
-                ret = s.get(url, params=params)
+                # workaround to address the follwing issues where 'verify' is overridden in Requests:
+                #   - https://github.com/psf/requests/issues/3829
+                #   - https://github.com/psf/requests/issues/5209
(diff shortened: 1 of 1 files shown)


evgeni:
I would have expected https://github.com/theforeman/foreman-ansible-modules/blob/1bd19e528d547e861ac4cf12d9e7e43653d72469/plugins/inventory/foreman.py#L223 to be sufficient?

What would I need to reproduce this? Just point `REQUESTS_CA_BUNDLE` env var at some CA that hasn't singed my Foreman?

evgeni:
indeed, the following script allows to repro:

```
import requests

s = requests.Session()
s.verify = False

s.get('https://somehost.example.com/')
```

evgeni:
https://github.com/psf/requests/issues/5209 / https://github.com/psf/requests/issues/3829 is a known issue

@saito-hideki would you mind adding a reference to the requests issue as a comment to the code? it's good to be merged then!

saito-hideki:
@evgeni thank you for your review.  I have added references about Requests issues. please let me know if I have missed something.

evgeni (approved):

PR #1185, opened 2021-03-24
add bmc_proxy param to subnet

evgeni:
also clarify that subnet DNS proxies are about *reverse* DNS

### DIFF
--- a/plugins/modules/subnet.py
+++ b/plugins/modules/subnet.py
@@ -118,7 +118,7 @@ options:
     required: false
     type: str
   dns_proxy:
-    description: DNS Smart proxy for this subnet
+    description: Reverse DNS Smart proxy for this subnet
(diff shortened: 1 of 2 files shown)


plugins/modules/subnet.py
```diff
@@ -118,13 +118,17 @@
...
     required: false
     type: str
   template_proxy:
     description: Template Smart proxy for this subnet
     required: false
     type: str
+  bmc_proxy:
+    description: BMC Smart proxy for this subnet
+    required: false
+    type: str
```
mdellweg, line 131: "added_in"? Or what was it?

evgeni: version_added, yes. thanks

mdellweg (approved):

PR #1189, opened 2021-04-01
Add repository_set_info module

wbclark:


plugins/modules/repository_set_info.py
```diff
@@ -0,0 +1,166 @@
...
+      - Find all repositories in the repository set
+      - Required when I(repositories) is unset or an empty list
+    required: false
+    type: bool
+  enabled:
+    description:
+      - When true, only return repository sets which are already enabled. Defaults to false.
+    required: false
+    default: false
+    type: bool
```
wbclark: I'm not quite sure yet whether this will work but I thought it would be interesting to give it a try.

(As of today) Katello supports in scoped search: https://github.com/Katello/katello/blob/a41851ba5600cc8526edbfade6dae9058861393d/app/models/katello/product_content.rb#L22

Which initially looked promising for doing something like `search: 'enabled = true'`, BUT on closer inspection that seems to have `:rename => :enabled_by_default` which seems to be not what I am looking for.

However this looks promising: https://github.com/Katello/katello/blob/a41851ba5600cc8526edbfade6dae9058861393d/app/controllers/katello/api/v2/repository_sets_controller.rb#L24

Which leads me to believe my implementation here could work? I will write some tests soon and report back.

plugins/modules/repository_set_info.py
```diff
@@ -0,0 +1,166 @@
...
+        ],
+        required_if=[
+            ['all_repositories', False, ['repositories']],
+            ['repositories', [], ['all_repositories']],
+        ],
+    )
+
+    # KatelloInfoAnsibleModule automatically adds organization to the entity scope
+    # but repository sets are scoped by product (and these are org scoped)
+    module.foreman_spec['entity']['scope'].remove('organization')
```
wbclark, line 94: I copied this over from the `repository_info` module but I *think* it applies here as well. I will have a better idea after writing some tests.

wbclark:
Added tests, they are failing at the moment.

This also includes https://github.com/theforeman/foreman-ansible-modules/pull/1197 which is necessary for the tests, but still isn't working correctly. I can rebase to drop once that is worked out and merged.

ehelms:
Options and structurally this looks good to me.

plugins/modules/repository_set_info.py
```diff
@@ -0,0 +1,166 @@
...
+# along with this program.  If not, see <http://www.gnu.org/licenses/>.
+
+from __future__ import absolute_import, division, print_function
+__metaclass__ = type
+
+
+DOCUMENTATION = '''
+---
+module: repository_set_info
+version_added: 2.0.0
```
evgeni:
```suggestion
version_added: 2.1.0
```

plugins/modules/repository_set_info.py
```diff
@@ -0,0 +1,166 @@
...
+        type: str
+  all_repositories:
+    description:
+      - Find all repositories in the repository set
+      - Required when I(repositories) is unset or an empty list
+    required: false
+    type: bool
+  enabled:
+    description:
+      - When true, only return repository sets which are already enabled. Defaults to false.
```
evgeni:
```suggestion
      - When true, only return repository sets which are already enabled.
```

We have `default: false` below already.

tests/test_playbooks/repository_set_info.yml
```diff
@@ -0,0 +1,71 @@
...
+  gather_facts: false
+  vars_files:
+    - vars/server.yml
+  roles:
+    - role: content_rhel
+      vars:
+        foreman_organization: "Test Organization"
+        foreman_validate_certs: false
+        foreman_manifest_download: false
+        foreman_manifest_path: "{{ subscription_manifest_path }}"
```
evgeni, line 27: Where is this variable coming from? :)

plugins/modules/repository_set_info.py
```diff
@@ -0,0 +1,166 @@
...
+    module = KatelloRepositorySetInfo(
+        foreman_spec=dict(
+            product=dict(type='entity', scope=['organization']),
+            name=dict()
+            label=dict()
+            repositories=dict(type='list', elements='dict', options=dict(
+                basearch=dict(),
+                releasever=dict(),
+            )),
+            all_repositories=dict(type='bool'),
```
evgeni: I don't see any code using `repositories` and/or `all_repositories` and would argue that these are totally unnecessary in this info module, as as a user, I want to fetch information about the *whole* repo set, and see which repos of it are enabled and which not.

plugins/modules/repository_set_info.py
```diff
@@ -0,0 +1,166 @@
...
+    required: false
+    type: str
+  product:
+    description:
+      - Name of the parent product
+    required: false
+    type: str
+  label:
+    description:
+      - Label of the repository set. Can be used in place of I(name) & I(product)
```
evgeni: That's not what will happen with the current code:

https://github.com/theforeman/foreman-ansible-modules/blob/ab4faa1cdc2f56cf5eca6b27388018563b848156/plugins/module_utils/foreman_helper.py#L1365-L1369

You'll get *all* reposets as you didn't specify a name.

evgeni (changes requested):
see previous comments

plugins/modules/repository_set_info.py
wbclark: OK, I fixed this by using e.g. `search: 'label = "rhel-7-server-rpms"'` instead to find repository_sets by the label

tests/test_playbooks/repository_set_info.yml
wbclark: `foreman_manifest_path` is expected by the manifest role at https://github.com/theforeman/foreman-ansible-modules/blob/292dc967ee203660c4628cfcf3edf12fc0278590/roles/manifest/tasks/main.yml#L17

`subscription_manifest_path` comes from my `tests/test_playbooks/vars/server.yml` which is based on the template at https://github.com/theforeman/foreman-ansible-modules/blob/292dc967ee203660c4628cfcf3edf12fc0278590/tests/test_playbooks/vars/server.yml.example#L17

This seems to work fine and is consistent with what I've done for other tests in the past. Are you suggesting there is a reason to take a different approach?

wbclark:
This is now working the way that I expect it should. Please let me know if there is any further feedback -- thanks!

plugins/modules/repository_set_info.py
```diff
@@ -0,0 +1,103 @@
...
+
+class KatelloRepositorySetInfo(KatelloInfoAnsibleModule):
+    pass
+
+
+def main():
+    module = KatelloRepositorySetInfo(
+        foreman_spec=dict(
+            product=dict(type='entity', scope=['organization']),
+            name=dict(),
```
evgeni:
```suggestion
```

That's part of the base class, no need to define it here.

plugins/modules/repository_set_info.py
```diff
@@ -0,0 +1,103 @@
...
+short_description: Fetch information about Red Hat Repositories
+description:
+  - Fetch information about Red Hat Repositories
+author: "William Bradford Clark (@wbclark)"
+options:
+  name:
+    description:
+      - Name of the repository set
+    required: false
+    type: str
```
evgeni: that's part of the base `infomodule` doc, so you don't *need* to specify it (unless you want to override something):

https://github.com/theforeman/foreman-ansible-modules/blob/292dc967ee203660c4628cfcf3edf12fc0278590/plugins/doc_fragments/foreman.py#L331-L336

tests/test_playbooks/repository_set_info.yml
evgeni: No, I was just wondering that this is actually defined in our example XD (which you have shown is, so I am silent)

plugins/modules/repository_set_info.py
```diff
@@ -0,0 +1,103 @@
...
+    organization: "Default Organization"
+    search: 'label = "rhel-7-server-rpms"'
+'''
+
+RETURN = '''
+repository_sets:
+  description: List of repository sets.
+  returned: success
+  type: list
+  elements: dict
```
evgeni:
```suggestion
repository_set:
  description: Details about the found Red Hat Repository
  returned: success and I(name) was passed
  type: dict
repository_sets:
  description: List of all found Red Hat Repositories and their details
  returned: success and I(search) was passed
  type: list
  elements: dict
```

We return different things based on "show me one" and "get me the list" (as you see in your tests ;))

evgeni (changes requested):
tiny docs nits, otherwise LGTM!

wbclark:
Thanks @evgeni . I've updated with all recommendations. I appreciate the tips!

evgeni (approved):

PR #1190, opened 2021-04-04
Fixes #32124 - add check mode detail to callback report

yifatmakias:
Added check mode detail to the foreman.py callback and also make the tasks in pending status whenever the check mode is enabled.

### DIFF
--- a/plugins/callback/foreman.py
+++ b/plugins/callback/foreman.py
@@ -246,8 +246,12 @@ class CallbackModule(CallbackBase):
                     },
                     "logs": list(build_log(self.items[host])),
                     "reporter": "ansible",
+                    "check_mode": self.check_mode,
                 }
             }
+            if self.check_mode:
+                report['config_report']['status']['pending'] = total['changed']
+                report['config_report']['status']['applied'] = 0
             try:
                 response = self.session.post(url=url, json=report)
                 response.raise_for_status()
@@ -263,6 +267,7 @@ class CallbackModule(CallbackBase):
         value = result._result
         value['failed'] = failed
         self.items[host].append((name, value))
+        self.check_mode = result._task.check_mode
         if 'ansible_facts' in value:
             self.facts[host].update(value['ansible_facts'])
 


plugins/callback/foreman.py
```diff
@@ -249,8 +262,11 @@ def send_reports(self, stats):
                     },
                     "logs": list(build_log(self.items[host])),
                     "reporter": "ansible",
+                    "check_mode": self.check_mode,
                 }
             }
+            if self.check_mode == '1':
+                report['config_report']['status']['pending'] = total['ok']
```
ares: should we also se ok to 0? otherwise we'd double count tasks or not?

yifatmakias: Yes, I think you are right, I will add it.

yifatmakias: I am not sure this is necessary because there is no other use of the 'ok' tasks. the status focuses on the changed, failed, and skipped tasks only.

evgeni:
is that an API *ansible* provides/sets, or is that something that foreman_ansible "invented"?

xprazak2:
The data sent back is 'invented' by foreman_ansible.

plugins/callback/foreman.py
evgeni: "ok" sounds like the ones that are unchanged? and you want "changed"?

evgeni (changes requested):
Okay, after reading https://projects.theforeman.org/issues/32124 I still have *no idea* what this is supposed to do, so I am going to guess (based on https://github.com/theforeman/foreman_ansible/commit/e652ae2c5bc4015897d2ba1fb89ec41d9e22012c) , and you can yell at me if I'm wrong:

You want to run a playbook from `foreman_ansible` with `--check` and be able to visualize this in the report view.
To do so, you need the callback to report the "in checkmode" fact, as otherwise the config reports don't know that (as the callback is the only data source for them).
And in this PR this desire is implemented by a new parameter to the callback, which `foreman_ansible` will set when running Ansible.

I think this approach is wrong. Besides the fact that this is an invented custom interface, callbacks already know when tasks/plays run in checkmode. See for example the [`default`](https://github.com/ansible/ansible/blob/devel/lib/ansible/plugins/callback/default.py) callback from Ansible, which looks for [`play.check_mode`](https://github.com/ansible/ansible/blob/126e9244c2bbe12234d0cbd98c0802444ab103f5/lib/ansible/plugins/callback/default.py#L248-L253), [`result._task.check_mode`](https://github.com/ansible/ansible/blob/devel/lib/ansible/plugins/callback/default.py#L227-L230) and [`context.CLIARGS['check']`](https://github.com/ansible/ansible/blob/126e9244c2bbe12234d0cbd98c0802444ab103f5/lib/ansible/plugins/callback/default.py#L419-L420) to determine exactly that.

Please use the existing API instead of inventing your own.

plugins/callback/foreman.py
yifatmakias: @evgeni Thanks I did not know this information already exists, I will use this.
regarding the changed vs ok I think that is what we discussed in the ansible meeting. @ares @ezr-ondrej do you think the pending should be the changed tasks or all the ok tasks?

xprazak2: Pending should be the tasks that would have been changed without check mode I think.

ezr-ondrej: yeah, `changed` should mean pending, unchanged, should probably stay untouched, as those are not pending changes.

yifatmakias: Ok, I will fix this

xprazak2:
@evgeni, any additional concerns?

plugins/callback/foreman.py
```diff
@@ -264,7 +268,13 @@ def append_result(self, result):
         name = result._task.get_name()
         host = result._host.get_name()
         value = result._result
+        check_mode = result._task.check_mode
         self.items[host].append((name, value))
+        if check_mode:
+            self.check_mode = '1'
+            self._display.warning(u'Task run in check mode.')
```
evgeni: That will produce a warning for every task executed? Why?

evgeni (commented):
Yes, more concerns ;)

plugins/callback/foreman.py
yifatmakias: It is supposed to produce a warning only when the task run in check mode.
If a task runs in check mode, we would want to notify the user with a warning but if you think it is redundant, I can remove it.

evgeni: I'd argue the users knows: they asked it to be run in check mode.

Plus I'd think foreman visualizes that fact somehow in the UI, now that it knows?

yifatmakias: Ok, I will remove it.

plugins/callback/foreman.py
```diff
@@ -262,7 +266,9 @@ def append_result(self, result, failed=False):
         host = result._host.get_name()
         value = result._result
         value['failed'] = failed
+        check_mode = result._task.check_mode
         self.items[host].append((name, value))
+        self.check_mode = '1' if check_mode else '0'
```
mdellweg: Interleaving the code here is kind of obfuscating.
```suggestion
        self.items[host].append((name, value))
        self.check_mode = '1' if result._task.check_mode else '0'
```

yifatmakias: I agree, changed it. Thanks :)

evgeni:
I wonder if `self.check_mode` should be a boolean, not an int, but ultimately don't mind too much.

yifatmakias:
> I wonder if `self.check_mode` should be a boolean, not an int, but ultimately don't mind too much.

Thanks @evgeni I changed it to a boolean.

plugins/callback/foreman.py
```diff
@@ -263,6 +267,7 @@ def append_result(self, result, failed=False):
         value = result._result
         value['failed'] = failed
         self.items[host].append((name, value))
+        self.check_mode = True if result._task.check_mode else False
```
evgeni: isn't that the same as
```suggestion
        self.check_mode = result._task.check_mode
```

Or can `task.check_mode` be anything else? (I don't know)

mdellweg: In that case `bool(result._task.check_mode)` should do the trick.

yifatmakias: You are definitely right, I fixed it.

evgeni (approved):

evgeni:
@yifatmakias @xprazak2 @ezr-ondrej is there anything that needs tested in terms of "playing together with foreman_ansible" from the most recent changes Yifat did? Otherwise :shipit:

yifatmakias:
> @yifatmakias @xprazak2 @ezr-ondrej is there anything that needs tested in terms of "playing together with foreman_ansible" from the most recent changes Yifat did? Otherwise :shipit:

This PR: https://github.com/theforeman/foreman_ansible/pull/402# is waiting for the callback change to be merged. so this should be tested together. I tested it together of course.

PR #1197, opened 2021-04-12
content_rhel role: add products to sync plan only when enabled

wbclark:
`foreman_content_rhel_enable_rhel8: false` leads to an error message:         `"displayMessage": "Validation failed: Can not add product Red Hat Enterprise Linux for x86_64 because it is disabled.",`

### DIFF
--- a/roles/content_rhel/tasks/main.yml
+++ b/roles/content_rhel/tasks/main.yml
@@ -64,8 +64,8 @@
         cron_expression: "{{ foreman_sync_plan_cron_expression | default(omit) }}"
         sync_date: "{{ foreman_sync_plan_sync_date | default('2020-01-01 00:00:00 UTC') }}"
         products:
-          - Red Hat Enterprise Linux Server
-          - Red Hat Enterprise Linux for x86_64
+          - "{{ foreman_content_rhel_enable_rhel7 | ternary('Red Hat Enterprise Linux Server', omit) }}"
+          - "{{ foreman_content_rhel_enable_rhel8 | ternary('Red Hat Enterprise Linux for x86_64', omit) }}"
 
 - name: "Create Activation Key"
   include_role:


ehelms (approved):

wbclark:
Heads up @ehelms , I don't believe this works in the form that was merged. I had a note about that on the other PR but should have had this one marked as a draft with an explanation as well. My apologies for that.

wbclark:
Or maybe I am not remembering correctly where I left off, since I previously had this as an empty string (didn't work), a `select()` filter (also didn't work), before making the change to `omit` (don't recall if I fully tested it yet).

I will confirm shortly whether there is still an issue.

ehelms:
@wbclark shall we then open a PR to revert?

wbclark:
@ehelms technically it doesn't break anything that wasn't already broken, but yeah I confirmed that this still doesn't work:

```
failed: [localhost] (item={'name': 'RHEL Sync Plan', 'interval': 'daily', 'cron_expression': '__omit_place_holder__2b0d5eb1f19f9e2fcc13d731ca4aad1402a0f433', 'sync_date': '2020-01-01 00:00:00 UTC', 'products': ['Red Hat Enterprise Linux Server', '__omit_place_holder__2b0d5eb1f19f9e2fcc13d731ca4aad1402a0f433']}) => {
    "ansible_loop_var": "item",
    "changed": false,
    "invocation": {
        "module_args": {
            "cron_expression": null,
            "description": null,
            "enabled": true,
            "interval": "daily",
            "name": "RHEL Sync Plan",
            "organization": "Test Organization",
            "products": [
                "Red Hat Enterprise Linux Server",
                "__omit_place_holder__2b0d5eb1f19f9e2fcc13d731ca4aad1402a0f433"
            ],
            "server_url": "https://katello.example.com",
            "state": "present",
            "sync_date": "2020-01-01 00:00:00 UTC",
            "username": "myuser",
            "validate_certs": false
        }
    },
    "item": {
        "cron_expression": "__omit_place_holder__2b0d5eb1f19f9e2fcc13d731ca4aad1402a0f433",
        "interval": "daily",
        "name": "RHEL Sync Plan",
        "products": [
            "Red Hat Enterprise Linux Server",
            "__omit_place_holder__2b0d5eb1f19f9e2fcc13d731ca4aad1402a0f433"
        ],
        "sync_date": "2020-01-01 00:00:00 UTC"
    },
    "msg": "Found no results while searching for products with name=\"__omit_place_holder__2b0d5eb1f19f9e2fcc13d731ca4aad1402a0f433\""
}
```

In that case, Is a PR to revert the commit still the best approach when we didn't break anything, we only failed to fix the thing that was already broken?

PR #1201, opened 2021-04-13
Add a setting_info module

ehelms:


mdellweg (approved):
This all looks very good to me. The only thing i do not understand, is why is the `tests/test_playbooks/fixtures/setting_info-0.yml` so humongous.
Did the we-provide-the-api-docs-as-fixture stop to work?

ehelms:
> This all looks very good to me. The only thing i do not understand, is why is the `tests/test_playbooks/fixtures/setting_info-0.yml` so humongous.
> Did the we-provide-the-api-docs-as-fixture stop to work?

That fixture does seem to include:

```
    method: GET
    uri: https://foreman.example.org/apidoc/v2.json
```

Which is unexpected right? It should not need to fetch the apidoc?

mdellweg:
Yes, that is what i tried to get at.

evgeni:
Yeah, I think this happens if the apidoc of the machine you're recording against doesn't match the apidoc in the fixtures.

A workaround is to add a `hosts: localhost` step in the playbook, so the cache is pre-warmed. Allmost all other test playbooks do that anyways as they need a "setup" step, but settings do not ;)

ehelms:
Thanks for pointing that out @mdellweg and the key to the answer @evgeni. Things look much cleaner now

evgeni:
You don't need a changelog, it's autogenerated for new modules

ehelms:
> You don't need a changelog, it's autogenerated for new modules

How do I get this for roles? :)

evgeni:
> > You don't need a changelog, it's autogenerated for new modules
> 
> How do I get this for roles? :)

https://github.com/ansible-community/antsibull-changelog/pulls

evgeni (approved):

PR #1205, opened 2021-04-22
Drop requests version check in foreman callback

adamruzicka:
Before this commit we briefly required to have requests >=2.14 by mistake. FAM
already requires requests >=2.4.2 which should contain everything the callback
needs.

### DIFF
--- a/plugins/callback/foreman.py
+++ b/plugins/callback/foreman.py
@@ -159,11 +159,7 @@ class CallbackModule(CallbackBase):
         ssl_cert = self.get_option('client_cert')
         ssl_key = self.get_option('client_key')
 
-        if HAS_REQUESTS:
-            requests_version = tuple(map(int, requests.__version__.split('.')))
-            if requests_version < (2, 14):
-                self._disable_plugin(u'The `requests` python module is older than 2.14.0.')
-        else:
+        if not HAS_REQUESTS:
             self._disable_plugin(u'The `requests` python module is not installed')
 
         if self.foreman_url.startswith('https://'):


ekohl (approved):

mdellweg:
Should we keep the version check, just with adjusted numbers? Now that we know exactly what minimal version we need?

adamruzicka:
Is shouldn't be necessary when fam itself requires a version which has everything we need. For context a bit of IRC logs

```
[12:32] <Zhenech> aruzicka, but I also question whether it needs to do all that checking at all, as the collection already depends on requests
[12:32] <Zhenech> (>= 2.4.2, for the known reasons)
[12:33] <Zhenech> the check made sense while it was in ansible.git, as *ansible* doesn't depend on requests…
[12:34] <Zhenech> (I'd still do HAS_REQUESTS, as that's mandatory as per ansible policy, but ditch the version checks)
```

mdellweg:
I may have missed something on the latest collection dependency innovations.
But i thought that installing the collection from galaxy will maybe install other collections as dependencies, but not necessary "binary" dependencies like python packages. At least not on all systems, with all supported versions of Ansible?
This is just me being cautious. Feel free to disagree and merge anyaway.
It's probably also rather complicated to accidentally install a version of requests older than 8y.

evgeni:
I feel like people who install the collection and ignore its `requirements.txt` can see it braking down just fine.

We don't check the version in `foreman_helper.py` either for the other modules.

:woman_shrugging:

PR #1208, opened 2021-04-23
content views role

### LINKED ISSUE #1111: content view role
##### SUMMARY
role for managing content views as variables

##### ISSUE TYPE
 - Feature Idea

evgeni:


roles/content_views/README.md
```diff
@@ -0,0 +1,64 @@
+theforeman.foreman.content_views
+=========================================
```
wbclark:
```suggestion
theforeman.foreman.content_views
================================
```

I suppose this doesn't affect the way it's rendered when viewing the README on GH or some other web page but does look nicer if the user browses docs locally in their editor.

roles/content_views/README.md
```diff
@@ -0,0 +1,64 @@
...
+Each content view also requires either a list of repositories or components (for a composite content view):
+- `repositories` - list of repositories to add to the content view, each repository needs the following fields:
+  - `name` - the name of the repository
+  - `product` - the product the repository belongs to
+- `components` - list of content views to add to the composite content view, each component needs the following fields:
+  - `content_view` - the name of the content view
+  - `content_view_version` - the version of the content view to add, *or*
+  - `latest` - if `true`, the latest version of the content view will be used
+
+Additionally you can pass any other parameter the `content_view` module accepts.
```
wbclark:
```suggestion
Additionally you can pass any other [parameters accepted by the content_view module](https://docs.ansible.com/ansible/latest/collections/theforeman/foreman/content_view_module.html#parameters).
```

Let's help them find this information :)

roles/content_views/README.md
```diff
@@ -0,0 +1,64 @@
...
+
+This role also allows to add Content View Filters to the content view, by listing those in `filters` of the content view.
```
wbclark:
```suggestion
This role also allows you to create Content View Filters and add them to the Content View by passing a list of `filters`:
- `filters` - List of filters to create and add to the content view. Each filter needs the following fields:
  - `name` - Name of the content view filter
  - `filter_type` - Content view filter type. The available types are `rpm`, `package_group`, `erratum`, or `docker`

Additionally you can pass any other [parameters accepted by the content_view_filter module](https://docs.ansible.com/ansible/latest/collections/theforeman/foreman/content_view_filter_module.html#parameters).
```

roles/content_views/README.md
```diff
@@ -0,0 +1,64 @@
...
+- `name` - the name of the content view
+
+Each content view also requires either a list of repositories or components (for a composite content view):
+- `repositories` - list of repositories to add to the content view, each repository needs the following fields:
+  - `name` - the name of the repository
+  - `product` - the product the repository belongs to
+- `components` - list of content views to add to the composite content view, each component needs the following fields:
+  - `content_view` - the name of the content view
+  - `content_view_version` - the version of the content view to add, *or*
+  - `latest` - if `true`, the latest version of the content view will be used
```
wbclark:
```suggestion
- `repositories` - List of repositories to add to the content view. Each repository requires the following fields:
  - `name` - The name of the repository
  - `product` - The product which the repository belongs to
- `components` - List of content views to add to the composite content view. Each component requires the following fields:
  - `content_view` - The name of the content view
  - `content_view_version` - The version of the content view to add, *or*
  - `latest` - If `true`, the latest version of the content view will be used
```

A few very small style nits here. "The product to which the repository belongs" would _technically_ be more formally correct English, but still sounds awkward even to a lot of native speakers. I settled on "The product which the repository belongs to" as a clear but less formal middle ground.

wbclark:
I added a review with some style suggestions for the docs as well as linking directly to other docs where appropriate (additional parameters accepted by content_view and content_view_filter modules)

Another thought I had was, since this is a role and we want to have a very user friendly API, for CCV components could we allow the user to pass `content_view_version: latest` instead of requiring them to do this with a separate parameter `latest: true` ? What do you think @evgeni ?

evgeni:
I've updated the docs based on @wbclark's suggestions, but I didn't add the links, as linking to `docs.ansible.com` felt wrong -- the version there is not necessarily the same as the user will be using.

As for `version: latest`. I think that'd be actually a great addition to the module directly, and would prefer to do that change there, so that both module and role users will benefit?

evgeni:
In other roles, we've been either linking to `https://theforeman.github.io/foreman-ansible-modules/develop/plugins/repository_module.html` or just writing "See module documentation for a list of available options"

ehelms (approved):
Ship it!

wbclark:
> I've updated the docs based on @wbclark's suggestions, but I didn't add the links, as linking to `docs.ansible.com` felt wrong -- the version there is not necessarily the same as the user will be using.
> 
> As for `version: latest`. I think that'd be actually a great addition to the module directly, and would prefer to do that change there, so that both module and role users will benefit?

Good points all around.

I suppose there is no reason not to do the linking in both the roles and modules. We can follow up on that (and look for similar opportunities) in another PR.

wbclark (approved):

evgeni:
🚢

PR #1210, opened 2021-04-29
Add role to publish a list of content views

ehelms:
This starts simple and basic to publish a list of content views by name. What, if any, other parameters would anyone want to see here as part of this?


evgeni:
only one of this or #1214 needs to win :)

ephracis (approved):

evgeni (approved):

PR #1211, opened 2021-05-03
Add Redfish to list of possible BMC providers of an interface

evgeni:


changelogs/fragments/host-redfish-bmc.yml
```diff
@@ -0,0 +1,2 @@
+monor_changes:
```
mdellweg:
```suggestion
mordor_changes:
```
```suggestion
mooron_changes:
```
```suggestion
minor_changes:
```

mdellweg (approved):
Assuming this red fish is not a red herring...

PR #1214, opened 2021-05-03
create content_view_publish role

bmarlow:
Per the previous conversation, breaking out roles into smaller chunks.

This role will take a dictionary and publish new content-view versions based on that dictionary.


ehelms:
I had opened https://github.com/theforeman/foreman-ansible-modules/pull/1210 which I think is a bit simpler model and allows more flexiblity. Thoughts?

bmarlow:
I'm looking at #1210 and I'd argue that its less flexible as it has to be separately instantiated for each org.

evgeni:
triage notes: I think we'd prefer not to have the added complexity of the organization in the data structure and rather make the user call the role twice if they need to manage CVs in different ORGs (as we think most users would actually have only one org and other roles already expect to be called multiple times in similar scenarios).

plus this requires the user to have permissions in all orgs, and will be rather slow if you need to check more entities (instead of having a different playbook that is only executed if the data in that specific org changed).

wbclark (commented):
elaborating on the discussion from triage:

1. other roles in the collection already don't support support multiple organizations in the data structure

2. multiple organizations could be handled with looping at the level of playbook, play, or task

3. the multi-tenancy model for interfaces (in this case, ansible data) should reflect the multi-tenancy model for the application. we expect that at least some types of data would be identical across organizations and this common data should exist in a single place, treating the organization as a target, while data that differs between organizations should be stored separately per organization.

in the enterprise, at scale, multi-org deployments are a valid and often necessary use case and should be supported by the collection. our thinking is that the best way to accomplish that is building these roles to support receiving data for a single organization at a time, and providing documentation and code examples of how to organize and loop over the data for multiple organizations.

roles/content_view_publish/meta/main.yml
```diff
@@ -0,0 +1,21 @@
+galaxy_info:
```
evgeni, line 1: we usually don't add this file, as those roles are not to be uploaded to galaxy individually

roles/content_view_publish/defaults/main.yml
```diff
@@ -0,0 +1,3 @@
+---
+# defaults file for content_view_publish
+FOREMAN_VALIDATE_CERTS: true
```
evgeni, line 3: nothing uses this variable, the playbook uses `foreman_validate_certs` as it is defined in the "common" variables

roles/content_view_publish/tasks/main.yml
```diff
@@ -0,0 +1,21 @@
...
+# only publish new version
+- name: "Publish new version of Content-View"
+  theforeman.foreman.content_view_version:
+    username: "{{ foreman_user }}"
+    password: "{{ foreman_password }}"
+    server_url: "{{ foreman_server_url }}"
+    organization: "{{ foreman_organization }}"
+    content_view: "{{ foreman_content_view }}"
+    validate_certs: "{{ foreman_validate_certs }}"
+  async: 3600 #allow async run for up to 1 hour for large content-views
```
evgeni, line 11: is the async needed, given we don't loop or anything in this playbook?

roles/content_view_publish/tasks/main.yml
```diff
@@ -0,0 +1,21 @@
+---
+# only publish new version
+- name: "Publish new version of Content-View"
+  theforeman.foreman.content_view_version:
+    username: "{{ foreman_user }}"
+    password: "{{ foreman_password }}"
+    server_url: "{{ foreman_server_url }}"
+    organization: "{{ foreman_organization }}"
+    content_view: "{{ foreman_content_view }}"
```
evgeni, line 9: In the other roles, we usually followed the rule that the role takes a list to work on, as otherwise there is little benefit over the normal module.

evgeni:
looking at this role, vs @ehelms' attempt in #1210, I would prefer the later.

PR #1186, opened 2021-03-25
Closes #1124 - Add documentation for role developers

wbclark:

PR #1216, opened 2021-05-03
create content_view_promotion role

bmarlow:
This role will promote all appropriate Lifecycle Environments based on an dictionary provided as an extra-var.

If, prior to promoting the LCE, it is determined that there is no new CV to promote to, a new version will be published.


roles/content_view_promote/README.md
```diff
@@ -0,0 +1,127 @@
...
+------------
+
+This role requires the theforeman.foreman module collection.
+
+Role Variables
+--------------
+
+The primary dictionary is the organizations dictionary, which is formatted as such:
+```
+organizations:
```
ehelms: I think this tends to break our general model of crafting roles that follow along tenet-lines. Is it common in your experience to be performing publish and promotes in different organizations from the same workflow?

@evgeni @mdellweg @wbclark your thoughts welcome

roles/content_view_promote/tasks/promote.yml
```diff
@@ -0,0 +1,87 @@
...
+  with_dict: "{{ environments }}"
+
+- name: "Check if publishing new version of Content-View is necessary"
+  set_fact:
+    publish_true: true
+  with_dict: "{{ new_environments }}"
+  when: item.value | int > highest_version | int
+
+# only publish new view if necessary (when the current view is at or above library)
+- name: "Publish new version of Content-View"
```
ehelms: I think you can drop this as I wouldn't expect a promote role to publish as well.

roles/content_view_promote/README.md
bmarlow: Why wouldn't we build the role for maximum flexibility?  

If they don't want to run it against more than one organization they don't have to.

If foreman is configured with 8 organizations but they only put one in the dictionary, that is the only one that will be operated against.

If you prefer, we can just drop the bit of documentation calling out multiple orgs, but leave the functionality.

roles/content_view_promote/tasks/promote.yml
bmarlow: If the LCE is already at the latest version of a CV then there is nowhere to promote it to.

Are you suggesting that we should have this fail as an alternative?

I can do that, it just seems a bit hostile, especially when its trivial to accommodate this way.

roles/content_view_promote/README.md
bmarlow: @ehelms another thought here:

If there are opinions or tenets that the foreman team wants contributors to follow, there needs to be some documentation stating such.  coming in after a commit and saying something isn't following an undocumented guideline makes for a pretty frustrating contributor experience

roles/content_view_promote/tasks/promote.yml
evgeni: we're working on this in https://github.com/theforeman/foreman-ansible-modules/pull/1186 :)

roles/content_view_promote/README.md
evgeni: we're working on this in https://github.com/theforeman/foreman-ansible-modules/pull/1186 :)

PR #1217, opened 2021-05-03
create content_view_rollback role

bmarlow:
A roll for rolling back LCE's in Content-Views


roles/content_view_rollback/tasks/version.yml
```diff
@@ -0,0 +1,8 @@
+---
+# reset the var value to the next lowest value
+# reversing the list and iterating downwards until first match would be more efficient
+- name: "Get closest lower version"
+  set_fact:
+    new_environments: "{{ new_environments | combine({version.key: item}) }}"
+  with_items: "{{ cv_versions | sort | list }}"
+  when: item | int < version.value | int
```
mdellweg, line 8: Suggestion: Maybe this can be turned into a filter plugin.

bmarlow: Which part do you think would lend itself best to being a filter plugin?

My main concern here, is that we are pretty much handling this with 3 lines right now, and (at least after my cursory glance) adding a filter plugin would also require more code and perhaps make this a bit more confusing.

However, filters for something like 'get_next_lowest' integer in a list and 'get_next_highest' integer in a list might be good ideas for core contributions.

If that's of interest I can always make a PR over there.

mdellweg: I was thinking exactly about that kind of arithmetic. It was my first thought when reading the task name, and was reinforced by seeing set_fact. I'm fine if we say, that's a task for another PR on a another day.

PR #1221, opened 2021-05-10
host_errata_info module

evgeni:


plugins/modules/host_errata_info.py
```diff
@@ -0,0 +1,103 @@
...
+
+
+def main():
+    module = ForemanHostErrataInfo(
+        foreman_spec=dict(
+            name=dict(invisible=True),
+            host=dict(type='entity', required=True),
+            content_view=dict(type='entity'),
+            environment=dict(invisible=True, type='entity', resource_type='lifecycle_environments', scope=['organization']),
+            lifecycle_environment=dict(),
```
evgeni: @mdellweg is there a better way to achieve this? the API expects a "environment_id", but I really prefer to call it "lifecycle_environment" in the params. also see the crude hack in L95+

mdellweg: Isn't `flat_name` doing exacly that?

Don't we have the same confusing names elsewhere?

evgeni: Wald, Bäume.

Thanks. That was it. I *knew* we have that, and couldn't think of it.

plugins/modules/host_errata_info.py
```diff
@@ -0,0 +1,99 @@
...
+      - Required together with I(content_view).
+    required: false
+    type: str
+extends_documentation_fragment:
+  - theforeman.foreman.foreman
+  - theforeman.foreman.foreman.infomodulewithoutname
+'''
+
+EXAMPLES = '''
+- name: "Find errata for host"
```
ehelms: I am supposing you will add an example using environment and content view? Would it make sense to keep this as simple as returning errata for a host and having a separate info module for applicable errata host info? That could keep things separate and simpler and make the playbook a user writes a bit clearer to understand, e.g.

```
host_errata_info
host_applicable_errata_info
```

Just thinking out loud.

evgeni: I can add examples, sure.

This essentially implements the API in https://theforeman.org/plugins/katello/3.18/api/apidoc/v2/host_errata/index.html

So with just hostname, you get "installable" errata (I think? the one you see in the UI when you click the content host) and with CV/LCE you get applicable?

I don't think we need two modules for that.

ehelms: > So with just hostname, you get "installable" errata (I think? the one you see in the UI when you click the content host) and with CV/LCE you get applicable?
> 
> I don't think we need two modules for that.

That is my understanding as well. And if I know anything about Katello, installable and applicable are easy to get oneself confused on. So my suggestion was how we could help make that clear for users when they have some inconsistency with their expectations and are staring at their code confused. Thinking back on it, I often wish we had made that distinction more clear in the API between installable and applicable.

Let's phone a friend.... @jlsherrill  from your perspective how do you think we should lay out the design of returning host errata information?

jlsherrill: yes, by default you get the errata that are in the host's CV/LCE which we typically call 'installable'.  If you pass in the Library LCE and default org view CV, you get what what we call 'applicable'.  So the content view and lifecycle environment options are how you select which set of errata to consider.  I think what is here is perfectly fine considering its what the api provides.  That said i think we could do a better job of letting the api match the verbage we use.  I.e. providing an option for 'all_applicable' that would auto-select default org view/Library

ehelms: Resolving this, thanks @jlsherrill for weighing in. Examples can be updated now

evgeni:
Examples are updated, tests are written, I'd say this is ready!

plugins/modules/host_errata_info.py
```diff
@@ -0,0 +1,108 @@
...
+
+- name: "List applicable errata for host"
+  theforeman.foreman.host_errata_info:
+    username: "admin"
+    password: "changeme"
+    server_url: "https://foreman.example.com"
+    host: "host.example.com"
+    lifecycle_environment: "Library"
+    content_view: "Default Organization View"
+'''
```
ehelms, line 73: Do these examples require an organization to fetch the info?

tests/test_playbooks/host_errata_info.yml
```diff
@@ -0,0 +1,77 @@
+---
+- hosts: localhost
```
ehelms, line 2: What part the playbook here sets up the host inside Foreman?

evgeni: None. I did that manually. And could document that in a comment?

plugins/modules/host_errata_info.py
evgeni: The "name" one doesn't (you can search for hosts w/o an org, and after that everything is host-scoped).

The LCE/CV one needs an org, good catch.

evgeni: updated docs and module constraints

tests/test_playbooks/host_errata_info.yml
evgeni: I have now added a play that would setup everything :)

tests/test_playbooks/host_errata_info.yml
```diff
@@ -0,0 +1,99 @@
...
+  vars_files:
+    - vars/server.yml
+  tasks:
+    - name: get errata info
+      host_errata_info:
+        username: "{{ foreman_username }}"
+        password: "{{ foreman_password }}"
+        server_url: "{{ foreman_server_url }}"
+        validate_certs: "{{ foreman_validate_certs }}"
+        organization: "Test Organization"
```
ehelms: Similar to my comment on subs info module, if we have an example without this and it's not needed for this particular search can we drop it from the tests to help ensure the example works?

mdellweg:
I think, this should be called `host_erratum_info`.

ehelms:
> I think, this should be called `host_erratum_info`.

I suppose you also want me to call gooses geese. ;) I do agree that technically `erratum` is more correct, users *know* errata though and use it everywhere. And the API is called errata.....I can live with either one.

evgeni:
Why not `host_erratae_info`? ;)

Honestly, Ii prefer errata as that's what's used everywhere in the UI and API

mdellweg:
@ehelms I yield. Supposedly it spits out many errata for the single host.

tests/test_playbooks/host_errata_info.yml
evgeni: done

ehelms (approved):

ehelms:
Thanks @evgeni !

PR #1223, opened 2021-05-11
subscription_info module

evgeni:


plugins/modules/subscription_info.py
```diff
@@ -0,0 +1,81 @@
...
+  - theforeman.foreman.foreman.infomodule
+'''
+
+EXAMPLES = '''
+- name: "Show a subscription"
+  theforeman.foreman.subscription_info:
+    username: "admin"
+    password: "changeme"
+    server_url: "https://foreman.example.com"
+    name: "Red Hat Satellite Infrastructure Subscription"
```
ehelms, line 42: No organization context is needed to fetch sub info?

plugins/modules/subscription_info.py
```diff
@@ -0,0 +1,81 @@
...
+
+- name: "Show all subscriptions with a certain name"
+  theforeman.foreman.subscription_info:
+    username: "admin"
+    password: "changeme"
+    server_url: "https://foreman.example.com"
+    search: name="Red Hat Satellite Infrastructure Subscription"
```
ehelms, line 49: No organization context is needed to fetch sub info?

plugins/modules/subscription_info.py
evgeni: According to https://theforeman.org/plugins/katello/3.18/api/index.html - no.

ehelms: Can we have one or all of the tests reflect that so we are testing the example? (e.g. drop https://github.com/theforeman/foreman-ansible-modules/pull/1223/files#diff-26c401d211772be08e88a4e26eb7f42fc1c21f16b49ace60767e3113a237b49eR29)

evgeni: dropped from all tests

ehelms (approved):

PR #1224, opened 2021-05-11
[WIP] Add ansible variable module

cru90:
[WIP] Add a module to add, manage, change and delete variables for Ansible roles within foreman.

This module is thought to manage Ansible variables for already imported Ansible roles. It is a work in progress with some issues, but I think it might still be worth sharing. Perhaps someone can help me? 

At the moment, if two different Ansible roles posses variables with identical names, the module does not distinguish between these variables, but will only affect one.. This results in undesired behavior like assigning a value to the wrong variable.

Furthermore, the module seems to loose its idempotency, if variable data types array, hash, yaml or json are used. Even if no changes are applied, the corresponding tasks are always marked as 'changed'.

### DIFF
--- a/plugins/modules/ansible_variable.py
+++ b/plugins/modules/ansible_variable.py
@@ -1,0 +1,359 @@
+#!/usr/bin/python
+# -*- coding: utf-8 -*-
+# (c) 2021 Martin Grundei (ATIX AG)
+#
+# This program is free software: you can redistribute it and/or modify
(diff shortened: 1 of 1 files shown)


mdellweg:
Good start. I don't see right away what is happening with the non string content types. And i would suggest that the handling should be similar to that of global_parameters.
To bring this forward. Can you add a test playbook please? I can vastly help to understand and reproduce what's going on. And in the end you need it anyway.

PR #1227, opened 2021-05-12
add `sub_id` parameter for handling the Azure Subscription ID

evgeni:
instead of the `user` parameter, which still works, but sub_id is
preferred

### DIFF
--- a/changelogs/fragments/azure-subid.yml
+++ b/changelogs/fragments/azure-subid.yml
@@ -1,0 +1,2 @@
+minor_changes:
+  - compute_resource - add ``sub_id`` parameter for handling the Azure Subscription ID instead of the ``user`` parameter
--- a/plugins/modules/compute_resource.py
+++ b/plugins/modules/compute_resource.py
@@ -122,6 +122,11 @@ options:
           - azurechina
           - azuregermancloud
         version_added: 2.1.0
+      sub_id:
+        description:
+          - Subscription ID for I(provider=AzureRm)
+        type: str
+        version_added: 2.1.0
       ssl_verify_peer:
         description:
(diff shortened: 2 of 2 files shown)


mdellweg (approved):
Because we need `user` for other entities, we cannot make it an alias for `sub_id`. Correct?
Should we add a validation that not both are provided at the same time?

evgeni:
> Because we need `user` for other entities, we cannot make it an alias for `sub_id`. Correct?

It's "complicated". Most CRs use "user" as they login to an user account. In the Azure context, you don't have a user, but access to a subscription (or multiple), and this is what sub_id selects.

Now *internally* foreman_azure_rm stores the sub id in the user field, so the old way of setting `user` worked perfectly (but not idempotently, as the api will return sub_id:X, not user:X).

> Should we add a validation that not both are provided at the same time?

Could. I was actually more thinking along the lines of: if user is provided, rename it to sub_id, silently, and move on.

mdellweg:
> > Because we need `user` for other entities, we cannot make it an alias for `sub_id`. Correct?
> 
> It's "complicated". Most CRs use "user" as they login to an user account. In the Azure context, you don't have a user, but access to a subscription (or multiple), and this is what sub_id selects.
> 
> Now _internally_ foreman_azure_rm stores the sub id in the user field, so the old way of setting `user` worked perfectly (but not idempotently, as the api will return sub_id:X, not user:X).
> 
> > Should we add a validation that not both are provided at the same time?
> 
> Could. I was actually more thinking along the lines of: if user is provided, rename it to sub_id, silently, and move on.

... and if both are provided, tell the user he is stu... ahm i mean, educate about a better way?

mdellweg (approved):

PR #1232, opened 2021-05-25
Fixes #32189 - add task action to callback

yifatmakias:

PR #1237, opened 2021-05-31
Added integration tests for callbacks

lzap:
I am planning to do some changes to the callback plugin but I am not comfortable without some tests. I looked into the Ansible codebase and it has many integration tests and several of them for callbacks: https://github.com/ansible/ansible/tree/devel/test/integration/targets/callback_default

I would like to take the same approach, calling ansible with parameters to generate and compare JSON results with fixtures. Let me know how you like this.


tests/integration/runme.sh
```diff
@@ -0,0 +1,15 @@
+#!/usr/bin/env bash
+export ANSIBLE_STDOUT_CALLBACK=foreman
+export ANSIBLE_LOAD_CALLBACK_PLUGINS=1
+export ANSIBLE_FORCE_COLOR=0
+export ANSIBLE_NOCOLOR=1
+# No connection is actually performed during the test
+export FOREMAN_URL=http://localhost
+export FOREMAN_SSL_CERT=/dev/zero
+export FOREMAN_SSL_KEY=/dev/zero
+export FOREMAN_DIR_STORE=/tmp
```
lzap: TODO: This needs to be a temporary directory created + shell trap hook.

Also the generated fixtures (JSON documents) are missing from the patch, TBD.

tests/integration/runme.sh
```diff
@@ -0,0 +1,15 @@
...
+
+ansible-playbook -i inventory test.yml
+
+sed -E -i 's/"[0-9]+-[0-9]+-[0-9]+[ T][0-9]+:[0-9]+:[0-9]+\.[0-9]+/"2000-01-01 12:00:00.0000/g' /tmp/*.json
+sed -E -i 's/"[0-9]+:[0-9]+:[0-9]+\.[0-9]+/"0:00:00.01/g' /tmp/*.json
```
lzap: These statements are to clear date and time strings so fixtures can be compared. Probably an ENV variable could be used to generate new fixtures.

There is no test code yet, TBD.

tests/integration/test.yml
```diff
@@ -0,0 +1,71 @@
...
+      changed_when: false
+
+    - name: Failed task
+      fail:
+        msg: no reason
+      ignore_errors: yes
+
+    - name: Skipped task
+      command: echo foo
+      when: false
```
lzap: Example taken from Ansible core, simplified. Note this play has three hosts so 3 reports will be created.

tests/integration/test.yml
```diff
@@ -0,0 +1,71 @@
...
+      listen: test handlers
+
+    - name: Test handler 3
+      command: echo foo
+      listen: test handlers
+
+- hosts:
+  - testhostA
+  - testhostB
+  gather_facts: no
```
lzap: Facts can be tested the same way, TBD.

lzap:
The reason why I am showing this draft so early is that this is an integration test - it literally calls "ansible" command. I am not sure how feasible is this on the CI (Github Actions in this case).

mdellweg:
> The reason why I am showing this draft so early is that this is an integration test - it literally calls "ansible" command. I am not sure how feasible is this on the CI (Github Actions in this case).

To be fair, all our other tests call ansible-playbook too. They just try to run off of recorded foreman rest api responses. I guess for the callback, the latter not suitable.

lzap:
> all our other tests call ansible-playbook too

By our do you mean tests in this repository? I am unable to grep `ansible-playbook` in there :-)

mdellweg:
> > all our other tests call ansible-playbook too
> 
> By our do you mean tests in this repository? I am unable to grep `ansible-playbook` in there :-)

You are right, it is a bit garbled:
https://github.com/theforeman/foreman-ansible-modules/blob/eeb93d566abb3d12b606e0ba7ef543e83fb98944/tests/test_crud.py#L82

This is running the guts of `ansible-playbook` in the pytest python process.

lzap:
Any idea why this fails: FAILED tests/test_crud.py::test_crud[content_credentials_role] ?

lzap:
Evgeni has a solution: https://github.com/theforeman/foreman-ansible-modules/pull/1242

tests/fixtures/1-testhost.json
```diff
@@ -0,0 +1,108 @@
+{
+  "config_report": {
+    "host": "testhost",
+    "reported_at": 2000-01-01 12:00:00.0000",
```
lzap: A bug in the regexp

plugins/callback/foreman.py
```diff
@@ -195,13 +211,35 @@ def _ssl_verify(self, option):
...
+            with open(filename, 'w') as f:
+                f.write(json.dumps(data, indent=2))
+            self.report_number += 1
+        else:
+            try:
+                response = self.session.post(url=url, json=data)
+                response.raise_for_status()
+            except requests.exceptions.RequestException as err:
+                self._display.warning(u'Sending data to Foreman at {url} failed for {host}: {err}'.format(
+                    host=to_text(host), err=to_text(err), url=to_text(self.foreman_url)))
```
lzap, line 232: Continuation line over-indented for hanging indent (E126)

I do not understand, the indentation is exactly extra one tab.

lzap:
Hmmm it failed for some  Python versions. That's weird.

evgeni:
On 2.7, it doesn't like your `os.listdir`: `TypeError: coercing to Unicode: need string or buffer, LocalPath found`

the others are due to the linter:

```
plugins/callback/foreman.py:224:21: E126 continuation line over-indented for hanging indent
tests/test_callback.py:2:1: F401 'sys' imported but unused
tests/test_callback.py:6:1: E302 expected 2 blank lines, found 1
tests/test_callback.py:20:1: E302 expected 2 blank lines, found 1
```

plugins/callback/foreman.py
evgeni: this is not the line it's complaining about

plugins/callback/foreman.py
```diff
@@ -195,13 +211,35 @@ def _ssl_verify(self, option):
...
+        if endpoint == 'facts':
+            url = self.foreman_url + '/api/v2/hosts/facts'
+        elif endpoint == 'report':
+            url = self.foreman_url + '/api/v2/config_reports'
+        else:
+            self._display.warning(u'Unknown endpoint type: {type}'.format(type=endpoint))
+
+        if len(self.dir_store) > 0:
+            filename = u'{number}-{host}.json'.format(
+                    number=self.report_number, host=to_text(host))
```
```suggestion
                number=self.report_number, host=to_text(host))
```

4 spaces, not 8 :)

plugins/callback/foreman.py
lzap: Thanks I just found out how to run flake8 locally, good.

lzap:
All checks have passed!

tests/test_callback.py
```diff
@@ -0,0 +1,37 @@
...
+    kwargs['playbook'] = os.path.join(os.getcwd(), 'tests', 'callback', 'three_hosts.yml')
+    kwargs['inventory'] = os.path.join(os.getcwd(), 'tests', 'callback', 'three_hosts')
+    kwargs['verbosity'] = 4
+    return ansible_runner.run(**kwargs)
+
+
+def test_callback(tmpdir, vcrmode):
+    run = run_playbook(tmpdir)
+    assert run.rc == 0
+    for file in os.listdir(str(tmpdir)):
```
evgeni: this means that if nothing was written, no assertions happen.

lzap: Is that a bad assumption?

lzap: I mean, FOREMAN_DIR_STORE variable makes Ansible to always write results.

evgeni: Well, I managed to break it, and wondered why tests were still green ;)

evgeni: more exactly, the folder is also empty when I run your code verbatim, so the asserts never trigger

lzap: OMG I built a car without wheels. Lemme fix this.

tests/test_callback.py
```diff
@@ -0,0 +1,37 @@
+import os
+import re
+import ansible_runner
+
+
+def run_playbook(tmpdir):
+    os.environ['ANSIBLE_STDOUT_CALLBACK'] = "foreman"
```
evgeni:
```suggestion
    os.environ['ANSIBLE_STDOUT_CALLBACK'] = "foreman"
    os.environ['ANSIBLE_CALLBACK_WHITELIST'] = "foreman"
```

Otherwise it will never allow that callback

tests/test_callback.py
```diff
@@ -0,0 +1,37 @@
...
+            contents = re.sub(r"\d+:\d+:\d+\.\d+", "12:00:00.0000", contents)
+        fixture = os.path.join(os.getcwd(), 'tests', 'fixtures', file)
+        if vcrmode == "record":
+            print("Writing: ", str(fixture))
+            with open(fixture, 'w') as f:
+                f.write(contents)
+        else:
+            with open(fixture, 'r') as f:
+                expected_contents = f.read()
+                assert expected_contents, contents
```
evgeni:
```suggestion
                assert expected_contents == contents
```

this is not ruby :)

evgeni (changes requested):

tests/test_callback.py
```diff
@@ -0,0 +1,37 @@
...
+            contents = re.sub(r"\d+-\d+-\d+[ T]\d+:\d+:\d+\.\d+", "2000-01-01 12:00:00.0000", contents)
+            contents = re.sub(r"\d+:\d+:\d+\.\d+", "12:00:00.0000", contents)
+        fixture = os.path.join(os.getcwd(), 'tests', 'fixtures', file)
+        if vcrmode == "record":
+            print("Writing: ", str(fixture))
+            with open(fixture, 'w') as f:
+                f.write(contents)
+        else:
+            with open(fixture, 'r') as f:
+                expected_contents = f.read()
```
evgeni: and I think I'd do:

```python
             with open(fixture, 'r') as f:
                 expected_contents = json.load(f)
                 real_contents = json.loads(contents)
                 assert expected_contents == real_contents
```

to compare dicts, instead of super big strings

evgeni:
You need to re-record the fixtures after #1232 was merged and changed the output

tests/test_callback.py
evgeni: if you use `theforeman.foreman.foreman` here, you don't need to make the symlink for the `foreman.py` file, which I think is nicer.

evgeni:
If you're curious, I've pushed my play-branch to https://github.com/evgeni/foreman-ansible-modules/tree/callback-tests

lzap:
Holy crap, thanks. Can I just pull that in and squash that?

tests/test_callback.py
lzap: Oh its a keyword! https://docs.python.org/3/reference/simple_stmts.html#grammar-token-assert-stmt

So much time spent with Ruby DSL. In Ruby there are almost no keywords :-)

lzap:
I squashed thanks, since the error message is not very helpful (not sure why I do not see what is advertised in pytest documentation showing difference in dictionaries) and because we have an easy way to regenerate fixtures:

```diff
diff --git a/tests/test_callback.py b/tests/test_callback.py
index 6bfb7afc..7e069173 100644
--- a/tests/test_callback.py
+++ b/tests/test_callback.py
@@ -37,4 +37,4 @@ def test_callback(tmpdir, vcrmode):
             with open(fixture, 'r') as f:
                 expected_contents = json.load(f)
                 real_contents = json.loads(contents)
-                assert expected_contents == real_contents
+                assert expected_contents == real_contents, "Fixture {file} differs, run pytest tests/test_callback.py --vcrmode record".format(file=file)
```

mdellweg:
> I squashed thanks, since the error message is not very helpful (not sure why I do not see what is advertised in pytest documentation showing difference in dictionaries) and because we have an easy way to regenerate fixtures:
> 
> ```diff
> diff --git a/tests/test_callback.py b/tests/test_callback.py
> index 6bfb7afc..7e069173 100644
> --- a/tests/test_callback.py
> +++ b/tests/test_callback.py
> @@ -37,4 +37,4 @@ def test_callback(tmpdir, vcrmode):
>              with open(fixture, 'r') as f:
>                  expected_contents = json.load(f)
>                  real_contents = json.loads(contents)
> -                assert expected_contents == real_contents
> +                assert expected_contents == real_contents, "Fixture {file} differs, run pytest tests/test_callback.py --vcrmode record".format(file=file)
> ```

Hmm, not sure if this is always the correct consequence. If the fixture do not match `real_contents`, it can also be a legit bug.

evgeni:
you need to call `pytest` with like `-vvvv` and that would produce a diff like output

lzap:
I think I am running into a 2vs3 issue:

```
[gw3] linux2 -- Python 2.7.18 /opt/hostedtoolcache/Python/2.7.18/x64/bin/python
tmpdir = local('/tmp/pytest-of-runner/pytest-1/popen-gw3/test_callback0')
vcrmode = 'replay'

    def test_callback(tmpdir, vcrmode):
        run = run_playbook_callback(tmpdir)
        assert run.rc == 0
        for file in os.listdir(tmpdir.strpath):
>           with open(os.path.join(tmpdir, file), 'r') as f:

tests/test_callback.py:27: 
_ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ 

a = local('/tmp/pytest-of-runner/pytest-1/popen-gw3/test_callback0')
p = ('2-testhostB.json',)
path = local('/tmp/pytest-of-runner/pytest-1/popen-gw3/test_callback0')
b = '2-testhostB.json'

    def join(a, *p):
        """Join two or more pathname components, inserting '/' as needed.
        If any component is an absolute path, all previous path components
        will be discarded.  An empty last part will result in a path that
        ends with a separator."""
        path = a
        for b in p:
            if b.startswith('/'):
                path = b
>           elif path == '' or path.endswith('/'):
E           AttributeError: 'LocalPath' object has no attribute 'endswith'
```

lzap:
I think I need to call `strpath` as in other cases, let me blindly do that.

Corrected the conclusion.

lzap:
Damn I think I need to create fixtures for separate ansible versions. Not sure about different Python versions, I guess I will do that as well. I see diffs. Makes sense.

evgeni:
> Damn I think I need to create fixtures for separate ansible versions. Not sure about different Python versions, I guess I will do that as well. I see diffs. Makes sense.

We have this in other places with nice workarounds. I'll show you tomorrow :)

lzap:
> We have this in other places with nice workarounds. I'll show you tomorrow :)

Thanks because now that I am figuring it out, it would be pretty tough for people to generate fixtures for ancient Ansible/Python versions. A typical contributor usually have one python and ansible version locally.

lzap:
From IRC discussion: messages do seem to be encoded as JSON strings:

https://github.com/theforeman/foreman-ansible-modules/pull/1237/files?file-filters%5B%5D=.json&file-filters%5B%5D=.md&file-filters%5B%5D=.py&file-filters%5B%5D=.yml#diff-9564a75a121b933e2f4a3a216ce5c36888438468a382fc200e5bf4c245ec4429R22

Is that a feature or a bug?

https://github.com/theforeman/foreman-ansible-modules/blob/develop/plugins/callback/foreman.py#L108-L118

@ares or @xprazak2 or @adamruzicka know ?

From the git history, I can say, this has been there from the day one. I wonder if this is there for performance reasons, we are planning to do the same for importing reports into Foreman.

lzap:
Oh I think this is a feature because Ansible needed to comply with Puppet report:

https://github.com/theforeman/foreman_ansible/blob/master/test/fixtures/report.json

@evgeni just ignore it for now, after I will do refactoring it will look normal

lzap:
Allright pushed all your changes, I am not squashing because something tells me I am not doing this for the last time.

lzap:
Okay I see several differences across output for the command module (keys: msg and warn).

I am adding more regexps, these are temporary once we refactor the foreman legacy format these can go away. Simplifying the legacy output is on the short-term radar, I think @yifatmakias is already working on that.

yifatmakias:
> Okay I see several differences across output for the command module (keys: msg and warn).
> 
> I am adding more regexps, these are temporary once we refactor the foreman legacy format these can go away. Simplifying the legacy output is on the short-term radar, I think @yifatmakias is already working on that.

The changes I made were in the foreman ansible side. I am not sure what other changes you are referring to

lzap:
> The changes I made were in the foreman ansible side. I am not sure what other changes you are referring to

Oh sorry, I thought you were working on some "simplification of reports", Marek mentioned we are sending too much data and that we are going to strip some of it.

yifatmakias:
> > The changes I made were in the foreman ansible side. I am not sure what other changes you are referring to
> 
> Oh sorry, I thought you were working on some "simplification of reports", Marek mentioned we are sending too much data and that we are going to strip some of it.

Oh yes, eventually I decided to deal with the data in the foreman ansible and did not changed mush in the callback report side.

evgeni:
My branch should have the warn issue fixed, but not the msg one.

lzap:
I am fixing the total which is time-based probably. Setting it to 42. Let's see what is missing.

evgeni:
Yeah, metrics.total seems to be time related and can vary.

https://github.com/theforeman/foreman-ansible-modules/blob/762e0773bf80f17931fa1d672167e4b5e8ed26f2/plugins/callback/foreman.py#L239

tests/test_callback.py
```diff
@@ -37,6 +37,10 @@ def test_callback(tmpdir, vcrmode):
         contents = real_file.read()
         contents = re.sub(r"\d+-\d+-\d+[ T]\d+:\d+:\d+\.\d+", "2000-01-01 12:00:00.0000", contents)
         contents = re.sub(r"\d+:\d+:\d+\.\d+", "12:00:00.0000", contents)
+        # cover differencess across Ansible versions for the legacy Foreman output
+        # contents = re.sub(r"\\\"warn\\\": (true|false)", '\\"warn\\": false', contents)
```
lzap: Tests are green, hopefully, it works this time. Any other comments before I remove this line?

evgeni: This commented one? Nah, away with it.

evgeni:
I really don't like that we edit out the msg entry. Need to dig deeper why that is.

Another thing: did you consider editing the actual dict that is parsed from the JSON, instead of the string? Using regex here seems like quite a hammer.

lzap:
> Another thing: did you consider editing the actual dict that is parsed from the JSON, instead of the string? Using regex here seems like quite a hammer.

I might surprise you, but I did. This is what I consider the better approach, because then I would probably need to do all the replacements this way. Drilling down the tree will be totally more complex than a single line. Also, the msg is in the JSON-encoded part, so I'd need to parse it for the second time. This is all temporary, we want to drop the legacy report format in the future.

Having said all of that, if you want the Python dictionary way, I can totally do that.

evgeni:
> I might surprise you, but I did. This is what I consider the better approach, because then I would probably need to do all the replacements this way. Drilling down the tree will be totally more complex than a single line. Also, the msg is in the JSON-encoded part, so I'd need to parse it for the second time. This is all temporary, we want to drop the legacy report format in the future.
> 
> Having said all of that, if you want the Python dictionary way, I can totally do that.

My main concern with regex is what if they match too much?

I've pushed to my branch a solution which uses regex for the date/time, but proper python for the metrics thing. What do you think?

Plus I've updated the callback to use `json.dump(file)` over `file.write(json.dumps())` which saves us from storing a big-ish string in memory.

docs/testing.md
```diff
@@ -7,6 +7,10 @@ Foreman Ansible Modules are tested in different ways:
 2. Integration tests
 3. Ansible Sanity
 
+Before you start, ensure all required dependencies are installed. Typically this is `python, python-devel, popt-devel`. To create a virtual environment, do:
+
+    python3 -m venv venv --system-site-packages
+
```
evgeni: this *shouldn't* be necessary, if it was for you, we shall debug that together, outside this PR

evgeni:
As we've now gotten this to a nice working state, I wanted to share a few further thoughts.

This PR is called "Added integration tests for callbacks", but in reality it adds a new feature to the callback and tests that *that* feature works. I'm okay with that, but then the PR should be named differently.

We could *also* try to split this up into 2 (or 3) PRs:
- add the basic skeleton of running tests for callbacks
  - all the code movement from `test_crud.py` to `conftest.py`
  - basic `test_callback.py` which can (but doesn't yet) execute tests
- add tests for the existing functionality
  - this would require extending the current `test_callback.py` to actually also work with VCR as `test_crud.py` does
- add tests for the directory function

But we don't have to -- the addition of "old" tests can also happen later.

tests/test_callback.py
```diff
@@ -0,0 +1,54 @@
...
+    for real_file in tmpdir.listdir(sort=True):
+        contents = real_file.read()
+        contents = re.sub(r"\d+-\d+-\d+[ T]\d+:\d+:\d+\.\d+", "2000-01-01 12:00:00.0000", contents)
+        contents = re.sub(r"\d+:\d+:\d+\.\d+", "12:00:00.0000", contents)
+        # cover differencess across Ansible versions for the legacy Foreman output
+        # contents = re.sub(r"\\\"warn\\\": (true|false)", '\\"warn\\": false', contents)
+        contents = re.sub(r'"total": \d+', '"total": 42', contents)
+        contents = re.sub(r", \\\"msg\\\": \\\"\\\"", "", contents)
+        fixture_name = real_file.basename
+        fixture = os.path.join(os.getcwd(), 'tests', 'fixtures', fixture_name)
```
evgeni: I think the fixtures for this should live in `tests/fixtures/callback/dir_store/<file>.json`

lzap:
I fixed the syntax error and pushed it to see how it works on CI.

lzap:
> I've pushed to my branch a solution which uses regex for the date/time, but proper python for the metrics thing. What do you think?

Looks better, yeah.

> in reality it adds a new feature to the callback

Well, it does, but the feature is only for testing itself. I'd consider this a part of introducing integration test.

> We could _also_ try to split this up into 2 (or 3) PRs:

Only if you insist, this looks fine to me. Unless you plan to backport something.

> the addition of "old" tests can also happen later.

That was the plan, yeah.

docs/testing.md
lzap: Removing the hunk then.

tests/test_callback.py
lzap: Okay, fixing but please fetch from my branch if you plan adding more commits :-)

lzap:
Damn my fingers were too fast, I amended the documentation change into b0760eb22ef18be4670eea04d2ad9b99b72c9c15 sorry. There was a short time window where you could merge the old commit, please make sure you have this one before starting doing any changes.

tests/test_callback.py
```diff
@@ -0,0 +1,51 @@
...
+    return run_playbook(playbook, inventory=inventory, extra_env=extra_env)
+
+
+def test_callback(tmpdir, vcrmode):
+    run = run_playbook_callback(tmpdir)
+    assert run.rc == 0
+    for real_file in tmpdir.listdir(sort=True):
+        contents = real_file.read()
+        contents = re.sub(r"\d+-\d+-\d+[ T]\d+:\d+:\d+\.\d+", "2000-01-01 12:00:00.0000", contents)
+        contents = re.sub(r"\d+:\d+:\d+\.\d+", "12:00:00.0000", contents)
```
evgeni, line 39: either you or me (probably me), killed the `contents = re.sub(r", \\\"msg\\\": \\\"\\\"", "", contents)` which is sadly still needed :(

evgeni: fixed in my branch

tests/test_callback.py
evgeni: *whistle*

tests/test_callback.py
lzap: thanks cherry picked

lzap:
It is green!

lzap:
Anything else? Do you want me to split it?

evgeni:
> Anything else? Do you want me to split it?

besides my time? nah.

One question: do you ever intend `dir_store` to be also used for production deployments? like when you want to dump data to disk and then submit it async?

lzap:
> One question: do you ever intend `dir_store` to be also used for production deployments? like when you want to dump data to disk and then submit it async?

Nope, this is also noted in the documentation - not for production use.

The refactoring which is to be filed after we merge this, will be just simplification of the JSON payload.

evgeni:
I am a bit hesitant adding an option that's only used for debugging, because I *know* people will abuse it for other things and think it's a supported API, regardless of how well we document it or not.

Let me think about it for a moment, whether there is a different way to solve this.

lzap:
What is the status? How about not documenting it? Or to use an ENV variable to do this? But I think the result will be the same, if someone wants to abuse, they will do it. Let's not be Apple, let our users do crazy stuff if that's what they want to do.

evgeni (approved):
Okay, I did figure out how to pipe all this mess though VCR, but it doesn't fail as failing callbacks are just silently disabled by ansible.

mdellweg:
> I am a bit hesitant adding an option that's only used for debugging, because I _know_ people will abuse it for other things and think it's a supported API, regardless of how well we document it or not.
> 
> Let me think about it for a moment, whether there is a different way to solve this.

Just for curiosity, did you call the option `Abgastest`?

lzap:
Thank you so much for the huge help with this!

PR #1238, opened 2021-06-01
Move CHANGELOG to bottom of docs index

quba42:
Normally changelogs are at the bottom of the index, and I managed to completely miss the changelog because it was not.

However, I am agnostic as to whether the changelog should just be at the bottom of the "User documentation", the bottom of the "Developer documentation" or within it's own section (as I now have it in this PR).

Strong opinions very welcome!

### DIFF
--- a/docs/index.rst
+++ b/docs/index.rst
@@ -11,7 +11,6 @@ Welcome to Foreman Ansible Modules' documentation!
    :caption: User documentation
 
    README
-   Changelog <CHANGELOG>
    plugins/index
    Filters <filters>
    roles/index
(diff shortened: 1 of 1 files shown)


docs/index.rst
```diff
@@ -11,7 +11,6 @@ Welcome to Foreman Ansible Modules' documentation!
    :caption: User documentation
 
    README
-   Changelog <CHANGELOG>
```
quba42, line 14: The simple version would be to move this to below `cvmanager` or possibly past `releasing`, but without adding the extra toctree.

evgeni (approved):

evgeni:
@quba42 triage says we agree, so if you undraft this, we can merge!

quba42:
Sorry, I was on holiday last week.

PR #1239, opened 2021-06-04
always lookup the ID of the usergroup

evgeni:
the API *should* take names, but sometimes it fails to do so.
and while this is an API bug as such, we can be nice and just submit the
ID.

This adds an additional API requests, which we originally wanted to
avoid with the old code. See
https://github.com/theforeman/foreman-ansible-modules/pull/538#discussion_r337341746
for the original discussion around that.


evgeni:
will probably need to re-record fixtures, but I am lazy right now

evgeni:
re-recorded, the test failure is unrelated and transient

mdellweg (approved):

PR #1240, opened 2021-06-07
don't override already set parameters when passing an activation key

evgeni:
draft, I want to add tests for that.

### DIFF
--- a/plugins/module_utils/foreman_helper.py
+++ b/plugins/module_utils/foreman_helper.py
@@ -205,7 +205,7 @@ class ParametersMixin(ParametersMixinBase):
 
         self.validate_parameters()
 
-    def run(self, **kwargs):
+    def update_parameters(self):
         entity = self.lookup_entity('entity')
         if not self.desired_absent:
             if entity and 'parameters' in entity:
@@ -213,7 +213,12 @@ class ParametersMixin(ParametersMixinBase):
             parameters = self.foreman_params.get('parameters')
             if parameters is not None:
                 self.foreman_params['parameters'] = parameters_list_to_str_list(parameters)
+            elif entity and 'parameters' in entity:
+                # this allows us to merge special named parameters like kt_activation_keys later
+                self.foreman_params['parameters'] = entity['parameters']
 
+    def run(self, **kwargs):
+        self.update_parameters()
         return super(ParametersMixin, self).run(**kwargs)
 
 
@@ -313,6 +318,9 @@ class HostMixin(ParametersMixin):
         mutually_exclusive = kwargs.pop('mutually_exclusive', []) + [['medium', 'kickstart_repository']]
         super(HostMixin, self).__init__(foreman_spec=foreman_spec, required_plugins=required_plugins, mutually_exclusive=mutually_exclusive, **kwargs)
 
+    def run(self, **kwargs):
+        self.update_parameters()
+
         if 'activation_keys' in self.foreman_params:
             if 'parameters' not in self.foreman_params:
                 self.foreman_params['parameters'] = []
@@ -320,6 +328,8 @@ class HostMixin(ParametersMixin):
             self.foreman_params['parameters'].append(ak_param)
 
         self.validate_parameters()
+
+        return super(HostMixin, self).run(**kwargs)
 
 
 class ForemanAnsibleModule(AnsibleModule):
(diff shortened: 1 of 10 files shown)


evgeni:
Actually, while writing tests, I found a bug 😅

plugins/module_utils/foreman_helper.py
```diff
@@ -203,7 +205,7 @@ def __init__(self, **kwargs):
 
         self.validate_parameters()
 
-    def run(self, **kwargs):
+    def update_parameters(self):
```
wbclark: Just curious -- is there a direct technical benefit to separating the logic into separate `update_parameters` and `run` methods here, or is the benefit "limited to" better conceptual clarity and readability?

evgeni: If you look at the very first commit of the series, it added code to this function and I needed that code to run at the very beginning of HostMixin.run, not at the end (as it would happen when HM.run calls ParameterMixin.run). With the latest iteration it's probably not required anymore and can be reverted back.

plugins/module_utils/foreman_helper.py
```diff
@@ -313,14 +313,26 @@ def __init__(self, **kwargs):
...
-            self.foreman_params['parameters'].append(ak_param)
+    def run(self, **kwargs):
+        entity = self.lookup_entity('entity')
+
+        if not self.desired_absent:
+            if 'activation_keys' in self.foreman_params:
+                if 'parameters' not in self.foreman_params:
+                    parameters = [param for param in (entity or {}).get('parameters', []) if param['name'] != 'kt_activation_keys']
+                else:
+                    parameters = self.foreman_params['parameters']
```
mdellweg, line 324: We should make sure "kt_activation_keys" is not in this, too.
Maybe we just say, it's a wron use of the module to have it in both places.

plugins/module_utils/foreman_helper.py
```diff
@@ -313,14 +313,26 @@ def __init__(self, **kwargs):
...
+                ak_param = {'name': 'kt_activation_keys', 'parameter_type': 'string', 'value': self.foreman_params.pop('activation_keys')}
+                self.foreman_params['parameters'] = parameters + [ak_param]
+            elif 'activation_keys' not in self.foreman_params and 'parameters' in self.foreman_params and entity is not None:
```
mdellweg: The first part of that condition is already established by the else in elif.

mdellweg (commented):
I don't see the case covered, where parameters is specified, but activation_keys is not.
I think it will remove and ak's from the host(group).

evgeni:
> I don't see the case covered, where parameters is specified, but activation_keys is not.
> I think it will remove and ak's from the host(group).

Like https://github.com/theforeman/foreman-ansible-modules/pull/1240/files#diff-2cd196e925cea5ae57ca3d16cd55bd7b81febaa8c6de5dfe1f1e33e4e215ae70R89?

plugins/module_utils/foreman_helper.py
evgeni: True, that's from an older attempt of mine.

plugins/module_utils/foreman_helper.py
evgeni: Validate Params will do that for us

mdellweg: Because we add a second instance?
:facepalm:

evgeni: Correct.

mdellweg:
> > I don't see the case covered, where parameters is specified, but activation_keys is not.
> > I think it will remove and ak's from the host(group).
> 
> Like https://github.com/theforeman/foreman-ansible-modules/pull/1240/files#diff-2cd196e925cea5ae57ca3d16cd55bd7b81febaa8c6de5dfe1f1e33e4e215ae70R89?

I must have looked at it the wrong way. It is in the condition with "parameters in foreman_params and entity not None".
And it's covered by the test. I must have been confused by the extra check for kt_activation_keys

plugins/module_utils/foreman_helper.py
mdellweg:
```suggestion
            elif 'parameters' in self.foreman_params and entity is not None:
```

mdellweg (approved):

Fodoj:
After this change, what's the correct way to set activate key for the hostgroup? kt_activation_key won't work anymore, because it will be used inside "Activation Key", but then still handled as a params from "foreman_params", resulting in `There are duplicate keys in 'parameters': {'kt_activation_keys'}` on playbook re-runs

evgeni:
@Fodoj correct, this was fixed in https://github.com/theforeman/foreman-ansible-modules/commit/87a7d4c75e625278f4a55f6e965ce13deb5bc231 ;)

Fodoj:
@evgeni ahh, missed that, thanks - when is the new release planned?

evgeni:
If you ask nicely, today? :)

Fodoj:
That would be most fantastic, @evgeni, please

evgeni:
https://github.com/theforeman/foreman-ansible-modules/pull/1277

PR #1241, opened 2021-06-08
Install python38-requests for execution envionments

pabelanger:
Rather then use requests from pypi, use the distro version. This is
related to downstream builds for ansible automation platform.

Signed-off-by: Paul Belanger <pabelanger@redhat.com>

### DIFF
--- a/bindep.txt
+++ b/bindep.txt
@@ -1,2 +1,3 @@
 python3-rpm [(platform:redhat platform:base-py3)]
 rpm-python [(platform:redhat platform:base-py2)]
+python38-requests [platform:centos-8 platform:rhel-8]


evgeni:
My understanding was that bindep is only for dependencies that cannot be satisfied with pure Python (and thus requirements.txt).

Can you point me at the place in the spec that says otherwise?

pabelanger:
> My understanding was that bindep is only for dependencies that cannot be satisfied with pure Python (and thus requirements.txt).
> 
> Can you point me at the place in the spec that says otherwise?

No, it is usually the other way around. For downstream, we prefer distro packages, that exist in rhel-8 (for support) and things that are not packaged will be pulled via cachito from pypi.

evgeni:
But if I list requests in bindep now, this will *also* pull it in via requirements.txt if we don't remove it there? And I think that feels wrong?

pabelanger:
We could remove them from requirements.txt if we wanted too, the way ansible-builder works is we first install bindep.txt packages, then pip install. pip will not pull it from pypi because requests will be already found on disk, so pip skips the dependencies.

evgeni:
Interesting. Would you mind updating https://ansible-builder.readthedocs.io/en/latest/collection_metadata.html with the current best practices/preferences?

pabelanger:
> Interesting. Would you mind updating [ansible-builder.readthedocs.io/en/latest/collection_metadata.html](https://ansible-builder.readthedocs.io/en/latest/collection_metadata.html) with the current best practices/preferences?

Sure, if you want to open a PR there, and link to this we can triage and try to make it a little clearer.

bindep.txt
```diff
@@ -1,2 +1,3 @@
 python3-rpm [(platform:redhat platform:base-py3)]
 rpm-python [(platform:redhat platform:base-py2)]
+python38-requests [platform:centos-8 platform:rhel-8]
```
evgeni, line 3: how does this differ to `(platform:redhat platform:base-py3)`?
also, should this be python3-requests, to allow both 3.6 and 3.8, depending on the rest of the environment?

I would prefer not to dictate what python env you use in the execution container?

pabelanger: I am unsure why you are using base-py2 / base-py3, must be something specific for your tooling.

For EEs, that are build on rhel-8 / centos-8 we've decided on using python38, so that is already decided. bindep in itself is opinionated sadly, so everything today downstream will only be python38

evgeni: We don't have any tooling. Nor do we produce EEs ourself.

At some point we were pinged to add system deps, and after reading documentation of `bindep`, this seemed like the most straight forward way to say:
* on el-based distributions with py2 as the default (centos7, rhel7, oel7, etc): use rpm-pyhon
* on el-based distributions with py3 as default (el8…) use python3-rpm

evgeni: So, this is still open, and I would like to get some closure here.

If for EE, you decided to go with Python 3.8, where does the user get the `rpm` module from? There is no `python38-rpm`.
So generating an EE with Py38 won't work for all modules of this collection?

For a moment I thought we should add both `python3-requests` and `python38-requests`, so it at least also works for both Pythons, but then I remembered that 3.6 is going away soon from Ansible anyways… But it's probably still the most "correct" solution, for now.

pabelanger: This is specifically for downstream packaging in brew, where we prefer to use the packaged version of python38-requests over the pip version.

For rhel-8 / centos-8 python38-requests is from the app stream: https://centos.pkgs.org/8/centos-appstream-x86_64/python38-requests-2.22.0-9.module_el8.4.0+647+0ba99ce8.noarch.rpm.html which we use.

if for centos-8 you want to use pip, that is fine, but for rhel-8 we prefer the distro version for support reasons.

evgeni (approved):

PR #1242, opened 2021-06-08
fix content_credentials role test

evgeni:
latest is not a good idea as it can change


lzap (approved):
Fixes tests, I need this, thank you!

PR #1243, opened 2021-06-08
include *which* param is duplicated in validator

evgeni:


plugins/module_utils/foreman_helper.py
```diff
@@ -174,8 +174,10 @@ class ParametersMixinBase(object):
     def validate_parameters(self):
         parameters = self.foreman_params.get('parameters')
         if parameters is not None:
-            if len(parameters) != len(set(param['name'] for param in parameters)):
-                self.fail_json(msg="There are duplicate keys in 'parameters'.")
+            parameter_names = [param['name'] for param in parameters]
+            duplicate_params = set([x for x in parameter_names if parameter_names.count(x) > 1])
```
mdellweg, line 178: Looks a bit complicated...
But using a `defaultdict(int)` is probably no better.

mdellweg (approved):

plugins/module_utils/foreman_helper.py
evgeni: If you think that's complicated, you need to look at the other answers I've found on StackOverflow 😅

PR #1254, opened 2021-06-29
explicitly assert which versions are kept

evgeni:
only checking `length` can lead to the wrong versions (but the right
number of them) being removed, and the tests not noticing that


mdellweg (approved):
nice!

PR #1255, opened 2021-06-29
BZ#1974314 - make cv cleanup work with keep=0

evgeni:


evgeni:
:green_apple:

roles/content_view_version_cleanup/tasks/find_and_delete_unused_cv_versions.yml
```diff
@@ -17,4 +17,4 @@
     cv_name: "{{ cv.name }}"
     cv_versions: "{{ (versions.resources | rejectattr('environments') | rejectattr('composite_content_view_ids') |
       rejectattr('published_in_composite_content_view_ids') | map(attribute='version') | map('float') | sort |
-      map('string') | list )[:-foreman_content_view_version_cleanup_keep] }}"
+      map('string') | reverse | list )[foreman_content_view_version_cleanup_keep:] }}"
```
evgeni, line 20: if you're curious what happen(s|ed) here:

`alist[:-5]` gives all but the last 5 entries of the list, while `alist[:-0]` is the same as `alist[:0]` which is sadly `[]`.

now in normal python, I'd probably just do something like `alist[:len(alist)-n]`, but given `alist` is really a super long thing we generate on the fly, and I really hate using `set_fact`, I didn't.

another (clever) solution would be to use `alist[:-n or None]`, as `-n or None` evaluates to `None` for `n=0`, and then `alist[:None]` becomes `alist[:]` which is `alist` (well, technically a copy of it, but who cares). and I don't like clever too much.

now, given that we don't care about the order the CVs are removed, we can reverse the list, and start slicing at the beginning, where using 0 works just fine. still a bit clever, but less than the above, imho.

and because users, I've added a guard to disallow `keep` to be < 0.

ehelms (approved):

PR #1256, opened 2021-06-29
pin projects that dropped Python 2 support

evgeni:


mdellweg (approved):

evgeni:
`content_rhel` tests :roll_eyes:

PR #1257, opened 2021-06-30
when resolving values in vm_attrs, compare them as strings

### LINKED ISSUE #1245: cluster value in vm_attr for compute profiles expect string
##### SUMMARY

If we want to create compute profiles for vsphere compute resource the value of `cluster` within `vm_attrs` can't be an integer. We need to put an integer in quotes. 
As we can use numbers for cluster and or datacenter in Sphere configuration the module should also accept integers and may be do a type casting inside the module.

##### ISSUE TYPE
 - Bug Report

##### ANSIBLE VERSION

ansible >= 2.9.0, <= 2.9.10

##### COLLECTION VERSION

theforeman.foreman:2.1.0

##### KATELLO/FOREMAN VERSION

foreman-server 2.1.4

##### STEPS TO REPRODUCE

This code should reproduce the issue.

```yaml
- name: setup compute profiles
  vars:
    compute_profile:
      name: "generic"
      compute_attributes:
        - compute_resource: "vsphere.example.com"
          vm_attrs:
            cpus: 1
            sockets: 1
            memory_mb: '2048'
            firmware: automatic
            cluster: 1337
            resource_pool: Resources
            path: "/Datacenters/1337/vm/1337_VM"
            guest_id: otherGuest64
            hardware_version: Default
            memoryHotAddEnabled: '0'
            cpuHotAddEnabled: '0'
            add_cdrom: '0'
            boot_order:
              - network
              - disk
            annotation: 'Foreman controlled vm'
            scsi_controllers:
              - type: VirtualLsiLogicController
                key: 1000
            interfaces_attributes:
              '0':
                type: VirtualE1000
                network: 1337_Service_PROD_1200
            volumes_attributes:
              '0':
                name: Hard disk
                thin: true
                mode: persistent
                controller_key: 1000
                size_gb: 10
                datastore: 1337_DS01
  theforeman.foreman.compute_profile:
    username: "{{ foreman_admin }}"
    password: "{{ foreman_admin_password }}"
    validate_certs: "{{ validate_certs }}"
    name: "{{ compute_profile.name }}"
    compute_attributes: "{{ compute_profile.compute_attributes | default(omit) }}"
    state: "present"

```

##### EXPECTED RESULTS

Foreman should manage compute profile with a cluster name with integers only without complaining about.

##### ACTUAL RESULTS

```
TASK [config_foreman : Manage compute profile generic] **************************
70 fatal: [localhost]: FAILED! => {"changed": false, "msg": "Could not find clusters '1894' on compute resource 'vcenter.example.com'."}
```

evgeni:
Fixes: #1245

### DIFF
--- a/changelogs/fragments/1245-search_vm_attr_as_string.yml
+++ b/changelogs/fragments/1245-search_vm_attr_as_string.yml
@@ -1,0 +1,2 @@
+bugfixes:
+  - host, compute_profile - when resolving cluster and other values in vm_attrs, compare them as strings (https://github.com/theforeman/foreman-ansible-modules/issues/1245)
--- a/plugins/module_utils/foreman_helper.py
+++ b/plugins/module_utils/foreman_helper.py
@@ -835,7 +835,7 @@ class ForemanAnsibleModule(AnsibleModule):
         api_name = 'available_{0}'.format(part_name)
         available_parts = self.resource_action('compute_resources', api_name, params=additional_params,
                                                ignore_check_mode=True, record_change=False)['results']
-        part = next((part for part in available_parts if part['name'] == name or part['id'] == name), None)
+        part = next((part for part in available_parts if str(part['name']) == str(name) or str(part['id']) == str(name)), None)
         if part is None:
             err_msg = "Could not find {0} '{1}' on compute resource '{2}'.".format(part_name, name, compute_resource.get('name'))
             self.fail_json(msg=err_msg)


evgeni:
@cmeissner would you mind testing this patch please?

cmeissner:
> testing this patch please?

The patch works for me. I can create a compute profile with bare int as value for `cluster` parameter.

mdellweg (approved):

PR #1259, opened 2021-07-07
add envrionmental settings to gitignore list

cmeissner:
To made it possible to each developer/contributor to manage its own development environment this change add some places (files/directories) to gits ignore list.

* add .vscode directory
* add .python-version for python environment handling

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -14,3 +14,7 @@ venv/
 docs/plugins/
 docs/_build/
 changelogs/.plugin-cache.yaml
+
+# environmental settings
+.vscode/*
+.python-version


evgeni:
Personally, I think such things belong in the users git ignore file (`~/.config/git/ignore`) and not in the project one.

See https://git-scm.com/docs/gitignore

But I won't block this PR if anyone else thinks we should add it.

cmeissner:
> Personally, I think such things belong in the users git ignore file (`~/.config/git/ignore`) and not in the project one.

Ah, you never learn enough. In that case this PR is not really needed. I personally prefer a more or less complete `.gitignore` in repo so that all developers start with the same environment. But your suggestion is also valid for me.

mdellweg:
I’m with Evgeni on this. We cannot compensate for all eventual Dev envs. And we should not.

PR #1261, opened 2021-07-07
Specify to use repository content_label for activation key content ov…

### LINKED ISSUE #1260: Update content_overrides label description with location for activation_key module docs
##### SUMMARY

Update activation_key module documentation with the best option for finding content_label for content_overrides

##### ISSUE TYPE
 - Feature Idea

wbclark:
…errides

Most entities in Katello have a "label", for example an Organization with name "ACME Organization" could have a label "ACME_Organization"

This is separate from the concept of a label in yum/dnf, which is referred to as a content_label for the repository in Satellite. In particular, the repository_info module will return the content_label for the repository, and this is the correct one to use when setting content_overrides.

### DIFF
--- a/plugins/modules/activation_key.py
+++ b/plugins/modules/activation_key.py
@@ -80,18 +80,20 @@ options:
     elements: str
   content_overrides:
     description:
-      - List of content overrides that include label and override state ('enabled', 'disabled' or 'default')
+      - List of content overrides that include label and override state
+      - Label refers to repository content_label, e.g. rhel-7-server-rpms
+      - Override state ('enabled', 'disabled', or 'default') sets initial state of repository for newly registered hosts
     type: list
     elements: dict
     suboptions:
       label:
         description:
-          - Label of the content override
+          - Repository content_label to override when registering hosts with the activation key
         type: str
         required: true
       override:
         description:
-          - Override value
+          - Override value to use for the repository when registering hosts with the activation key
         choices:
           - enabled
           - disabled
--- a/plugins/modules/activation_key.py
+++ b/plugins/modules/activation_key.py
@@ -81,14 +81,14 @@ options:
   content_overrides:
     description:
       - List of content overrides that include label and override state
-      - Label refers to repository content_label, e.g. rhel-7-server-rpms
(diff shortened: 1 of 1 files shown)


jamesmarshall24:
@wbclark maybe calling out `hammer activation-key product-content --name Activation_Key --organization-id 1` to get the content_label from org id 1, link to the KCS [How to override enable or disable a repository in an activation key using hammer command in Red Hat Satellite 6](https://access.redhat.com/solutions/3433551), or an example under "examples" on how to pull the label using repository_info -> activation_key would be really helpful for #1260

wbclark:
@jamesmarshall24 I like the idea. Let's see what the team thinks when we discuss it in the next issue triage.

evgeni:
Couple of things as food for thought:
* I prefer not to link to RH KCS, as those require an RH sub, which people might not have (or don't want to pick up a free devel sub)
* `content_label` is not a word, should we format it like [`C(content_label)`](https://docs.ansible.com/ansible/latest/dev_guide/developing_modules_documenting.html#linking-and-other-format-macros-within-module-documentation) to point out that it refers to a label (heh) in the code?

plugins/modules/activation_key.py
```diff
@@ -80,18 +80,20 @@
     elements: str
   content_overrides:
     description:
-      - List of content overrides that include label and override state ('enabled', 'disabled' or 'default')
+      - List of content overrides that include label and override state
+      - Label refers to repository content_label, e.g. rhel-7-server-rpms
```
```suggestion
      - Label refers to repository C(content_label), e.g. rhel-7-server-rpms
```

plugins/modules/activation_key.py
```diff
@@ -80,18 +80,20 @@
...
+      - Override state ('enabled', 'disabled', or 'default') sets initial state of repository for newly registered hosts
     type: list
     elements: dict
     suboptions:
       label:
         description:
-          - Label of the content override
+          - Repository content_label to override when registering hosts with the activation key
```
```suggestion
          - Repository C(content_label) to override when registering hosts with the activation key
```

evgeni (approved):

PR #1266, opened 2021-07-14
Update repository.py

igramic:
add support for filtering repositories by OS version based on API feature apidoc/v2/repositories/create.html

### DIFF
--- a/plugins/modules/repository.py
+++ b/plugins/modules/repository.py
@@ -332,7 +332,7 @@ def main():
             module.fail_json(msg="({0}) can only be used with content_type 'ansible_collection'".format(",".join(invalid_list)))
 
     if module.foreman_params['content_type'] != 'yum':
-        invalid_list = [key for key in ['ignorable_content'] if key in module.foreman_params]
+        invalid_list = [key for key in ['ignorable_content', 'os_versions'] if key in module.foreman_params]
         if invalid_list:
             module.fail_json(msg="({0}) can only be used with content_type 'yum'".format(",".join(invalid_list)))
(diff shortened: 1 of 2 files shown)


plugins/modules/repository.py
```diff
@@ -203,6 +203,16 @@
       - repositories will be automatically enabled on a registered host subscribed to this product
     type: bool
     required: false
+  os_versions:
+    description:
+      - Identifies whether the repository should be disabled on a client with a non-matching OS version
```
evgeni:
```suggestion
      - Identifies whether the repository should be disabled on a client with a non-matching OS version.
      - A maximum of one OS version can be selected.
      - Set to C([]) to disable filtering again.
```

evgeni (changes requested):
The change itself looks great, thanks!

There is a documentation suggestion inline, plus:
* could you please make the commit message more descriptive (something like "add support for filtering repositories by OS version")?
* could you add a changelog snippet in `changelogs/fragments`? it's supposed to be a `.yml` file with a content like this:
```yaml
minor_changes:
  - repository - <explanation>
```
* as far as I understand, this only works for YUM repos, so probably a good idea to add this parameter to the list in
https://github.com/theforeman/foreman-ansible-modules/blob/070dc2ec059fa65382b6b851c127498062b82022/plugins/modules/repository.py#L321-L324

evgeni:
changes look good to me, Igor wanted to do a few tests with it tomorrow, and then we can merge

evgeni (approved):

PR #1267, opened 2021-07-23
Host Report Format

lzap:
Introduces new Host Reports format for Smart Proxy Host Report plugin. It is a complete rewrite of Foreman configuration management reports where Ansible is now a native format. There is no need to mimic a Puppet report, therefore this patch drops some things which seems unnecessary (log as hash of entries with sources/messages records and encoding of JSON as string).

For more info:

* https://github.com/theforeman/foreman_host_reports
* https://github.com/theforeman/smart_proxy_host_reports

Example report before the change (legacy format - you need to scroll to the left to see contents of the message string):

```json
{
  "config_report": {
    "host": "localhost",
    "reported_at": "2021-09-10T13:31:32.681238",
    "metrics": {
      "time": {
        "total": 0
      }
    },
    "status": {
      "applied": 0,
      "failed": 0,
      "skipped": 0
    },
    "logs": [
      {
        "log": {
          "sources": {
            "source": "ansible.builtin.copy"
          },
          "messages": {
            "message": "{\"_ansible_no_log\": false, \"changed\": false, \"checksum\": \"6021762040fb5fe4aae926b12f196a5cf379b6c5\", \"dest\": \"/tmp/hosts\", \"diff\": {\"after\": {\"path\": \"/tmp/hosts\"}, \"before\": {\"path\": \"/tmp/hosts\"}}, \"failed\": false, \"gid\": 1001, \"group\": \"lzap\", \"invocation\": {\"module_args\": {\"_diff_peek\": null, \"_original_basename\": \"hosts\", \"access_time\": null, \"access_time_format\": \"%Y%m%d%H%M.%S\", \"attributes\": null, \"backup\": null, \"content\": null, \"delimiter\": null, \"dest\": \"/tmp/hosts\", \"directory_mode\": null, \"follow\": true, \"force\": false, \"group\": null, \"mode\": null, \"modification_time\": null, \"modification_time_format\": \"%Y%m%d%H%M.%S\", \"owner\": null, \"path\": \"/tmp/hosts\", \"recurse\": false, \"regexp\": null, \"remote_src\": null, \"selevel\": null, \"serole\": null, \"setype\": null, \"seuser\": null, \"src\": null, \"state\": \"file\", \"unsafe_writes\": null}}, \"mode\": \"0644\", \"module\": \"ansible.builtin.copy\", \"owner\": \"lzap\", \"path\": \"/tmp/hosts\", \"secontext\": \"unconfined_u:object_r:user_home_t:s0\", \"size\": 425, \"state\": \"file\", \"uid\": 1001}"
          },
          "level": "info"
        }
      }
    ],
    "reporter": "ansible",
    "check_mode": false
  }
}
```

Example report output of the new format:

```json
{
  "host": "localhost",
  "reported_at": "2021-09-10T13:56:23.103318",
  "uuid": "e207233c-123e-11ec-8c97-1c697a0bae03",
  "metrics": {
    "time": {
      "total": 0
    }
  },
  "status": {
    "applied": 0,
    "failed": 0,
    "skipped": 0
  },
  "logs": [
    {
      "diff": {
        "before": {
          "path": "/tmp/hosts"
        },
        "after": {
          "path": "/tmp/hosts"
        }
      },
      "path": "/tmp/hosts",
      "changed": false,
      "uid": 1001,
      "gid": 1001,
      "owner": "lzap",
      "group": "lzap",
      "mode": "0644",
      "state": "file",
      "secontext": "unconfined_u:object_r:user_home_t:s0",
      "size": 425,
      "invocation": {
        "module_args": {
          "dest": "/tmp/hosts",
          "_original_basename": "hosts",
          "recurse": false,
          "state": "file",
          "path": "/tmp/hosts",
          "force": false,
          "follow": true,
          "modification_time_format": "%Y%m%d%H%M.%S",
          "access_time_format": "%Y%m%d%H%M.%S",
          "_diff_peek": null,
          "src": null,
          "modification_time": null,
          "access_time": null,
          "mode": null,
          "owner": null,
          "group": null,
          "seuser": null,
          "serole": null,
          "selevel": null,
          "setype": null,
          "attributes": null,
          "content": null,
          "backup": null,
          "remote_src": null,
          "regexp": null,
          "delimiter": null,
          "directory_mode": null,
          "unsafe_writes": null
        }
      },
      "checksum": "6021762040fb5fe4aae926b12f196a5cf379b6c5",
      "dest": "/tmp/hosts",
      "_ansible_no_log": false,
      "failed": false,
      "module": "ansible.builtin.copy"
    }
  ],
  "check_mode": false
}
```

Also this is a chance to cleanup some things, it looks like many reports do appear to carry a lot of `null` values in `module_args` dictionary. These can be dropped as they do not provide any value.

**Configuration**

This is the most challenging task - Ansible callback sends both facts and reports, however, in the first phase we are only looking into rewriting reports. So two endpoints must be recognized - foreman and proxy. Instead of just:

* url
* client_cert
* client_key

We need a few more:

* report_type - one of "foreman" or "proxy"
* fact_type - one of "foreman" or "proxy" 
* proxy_url
* proxy_client_cert
* proxy_client_key

This way users can migrate reports to the smart proxy endpoint while keeping facts on foreman core until we also move them. They should probably be named consistently "foreman_url" and "proxy_url" but this would require users to migrate and since the legacy ("foreman") will be dropped at some point, I think we can just live with "url" and "proxy_url" for couple of releases. Unless there is a better idea how to figure out configuration.

Implementation detail: Dictionary `items` used to be array of tuples (module name, hash of data), but since the module name is actually also stored in the hash under key named "module", this patch simplifies that and only appends the hash to the array. Method `build_log_legacy` was refactored to use module name correctly.

TODO:

* [x] Tests
* [x] Handling of client X509 certs


plugins/callback/foreman.py
```diff
@@ -19,6 +19,16 @@
...
+      type:
+        description:
+          - endpoint type: foreman_legacy or proxy_host_report
+        env:
+          - name: FOREMAN_TYPE
+        required: True
+        default: foreman_legacy
+        ini:
+          - section: callback_foreman
+            key: type
```
lzap: Unsure about naming here, provide a preference please.

plugins/callback/foreman.py
```diff
@@ -250,10 +263,43 @@ def send_facts(self):
...
+        Send reports to Foreman Smart Proxy running Host Reports
+        plugin. The format is native Ansible report without any
+        changes.
+        """
+        for host in stats.processed.keys():
+            total = stats.summarize(host)
+            report = {
+                "host": host,
+                "reported_at": get_now(),
+                "uuid_XXX": "XXXXXXXXXXXXXXXXXXXXXXXXXXX",
```
lzap: I am not sure if there is an unique UUID per report per host, we might need to generate one.

plugins/callback/foreman.py
```diff
@@ -19,6 +19,16 @@
...
+        env:
+          - name: FOREMAN_TYPE
+        required: True
+        default: foreman_legacy
+        ini:
+          - section: callback_foreman
+            key: type
       url:
         description:
           - URL of the Foreman server.
```
ofedoren, line 33: I guess you would also need to mention that in case of `proxy_host_report` endpoint type, there will be a need to use Proxy URL instead of Foreman server URL.

UPD: Also, if proxy URL is being specified then the facts are being sent there as well instead of sending them to the Foreman API. I guess we need to use two separate URLs here...

plugins/callback/foreman.py
ofedoren: Maybe `endpoint`? But no strong opinion here...

plugins/callback/foreman.py
lzap: DAMN! You are right, I forgot about facts.

lzap:
I vastly improved description for @evgeni and rebased. Hope it makes more sense now, apologies for an ugly description.

plugins/callback/foreman.py
```diff
@@ -281,17 +328,25 @@ def send_reports(self, stats):
...
+            self.send_reports_proxy_host_report(stats)
+        else:
+            self._display.warning(u'Unknown foreman endpoint type: {type}'.format(type=self.report_type))
+
     def append_result(self, result, failed=False):
         name = result._task.get_name()
         host = result._host.get_name()
         value = result._result
         value['failed'] = failed
         value['module'] = result._task.action
```
lzap, line 348: Shell we copy the dictionary? Because if the `_result` method does not return a copy, we are modifying the source dictionary here which is not nice.

lzap:
This is now ready for review @evgeni ping me if you need explanation, I will add new format fixture tests once the code is ready, I am sure some changes will be required :-)

plugins/callback/foreman.py
```diff
@@ -19,6 +20,26 @@
       - whitelisting in configuration
       - requests (python library)
     options:
+      report_type:
+        description:
+          - endpoint type for reports: foreman or proxy
+        env:
+          - name: FOREMAN_REPORT_TYPE
+        required: True
+        default: foreman
```
evgeni: `required` and `default` together doesn't make much sense IMHO, just `default` seems sensible here, as this makes it default to the old, Foreman-based report type, which is what users used before this patch.

plugins/callback/foreman.py
```diff
@@ -31,6 +52,16 @@
         ini:
           - section: callback_foreman
             key: url
+      proxy_url:
+        description:
+          - URL of the Foreman Smart Proxy server.
+        env:
+          - name: FOREMAN_PROXY_URL
+        required: True
+        default: http://localhost:8000
```
evgeni: This is not a sensible default. Nobody will have the proxy running w/o auth on localhost:8000 (but dev, but that's not what we're optimizing for).

We do have the silly default for the `foreman_url`, but just to not break compatibility (yay, contracts)

plugins/callback/foreman.py
```diff
@@ -55,6 +86,30 @@
           - section: callback_foreman
             key: client_key
         aliases: [ ssl_key ]
+      proxy_client_cert:
+        description:
+          - X509 certificate to authenticate to Foreman Smart Proxy if https is used
```
evgeni: Will the key be really *different* in some situations? The default points at the same key/cert, so I'd guess in most configurations they would actually be the same? And not having to configure another key sounds like a win to me?

plugins/callback/foreman.py
```diff
@@ -248,12 +323,45 @@ def send_facts(self):
...
+        Send reports to Foreman Smart Proxy running Host Reports
+        plugin. The format is native Ansible report without any
+        changes.
+        """
+        for host in stats.processed.keys():
+            total = stats.summarize(host)
+            report = {
+                "host": host,
+                "reported_at": get_now(),
+                "uuid": str(uuid.uuid1()),
```
evgeni: why is that needed? what happens if I maliciously submit multiple reports with the same UUID?

plugins/callback/foreman.py
```diff
@@ -169,16 +223,20 @@ def set_options(self, task_keys=None, var_options=None, direct=None):
         if self.get_option('disable_callback'):
             self._disable_plugin('Callback disabled by environment.')
 
+        self.report_type = self.get_option('report_type')
+        self.fact_type = self.get_option('fact_type')
         self.foreman_url = self.get_option('url')
+        self.proxy_url= self.get_option('proxy_url')
```
```suggestion
        self.proxy_url = self.get_option('proxy_url')
```

plugins/callback/foreman.py
```diff
@@ -19,6 +20,26 @@
...
+        env:
+          - name: FOREMAN_REPORT_TYPE
+        required: True
+        default: foreman
+        ini:
+          - section: callback_foreman
+            key: report_type
+      fact_type:
+        description:
+          - endpoint type for facts: foreman (proxy setting is not yet supported)
```
```suggestion
          - "endpoint type for facts: foreman (proxy setting is not yet supported)"
```

YAML is picky if you have colons in your strings

plugins/callback/foreman.py
```diff
@@ -19,6 +20,26 @@
       - whitelisting in configuration
       - requests (python library)
     options:
+      report_type:
+        description:
+          - endpoint type for reports: foreman or proxy
```
```suggestion
          - "endpoint type for reports: foreman or proxy"
```

YAML is picky if you have colons in your strings

plugins/callback/foreman.py
evgeni: On the other hand, given this is not yet supported, why introducing it here?

plugins/callback/foreman.py
```diff
@@ -248,12 +323,45 @@ def send_facts(self):
...
+                "status": {
+                    "applied": total['changed'],
+                    "failed": total['failures'] + total['unreachable'],
+                    "skipped": total['skipped'],
+                },
+                "logs": self.items[host],
+                "check_mode": self.check_mode,
+            }
+            if self.check_mode:
+                report['config_report']['status']['pending'] = total['changed']
```
ofedoren: This will fail if run ansible with check mode. I guess `['config_report']` is a leftover after copy-paste :)

plugins/callback/foreman.py
```diff
@@ -106,30 +161,29 @@
...
-            "log": {
-                'sources': {
-                    'source': source,
-                },
-                'messages': {
-                    'message': json.dumps(msg, sort_keys=True),
-                },
-                'level': level,
-            }
+    if data.get('failed'):
```
ofedoren: This and all `.get` on `data` will fail since the `data` will be still a list (array): https://github.com/theforeman/foreman-ansible-modules/blob/ea8596f90fe3d48083cf817279364e582f575fbb/plugins/callback/foreman.py#L215

Needs some further refactoring, or you can simply use `data[0]...`.

plugins/callback/foreman.py
```diff
@@ -106,30 +161,29 @@
...
+        level = 'err'
+    elif data.get('changed'):
+        level = 'notice'
+    else:
+        level = 'info'
+
+    yield {
+        "log": {
+            'sources': {
+                'source': data.get('module'),
```
ofedoren: Previously this was a task name (e.g. `Create /tmp/file file`), but with this change it will be the module name (e.g. `file`). Since we don't want to change existing behaviour, maybe we shouldn't change this?

plugins/callback/foreman.py
```diff
@@ -281,17 +389,37 @@ def send_reports(self, stats):
...
+                dd[k] = self.drop_nones(v)
+            elif isinstance(v, (list, set, tuple)):
+                dd[k] = type(v)(self.drop_nones(vv) if isinstance(vv, dict) else vv
+                                for vv in v)
+            elif v is not None:
+                dd[k] = v
+        return dd
+
     def append_result(self, result, failed=False):
-        name = result._task.get_name()
```
ofedoren, line 289: I guess for better UI we could keep the name of a task, but it's up to you how to send it exactly: put this into `result._result` or send it in the old way.

ofedoren (changes requested):
@lzap, I've tested it a bit and with some small fixes it works as expected for the new (proxy) way. Facts are being updated as well.

My concerns are inline and also I didn't test it with SSL enabled, and I've noticed that only `gather_facts` entry/message is present in the old report format in the database (this may be because of my poor adjustments to this script though).

plugins/callback/foreman.py
lzap: Yup.

plugins/callback/foreman.py
lzap: I am not sure what to set tho. To me `localhost` sounds about right, perhaps 9090 or 8443?

plugins/callback/foreman.py
lzap: I was thinking the same but I was not sure. But makes sense since both share (always?) the same CA? Yeah removing, I can add later if this is not true but I think you are right.

plugins/callback/foreman.py
lzap: This is just an identifier used to find same reports uploaded multiple times (e.g. via bug). There is no logic on backend that would process that in any way. Puppet reports do have unique UUIDs so I thought it would be good idea to require that. Now I am thinking about this, let's just drop it, there is no reason to generate this.

plugins/callback/foreman.py
lzap: True, removed.

plugins/callback/foreman.py
lzap: Indeed, fixing.

plugins/callback/foreman.py
lzap: Great find! Yeah I was confused I thought it is the same as module name but it can be different.

I stored it into the hash under "name" key, I hope there is not such key in use (in my few examples I haven't seen such key used).

plugins/callback/foreman.py
lzap: Not sure what you mean, items is a list but it contains dicts now instead of lists. I grabbed my chance to simplify this in memory. Previously it was a list of lists, now it is a list of dicts. Is there an error or exception you see?

plugins/callback/foreman.py
lzap: Good find, I have created "name" key that will hold the previous value. And I am changing this to "name" as well.

lzap:
Okay, integrated all your comments, let's see if tests are passing first, including fixtures which is very important (not to break existing users). Please re-review and then I am adding new set of fixtures for the new format.

For the reference, this is how report now looks like when it reaches proxy (ruby format):

```ruby
{"host"=>"localhost",
 "reported_at"=>"2021-09-21T13:23:32.250078",
 "metrics"=>{"time"=>{"total"=>0}},
 "status"=>{"applied"=>0, "failed"=>0, "skipped"=>0},
 "logs"=>
  [{"diff"=>
     {"before"=>{"path"=>"/tmp/hosts"}, "after"=>{"path"=>"/tmp/hosts"}},
    "path"=>"/tmp/hosts",
    "changed"=>false,
    "uid"=>1001,
    "gid"=>1001,
    "owner"=>"lzap",
    "group"=>"lzap",
    "mode"=>"0644",
    "state"=>"file",
    "secontext"=>"unconfined_u:object_r:user_home_t:s0",
    "size"=>446,
    "invocation"=>
     {"module_args"=>
       {"dest"=>"/tmp/hosts",
        "_original_basename"=>"hosts",
        "recurse"=>false,
        "state"=>"file",
        "path"=>"/tmp/hosts",
        "force"=>false,
        "follow"=>true,
        "modification_time_format"=>"%Y%m%d%H%M.%S",
        "access_time_format"=>"%Y%m%d%H%M.%S"}},
    "checksum"=>"3a79e40c022cabdde69982e706520194f319500f",
    "dest"=>"/tmp/hosts",
    "_ansible_no_log"=>false,
    "failed"=>false,
    "module"=>"ansible.builtin.copy",
    "name"=>"ansible.builtin.copy"}],
 "check_mode"=>false}
```

plugins/callback/foreman.py
evgeni: Just don't set it? You declare it as `required: True`, so the user needs to set it.

evgeni:
CI fuckup unrelated, fixing in https://github.com/theforeman/foreman-ansible-modules/pull/1288

evgeni:
CI should work again, would you please rebase?

plugins/callback/foreman.py
ofedoren: I still see this error:

```
[WARNING]: Failure using method (v2_playbook_on_stats) in callback plugin (<ansible.plugins.callback.foreman.CallbackModule object at 0x7fe990b01810>): 'list' object has no attribute 'get'
Callback Exception: 
  File "/usr/lib/python2.7/site-packages/ansible/executor/task_queue_manager.py", line 381, in send_callback
    method(*new_args, **kwargs)
   File "/home/vagrant/foreman-ansible-modules/plugins/callback/foreman.py", line 386, in v2_playbook_on_stats
    self.send_reports(stats)
   File "/home/vagrant/foreman-ansible-modules/plugins/callback/foreman.py", line 341, in send_reports
    self.send_reports_foreman_legacy(stats)
   File "/home/vagrant/foreman-ansible-modules/plugins/callback/foreman.py", line 327, in send_reports_foreman_legacy
    "logs": list(build_log_legacy(self.items[host])),
   File "/home/vagrant/foreman-ansible-modules/plugins/callback/foreman.py", line 133, in build_log_legacy
    if data.get('failed'):
```

plugins/callback/foreman.py
```diff
@@ -281,17 +333,37 @@ def send_reports(self, stats):
...
 
+    def send_reports(self, stats):
+        if self.report_type == "foreman":
+            self.send_reports_foreman_legacy(stats)
+        elif self.report_type == "proxy":
+            self.send_reports_proxy_host_report(stats)
+        else:
+            self._display.warning(u'Unknown foreman endpoint type: {type}'.format(type=self.report_type))
+
+    def drop_nones(self, d: dict) -> dict:
```
ofedoren: AFAIU this is not supported by python 2.7.5 or at least my (ansible 2.9.18) version. Do we want to make this compatible with older ansible versions?

evgeni: This *has* to run on Python 2.7 and Ansible 2.9.

ofedoren: damn...

lzap: Yep you are right, I rebased and it fails for 2.7. Damn, I am not good in Python. Let me try to figure this out.

lzap: Ok thanks for help, rebased with 2.7 compatibility :-)

plugins/callback/foreman.py
lzap: When do you see this error? When in foreman (legacy) mode? It works for me.

ofedoren: Yeap, still the same error. Used `report_type = foreman` and used both custom playbook and `-m ansible.builtin.copy -a "src=/etc/hosts dest=/tmp/hosts"`, same results.

lzap:
Can you re-review @evgeni ? I am going to add new snapshot/fixture tests now.

lzap:
This is weird, I can't find why fixtures tests do not work on this branch. They just pass but not checking any files as the ansible creates no files in the temporary directory for some reason:

```diff
diff --git a/tests/test_callback.py b/tests/test_callback.py
index 5a4a5fc5..370aaafa 100644
--- a/tests/test_callback.py
+++ b/tests/test_callback.py
@@ -22,6 +22,7 @@ def run_playbook_callback(tmpdir):
     extra_env['ANSIBLE_LOAD_CALLBACK_PLUGINS'] = "1"
     # No connection is actually performed during the test
     extra_env['FOREMAN_URL'] = "http://localhost"
+    extra_env['FOREMAN_PROXY_URL'] = "http://localhost"
     extra_env['FOREMAN_SSL_CERT'] = "/dev/zero"
     extra_env['FOREMAN_SSL_KEY'] = "/dev/zero"
     extra_env['FOREMAN_DIR_STORE'] = tmpdir.strpath
@@ -33,6 +34,7 @@ def run_playbook_callback(tmpdir):
 def test_callback(tmpdir, vcrmode):
     run = run_playbook_callback(tmpdir)
     assert run.rc == 0
+    assert len(tmpdir.listdir()) > 0 # this fails
     for real_file in tmpdir.listdir(sort=True):
         contents = real_file.read()
         contents = re.sub(r"\d+-\d+-\d+[ T]\d+:\d+:\d+\.\d+", "2000-01-01 12:00:00.0000", contents)
```

For the record, we have decided also to create smart proxy endpoint for facts, so it will be either foreman or proxy for both endpoints. This will land as a separate PR tho, I want to solve reports first.

plugins/callback/foreman.py
lzap: Ok now I am getting it. Will fix.

lzap:
FINALLY it is now passing both for legacy format and the new format. To check how it looks like for both options, check `test/fixtures/callback/dir_store/{foreman,proxy}` and compare. It is a bit more clean I believe.

This is ready for the final blessing and merge! Thanks for patience.

ofedoren (approved):
Thanks, @lzap! LGTM :)

UPD: Works great for me, but there are failed tests :(

lzap:
There are failures tho :-( `'CallbackModule' object has no attribute 'report_type'`

I think this is because the test sets `FOREMAN_REPORT_TYPE` env variable and the test is against some different Ansible version that has old callback?! Not sure.

lzap:
My only idea is that in some cases, `set_options` is not called and therefore `self.foreman_type` is undefined? Trying to ensure the default value is present in constructor to see if it fixes it.

lzap:
Fixed lint checker, crossing my fingers.

lzap:
Almost there, only Ansible 2.8 is failing. There is a warning:

```
[WARNING]: Failure using method (v2_playbook_on_stats) in callback plugin
(<ansible.plugins.callback.theforeman.foreman.foreman.CallbackModule object at
0x7f766d231970>): 'CallbackModule' object has no attribute 'report_type'
```

Then the test fails because the default value is "foreman" therefore test fixtures are written in a different temporary directory and the one that is expected ("proxy") is empty.

For some reason, this Ansible version does not like this option. There is a common in the description but @evgeni told me to enclose it in quotes, which I did.

plugins/callback/foreman.py
```diff
@@ -281,17 +334,40 @@ def send_reports(self, stats):
...
 
+    def send_reports(self, stats):
+        if self.report_type == "foreman":
+            self.send_reports_foreman_legacy(stats)
+        elif self.report_type == "proxy":
+            self.send_reports_proxy_host_report(stats)
+        else:
+            self._display.warning(u'Unknown foreman endpoint type: {type}'.format(type=self.report_type))
+
+    def drop_nones(self, d):
```
evgeni, line 350: Is that the same as `_recursive_dict_without_none` in `plugins/module_utils/foreman_helper.py`?

https://github.com/theforeman/foreman-ansible-modules/blob/db9b4a5e61e0f0c54c8e994f91fee0586aee3bd5/plugins/module_utils/foreman_helper.py#L1675-L1693

evgeni:
Well, turns out Ansible 2.8 (actually, anything before 2.9.3) is broken with collection callbacks that want a config:
* Fix issue with callbacks set_options method that was not called with collections (from https://github.com/ansible/ansible/blob/stable-2.9/changelogs/CHANGELOG-v2.9.rst#bugfixes-23)

Given we're planing to have breaking changes anyways soon, I guess we can skip 2.8 tests now and drop 2.8 support soon.

lzap:
Allright implemented "if ansible_version < 2.9".

lzap:
It is green, except the two irrelevant tests! Yay!

plugins/callback/foreman.py
```diff
@@ -31,6 +40,16 @@
         ini:
           - section: callback_foreman
             key: url
+      proxy_url:
+        description:
+          - URL of the Foreman Smart Proxy server.
+        env:
+          - name: FOREMAN_PROXY_URL
+        required: True
+        default: https://localhost:8443
```
```suggestion
```

it's `required: True`, please don't set a default (which won't work, as the cert doesn't validate for localhost, anyways)

tests/test_callback.py
```diff
@@ -30,23 +34,54 @@ def run_playbook_callback(tmpdir):
...
+def drop_incompatible_items(d):
+    """Recursively drop incompatible items before assertion"""
+    dd = {}
+    for k, v in d.items():
+        if isinstance(v, dict):
+            dd[k] = drop_incompatible_items(v)
+        elif isinstance(v, (list, set, tuple)):
+            dd[k] = type(v)(drop_incompatible_items(vv) if isinstance(vv, dict) else vv
+                            for vv in v)
+        elif re.match(r'^(?!msg|start|end|delta)', k):
```
evgeni: why a `re.match`? wouldn't `k not in ['msg', 'start', 'end', 'delta']` be the same and much more readable?

tests/test_callback.py
```diff
@@ -30,23 +34,54 @@ def run_playbook_callback(tmpdir):
     return run_playbook(playbook, inventory=inventory, extra_env=extra_env)
 
 
-def test_callback(tmpdir, vcrmode):
-    run = run_playbook_callback(tmpdir)
+def drop_incompatible_items(d):
+    """Recursively drop incompatible items before assertion"""
```
evgeni: incompatible with what? isn't this more "remove report items that vary on individual invocations and can't be reasonable fixed"?

tests/test_callback.py
```diff
@@ -30,23 +34,54 @@ def run_playbook_callback(tmpdir):
...
+                            for vv in v)
+        elif re.match(r'^(?!msg|start|end|delta)', k):
+            dd[k] = v
+    return dd
+
+
+def run_callback(tmpdir, endpoint_type, vcrmode):
+    run = run_playbook_callback(tmpdir, endpoint_type)
     assert run.rc == 0
+    assert len(tmpdir.listdir()) > 0
```
evgeni: *technically* an empty list is falsy, so you could `assert tmpdir.listdir()`, but I am not sure what drives the message "the result directory is not empty" more

tests/test_callback.py
```diff
@@ -30,23 +34,54 @@ def run_playbook_callback(tmpdir):
...
     for real_file in tmpdir.listdir(sort=True):
         contents = real_file.read()
         contents = re.sub(r"\d+-\d+-\d+[ T]\d+:\d+:\d+\.\d+", "2000-01-01 12:00:00.0000", contents)
         contents = re.sub(r"\d+:\d+:\d+\.\d+", "12:00:00.0000", contents)
-        contents = re.sub(r", \\\"msg\\\": \\\"\\\"", "", contents)
+        if endpoint_type == "foreman":
+            # drop_incompatible_items cannot be used for the legacy format
```
evgeni, line 64: why not? you're doing it on line 77/78 (albeit on the JSON representation, not on the string one)

plugins/callback/foreman.py
evgeni: this comment is still open ;)

plugins/callback/foreman.py
```diff
@@ -19,6 +19,15 @@
       - whitelisting in configuration
       - requests (python library)
     options:
+      report_type:
```
evgeni, line 22: I wonder if this is even needed, given you need to configure a different URL.

Shouldn't "proxy_url is configured, send proxy format, otherwise send foreman format" be sufficient?

plugins/callback/foreman.py
```diff
@@ -281,17 +334,40 @@ def send_reports(self, stats):
...
-            self._send_data('report', host, report)
-
+            self._send_data('report', 'foreman', host, report)
             self.items[host] = []
 
+    def send_reports(self, stats):
+        if self.report_type == "foreman":
+            self.send_reports_foreman_legacy(stats)
+        elif self.report_type == "proxy":
+            self.send_reports_proxy_host_report(stats)
```
evgeni: why is the one called `foreman_legacy` and the other `proxy_host_report`? wouldn't `foreman` and `proxy` be sufficient? or will there also be a new foreman type in the future?

certainly not blocking, but wondering why the methods aren't just called `send_reports_<report_type>`, which would make reading the code easier

tests/fixtures/callback/dir_store/proxy/testhost.json
```diff
@@ -0,0 +1,278 @@
...
+        "module_args": {
+          "_raw_params": "echo foo",
+          "_uses_shell": false,
+          "stdin_add_newline": true,
+          "strip_empty_ends": true,
+          "warn": false
+        }
+      },
+      "module": "command",
+      "name": "Changed task",
```
evgeni: not sure if that is also true for the old format (think not, as there everything is folded into the `messages` subentry), but I'm seeing it here, so commenting here.

what if the module *returns* a `name`, would that overwrite this entry? same for `module`. I think it's wrong to mix data coming from Ansible and the metadata *you* generate.

evgeni (changes requested):

plugins/callback/foreman.py
lzap: Shall I delete the default value also from foreman_url?

Hmm I am wondering - should it be required since there are two options now. Previously it made sense, there was just a one URL, now there are two. Although the plan is to move facts too into a plugin. 

I think let's just make both required, we should be working on facts in this release.

tests/test_callback.py
lzap: Indeed.

tests/test_callback.py
lzap: Sure let me explain this better.

tests/test_callback.py
lzap: Actually no, this is still needed. The `drop_incompatible_method` does not drill down to JSON-encoded strings :-)

tests/test_callback.py
lzap: Okay, let me add a nicer assertion message then.

plugins/callback/foreman.py
lzap: Sure, yeah that is better.

plugins/callback/foreman.py
evgeni: please don't touch `foreman_url`, it's still there for not breaking backwards compatibility (i.e.: not changing defaults).

but you raise a good question, it should probaly not be `required: True`

tests/fixtures/callback/dir_store/proxy/testhost.json
lzap: Good point, we actually do the same for "name", "failed" and "module". Since this is the best time for incompatible change, let's perhaps introduce a subhash named "foreman" and store everything in there. Do you think "foreman" is safe enough? I think so, but I could also do "_foreman" instead.

evgeni: nothing *forbids* a module to return a `foreman` value, but the chances that this happens is way lower. and `_foreman` is probably even better.

lzap: Actually this made me to completely change the format from scratch. It's now like this:

```json
 "results": [
    {
      "failed": false,
      "result": {
        "_ansible_no_log": false,
        "changed": true,
        "cmd": [
          "echo",
          "foo"
        ]
      },
      "task": {
        "action": "command",
        "any_errors_fatal": false,
        "async": 0,
        "async_val": 0,
        "uuid": "1c697a0b-ae03-9772-67e4-00000000000e"
      }
    }
  ]
```

I shortened the example.

Root of each array element is ours, we can create any key we want, but subdicts "result" and "task" are taken directly from Ansible, no new keys added there. We just filter out None values, empty arrays/dics and drop some too verbose dicts (args).

plugins/callback/foreman.py
lzap: Oh I just changed the method to also drop empty arrays/dicts. But my implementation is not perfect - it does not detect empty dict in array (`[{}]`) but I think it is still useble.

Should I move it here?

Edit: Oh I thought this is some kind of helper (utility), looks like this is a base AnsibleModule class. Not sure how to approach this.

lzap: So I can perhaps move the method to foreman_helper and import it from the callback, do you want me to do this? I am not sure how useful this will be. Maybe if you won't like dropping empty dicts/arrays then I could reuse the existing implementation.

lzap:
Amended all your comments, the remaining open question is to reuse or not reuse the existing method from foreman_helper. Mine is slightly different but its not perfect, I don't mind just dropping None values (not empty arrays/dicsts).

lzap:
Wow all green!

ofedoren (commented):
Hmm, @lzap, I've tried again and now I see:
```
Callback Exception: 
  File "/usr/lib/python2.7/site-packages/ansible/executor/task_queue_manager.py", line 381, in send_callback
    method(*new_args, **kwargs)
   File "/home/vagrant/foreman-ansible-modules/plugins/callback/foreman.py", line 398, in v2_runner_on_ok
    self.append_result(result)
   File "/home/vagrant/foreman-ansible-modules/plugins/callback/foreman.py", line 377, in append_result
    if 'ansible_facts' in result:

```

And also I don't see any updated facts by Ansible :(

lzap:
I fixed Ansible facts, also the drop method now also recognizes `[{}]` so the output is nice and clean.

lzap:
@evgeni could you re-review please? Thanks!

plugins/callback/foreman.py
lzap: Can do a different PR if you want me to remove the required flag. But this should be deprecated pretty soon - I plan working also on the facts endpoint so everything should go through the proxy in the future.

plugins/callback/foreman.py
lzap: There will be some period when both endpoints will exist (2 for reports, 2 for facts) and user will need to decide which one to use. I actually plan to introduce also `fact_type` with similar semantics.

But after foreman_url is deprecated AND removed, we can drop the type completely. Unless you have a better plan.

ofedoren (approved):
Re-approving from integration with host_reports perspective.

tests/test_callback.py
evgeni: I think the correct comment would be "drop_incompatible_items can't filter *all* problems in the legacy format", but that's cosmetics

plugins/callback/foreman.py
evgeni: So I was thinking about importing `_recursive_dict_without_none` and using that instead of your method. but we can play around with that part later.

plugins/callback/foreman.py
evgeni: My trail of thought was like this:
* let's make `foreman_url` and `proxy_url` optional
* if the user defines `foreman_url` the report type is automatically Foreman
* if the user defines `proxy_url` the report type is automatically Proxy
* if the user defines both… not sure :)

evgeni: or wont that work because you might want/need to send reports one way, and facts another?

plugins/callback/foreman.py
```diff
@@ -210,11 +233,13 @@ def _ssl_verify(self, option):
 
         return verify
 
-    def _send_data(self, endpoint, host, data):
-        if endpoint == 'facts':
+    def _send_data(self, type, endpoint, host, data):
```
evgeni: `type` is a python builtin, and you're shadowing it, please use something else :)

plugins/callback/foreman.py
```diff
@@ -31,6 +40,15 @@
         ini:
           - section: callback_foreman
             key: url
+      proxy_url:
+        description:
+          - URL of the Foreman Smart Proxy server.
+        env:
+          - name: FOREMAN_PROXY_URL
+        required: True
```
evgeni: I still think this shouldn't be required, as the `report_type` defaults to `foreman`.

plugins/callback/foreman.py
lzap: Renamed weird that flake8 did not complain about it.

plugins/callback/foreman.py
lzap: Removed require statement.

lzap:
The failed test is the RHEL role - irrelevant I believe.

plugins/callback/foreman.py
```diff
@@ -210,13 +232,15 @@ def _ssl_verify(self, option):
 
         return verify
 
-    def _send_data(self, endpoint, host, data):
-        if endpoint == 'facts':
+    def _send_data(self, report_type, endpoint, host, data):
```
evgeni: in the rest of the code, `report_type` is `foreman` or `proxy`, but here it's `facts` or `report`.

how about this?
```suggestion
    def _send_data(self, data_type, report_type, host, data):
```

plugins/callback/foreman.py
```diff
@@ -210,13 +232,15 @@ def _ssl_verify(self, option):
...
+        if report_type == 'facts':
             url = self.foreman_url + '/api/v2/hosts/facts'
-        elif endpoint == 'report':
+        elif report_type == 'report' and endpoint == 'foreman':
             url = self.foreman_url + '/api/v2/config_reports'
+        elif report_type == 'report' and endpoint == 'proxy':
+            url = self.proxy_url + '/host_reports/ansible'
         else:
-            self._display.warning(u'Unknown endpoint type: {type}'.format(type=endpoint))
+            self._display.warning(u'Unknown endpoint type: {report_type}'.format(report_type=endpoint))
```
evgeni: What's an "endpoint type", the user has never before heard this word ;)

tests/test_callback.py
```diff
@@ -21,7 +23,9 @@ def run_playbook_callback(tmpdir):
     extra_env['ANSIBLE_STDOUT_CALLBACK'] = "theforeman.foreman.foreman"
     extra_env['ANSIBLE_LOAD_CALLBACK_PLUGINS'] = "1"
     # No connection is actually performed during the test
+    extra_env['FOREMAN_REPORT_TYPE'] = endpoint_type
     extra_env['FOREMAN_URL'] = "http://localhost"
+    extra_env['FOREMAN_PROXY_URL'] = "http://localhost"
```
evgeni: can you only inject this for `endpoint_type == 'proxy'` so the non-requiredness of the parameter is ensured?

btw, `report_type`, please.

plugins/callback/foreman.py
lzap: Makes sense, I blindly renamed the attribute.

plugins/callback/foreman.py
lzap: Right, renamed. Once facts are implemented, I can improve the message further.

tests/test_callback.py
lzap: Yeah sure, better name for the future.

lzap:
Done, made only the requested changes. Tests are green! And I tested it.

evgeni (approved):

lzap:
Thank you so much for your help!

PR #1271, opened 2021-08-10
don't accidentally duplicate `kt_activation_keys` param

### LINKED ISSUE #1268: kt_activation_keys in parameters get duplicated
##### SUMMARY

if you pass

```
parameters:
  - name: kt_activation_keys
    value: lol
```

to `hostgroup`, you end up with the duplicate param error on the second run of the playbook, as the below code doesn't check if the param actually needs to be appended again

https://github.com/theforeman/foreman-ansible-modules/blob/604f60b884707ee25282155c1fe5b5f0146a943d/plugins/module_utils/foreman_helper.py#L320-L330

##### ISSUE TYPE
 - Bug Report

##### ANSIBLE VERSION

```

```

##### COLLECTION VERSION

```

```

##### KATELLO/FOREMAN VERSION

```

```

##### STEPS TO REPRODUCE

```yaml

```

##### ACTUAL RESULTS

```

```

evgeni:
Fixes: #1268

### DIFF
--- a/plugins/module_utils/foreman_helper.py
+++ b/plugins/module_utils/foreman_helper.py
@@ -325,9 +325,10 @@ class HostMixin(ParametersMixin):
                 ak_param = {'name': 'kt_activation_keys', 'parameter_type': 'string', 'value': self.foreman_params.pop('activation_keys')}
                 self.foreman_params['parameters'] = parameters + [ak_param]
             elif 'parameters' in self.foreman_params and entity is not None:
-                ak_param = next((param for param in entity.get('parameters') if param['name'] == 'kt_activation_keys'), None)
-                if ak_param:
-                    self.foreman_params['parameters'].append(ak_param)
+                current_ak_param = next((param for param in entity.get('parameters') if param['name'] == 'kt_activation_keys'), None)
+                desired_ak_param = next((param for param in self.foreman_params['parameters'] if param['name'] == 'kt_activation_keys'), None)
+                if current_ak_param and desired_ak_param is None:
+                    self.foreman_params['parameters'].append(current_ak_param)
 
         self.validate_parameters()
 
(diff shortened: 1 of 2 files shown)


plugins/module_utils/foreman_helper.py
```diff
@@ -325,9 +325,10 @@ def run(self, **kwargs):
                 ak_param = {'name': 'kt_activation_keys', 'parameter_type': 'string', 'value': self.foreman_params.pop('activation_keys')}
                 self.foreman_params['parameters'] = parameters + [ak_param]
             elif 'parameters' in self.foreman_params and entity is not None:
-                ak_param = next((param for param in entity.get('parameters') if param['name'] == 'kt_activation_keys'), None)
-                if ak_param:
-                    self.foreman_params['parameters'].append(ak_param)
+                entity_ak_param = next((param for param in entity.get('parameters') if param['name'] == 'kt_activation_keys'), None)
+                user_ak_param = next((param for param in self.foreman_params['parameters'] if param['name'] == 'kt_activation_keys'), None)
```
mdellweg: we used "current" and "desired" as prefixes in other places.

plugins/module_utils/foreman_helper.py
```diff
@@ -325,9 +325,10 @@ def run(self, **kwargs):
                 ak_param = {'name': 'kt_activation_keys', 'parameter_type': 'string', 'value': self.foreman_params.pop('activation_keys')}
                 self.foreman_params['parameters'] = parameters + [ak_param]
```
mdellweg, line 326: I think, it can still happen here, that you end up with doubled "kt_activation_keys" if the user specified it in both ways.
Arguably, this is not a proper way to call the module.

evgeni: If the user specified it in both places, they have earned the error.

plugins/module_utils/foreman_helper.py
evgeni: See, that's what two weeks of PTO do to your brain.

plugins/module_utils/foreman_helper.py
mdellweg: :trumpet:

mdellweg (approved):

PR #1277, opened 2021-08-24
release 2.2.0

evgeni:


Fodoj:
image

mdellweg (approved):

mdellweg:
That's a hell of an error message: `ERROR! Unexpected Exception, this is probably a bug: 'results'`
Can we call it a misfire and just move on?

evgeni:
>  Can we call it a misfire and just move on?

that's exactly what I was about to do!

evgeni:
released @ https://galaxy.ansible.com/theforeman/foreman

FTR, there will be galaxy maintenance in like 5 minutes, for up to 3 hours, so things might not work :)

PR #1280, opened 2021-08-26
Remove content view version from environments if empty

ehelms:
Probably needs some testing added, did not want to forget about this change I encountered doing some cleanup tasks.

### DIFF
--- a/plugins/modules/content_view_version.py
+++ b/plugins/modules/content_view_version.py
@@ -252,13 +252,19 @@ def main():
                     content_view_version = {'id': -1, 'environments': []}
 
             if 'lifecycle_environments' in module.foreman_params:
-                promote_content_view_version(
-                    module,
-                    content_view_version,
-                    module.foreman_params['lifecycle_environments'],
-                    force=module.foreman_params['force_promote'],
-                    force_yum_metadata_regeneration=module.foreman_params['force_yum_metadata_regeneration'],
-                )
+                if module.foreman_params['lifecycle_environments'] == [] and content_view_version['environments']:
+                    current_environment_ids = {environment['id'] for environment in content_view_version['environments']}
+
+                    for environment_id in current_environment_ids:
+                        response = module.resource_action('content_views', 'remove_from_environment', params={'id': content_view['id'], 'environment_id': environment_id})
+                else:
+                    promote_content_view_version(
+                        module,
+                        content_view_version,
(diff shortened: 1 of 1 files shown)


plugins/modules/content_view_version.py
```diff
@@ -252,13 +252,19 @@ def main():
...
-                    content_view_version,
-                    module.foreman_params['lifecycle_environments'],
-                    force=module.foreman_params['force_promote'],
-                    force_yum_metadata_regeneration=module.foreman_params['force_yum_metadata_regeneration'],
-                )
+                if module.foreman_params['lifecycle_environments'] == [] and content_view_version['environments']:
+                    current_environment_ids = {environment['id'] for environment in content_view_version['environments']}
+
+                    for environment_id in current_environment_ids:
+                        response = module.resource_action('content_views', 'remove_from_environment', params={'id': content_view['id'], 'environment_id': environment_id})
```
ehelms, line 259: On newer releases of Katello, I *think* this can be done in one call via:

```
response = module.resource_action('content_views_versions', 'remove_from_environments', params={'id': content_view['id'], 'environment_ids': current_environment_ids})
```

evgeni: Newer is 4.0+? If there is a possible perfomance gain, I'd be in favor of an if/else?

BTW, the linter says this line is too damn long (170>160 chars) ;)

plugins/modules/content_view_version.py
```diff
@@ -252,13 +252,19 @@ def main():
...
 
             if 'lifecycle_environments' in module.foreman_params:
-                promote_content_view_version(
-                    module,
-                    content_view_version,
-                    module.foreman_params['lifecycle_environments'],
-                    force=module.foreman_params['force_promote'],
-                    force_yum_metadata_regeneration=module.foreman_params['force_yum_metadata_regeneration'],
-                )
+                if module.foreman_params['lifecycle_environments'] == [] and content_view_version['environments']:
```
evgeni, line 255: Should this happen only if the user says `[]` or always when the list doesn't match?

The later has some complications as we never list `Library` and I think the current user expectation is "promote to these envs, don't touch the others"?

plugins/modules/content_view_version.py
evgeni: I can't find this in the Katello API?

Issue #1231, opened 2021-05-20
hostgroup update with puppetclasses or description


achevalet:
##### SUMMARY
Hi, I've some issues with the hostgroup module:
- it seems that `puppetclasses` cannot be updated
- updating `description` when passing `puppetclasses` fails
- `puppetclass_ids` is not returned at creation

##### ISSUE TYPE
 - Bug Report

##### ANSIBLE VERSION

```
ansible 2.10.6
  config file = None
  configured module search path = ['/home/neo/.ansible/plugins/modules', '/usr/share/ansible/plugins/modules']
  ansible python module location = /usr/local/lib/python3.6/site-packages/ansible
  executable location = /usr/local/bin/ansible
  python version = 3.6.8 (default, Nov 16 2020, 16:55:22) [GCC 4.8.5 20150623 (Red Hat 4.8.5-44)]

```

##### COLLECTION VERSION

```
# /usr/local/lib/python3.6/site-packages/ansible_collections
Collection         Version
------------------ -------
theforeman.foreman 2.1.0  

```

##### KATELLO/FOREMAN VERSION

```
tfm-rubygem-katello-3.14.1-1.el7.noarch
foreman-1.24.3-1.el7.noarch

```

##### STEPS TO REPRODUCE

```yaml
---
- hosts: localhost
  collections:
    - theforeman.foreman
  gather_facts: false
  tasks:
    - name: Ensure Hostgroup '{{ name }}' is {{ state }}
      hostgroup:
        name: "{{ name }}"
        updated_name: "{{ updated_name | default(omit) }}"
        parent: "{{ parent | default(omit) }}"
        description: "{{ description | default(omit) }}"
        locations: "{{ locations | default(omit) }}"
        organizations: "{{ organizations | default(omit) }}"
        compute_resource: "{{ compute_resource | default(omit) }}"
        compute_profile: "{{ compute_profile | default(omit) }}"
        domain: "{{ domain | default(omit) }}"
        subnet: "{{ subnet | default(omit) }}"
        subnet6: "{{ subnet6 | default(omit) }}"
        realm: "{{ realm | default(omit) }}"
        architecture: "{{ architecture | default(omit) }}"
        operatingsystem: "{{ operatingsystem | default(omit) }}"
        medium: "{{ installation_medium | default(omit) }}"
        ptable: "{{ ptable | default(omit) }}"
        pxe_loader: "{{ pxe_loader | default(omit) }}"
[...]

PR #1281, opened 2021-08-27
really fix ensure_puppetclasses

evgeni:


plugins/module_utils/foreman_helper.py
```diff
@@ -1096,14 +1096,14 @@ def _update_entity(self, resource, desired_entity, current_entity, params, forem
...
-            payload['id'] = current_entity['id']
+            payload['id'] = current_flat_entity['id']
             if not self.check_mode:
                 if params:
                     payload.update(params)
                 return self.resource_action(resource, 'update', payload)
             else:
                 # In check_mode we emulate the server updating the entity
-                fake_entity = current_entity.copy()
+                fake_entity = current_flat_entity.copy()
```
evgeni, line 1106: this one is "interesting", as in theory we need the `current_entity`, but then I am not sure that `update(payload)` would actually do something useful :(

evgeni:
Heh, so @mdellweg, this now fails the fixtures for `host`, because :drum: you fixed the "update host puppetclasses" branch, but it still wasn't just totally right :laughing:

tests/test_playbooks/fixtures/host-0.yml
```diff
@@ -33,7 +33,7 @@ interactions:
       Foreman_current_organization:
       - ; ANY
       Foreman_version:
-      - 2.3.0
+      - 3.0.0-rc1
```
evgeni: so technically this is not run against a plain foreman, but foreman+foreman_puppet now… let's see if old apidocs still match :D

tests/test_playbooks/fixtures/host-26.yml
```diff
@@ -114,6 +114,66 @@ interactions:
...
+      Accept:
+      - application/json;version=2
+      Accept-Encoding:
+      - gzip, deflate
+      Connection:
+      - keep-alive
+      User-Agent:
+      - apypie (https://github.com/Apipie/apypie)
+    method: GET
+    uri: https://foreman.example.org/foreman_puppet/api/environments?search=name%3D%22production%22&per_page=4294967296
```
evgeni: meh, this won't work, I bet

evgeni:
this is missing:
- [x] changelog
- [x] test for the actual issue in #1231, setting `description` (or really, any other normal attribute) while also providing (but not changing) `puppetclasses`

plugins/module_utils/foreman_helper.py
evgeni: huh, how about

```
fake_entity = current_entity.copy()
fake_entity.update(current_flat_entity)
fake_entity.update(payload)
```

evgeni: @mdellweg what did we end up deciding here? I cant' remember :(

evgeni:
CI errors unrelated

plugins/module_utils/foreman_helper.py
mdellweg: As the payload is only updating the flattened values, copying the `current_entity` is probably dangerous / misleading.
So i think, we should leave it as is.

plugins/module_utils/foreman_helper.py
```diff
@@ -1779,10 +1779,11 @@ def build_fqn(name, parent=None):
 
 # Helper for puppetclasses
 def ensure_puppetclasses(module, entity_type, entity, expected_puppetclasses=None):
-    if expected_puppetclasses:
+    if expected_puppetclasses is not None:
         puppetclasses_resource = '{0}_classes'.format(entity_type)
         expected_puppetclasses = module.find_puppetclasses(expected_puppetclasses, environment=entity['environment_id'], thin=True)
-        current_puppetclass_ids = entity.get('puppetclass_ids', [])
+        current_puppetclasses = entity.get('puppetclasses', [])
```
mdellweg, line 1785: Will this be available in the `fake_entity` created from the `flattened_current_entity`?

mdellweg (approved):
If i understand the change correctly, we may report slightly different (but still correct on their own) results in "normal" and "check mode" operation. However, this may not even be new.
This solves a bug; we should just take it.

plugins/module_utils/foreman_helper.py
evgeni: OK

evgeni:
> If i understand the change correctly, we may report slightly different (but still correct on their own) results in "normal" and "check mode" operation. However, this may not even be new.

Yeah, the check mode output was always a bit opportunistic, as we never ask the server how it would change it.

plugins/module_utils/foreman_helper.py
evgeni: No.

PR #1282, opened 2021-09-07
add arch parameter to repository module

### LINKED ISSUE #1265: Add arch option to theforeman.foreman.repository module
##### SUMMARY
In Satellite GUI when we create product repository we can restrict that repository on specific OS version (rhel8, rhel7 ...) and architecture
It would be awesome if we could do same via theforeman.foreman.repository

##### ISSUE TYPE
 - Feature Idea

evgeni:
Fixes: #1265

### DIFF
--- a/changelogs/fragments/1265-repository-arch.yml
+++ b/changelogs/fragments/1265-repository-arch.yml
@@ -1,0 +1,2 @@
+minor_changes:
+  - repository - add ``arch`` parameter to limit architectures of the repository (https://github.com/theforeman/foreman-ansible-modules/issues/1265)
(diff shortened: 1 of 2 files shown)


evgeni:
I wonder if this should be limited to `yum` repos only, talking to Katello about that.

evgeni:
I talked to @jlsherrill and it seems this *should* work for YUM and DEB, but really only works for YUM right now.

mdellweg (approved):

evgeni:
Given that katello *accepts* (and ignores) this on other repo types, we're going to merge as-is, thanks

PR #1283, opened 2021-09-08
[WIP] Module for generating registration command

stejskalleos:


wbclark (commented):
interesting concept @stejskalleos ... thanks for the PR!

understanding that this is a WIP, right now the role looks like a wrapper that provides the same features as the module itself. thanking about what users might like to see from that, how about registering the generated commands in a variable (so that other ansible may refer to it) and giving the role options to output the generated commands in the terminal and/or write them into a file or script on the ansible target host.

let us know please what you envision in terms of what it will look like for users to use this... looking forward to hearing from you

stejskalleos:
>  right now the role looks like a wrapper that provides the same features as the module itself

Yeah, after some thinking I decided to remove the role. This module will be part of another role for converting centos machines to rhel (WIP), so having role just for generating registration commands doesn't make sense.

wbclark:
> Yeah, after some thinking I decided to remove the role. This module will be part of another role for converting centos machines to rhel (WIP), so having role just for generating registration commands doesn't make sense.

Understood. Since you don't need the role right now for your particular goal, but the basic core of a role is still present in the commit history: I could use what is there to create a separate WIP PR so that it's not lost if this gets squashed on merge, meanwhile you would not need to write additional tests for the role, unrelated to your actual goal.

So as to not bog down this PR with further discussion that's not related to your immediate goal, I created an issue #1284 to discuss how a primary Ansible user should interact with the global registration feature. That said, please feel free to weigh in if you have any ideas about how that should work.

wbclark:
Also created #1285 so the initial work on a role is not lost -- thanks @stejskalleos for that

play.yml
```diff
@@ -0,0 +1,37 @@
+# - hosts: localhost
```
wbclark: Heads up, it looks like `play.yml` was committed by accident.

wbclark (commented):
Am I correct in understanding that global registration should use only the Foreman API? So `tests/fixtures/apidoc/registration_command.json` should work as a symlink to `foreman.json` instead of a separately populated file?

stejskalleos:
Thanks for creating https://github.com/theforeman/foreman-ansible-modules/pull/1285,
PR updated, I removed  files that shouldn't be part of the commit and`tests/fixtures/apidoc/registration_command.json` is link to `foreman.json`.

plugins/modules/registration_command.py
```diff
@@ -0,0 +1,89 @@
...
+# You should have received a copy of the GNU General Public License
+# along with this program.  If not, see <http://www.gnu.org/licenses/>.
+
+from __future__ import absolute_import, division, print_function
+__metaclass__ = type
+
+DOCUMENTATION = '''
+---
+module: registration_command
+version_added: 1.0.0
```
wbclark, line 24:
```suggestion
version_added: 2.3.0
```

plugins/modules/registration_command.py
```diff
@@ -0,0 +1,89 @@
...
+            packages=dict(type='str'),
+            update_packages=dict(type='bool'),
+            repo=dict(type='str'),
+            repo_gpg_key_url=dict(type='str'),
+            jwt_expiration=dict(type='int', default=4),
+            remote_execution_interface=dict(type='str'),
+            lifecycle_environment=dict(type='entity', flat_name='lifecycle_environment_id'),
+            ignore_subman_errors=dict(type='bool'),
+            force=dict(type='bool'),
+            activation_keys=dict(type='list', elements='str'),
```
wbclark, line 80: Could you please document these in the `DOCUMENTATION` string above, under `options:`? The activation key module has a nice example of what it looks like: https://github.com/theforeman/foreman-ansible-modules/blob/5503dc8033129fe694f88c41cc75466b5a519d72/plugins/modules/activation_key.py#L30-L157

tests/test_playbooks/tasks/registration_command.yml
```diff
@@ -0,0 +1,21 @@
...
+    rc_location: Test Location
+    rc_state: present
+  registration_command:
+    username: "{{ foreman_username }}"
+    password: "{{ foreman_password }}"
+    server_url: "{{ foreman_server_url }}"
+    validate_certs: "{{ foreman_validate_certs }}"
+    organization: "{{ rc_organization }}"
+    location: "{{ rc_location }}"
+    state: "{{ rc_state }}"
```
wbclark, line 14: Is passing some `state` necessary for this module? What would it mean to ensure state for a registration command?

The documentation in the module has no state in the example:

```
EXAMPLES = '''
- name: "Generate registration command"
  theforeman.foreman.registration_command:
    username: "admin"
    password: "changeme"
'''
```

wbclark (commented):
Hey @stejskalleos could you also add a changelog fragment? Similar to https://github.com/theforeman/foreman-ansible-modules/commit/f4ab021e9bdf704a85a898f47853e488405b2a0e

plugins/module_utils/foreman_helper.py
```diff
@@ -237,7 +237,7 @@ def __init__(self, **kwargs):
 
     def run(self, **kwargs):
         new_entity = super(NestedParametersMixin, self).run(**kwargs)
-        if new_entity:
+        if new_entity and 'id' in new_entity:
```
mdellweg, line 240: Is this another hint, that it really should not be an entity module?

evgeni:
> Hey @stejskalleos could you also add a changelog fragment? Similar to [f4ab021](https://github.com/theforeman/foreman-ansible-modules/commit/f4ab021e9bdf704a85a898f47853e488405b2a0e)

For new modules, no changelog entry is needed, it is autogenerated

wbclark:
> For new modules, no changelog entry is needed, it is autogenerated

Interesting... thank you

PR #1285, opened 2021-09-09
[WIP] Add corresponding role for registration_commands module

wbclark:
Initial work by @stejkalleos

This was dropped from #1283 , therefore creating a separate PR so that the basic work is not lost.

Requires #1283


evgeni:
We now have a `registration_command` module via https://github.com/theforeman/foreman-ansible-modules/pull/1683, but I fail to see which benefit this role gives over calling the module directly?

@wbclark is this still a thing you'd think is needed?

PR #1286, opened 2021-09-20
Convert2RHEL role

stejskalleos:
`Conver2RHEL` role creates a basic configuration & data for
registering and converting CentOS clients to RHEL.
For more description see `roles/convert2rhel/README.md`
**Sample playbook**
```
- hosts: localhost
  roles:
    - role: theforeman.foreman.convert2rhel
      vars:
        foreman_server_url: "https://centos7-katello-devel-stable.example.com"
        foreman_username: "admin"
        foreman_password: "changeme"
        foreman_validate_certs: false
        foreman_organization: "Default Organization"
        foreman_content_rhel_wait_for_syncs: false
        foreman_content_rhel_enable_rhel7: true
        foreman_content_rhel_enable_rhel8: true
        foreman_content_rhel_wait_for_syncs: false
        foreman_content_rhel_rhel8_releasever: 8.4
        foreman_convert2rhel_rhel7_subs:
          - upstream_pool_id: ID
          - upstream_pool_id: ID
        foreman_convert2rhel_rhel8_subs:
          - upstream_pool_id: ID
          - upstream_pool_id: ID
```


roles/content_rhel/README.md
```diff
@@ -30,6 +30,8 @@ It also supports customizing the included roles with:
 
 `foreman_sync_plan_sync_date`: Initial sync date for the sync plan, formatted as 'YYYY-MM-DD HH:MM:SS UTC'.
 
+`foreman_content_rhel_create_ack`: Create activation key. Default `true`.
```
ogajduse: I am a fan of using `ak` instead of `ack`. I slightly remember that `ak` is already used in some of our repositories.

roles/convert2rhel/tasks/c2r_repos.yml
```diff
@@ -0,0 +1,70 @@
+---
+- name: "Create credentials for 'Convert2RHEL' repository"
```
ares: This may not be needed as the CA cert should be preinstalled on RHEL systems. Worth testing though. The less we hardcode the better :-)

requirements.txt
```diff
@@ -1,3 +1,4 @@
 requests>=2.4.2
 ipaddress; python_version < '3.3'
 PyYAML
+jmespath
```
evgeni: We should not depend on that.

roles/content_rhel/README.md
evgeni: any reason not to use `…create_acvtivation(_)key`?

roles/convert2rhel/README.md
```diff
@@ -0,0 +1,42 @@
+theforeman.foreman.convert2rhel
+===========================
+
+This role creates a basic configuration for everything needed to register and convert `CentOS` clients to `Red hat Enterprise Linux`.
```
evgeni: I don't think we should format CentOS an RHEL as code here?

roles/convert2rhel/README.md
```diff
@@ -0,0 +1,42 @@
...
+Role Variables
+--------------
+
+This role supports the [Common Role Variables](https://github.com/theforeman/foreman-ansible-modules/blob/develop/README.md#common-role-variables).
+
+- `foreman_c2r_simple_content_access`: TODO: Simple content access enabled.
+- `foreman_c2r_manifest_path`: Path to manifest, default: `~/manifest.zip`.
+- `foreman_c2r_enable_rhel7`: Enable & sync RHEL 7 repositories, default: `true`.
+- `foreman_c2r_enable_rhel8`: Enable & sync RHEL 8 repositories, default: `true`.
+- `foreman_c2r_rhel8_releasever`: RHEL 8 repository version, required if `foreman_c2r_enable_rhel8` is `true`.
```
evgeni: is there a reason to have these "redefined" instead of using the defaults of `content_rhel` and let the user override the `conent_rhel` values directly?

roles/convert2rhel/README.md
```diff
@@ -0,0 +1,42 @@
...
+First is upload of manifest & sync of `RHEL 7` & `RHEL 8` repositories, for more detail see [content_rhel Role](https://github.com/theforeman/foreman-ansible-modules/blob/develop/roles/content_rhel/README.md).
+
+The role creates then Convert2RHEL products & repositories (with synchronization), activation keys and host groups for each OS.
+
+Role Variables
+--------------
+
+This role supports the [Common Role Variables](https://github.com/theforeman/foreman-ansible-modules/blob/develop/README.md#common-role-variables).
+
+- `foreman_c2r_simple_content_access`: TODO: Simple content access enabled.
```
evgeni: I'd prefer if the prefix would be `foreman_convert2rhel`, not `…_c2r`

roles/convert2rhel/tasks/c2r_repos.yml
evgeni: Pretty sure this is not needed, neither on CentOS nor RHEL.

roles/convert2rhel/tasks/host_groups.yml
```diff
@@ -0,0 +1,29 @@
...
+    password: "{{ foreman_password }}"
+    server_url: "{{ foreman_server_url }}"
+    validate_certs: "{{ foreman_validate_certs | default(omit) }}"
+    organization: "{{ foreman_organization }}"
+    name: "{{ foreman_c2r_hostgroup7 }}"
+    lifecycle_environment: "{{ foreman_c2r_lifecycle_env }}"
+    content_view: "{{ foreman_c2r_content_view }}"
+    parameters:
+      - name: "kt_activation_keys"
+        value: "{{ foreman_c2r_ack_centos7 }}"
```
evgeni: not strictly required, but you could use `activation_keys: "{{ foreman_c2r_ack_centos7 }}"` instead, which will set the right parameter automatically.

roles/convert2rhel/tasks/activation_keys_sca_disabled.yml
```diff
@@ -0,0 +1,68 @@
...
+    username: "{{ foreman_username }}"
+    password: "{{ foreman_password }}"
+    server_url: "{{ foreman_server_url }}"
+    validate_certs: "{{ foreman_validate_certs | default(omit) }}"
+    organization: "{{ foreman_organization }}"
+    name: "{{ foreman_c2r_ack_rhel7 }}"
+    lifecycle_environment: "{{ foreman_c2r_lifecycle_env }}"
+    content_view: "{{ foreman_c2r_content_view }}"
+    auto_attach: false
+    subscriptions: "{{ subscriptions['subscriptions'] | map('json_query','{ name: name } ') | list }}"
```
evgeni: `json_query` needs `jmespath`, and we can't depend on that.

I think `map(attribute='name')` should do the same?

roles/convert2rhel/tasks/activation_keys_sca_enabled.yml
```diff
@@ -0,0 +1,48 @@
+---
```
evgeni: the only difference between this file, and the `_sca_disabled` one is, that this is not adding any subs to the keys, and the other one is, right?

evgeni (changes requested):

roles/convert2rhel/tasks/c2r_repos.yml
stejskalleos: To my surprise it actually doesn't work, after removing this task sync fails with following error:
```
fatal: [localhost]: FAILED! => Task information: 
['Cannot connect to host cdn.redhat.com:443 ssl:default [[SSL: CERTIFICATE_VERIFY_FAILED] certificate verify failed (_ssl.c:877)]']
```

Foreman machine:
```
/etc/os-release
NAME="Red Hat Enterprise Linux Server"
VERSION="7.9 (Maipo)"
ID="rhel"
ID_LIKE="fedora"
VARIANT="Server"
VARIANT_ID="server"
VERSION_ID="7.9"
PRETTY_NAME=RHEL
ANSI_COLOR="0;31"
CPE_NAME="cpe:/o:redhat:enterprise_linux:7.9:GA:server"
HOME_URL="https://www.redhat.com/"
BUG_REPORT_URL="https://bugzilla.redhat.com/"

REDHAT_BUGZILLA_PRODUCT="Red Hat Enterprise Linux 7"
REDHAT_BUGZILLA_PRODUCT_VERSION=7.9
REDHAT_SUPPORT_PRODUCT="Red Hat Enterprise Linux"
REDHAT_SUPPORT_PRODUCT_VERSION="7.9"
```

roles/convert2rhel/tasks/activation_keys_sca_enabled.yml
stejskalleos: Exactly

roles/convert2rhel/tasks/activation_keys_sca_disabled.yml
stejskalleos: TBH I do not understand much how and why this Ansible stuff works like it works and this was the only way that it worked for me.
When I try `{{ subscriptions['subscriptions'] | map(attribute='name') | list }}` I'll get error:
`fatal: [localhost]: FAILED! => {"changed": false, "msg": "dictionary requested, could not parse JSON or key=value"}`

requirements.txt
stejskalleos: I couldn't find the way how to do it without `jmespath`. Suggestions are more than welcomed :)

roles/content_rhel/README.md
ogajduse: > any reason not to use `…create_acvtivation(_)key`?

The only reason I see is a long variable name.
But I am not strongly concerned about it.

roles/convert2rhel/tasks/repositories.yml
```diff
@@ -0,0 +1,70 @@
...
+- name: "Create credentials for 'Convert2RHEL' repository"
+  theforeman.foreman.content_credential:
+    username: "{{ foreman_username }}"
+    password: "{{ foreman_password }}"
+    server_url: "{{ foreman_server_url }}"
+    validate_certs: "{{ foreman_validate_certs | default(omit) }}"
+    organization: "{{ foreman_organization }}"
+    name: "Convert2RHEL CA"
+    content_type: cert
+    content: "{{ lookup('file', 'cdn-redhat-com-chain.pem') }}"
```
ares: I think we don't have to distribute this cert ourselves with the role, instead we could rely on `/etc/rhsm/ca/redhat-uep.pem` being present on RHEL and take the certificate from there

subscription-manager from CentOS won't have it I think but if we leave a note for them to get it from https://raw.githubusercontent.com/candlepin/subscription-manager/main/etc-conf/redhat-uep.pem I think we're fine

ogajduse: How about users with disconnected/air-gapped setups? Would they be fine with the suggested approach?

roles/content_rhel/tasks/main.yml
```diff
@@ -2,6 +2,8 @@
 - name: "Subscription Manifest"
   include_role:
     name: theforeman.foreman.manifest
+  vars:
+    state: refreshed
```
evgeni: the role currently does not support passing `state`, so this definition is a noop.
and if you want to add support for that, please call the var `foreman_manifest_state`

roles/convert2rhel/defaults/main.yml
```diff
@@ -0,0 +1,18 @@
+---
+# Convert2RHEL Products & repos
+convert2rhel_rhel7_product: "Convert2RHEL7"
```
evgeni: I think you accidentally dropped the `foreman_` from the prefix.

roles/convert2rhel/tasks/activation_keys_sca_disabled.yml
evgeni: Interesting. Worked for me just fine?

```yaml
---
- hosts: localhost
  tasks:
    - name: fetch subs
      theforeman.foreman.subscription_info:
        username: admin
        password: changeme
        organization: ACME
        search: "name ~ Infrastructure"
      register: subs
    - name: debug
      debug:
        msg: "{{ subs['subscriptions'] | map(attribute='name') | list }}"
```

Resulted in
```
ok: [localhost] => {
    "msg": [
        "Red Hat Satellite Infrastructure Subscription"
    ]
}
```

roles/convert2rhel/tasks/activation_keys_sca_enabled.yml
evgeni: Can we unify those two files then?

I think the following should work:

```yaml
- name: prepare non-SCA data
  when: not organization['organization']['simple_content_access']
  block:
    - name: "Find subscriptions"
      theforeman.foreman.subscription_info:
        …
    - name: "Set non SCA facts"
      set_fact:
        foreman_convert2rhel_rhel_subs: "{{ subscriptions['subscriptions'] | map(attribute='name') | list }}"
        foreman_convert2rhel_centos7_subs:
          - "Convert2RHEL7"
        foreman_convert2rhel_centos8_subs:
          - "Convert2RHEL8"

- name: "Set SCA facts"
  set_fact:
    foreman_convert2rhel_rhel_subs: "{{ omit }}"
    foreman_convert2rhel_centos7_subs: "{{ omit }}"
    foreman_convert2rhel_centos8_subs: "{{ omit }}"
  when: organization['organization']['simple_content_access']

- name: "Create 'CentOS 7' activation key"
  theforeman.foreman.activation_key:
    validate_certs: "{{ foreman_validate_certs | default(omit) }}"
    organization: "{{ foreman_organization }}"
    name: "{{ convert2rhel_key_centos7 }}"
    lifecycle_environment: "{{ convert2rhel_lifecycle_env }}"
    content_view: "{{ convert2rhel_content_view }}"
    auto_attach: false
    subscriptions: "{{ foreman_convert2rhel_centos7_subs }}"
  when: foreman_content_rhel_enable_rhel7
…
```

roles/convert2rhel/tasks/repositories.yml
evgeni: I honestly don't understand why they decided to host c2r on cdn.redhat.com and not ftp.redhat.com, both support SSL, but the later doesn't require the user to import a dedicated certificate, and it's not like you need a subscription to access the c2r repo…

But we're here, and we want this to work. I don't see a big problem for disconnected/air-gapped setups, those are used to pain (and most would run RHEL anyways). But for Katello users on CentOS, dropping the cert here would produce more pain than needed.

How about `lookup('url', 'https://ftp.redhat.com/redhat/convert2rhel/redhat-uep.pem')` instead, which is the [documented way to obtain this cert for C2R](https://access.redhat.com/documentation/en-us/red_hat_enterprise_linux/8/html-single/converting_from_an_rpm-based_linux_distribution_to_rhel/index#proc_preparing-for-a-rhel-conversion_converting-from-a-linux-distribution-to-rhel)?

roles/convert2rhel/tasks/repositories.yml
```diff
@@ -0,0 +1,70 @@
...
+    server_url: "{{ foreman_server_url }}"
+    validate_certs: "{{ foreman_validate_certs | default(omit) }}"
+    organization: "{{ foreman_organization }}"
+    name: "{{ convert2rhel_rhel7_repo }}"
+    content_type: "yum"
+    product: "{{ convert2rhel_rhel7_product }}"
+    url: "https://cdn.redhat.com/content/public/convert2rhel/7/x86_64/os/"
+    # mirror_on_sync: true
+    verify_ssl_on_sync: true
+    download_policy: immediate
```
evgeni: why?

roles/convert2rhel/tasks/repositories.yml
```diff
@@ -0,0 +1,70 @@
...
+- name: "Synchronize 'Convert2RHEL 7' repository"
+  theforeman.foreman.repository_sync:
+    username: "{{ foreman_username }}"
+    password: "{{ foreman_password }}"
+    server_url: "{{ foreman_server_url }}"
+    validate_certs: "{{ foreman_validate_certs | default(omit) }}"
+    organization: "{{ foreman_organization }}"
+    repository: "{{ convert2rhel_rhel7_repo }}"
+    product: "{{ convert2rhel_rhel7_product }}"
+  when: foreman_content_rhel_enable_rhel7
```
evgeni: the `content_rhel` role has a `foreman_content_rhel_sync_now` variable, to allow the user to decide whether they want a sync, or not.

I *think* we also should honor this here. @wbclark what do you think?

```suggestion
  when: foreman_content_rhel_enable_rhel7 and foreman_content_rhel_sync_now
```

roles/convert2rhel/tasks/main.yml
```diff
@@ -0,0 +1,14 @@
+---
+- name: "Upload Subscription Manifest and sync RHEL repository"
+  include_role:
+    name: theforeman.foreman.content_rhel
+  vars:
+    foreman_content_rhel_wait_for_syncs: false
```
evgeni: not sure how I feel about overriding the default of the `content_rhel` role.

@wbclark thoughts?

roles/convert2rhel/tasks/main.yml
```diff
@@ -0,0 +1,14 @@
...
+- import_tasks: products.yml
+- import_tasks: repositories.yml
```
evgeni: should those two be replaced by the `theforeman.foreman.repositories` role instead?

roles/convert2rhel/tasks/main.yml
evgeni: actually, I think we should not include the role here at all and document that the user needs to call both roles if they don't have rhel content synced yet.

roles/convert2rhel/tasks/activation_keys_sca_disabled.yml
```diff
@@ -0,0 +1,68 @@
+---
+- name: "Find subscriptions"
+  theforeman.foreman.subscription_info:
+    username: "{{ foreman_username }}"
+    password: "{{ foreman_password }}"
+    server_url: "{{ foreman_server_url }}"
+    validate_certs: "{{ foreman_validate_certs | default(omit) }}"
+    organization: "{{ foreman_organization }}"
+    search: "name != {{ convert2rhel_rhel7_product }} and name != {{ convert2rhel_rhel8_product }} || "
```
evgeni: this will list *all* available subscriptions, and add *all* of them to the AK later, which I think is wrong, as the resulting system only needs a single RHEL sub.

additionally, if Katello has multiple subs with the same name (but pool id, e.g. because they were bought at different times), you end up with an error as the "add subs to AK" doesn't know which sub to add (the selection by name is not unique).

I think, instead of trying to guess which subscription the user might have, we should provide a way for them to tell us.

```
foreman_convert2rhel_rhel_sub_name: "Red Hat Enterprise Linux Server, Standard (Physical or Virtual Nodes)"
```

And allow the user to provide a pool id instead, via `foreman_convert2rhel_rhel_sub_pool_id` (or similar)

roles/convert2rhel/tasks/repositories.yml
stejskalleos: So you mean to set `foreman_content_rhel_sync_now` on `convert2rhel` role too? Or can I get value of `foreman_content_rhel_sync_now` from `content_rhel` role?

stejskalleos: Just note: the sync takes few seconds, repo is really small

evgeni: I think (not tested) that if you `include_role`, the variables of that role are still available later in the play, so if the `include_role` remains (see my other comment), you shouldn't need to do anything

But yeah, if the repo is tiny, this might be over-optimization, I agree

roles/convert2rhel/tasks/main.yml
stejskalleos: How does it help? `theforeman.foreman.repositories` still requires to create products before, so `products.yml` is still needed.

evgeni: No, it creates the products for you. See the example in the readme: https://github.com/theforeman/foreman-ansible-modules/tree/develop/roles/repositories#role-variables

evgeni: So the only additional step is to create the content credential. which we also have a role for: https://github.com/theforeman/foreman-ansible-modules/tree/develop/roles/content_credentials

stejskalleos: Hmm, interesting, it didn't worked for me, I had to create products manually ... need to try it again

roles/convert2rhel/tasks/main.yml
stejskalleos: What would be the benefit of removing it and force user to setup & run two playbooks?

roles/convert2rhel/tasks/main.yml
stejskalleos: Found it, note for myself: `repositories:` vs `repository_sets` !!!

stejskalleos:
**Rebased & updated**
- removed `jmespath`
- fixed `convert2rhel_*` vars to `foreman_convert2rhel_*`
- `cdn-redhat-com-chain.pem` is now downloaded from server
- New simple logic for activation keys - if SCA is disabled, user have to manually set `foreman_convert2rhel_rhel7_subs` & `foreman_convert2rhel_rhel8_subs` vars to add subscriptions to the keys
- Using `theforeman.foreman.repositories` for products & repos

roles/convert2rhel/tasks/repositories.yml
ares: ftp.redhat.com had some issue with publishing. While we can download the cert online, why wouldn't we take the one from local FS which is GPG signed? Note that the whole convert2rhel story is aimed to make this easily possible in disconnected environments.

ares: Would it perhaps be possible to first try to get it from FS and if the file is not deployed (e.g. centos) then try to fetch it from ftp.redhat.com?

evgeni: So you say we can fetch RPMs from `cdn.redhat.com`, but we can't fetch a cert from `ftp.redhat.com` (which is the documented way) in an disconnected env? :smirk:

roles/convert2rhel/tasks/main.yml
evgeni: In my opinion, the `content_rhel` role is aimed at the "given a fresh Katello installation, make it possible to consume RHEL" scenario, while `convert2rhel` is something you'd add as a feature much later in the lifecycle of the Katello installation, thus most probably already have done the whole "basic content management" setup. And in this case re-running those steps seems superfluous.

but I also might be wrong in my "expectations what users do" XD

ares: One of the concern is, that customers are unable to select the right repos the conversion needs. That's why convert2rhel guys created their own role, the UI (especially with RHEL 8) is so confusing that users don't know how to find the repo (e.g. rhel 8.4) and which checkbox to click.

If we make this part optional, that may be the best compromise.

stejskalleos:
Updated after discussion with @ares:
- cacert for C2R credentials is first looked up in the `/etc/rhsm/ca/redhat-uep.pem` path and if not found then downloaded from
ftp
- removed vars `foreman_convert2rhel_rhel7_subs` and `foreman_convert2rhel_rhel8_subs` used for non-SCA keys. User have to add subscriptions manually

roles/convert2rhel/tasks/repositories.yml
ares: Well yeah, that does not make any sense :-) the original write up contained info about how to upload rpms manually, but the role of course only pulls the content from the repo. Ok, so fetching from the FS is nice to have, or not even that and feel free to ignore me.

ares:
ha you were faster than I thought :-) leaving up to maintainers to decide, whether the local FS is better or not :-)

ares:
Tests are green, yay, time for another (last? :-) review

roles/convert2rhel/README.md
```diff
@@ -0,0 +1,41 @@
+theforeman.foreman.convert2rhel
+===========================
```
evgeni:
```suggestion
===============================
```

roles/convert2rhel/README.md
```diff
@@ -0,0 +1,41 @@
...
+
+This role creates a basic configuration for everything needed to register and convert CentOS clients to Red hat Enterprise Linux.
+
+First is upload of manifest & sync of RHEL 7 & RHEL 8 repositories, for more detail see [content_rhel Role](https://github.com/theforeman/foreman-ansible-modules/blob/develop/roles/content_rhel/README.md).
```
evgeni: I can't exactly parse that sentence. Did you want to say "First it uploads a manifest and syncs RHEL, …"?

roles/convert2rhel/README.md
```diff
@@ -0,0 +1,41 @@
...
+
+The role creates then Convert2RHEL products & repositories (with synchronization), activation keys and host groups for each OS.
```
evgeni:
```suggestion
The role then creates Convert2RHEL products & repositories (and synchronizes them), activation keys and host groups for each OS.
```

roles/convert2rhel/README.md
```diff
@@ -0,0 +1,41 @@
...
+
+If simple content access is disabled, subscriptions and repositories for RHEL activation keys must be added manually.
+
+Role Variables
+--------------
+
+This role supports the [Common Role Variables](https://github.com/theforeman/foreman-ansible-modules/blob/develop/README.md#common-role-variables) and [Content RHEL variables](https://github.com/theforeman/foreman-ansible-modules/blob/develop/roles/content_rhel/README.md)
+
+- `convert2rhel_lifecycle_env`: Lifecycle environment for activation keys, default: Library.
+- `convert2rhel_content_view`: Content view for activation keys, default: Default Organization View.
```
evgeni:
```suggestion
- `foreman_convert2rhel_lifecycle_env`: Lifecycle environment for activation keys, default: Library.
- `foreman_convert2rhel_content_view`: Content view for activation keys, default: Default Organization View.
```

in `roles/convert2rhel/defaults/main.yml` there are more variables, should they also be documented?

roles/convert2rhel/tasks/activation_keys.yml
```diff
@@ -0,0 +1,82 @@
+---
+- name: "Get organization (SCA) info"
+  theforeman.foreman.organization_info:
+    username: "{{ foreman_username }}"
+    password: "{{ foreman_password }}"
+    server_url: "{{ foreman_server_url }}"
+    validate_certs: "{{ foreman_validate_certs | default(omit) }}"
+    name: "{{ foreman_organization }}"
+  register: org
```
evgeni: Ansible has this odd behavior, where it leaks variables/facts/anything after a role is done, so this `org` will be defined until the play has ended. Because of that, I'd prefer if we could use something longer, and less generic? Maybe `__foreman_convert2rhel_org_info` or something?

roles/convert2rhel/tasks/host_groups.yml
```diff
@@ -0,0 +1,26 @@
+---
+- name: "Create 'CentOS 7 converting' host group"
```
evgeni:
```suggestion
- name: "Create '{{ foreman_convert2rhel_hostgroup7 }}' host group"
```

roles/convert2rhel/tasks/host_groups.yml
```diff
@@ -0,0 +1,26 @@
...
+    server_url: "{{ foreman_server_url }}"
+    validate_certs: "{{ foreman_validate_certs | default(omit) }}"
+    organization: "{{ foreman_organization }}"
+    name: "{{ foreman_convert2rhel_hostgroup7 }}"
+    lifecycle_environment: "{{ foreman_convert2rhel_lifecycle_env }}"
+    content_view: "{{ foreman_convert2rhel_content_view }}"
+    activation_keys: "{{ foreman_convert2rhel_key_centos7 }}"
+  when: foreman_content_rhel_enable_rhel7
+
+- name: "Create 'CentOS 8 converting' host group"
```
evgeni:
```suggestion
- name: "Create '{{ foreman_convert2rhel_hostgroup8 }}' host group"
```

roles/convert2rhel/tasks/products_and_repos.yml
```diff
@@ -0,0 +1,75 @@
...
+    password: "{{ foreman_password }}"
+    server_url: "{{ foreman_server_url }}"
+    validate_certs: "{{ foreman_validate_certs | default(omit) }}"
+    organization: "{{ foreman_organization }}"
+    name: "Convert2RHEL CA"
+    content_type: cert
+    content: "{{ ct.stat.exists | ternary(lookup('file', '/etc/rhsm/ca/redhat-uep.pem', errors='ignore'), lookup('url', 'https://ftp.redhat.com/redhat/convert2rhel/redhat-uep.pem', split_lines=False)) }}"
+    state: present
+
+- name: "Create 'Convert2RHEL 7' product and repositories"
```
evgeni:
```suggestion
- name: "Create '{{ foreman_convert2rhel_rhel7_product }}' product and repositories"
```

roles/convert2rhel/tasks/products_and_repos.yml
```diff
@@ -0,0 +1,75 @@
...
+            product: "{{ foreman_convert2rhel_rhel7_product }}"
+            url: "https://cdn.redhat.com/content/public/convert2rhel/7/x86_64/os/"
+            mirror_on_sync: true
+            verify_ssl_on_sync: true
+            download_policy: immediate
+            ssl_ca_cert: "Convert2RHEL CA"
+            state: present
+  when: foreman_content_rhel_enable_rhel7
+
+- name: "Create 'Convert2RHEL 8' product and repositories"
```
evgeni:
```suggestion
- name: "Create '{{ foreman_convert2rhel_rhel8_product }}' product and repositories"
```

roles/convert2rhel/tasks/products_and_repos.yml
```diff
@@ -0,0 +1,75 @@
...
+            product: "{{ foreman_convert2rhel_rhel8_product }}"
+            url: "https://cdn.redhat.com/content/public/convert2rhel/8/x86_64/os/"
+            mirror_on_sync: true
+            verify_ssl_on_sync: true
+            download_policy: immediate
+            ssl_ca_cert: "Convert2RHEL CA"
+            state: present
+  when: foreman_content_rhel_enable_rhel8
+
+- name: "Synchronize 'Convert2RHEL 7' repository"
```
evgeni:
```suggestion
- name: "Synchronize '{{ foreman_convert2rhel_rhel7_repo }}' repository"
```

roles/convert2rhel/tasks/products_and_repos.yml
```diff
@@ -0,0 +1,75 @@
...
+    username: "{{ foreman_username }}"
+    password: "{{ foreman_password }}"
+    server_url: "{{ foreman_server_url }}"
+    validate_certs: "{{ foreman_validate_certs | default(omit) }}"
+    organization: "{{ foreman_organization }}"
+    repository: "{{ foreman_convert2rhel_rhel7_repo }}"
+    product: "{{ foreman_convert2rhel_rhel7_product }}"
+  when: foreman_content_rhel_enable_rhel7
+
+- name: "Synchronize 'Convert2RHEL 8' repository"
```
evgeni:
```suggestion
- name: "Synchronize '{{ foreman_convert2rhel_rhel8_repo }}' repository"
```

roles/convert2rhel/tasks/activation_keys.yml
```diff
@@ -0,0 +1,82 @@
...
+    server_url: "{{ foreman_server_url }}"
+    validate_certs: "{{ foreman_validate_certs | default(omit) }}"
+    organization: "{{ foreman_organization }}"
+    name: "{{ foreman_convert2rhel_key_centos7 }}"
+    lifecycle_environment: "{{ foreman_convert2rhel_lifecycle_env }}"
+    content_view: "{{ foreman_convert2rhel_content_view }}"
+    subscriptions: "{{ org['organization']['simple_content_access'] | ternary(omit, foreman_convert2rhel_centos7_subs) }}"
+  when: foreman_content_rhel_enable_rhel7
+
+- name: "Create 'CentOS 8' activation key"
```
evgeni:
```suggestion
- name: "Create '{{ foreman_convert2rhel_key_centos8 }}' activation key"
```

roles/convert2rhel/tasks/activation_keys.yml
```diff
@@ -0,0 +1,82 @@
...
+    validate_certs: "{{ foreman_validate_certs | default(omit) }}"
+    organization: "{{ foreman_organization }}"
+    name: "{{ foreman_convert2rhel_key_centos8 }}"
+    lifecycle_environment: "{{ foreman_convert2rhel_lifecycle_env }}"
+    content_view: "{{ foreman_convert2rhel_content_view }}"
+    auto_attach: false
+    subscriptions: "{{ org['organization']['simple_content_access'] | ternary(omit, foreman_convert2rhel_centos8_subs) }}"
+  when: foreman_content_rhel_enable_rhel8
+
+- name: "Create 'RHEL 7' activation key"
```
evgeni:
```suggestion
- name: "Create '{{ foreman_convert2rhel_key_rhel7 }}' activation key"
```

roles/convert2rhel/tasks/activation_keys.yml
```diff
@@ -0,0 +1,82 @@
...
+    server_url: "{{ foreman_server_url }}"
+    validate_certs: "{{ foreman_validate_certs | default(omit) }}"
+    organization: "{{ foreman_organization }}"
+    name: "{{ foreman_convert2rhel_key_rhel7 }}"
+    lifecycle_environment: "{{ foreman_convert2rhel_lifecycle_env }}"
+    content_view: "{{ foreman_convert2rhel_content_view }}"
+    auto_attach: false
+  when: foreman_content_rhel_enable_rhel7
+
+- name: "Create 'RHEL 8' activation key"
```
evgeni:
```suggestion
- name: "Create '{{ foreman_convert2rhel_key_rhel8 }}' activation key"
```

roles/convert2rhel/tasks/activation_keys.yml
```diff
@@ -0,0 +1,82 @@
...
+
+- name: "Set non SCA facts"
+  set_fact:
+    foreman_convert2rhel_centos7_subs:
+      - name: "Convert2RHEL7"
+    foreman_convert2rhel_centos8_subs:
+      - name: "Convert2RHEL8"
+  when: not org['organization']['simple_content_access']
+
+- name: "Create 'CentOS 7' activation key"
```
evgeni:
```suggestion
- name: "Create '{{ foreman_convert2rhel_key_centos7 }}' activation key"
```

roles/content_rhel/README.md
```diff
@@ -30,6 +30,7 @@ It also supports customizing the included roles with:
 
 `foreman_sync_plan_sync_date`: Initial sync date for the sync plan, formatted as 'YYYY-MM-DD HH:MM:SS UTC'.
 
+
```
evgeni: I think this empty line doesn't need to be here ;)

evgeni (commented):
Design wise, I am happy.

Few inline comments, mostly around unclear docs and better task descriptions, but nothing too bad ;)

One thing we discussed, but I don't see in the current code is making the `content_rhel` inclusion optional (or disableable), see https://github.com/theforeman/foreman-ansible-modules/pull/1286#discussion_r728879705 - I won't block the PR on this, but wanted to hear an explicit "we decided not to do this" and not "oh, we just forgot :see_no_evil:".

Another thing: this is missing tests. I must admit I am torn -- our role tests essentially repeat our module tests, and this seems rather superfluous, but here we are. I can help you record them, if you want :)

And the last thing: this is missing a changelog entry, something like https://github.com/theforeman/foreman-ansible-modules/blob/2045c5edef5792d3235eb5c2b234ef08e3fa96b3/changelogs/fragments/1097-content-rhel-role.yml should be sufficient

stejskalleos:
Rebased & updated:
- fixed review suggestions
- upload of manifest and repo sync is optional, it runs only when `foreman_convert2rhel_manage_subscription` is `true`, which is default value
- added changelog
- added 'refreshed' state `to roles/manifest/tasks/main.yml`. Without refresh of manifest role failed on missing RHELx repositories
-  added basic test

roles/manifest/tasks/main.yml
```diff
@@ -15,4 +15,4 @@
     validate_certs: "{{ foreman_validate_certs | default(omit) }}"
     organization: "{{ foreman_organization }}"
     manifest_path: "{{ foreman_manifest_path }}"
-    state: present
+    state: refreshed
```
evgeni: Please don't change defaults of existing roles!

This is why the tests fail.

```suggestion
    state: "{{ foreman_manifest_state | default('present') }}"
```

And then you tell users to set it if they want it refreshed.

But honestly, this really doesn't belong into this PR.
If you're uploading a manifest, why do you need a refresh? It should be fresh already.

evgeni (changes requested):

roles/manifest/tasks/main.yml
evgeni: https://github.com/theforeman/foreman-ansible-modules/pull/1286#discussion_r727861792 :smirk_cat:

evgeni (approved):

PR #1290, opened 2021-09-25
add ansible 2.12 to CI

evgeni:
also re-work how we define the matrix, making it much easier to understand.

the new logic is: test *all* Ansible branches on Python 3.8 and additionally test Ansible 2.11 on Python 2.7, 3.6 and 3.7, and Ansible devel on Python 3.9, which results in almost the same list as before, but with a much shorter definition

### DIFF
--- a/.github/workflows/docs.yml
+++ b/.github/workflows/docs.yml
@@ -21,7 +21,7 @@ jobs:
       - name: Set up Python
         uses: actions/setup-python@v2
         with:
-          python-version: "3.7"
+          python-version: "3.8"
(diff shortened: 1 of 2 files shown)


evgeni:
`ERROR: Package 'ansible-core' requires a different Python: 3.7.12 not in '>=3.8'`, well, okay, switching default python to 3.8 then :roll_eyes:

wbclark (approved):
Nice. LGTM

mdellweg (approved):

PR #1291, opened 2021-09-30
Changing default to true so the report is used 'Ansible - Ansible Inv…

pipopopo:
Getting the report to be used by default.

Signed-off-by: Kenny Tordeurs <ktordeur@redhat.com>

### DIFF
--- a/changelogs/fragments/1291-foreman-ansible-inventory-default.yml
+++ b/changelogs/fragments/1291-foreman-ansible-inventory-default.yml
@@ -1,0 +1,2 @@
+breaking_changes:
+  - Make use_reports_api the default 
--- a/changelogs/fragments/1291-foreman-ansible-inventory-default.yml
+++ b/changelogs/fragments/1291-foreman-ansible-inventory-default.yml
@@ -1,2 +1,2 @@
 breaking_changes:
-  - Make use_reports_api the default 
+  - Set use_reports_api default value to true for the inventory plugin
--- a/plugins/inventory/foreman.py
+++ b/plugins/inventory/foreman.py
@@ -88,7 +88,7 @@ DOCUMENTATION = '''
       use_reports_api:
         description: Use Reporting API.
         type: boolean
-        default: False
+        default: True
       foreman:
         description:
           - Foreman server related configuration, deprecated.


evgeni:
This should be safe, as there is a fallback to the normal API if the report one doesn't work (e.g. because the user doesn't have the Ansible plugin enabled)

https://github.com/theforeman/foreman-ansible-modules/blob/83ff30943dd1f00da498c4f63d128bb86c1c39c3/plugins/inventory/foreman.py#L434-L439

evgeni:
CI failure unrelated.

@theforeman/ansible-modules this changes a default, which is technically a backwards incompatible change, as I am not sure the output is identical with the two ways of obtaining data. what do you think

wbclark:
> @theforeman/ansible-modules this changes a default, which is technically a backwards incompatible change, as I am not sure the output is identical with the two ways of obtaining data. what do you think

My thinking: the reports API should be a better tool for this purpose, so let's use the better tool as a default.

Another way of looking at it: at some point we will make this breaking change, so we might as well do it now.

My only caveat would be if there are other breaking changes planned, to try to make them around the same time.

pipopopo:
Just for reference in case anyone is interested:

The difference between use_reports_api true and false on a somewhat larger inventory ~990 hosts :

~~~
use_reports_api: false
want_facts: true
want_hostcollections: true
want_params: true

765.924 seconds processing time
---------------------------------------------------------------------------------

use_reports_api: true
want_facts: true
want_hostcollections: true
want_params: true

59.037 seconds processing time
---------------------------------------------------------------------------------
~~~

ares:
In the old version of this script we already defaulted to the reports API, when it was migrated to this repo, a regression was introduced by taking the wrong version. So this is technically a fix. Anyway, :+1: on merging.

evgeni:
> In the old version of this script we already defaulted to the reports API, when it was migrated to this repo, a regression was introduced by taking the wrong version. So this is technically a fix. Anyway, +1 on merging.

A version with reports api support was never shipped in `ansible/ansible`, so it's not a regression.

https://github.com/ansible/ansible/blob/stable-2.9/lib/ansible/plugins/inventory/foreman.py

wbclark (approved):
re-running tests removed the one stupid ephemeral issue related to `CannotOverwriteExistingFixture` which shows up from time to time

evgeni:
This needs a changelog entry, documenting the change

evgeni:
Like this: https://github.com/theforeman/foreman-ansible-modules/blob/9f0e9806121c2f009af40aefee5c2d5af5987cf7/changelogs/fragments/1145-role_prefix.yml

changelogs/fragments/1291-foreman-ansible-inventory-default.yml
```diff
@@ -0,0 +1,2 @@
+breaking_changes:
+  - Make use_reports_api the default
```
wbclark:
```suggestion
  - Set use_reports_api default value to true for the inventory plugin
```

Thanks @pipopopo , I made a recommendation to be slightly more specific since this collection includes more than just the inventory plugin

PR #1294, opened 2021-10-05
add python 3.10 to CI matrix

evgeni:


evgeni:
using RC2 for now, as there is no final yet on the GHA setup (it was released today)

mdellweg (approved):

evgeni:
let's see, https://github.com/actions/python-versions/pull/115 should add 3.10 GA support

evgeni:
still green, yay

PR #1295, opened 2021-10-08
drop the default url from the callback

evgeni:
the url parameter is `required: True`, so if you don't define it in your
configuration, ansible yields an error and the default value is never
used anyways

this makes it a non-breaking change, not a change at all actually ;)


wbclark (approved):
lol

for due diligence purposes, I tried thinking about the pros/cons of removing `required: True` instead of the default value, for all of two seconds until I thought about what this parameter actually _does_.

nice find btw.

evgeni:
oh I will totally drop the `required: True` once @lzap is done re-writing all that reporting in Foreman, but not today :)

PR #1297, opened 2021-10-10
add host_init_config to TEMPLATE_KIND_LIST and os_default_template docs

Bonehead5338:
Unable to select 'Host initial configuration template' default template in operating system with os_default_template module.

image

foreman code reference for template type: https://github.com/theforeman/foreman/blob/3.0.0/app/models/template_kind.rb

### DIFF
--- a/plugins/module_utils/foreman_helper.py
+++ b/plugins/module_utils/foreman_helper.py
@@ -1841,6 +1841,7 @@ TEMPLATE_KIND_LIST = [
     'Bootdisk',
     'cloud-init',
     'finish',
+    'host_init_config',
     'iPXE',
     'job_template',
     'kexec',
--- a/plugins/modules/os_default_template.py
+++ b/plugins/modules/os_default_template.py
@@ -40,6 +40,7 @@ options:
       - Bootdisk
       - cloud-init
       - finish
+      - host_init_config
       - iPXE
       - job_template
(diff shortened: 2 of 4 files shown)


plugins/module_utils/foreman_helper.py
```diff
@@ -1840,6 +1840,7 @@ def ensure_puppetclasses(module, entity_type, entity, expected_puppetclasses=Non
     'script',
     'user_data',
     'ZTP',
+    'host_init_config',
```
mdellweg: Looks like they should be in alphabetical order

plugins/modules/os_default_template.py
```diff
@@ -53,6 +53,7 @@
       - script
       - user_data
       - ZTP
+      - host_init_config
```
mdellweg: same here

mdellweg:
This will need a changelog snippet too.

martin-schlossarek:
Whats the blocker for this PR? We need to set the default `host_init_config `template.

evgeni:
I took the liberty to adress @mdellweg's comments :)

evgeni (approved):

mdellweg (approved):

martin-schlossarek:
Cool, thanks! :)

Issue #1298, opened 2021-10-11
Misleading documentation for theforeman.foreman.host when using host groups


sscheib-rh:
##### SUMMARY
The documentation regarding `theforeman.foreman.host` is misleading when using host groups.
The [documentation](https://docs.ansible.com/ansible/latest/collections/theforeman/foreman/host_module.html) mentions:
`hostgroup: Name of related hostgroup.`

Actually it should mention that the *title* has to be used. This is especially important, when working with subgroups.

Please see the following example:
parent host group: `foo`
sub host group of `foo`: `bar`
complete path therefore: `for/bar`

When using the module with `bar` only, it would reference any parent level host group, which has the name `bar`, but not the intended host group `foo/bar`. To achieve this, one has to use `foo/bar` as `hostgroup` parameter.

##### ISSUE TYPE
 - Bug Report

##### STEPS TO REPRODUCE

This playbook would create a host in a top level host group `bar` (if existent):
```yaml
- name: "Create a host"
  theforeman.foreman.host:
    username: "admin"
    password: "changeme"
    server_url: "https://foreman.example.com"
    name: "new_host"
    hostgroup: bar
    state: present
```

To reference the sub group one has to do it like so:
```yaml
- name: "Create a host"
  theforeman.foreman.host:
    username: "admin"
    password: "changeme"
    server_url: "https://foreman.example.com"
    name: "new_host"
    hostgroup: 'foo/bar'
    state: present
```

##### EXPECTED RESULTS
The documentation should mention the use of 'title' instead of 'name'

pipopopo:
Suggested improvement: https://github.com/theforeman/foreman-ansible-modules/pull/1299

pipopopo:
I believe this issue can be closed out.

PR #1299, opened 2021-10-13
Fix for misleading documentation 1298

pipopopo:
Fix misleading documentation for theforeman.foreman.host with hostgroups.

Signed-off-by: Kenny Tordeurs <ktordeur@redhat.com>

### DIFF
--- a/plugins/modules/host.py
+++ b/plugins/modules/host.py
@@ -36,7 +36,8 @@ options:
     type: str
   hostgroup:
     description:
-      - Name of related hostgroup.
+      - Title of related hostgroup
+      - Example parent hostgroup I(foo) with a child hostgroup I(bar) would have the title I(foo/bar)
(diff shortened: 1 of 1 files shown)


pipopopo:
Fixes #1298

plugins/modules/host.py
```diff
@@ -36,7 +36,8 @@
     type: str
   hostgroup:
     description:
-      - Name of related hostgroup.
+      - Title of related hostgroup
+      - Example parent hostgroup I(foo) with a child hostgroup I(bar) would have the title I(foo/bar)
```
mdellweg:
```suggestion
      - "Example: A child hostgroup I(bar) within a parent hostgroup I(foo) would have the title I(foo/bar)."
```

evgeni (approved):

PR #1303, opened 2021-10-20
foreman_repository: Added download_concurrency

### LINKED ISSUE #1273: foreman_repository: Allow to configure download-concurrency
##### SUMMARY

Some repos may have resource restrictions that will cause timeouts with Pulp (possibly due to rate-limiting) when syncing repos.

The forum thread below suggested to set a download concurrency on those repos. However, the Ansible module currently is unable to do so.

https://community.theforeman.org/t/unable-to-sync-some-repositories-almalinux-in-this-case/24705/29

##### ISSUE TYPE
 - Feature Idea

Thulium-Drake:
This PR adds download_concurrency to foreman_repository

Fixes #1273

### DIFF
--- a/plugins/modules/repository.py
+++ b/plugins/modules/repository.py
@@ -118,6 +118,7 @@ options:
       - as the API does not return this value, this will break idempotence for this module
     required: false
     type: int
+    version_added: 3.0.0
   download_policy:
     description:
       - download policy for sync from upstream
--- a/plugins/modules/repository.py
+++ b/plugins/modules/repository.py
@@ -112,6 +112,11 @@ options:
     - Repository SSL client private key
     required: false
     type: str
+  download_concurrency:
+    description:
+      - download concurrency for sync from upstream
+    required: false
+    type: int
   download_policy:
     description:
       - download policy for sync from upstream
@@ -294,6 +299,7 @@ def main():
             ssl_client_cert=dict(type='entity', resource_type='content_credentials', scope=['organization']),
             ssl_client_key=dict(type='entity', resource_type='content_credentials', scope=['organization'], no_log=False),
             download_policy=dict(choices=['background', 'immediate', 'on_demand']),
+            download_concurrency=dict(type='int'),
             mirror_on_sync=dict(type='bool', default=True),
             verify_ssl_on_sync=dict(type='bool'),
             upstream_username=dict(),
--- a/plugins/modules/repository.py
+++ b/plugins/modules/repository.py
@@ -115,7 +115,7 @@ options:
   download_concurrency:
     description:
       - download concurrency for sync from upstream
-      - as the API does not return this value, this will break idempotency for this module
+      - as the API does not return this value, this will break idempotence for this module
     required: false
     type: int
   download_policy:
--- a/plugins/modules/repository.py
+++ b/plugins/modules/repository.py
@@ -115,6 +115,7 @@ options:
   download_concurrency:
     description:
       - download concurrency for sync from upstream
+      - as the API does not return this value, this will break idempotency for this module
     required: false
     type: int
   download_policy:
--- a/changelogs/fragments/1303-foreman-repository-download-concurrency.yml
+++ b/changelogs/fragments/1303-foreman-repository-download-concurrency.yml
@@ -1,0 +1,2 @@
+minor_changes:
+  - repository - new ``download_concurrency`` parameter (https://github.com/theforeman/foreman-ansible-modules/issues/1273)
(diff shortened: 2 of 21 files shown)


wbclark (commented):
From the issue discussion:

> FWIW, you can't see the current value for download_concurrency via Hammer or the API, but you can get it when you run the following SQL query on the foreman database:

How does Ansible confirm the current state of the repository then? Since these ansible modules use the API, wouldn't this break idempotence if true? Since the module would be unable to determine the current download_concurrency and should therefore always try to "change" it to the specified value.

mdellweg:
> From the issue discussion:
> 
> > FWIW, you can't see the current value for download_concurrency via Hammer or the API, but you can get it when you run the following SQL query on the foreman database:
> 
> How does Ansible confirm the current state of the repository then? Since these ansible modules use the API, wouldn't this break idempotence if true? Since the module would be unable to determine the current download_concurrency and should therefore always try to "change" it to the specified value.

I agree. If it's not in the api, we can only handle it non idempotent. We should at least claim that in the docs part (there are examples for this floating around in some modules).
Now that we have a discussion about idempotence here, should we add a test?

Thulium-Drake:
@wbclark You're right, it currently can't. However, looking at the other options that can be passed in the module, I feel it's 'just' a matter of making the API report it back and Ansible will then be able to decide whether or not it made a change.

I will add it to the docs :-)

@mdellweg I'm not entirely sure it makes a lot of sense to test it now, as there is no mention at all of this field in the API's response.

plugins/modules/repository.py
```diff
@@ -112,6 +112,12 @@
     - Repository SSL client private key
     required: false
     type: str
+  download_concurrency:
+    description:
+      - download concurrency for sync from upstream
+      - as the API does not return this value, this will break idempotency for this module
```
```suggestion
      - as the API does not return this value, this will break idempotence for this module
```
I guess, we don't have a version, where we expect this to be "fixed" in Katello?

Thulium-Drake: Not yet, I just asked it on the foreman support matrix, as I wasn't sure whether this API is provided by Katello/Foreman or Pulp :sweat_smile: 

Once I know that, I can file the enhancement :-)

mdellweg: These modules never talk to Pulp directly, so Katello needs to handle it (and hand it through).

wbclark:
> @mdellweg I'm not entirely sure it makes a lot of sense to test it now, as there is no mention at all of this field in the API's response.

The TDD / BDD point of view is that we should add the test, verify that it fails currently (matching our expectation), and then we can open a PR to katello to expose that field in the API.

At that point the tests can be run against a katello instance which has that support, and we should see passing tests.

Since we'd like to have the test before merging regardless, another way of looking at it is that there is no downside to creating them up front.

Thulium-Drake:
Test-wise this should do it right?

However, my development-test box died, and I don't currently have the capacity to make new VCRs for this PR... Can someone help me out there? :-) <3

mdellweg:
> Test-wise this should do it right?
> 
> However, my development-test box died, and I don't currently have the capacity to make new VCRs for this PR... Can someone help me out there? :-) <3

There you are:
https://github.com/mdellweg/foreman-ansible-modules/tree/dl_concurrency

Thulium-Drake:
Thanks! :-)

mdellweg:
I'm about to ack this. What are the rules for a changelog snippet here? @evgeni

plugins/modules/repository.py
```diff
@@ -112,6 +112,12 @@
     - Repository SSL client private key
     required: false
     type: str
+  download_concurrency:
+    description:
+      - download concurrency for sync from upstream
+      - as the API does not return this value, this will break idempotence for this module
+    required: false
+    type: int
```
evgeni, line 120: you can add a `version_added: 3.0.0` here, so it will render this information on the doc page too

evgeni:
rules? ;)

probably something like this should do?

```yaml
minor_changes:
  - repository - new ``download_concurrency`` parameter (https://github.com/theforeman/foreman-ansible-modules/issues/1273)
```

mdellweg:
rules like "We need one."

evgeni:
only new modules are picked up automatically by the tooling, everything else should have one

not that we've been very strict about it ;)

Thulium-Drake:
Well, I added one ;-)

plugins/modules/repository.py
Thulium-Drake: Done

mdellweg (approved):

Thulium-Drake:
@evgeni Is that failed test a fluke? Or did I break something with that changelog? ;-)

Also, do you need anything else before this can be merged? :-)

Thulium-Drake:
ping @evgeni  :)

evgeni (approved):

evgeni:
sorry, this slipped my radar somehow. yet it's a fluke

PR #1305, opened 2021-10-26
Do not send facts when proxy is set

lzap:
When report_type is set to proxy, facts are already part of request and smart proxy reports plugin will forward both report and facts to foreman. There is actually no need to send facts anywhere as they have been already sent.

This disables sending facts in this mode. Going forward, there will be only one HTTP request per callback and proxy will do the rest.

### DIFF
--- a/plugins/callback/foreman.py
+++ b/plugins/callback/foreman.py
@@ -260,6 +260,9 @@ class CallbackModule(CallbackBase):
         parser.  The default fact importer should import these facts
         properly.
         """
+        # proxy parses facts from report directly
+        if self.report_type == "proxy":
+            return
 
         for host, facts in self.facts.items():
             facts = {


evgeni (approved):
I would have done the "don't do anything" logic inside `send_facts`, much like we do different routing of the reports in `send_reports`, but won't block on this.

@lzap which way do you prefer?

lzap:
Sure I was thinking about this for a moment, then I decided. Eventually, we will remove this completely. Not a big deal, rebased to your recommendation. Thanks!

plugins/callback/foreman.py
```diff
@@ -260,6 +260,9 @@ def send_facts(self):
         parser.  The default fact importer should import these facts
         properly.
         """
+        # proxy parses facts from report directly
+        if self.report_type != "proxy":
```
```suggestion
        if self.report_type == "proxy":
```

evgeni (changes requested):

plugins/callback/foreman.py
lzap: These last minute changes, good catch :-)

evgeni (approved):

PR #1306, opened 2021-10-27
pin python-debian for 3.6 too

evgeni:
0.1.42 (accidentally) dropped support for 3.5/3.6 [1], and as we've
already had a pin for 3.5 and older, let's just extend it to also
include 3.6 -- 3.7 and newer can then use "latest"

[1] https://bugs.debian.org/cgi-bin/bugreport.cgi?bug=997857


evgeni:
:green_apple:

mdellweg (approved):

PR #1307, opened 2021-10-27
don't fail trying to remove children with absent parents

### LINKED ISSUE #1301: Remove of nested hostgroups not idempotent
##### SUMMARY
When I delete nested hostgroups via Ansible I get an error: 

Found no results while searching for hostgroups with title="child"

##### ISSUE TYPE
 - Bug Report

##### ANSIBLE VERSION

```
ansible 2.9.25
```

##### COLLECTION VERSION

```
theforeman.foreman 2.2.0
```

##### KATELLO/FOREMAN VERSION

```
tfm-rubygem-katello-3.18.1.37-1.el7sat.noarch
foreman-2.3.1.23-1.el7sat.noarch
```

##### STEPS TO REPRODUCE

```yaml
---                                                                                                     
- hosts: localhost 
  connection: local
  module_defaults: 
    theforeman.foreman.hostgroup: 
      username: "usr"                                                                            
      password: "psw" 
      server_url: "url" 
      validate_certs: "false"                                                                                                     
  tasks: 
    - name: "Remove hostgroups"                                                  
      theforeman.foreman.hostgroup:                                              
        name: "{{ item.name }}"                                                  
        parent: "{{ item.parent | default(omit) }}"                              
        organization: "{{ item.organization }}"                                 
        locations: "{{ item.locations }}"                                                                
        state: absent
      loop:                                                                      
        - name: "parent"                                                
          organization: "myorg"                                            
          locations: "mylocation"                                                
        - name: "child"                                                     
          parent: "parent"                                              
          organization: "myorg"                                            
          locations: "mylocation"
```

##### EXPECTED RESULTS

First run will throw an error, that parent can´t be deleted, because of nested group, thats okay, but in the second run I get an error, that he can´t find the child. I would expect in the second run, that i get a changed on the parent hostgroup and an ok for the child and not an error

##### ACTUAL RESULTS

```
First run will throw an error, that parent can´t be deleted, because of nested group, thats okay, but in the second run I get an error, that he can´t find the child.
```

evgeni:
Fixes: #1301

### DIFF
--- a/plugins/module_utils/foreman_helper.py
+++ b/plugins/module_utils/foreman_helper.py
@@ -1315,6 +1315,8 @@ class ForemanStatelessEntityAnsibleModule(ForemanAnsibleModule):
         if 'parent' in self.foreman_spec and self.foreman_spec['parent'].get('type') == 'entity':
             if 'resouce_type' not in self.foreman_spec['parent']:
                 self.foreman_spec['parent']['resource_type'] = self.foreman_spec['entity']['resource_type']
+            if 'failsafe' not in self.foreman_spec['parent']:
+                self.foreman_spec['parent']['failsafe'] = True
             current, parent = split_fqn(self.foreman_params[self.entity_key])
             if isinstance(self.foreman_params.get('parent'), six.string_types):
                 if parent:
@@ -1424,10 +1426,14 @@ class ForemanEntityAnsibleModule(ForemanStatelessEntityAnsibleModule):
     def run(self, **kwargs):
         """ lookup entities, ensure entity, remove sensitive data, manage parameters.
         """
+        parent_name = self.foreman_params.get('parent')
         if ('parent' in self.foreman_spec and self.foreman_spec['parent'].get('type') == 'entity'
-                and self.desired_absent and 'parent' in self.foreman_params and self.lookup_entity('parent') is None):
-            # Parent does not exist so just exit here
-            return None
+                and 'parent' in self.foreman_params and self.lookup_entity('parent') is None):
+            if self.desired_absent:
+                # Parent does not exist so just exit here
+                return None
+            else:
+                self.fail_json(msg="Couldn't find parent '{0}' for '{1}'.".format(parent_name, self.foreman_params['name']))
         if not self.desired_absent:
             self.auto_lookup_entities()
         entity = self.lookup_entity('entity')
--- a/tests/test_playbooks/tasks/location.yml
+++ b/tests/test_playbooks/tasks/location.yml
@@ -13,14 +13,14 @@
     parameters: "{{ location_parameters | default(omit) }}"
     state: "{{ location_state }}"
   register: result
-  ignore_errors: true
+  ignore_errors: "{{ expected_error is defined and expected_error }}"
 - assert:
     fail_msg: "Ensuring location is {{ location_state }} failed! (expected_change: {{ expected_change | default('unknown') }})"
     that:
       - result.changed == expected_change
   when: expected_change is defined
 - assert:
-    fail_msg: "Ensuring location is {{ location_state }} didn't fail! (expected_error: {{ expected_error | default('unknown') }})"
+    fail_msg: "Ensuring location is {{ location_state }} did fail! (expected_error: {{ expected_error | default('unknown') }})"
     that:
       - result is failed
   when:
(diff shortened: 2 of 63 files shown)


evgeni:
this certainly deserves a test, but wanted to show what I have for now

plugins/module_utils/foreman_helper.py
```diff
@@ -1315,6 +1315,8 @@ def __init__(self, **kwargs):
         if 'parent' in self.foreman_spec and self.foreman_spec['parent'].get('type') == 'entity':
             if 'resouce_type' not in self.foreman_spec['parent']:
                 self.foreman_spec['parent']['resource_type'] = self.foreman_spec['entity']['resource_type']
+            if 'failsafe' not in self.foreman_spec['parent']:
+                self.foreman_spec['parent']['failsafe'] = True
```
mdellweg, line 1319: This prevents the lookup failure...

plugins/module_utils/foreman_helper.py
```diff
@@ -1424,10 +1426,14 @@ def __init__(self, **kwargs):
...
         """
+        parent_name = self.foreman_params.get('parent')
         if ('parent' in self.foreman_spec and self.foreman_spec['parent'].get('type') == 'entity'
-                and self.desired_absent and 'parent' in self.foreman_params and self.lookup_entity('parent') is None):
-            # Parent does not exist so just exit here
-            return None
+                and 'parent' in self.foreman_params and self.lookup_entity('parent') is None):
+            if self.desired_absent:
+                # Parent does not exist so just exit here
+                return None
```
mdellweg, line 1434: ... this prevents that the missing parent is a problem ...

plugins/module_utils/foreman_helper.py
```diff
@@ -1424,10 +1426,14 @@ def __init__(self, **kwargs):
...
+            else:
+                self.fail_json(msg="Couldn't find parent '{0}' for '{1}'.".format(parent_name, self.foreman_params['name']))
```
mdellweg, line 1436: ... and this is the postponed failure.

Would it be possible to make the failsafe flag dependent on desired_absent in the first place?

plugins/module_utils/foreman_helper.py
```diff
@@ -1424,10 +1426,14 @@ def __init__(self, **kwargs):
     def run(self, **kwargs):
         """ lookup entities, ensure entity, remove sensitive data, manage parameters.
         """
+        parent_name = self.foreman_params.get('parent')
```
mdellweg, line 1429: This is only used in line 1436. Maybe move it there.

mdellweg (commented):
Looks good so far. I suggested another approach, but that may or may not create an even bigger mess.

plugins/module_utils/foreman_helper.py
evgeni: I thought about that, but `desired_absent` is only present in `ForemanEntityAnsibleModule`, but the `parent` logic is in `ForemanStatelessEntityAnsibleModule`, which doesn't have `state` XD

plugins/module_utils/foreman_helper.py
evgeni: No, as line 1430/1431 has `self.lookup_entity('parent')` which overwrites this with `None` ;)

plugins/module_utils/foreman_helper.py
mdellweg: Sounds like a bigger mess.

evgeni: Yes.

tests/test_playbooks/tasks/location.yml
```diff
@@ -21,6 +21,12 @@
   when: expected_change is defined
 - assert:
     fail_msg: "Ensuring location is {{ location_state }} didn't fail! (expected_error: {{ expected_error | default('unknown') }})"
+    that:
+      - result is succeeded
+  when:
+    - expected_error is not defined or not expected_error
```
mdellweg: Can we use `expected_error is defined` as a condition for ignore_errors?

evgeni: maybe? I'll try

evgeni: YES!

mdellweg (approved):

PR #1308, opened 2021-10-27
Convert2RHEL - support for Oracle 7

stejskalleos:
Followup to [PR Convert2RHEL role](https://github.com/theforeman/foreman-ansible-modules/pull/1286)

### DIFF
--- a/roles/convert2rhel/README.md
+++ b/roles/convert2rhel/README.md
@@ -17,6 +17,7 @@ This role supports the [Common Role Variables](https://github.com/theforeman/for
 - `foreman_convert2rhel_manage_subscription`: Run [content_rhel Role](https://github.com/theforeman/foreman-ansible-modules/blob/develop/roles/content_rhel/README.md) role, default: `true`
 - `foreman_convert2rhel_lifecycle_env`: Lifecycle environment for activation keys, default: Library.
 - `foreman_convert2rhel_content_view`: Content view for activation keys, default: Default Organization View.
+- `foreman_convert2rhel_enable_oracle7`: Create data for Oracle Linux 7 conversion, default: `false`
 
 Example Playbooks
 -----------------
@@ -39,4 +40,5 @@ Convert2RHEL
         foreman_content_rhel_wait_for_syncs: false
         foreman_convert2rhel_lifecycle_env: "Library"
         foreman_convert2rhel_content_view: "Default Organization View"
+        foreman_convert2rhel_enable_oracle7: true
 ```
--- a/roles/convert2rhel/defaults/main.yml
+++ b/roles/convert2rhel/defaults/main.yml
@@ -6,10 +6,13 @@ foreman_convert2rhel_rhel7_product: "Convert2RHEL7"
 foreman_convert2rhel_rhel7_repo: "Convert2RHEL7 main"
 foreman_convert2rhel_rhel8_product: "Convert2RHEL8"
 foreman_convert2rhel_rhel8_repo: "Convert2RHEL8 main"
+foreman_convert2rhel_oracle7_product: "Oracle Linux 7 Convert2RHEL"
+foreman_convert2rhel_oracle7_repo: "Oracle Linux 7 Convert2RHEL main"
 
 # Activation keys
 foreman_convert2rhel_key_centos7: "convert2rhel_centos7"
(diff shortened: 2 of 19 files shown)


stejskalleos:
@ares @evgeni

roles/convert2rhel/tasks/products_and_repos.yml
```diff
@@ -52,6 +52,24 @@
             state: present
   when: foreman_content_rhel_enable_rhel8
 
+- name: "Create 'UBI' product and repositories"
```
evgeni: what is this repo and why is this needed? (a doc link to make my brain happy is sufficient)

roles/convert2rhel/defaults/main.yml
```diff
@@ -16,3 +17,7 @@ foreman_convert2rhel_content_view: "Default Organization View"
 # Host groups
 foreman_convert2rhel_hostgroup7: "CentOS 7 converting"
 foreman_convert2rhel_hostgroup8: "CentOS 8 converting"
+foreman_convert2rhel_hg_ol7: "Oracle 7 converting"
```
evgeni: should this be `foreman_convert2rhel_hostgroup_ol7` to match the others?

or should the others also end in `_centosX`?

roles/convert2rhel/tasks/products_and_repos.yml
ares: OL does not have subscription-manager available in its repositories, but we need that for the registration, the best source for that is the [UBI](https://www.redhat.com/en/blog/introducing-red-hat-universal-base-image) repository. Convert2rhel installs the subscription-manager from there when it's not used with Satellite (and expects the Internet connectivity)

evgeni:
#1286 is in, this now needs a rebase and the comments adressed

roles/convert2rhel/tasks/products_and_repos.yml
evgeni: I think we should call the product/repos differently. While they come from UBI, this is not what we use them for and this will lead to confusion.

How about "Oracle Linux subscription-manager dependencies" or something? Or "Oracle Linux convert2rhel dependencies"?

stejskalleos:
@evgeni rebased and updated

I will update tests later when I solve the issue with oracle repo sync:
```
Cannot connect to host cdn-ubi.redhat.com:443 ssl:default [Network is unreachable]
```

roles/convert2rhel/README.md
```diff
@@ -17,6 +17,7 @@ This role supports the [Common Role Variables](https://github.com/theforeman/for
 - `foreman_convert2rhel_manage_subscription`: Run [content_rhel Role](https://github.com/theforeman/foreman-ansible-modules/blob/develop/roles/content_rhel/README.md) role, default: `true`
 - `foreman_convert2rhel_lifecycle_env`: Lifecycle environment for activation keys, default: Library.
 - `foreman_convert2rhel_content_view`: Content view for activation keys, default: Default Organization View.
+- `foreman_convert2rhel_enable_oracle7`: Create data for Oracle 7 conversion, default: `false`
```
evgeni:
```suggestion
- `foreman_convert2rhel_enable_oracle7`: Create data for Oracle Linux 7 conversion, default: `false`
```

roles/convert2rhel/defaults/main.yml
```diff
@@ -18,3 +21,7 @@ foreman_convert2rhel_content_view: "Default Organization View"
 # Host groups
 foreman_convert2rhel_hostgroup7: "CentOS 7 converting"
 foreman_convert2rhel_hostgroup8: "CentOS 8 converting"
+foreman_convert2rhel_hg_oracle7: "Oracle 7 converting"
```
evgeni: I still think that the hostgroup variables should have identical structure. And my vote would be:
* foreman_convert2rhel_hostgroup_centos7
* foreman_convert2rhel_hostgroup_centos8
* foreman_convert2rhel_hostgroup_oracle7

And you're converting Oracle *Linux*, so while I think it's fine to shorten it in the variable name, the HG name should be "Oracle Linux 7 converting"

evgeni (changes requested):
two minor nitpicks, LGTM otherwise and can be merged once nitpicks are adressed.

evgeni (approved):

PR #1321, opened 2021-11-24
fetch *all* facts, not only the first 250, when using the old Hosts API

evgeni:


ehelms (approved):

plugins/inventory/foreman.py
```diff
@@ -256,8 +256,20 @@ def _get_json(self, url, ignore_errors=None, params=None):
...
+                    if not isinstance(results, MutableMapping):
+                        results = {}
+
+                    # check for end of paging
+                    if len(json['results']) == 0:
+                        break
+
+                    for host, facts in json['results'].items():
+                        if host not in results:
+                            results[host] = {}
```
mdellweg, line 268: This could be prevented by using `results = defaultdict(dict)`.

evgeni: Sure, but I'm lazy :D

PR #1324, opened 2021-12-06
don't pin pip to <20

evgeni:
I don't know why this was done, but we really shouldn't


mdellweg (approved):
Wasn't that about the time they started to mess around with the dependency resolver?

evgeni:
looking at git history, it seems 20.0 was a bit borked

PR #1325, opened 2021-12-09
Send report summary unchanged

lzap:
I have realized that we drop some summary fields and do not send it all. In the new reports format, all the processing is done on the proxy and report format is now native, therefore, there is no need to create puppet-designed metrics. Let's send the summary as-is.

### DIFF
--- a/plugins/callback/foreman.py
+++ b/plugins/callback/foreman.py
@@ -283,7 +283,6 @@ class CallbackModule(CallbackBase):
         changes.
         """
         for host in stats.processed.keys():
-            total = stats.summarize(host)
             report = {
                 "host": host,
                 "reported_at": get_now(),
@@ -292,17 +291,10 @@ class CallbackModule(CallbackBase):
                         "total": int(get_time() - self.start_time)
                     }
                 },
-                "status": {
-                    "applied": total['changed'],
-                    "failed": total['failures'] + total['unreachable'],
-                    "skipped": total['skipped'],
-                },
+                "summary": stats.summarize(host),
                 "results": self.items[host],
                 "check_mode": self.check_mode,
             }
-            if self.check_mode:
-                report['status']['pending'] = total['changed']
-                report['status']['applied'] = 0
 
             self._send_data('report', 'proxy', host, report)
             self.items[host] = []
--- a/tests/fixtures/callback/dir_store/proxy/testhost.json
+++ b/tests/fixtures/callback/dir_store/proxy/testhost.json
@@ -436,9 +436,13 @@
       }
     }
   ],
-  "status": {
-    "applied": 5,
-    "failed": 0,
-    "skipped": 1
+  "summary": {
+    "changed": 5,
+    "failures": 0,
+    "ignored": 1,
+    "ok": 8,
+    "rescued": 0,
+    "skipped": 1,
+    "unreachable": 0
   }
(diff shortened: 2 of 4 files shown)


tests/test_callback.py
```diff
@@ -12,6 +12,7 @@ def run_playbook_callback(tmpdir, report_type):
     extra_env = {}
     ansible_version = get_ansible_version()
     if ansible_version is None:
+        print("Ansible version is None!")
```
evgeni: Shouldn't the next line have told you the same?

README.md
```diff
@@ -108,6 +108,20 @@ And install it with:
 
 `ansible-galaxy collection install ./theforeman-foreman-*.tar.gz`
 
+## Running tests
```
evgeni: We have https://github.com/theforeman/foreman-ansible-modules/blob/develop/docs/testing.md for that?

plugins/callback/foreman.py
```diff
@@ -292,17 +292,10 @@ def send_reports_proxy_host_report(self, stats):
                         "total": int(get_time() - self.start_time)
                     }
                 },
-                "status": {
-                    "applied": total['changed'],
-                    "failed": total['failures'] + total['unreachable'],
-                    "skipped": total['skipped'],
-                },
+                "summary": total,
```
evgeni: does that require any changes to the `report` plugin? or does it already understand this format?

README.md
lzap: Damn I knew it it is somewhere...

plugins/callback/foreman.py
lzap: Yes, however, we pulled the plugin from 3.1 (ehm) and there is no stable release yet, so changes will be made. I can link a PR (not up yet), I am waiting for our community to review my final proposal of the mapping:

https://community.theforeman.org/t/new-config-report-summary-columns/26531

tests/test_callback.py
lzap: A leftover.

lzap:
Dropped the two nonsense unrelated changes. Feel free to wait until mapping is confirmed or the reports plugin PR is up. I will implement it in a backward compatible way anyway so it is easier on devs to use it.

evgeni:
Aye!

The diff looks reasonable now.

Happy to wait until you get consensus.

lzap:
What are my chances getting this into the current stable version which ships in 3.1? Just probing...

evgeni:
If you ask nicely, we can get a release out with these changes and push that release into 3.1, sure

lzap:
I would appreciate your opinion on my final proposal for Ansible. I am keeping it simple for the first version:

https://community.theforeman.org/t/new-config-report-summary-columns/26531

If you are good with this, let's get this in!

plugins/callback/foreman.py
evgeni: Cool.

Btw, should the variable be renamed to `summary` too? It's the result of `summarize(host)` and you submit it as `summary` field…

evgeni (approved):

plugins/callback/foreman.py
lzap: Good point, I removed the variable completely, there are no other uses anymore.

evgeni (approved):

PR #1326, opened 2021-12-09
Fixes #33299 - upload support for ostree content

jjeffers:


plugins/modules/content_upload.py
```diff
@@ -67,6 +67,17 @@
...
+- name: "Upload ostree-archive.tar"
+  theforeman.foreman.content_upload:
+    username: "admin"
+    password: "changeme"
+    server_url: "https://foreman.example.com"
+    src: "ostree_archive.tar"
+    repository: "My OStree Repository"
+    product: "My Product"
+    organization: "Default Organization"
+    repository_name: "small"
```
evgeni:
```suggestion
    ostree_repository_name: "small"
```

plugins/modules/content_upload.py
```diff
@@ -128,6 +139,7 @@ def main():
             src=dict(required=True, type='path', aliases=['file']),
             repository=dict(required=True, type='entity', scope=['product'], thin=False),
             product=dict(required=True, type='entity', scope=['organization']),
+            ostree_repository_name=dict(required=False, type='str'),
```
evgeni, line 149: You'll have to add `ostree_repository_name` to the `DOCUMENTATION` section above.

tests/test_playbooks/tasks/upload.yml
```diff
@@ -14,6 +14,7 @@
     src: "{{ upload_src }}"
     repository: "{{ upload_repository }}"
     product: "{{ upload_product }}"
+    ostree_repository_name: "{{ upload_ostree_repository_name }}"
```
evgeni:
```suggestion
    ostree_repository_name: "{{ upload_ostree_repository_name | default(omit) }}"
```

that way it's not passed for non-ostree uploads

plugins/modules/content_upload.py
```diff
@@ -17,6 +17,8 @@
 # along with this program.  If not, see <http://www.gnu.org/licenses/>.
 
 from __future__ import absolute_import, division, print_function
+
+from ansible.module_utils.basic import AnsibleModule
```
evgeni: this shouldn't be needed, you get everything via the `foreman_helper` module

plugins/modules/content_upload.py
```diff
@@ -157,6 +171,11 @@ def main():
         elif module.foreman_params['repository']['content_type'] == 'file':
             query = 'name = "{0}" and checksum = "{1}"'.format(filename, checksum)
             content_unit = module.find_resource('file_units', query, params=repository_scope, failsafe=True)
+        elif module.foreman_params['repository']['content_type'] == 'ostree':
+            query = 'name = "{0}" and checksum = "{1}"'.format(filename, checksum)
+            ostree_type_scope = {'content_type': 'ostree_ref'}
```
evgeni: I would, for "safety" reasons, add the repo scope here. The search might work without, but if we set this, it will be limited to a specific repo, and should be faster.

```suggestion
            ostree_type_scope = {'content_type': 'ostree_ref'}
            ostree_type_scope.update(repository_scope)
```

tests/test_playbooks/content_upload_ostree.yml
```diff
@@ -0,0 +1,67 @@
...
+- hosts: tests
+  collections:
+    - theforeman.foreman
+  gather_facts: false
+  vars_files:
+    - vars/server.yml
+  tasks:
+
+    - name: upload ostree archive
+      include: tasks/upload.yml
```
evgeni:
```suggestion
      include_tasks: tasks/upload.yml
```

if you use `include_tasks` the `name` ("upload ostree archive") of the step is shown in the log, making the log nicer to read

tests/test_playbooks/content_upload_ostree.yml
```diff
@@ -0,0 +1,67 @@
...
+      vars:
+        upload_src: "{{ playbook_dir }}/data/fixtures_small_repo.tar"
+        upload_repository: "Test OStree Repository"
+        upload_ostree_repository_name: "small"
+        expected_change: true
+    - name: upload ostree archive again, no change
```
evgeni:
```suggestion
      include_tasks: tasks/upload.yml
```

if you use `include_tasks` the `name` ("upload ostree archive again, no change") of the step is shown in the log, making the log nicer to read

plugins/modules/content_upload.py
```diff
@@ -157,6 +175,12 @@ def main():
         elif module.foreman_params['repository']['content_type'] == 'file':
             query = 'name = "{0}" and checksum = "{1}"'.format(filename, checksum)
             content_unit = module.find_resource('file_units', query, params=repository_scope, failsafe=True)
+        elif module.foreman_params['repository']['content_type'] == 'ostree':
+            query = 'name = "{0}" and checksum = "{1}"'.format(filename, checksum)
```
evgeni: So, as discussed yesterday on IRC, this is what "fails" your idempotency test.

1. When I execute this search manually, the API tells me `Field 'checksum' not recognized for searching!` and is not finding anything.
2. Looking at the tarball you're using for this test, it has two "heads": `rawhide` and `stable`, so when I just list all known ostree refs, I get: 
    ```json
    "results":[
      {"id":2,"name":"rawhide","filename":null,"pulp_id":"/pulp/api/v3/content/ostree/refs/9ae5bf3d-525d-4a27-a527-d08f85f62786/","version":null,"content_type":"ostree_ref","additional_metadata":null},
      {"id":1,"name":"stable","filename":null,"pulp_id":"/pulp/api/v3/content/ostree/refs/9cf81424-9e75-4e15-b31f-a03d607d5890/","version":null,"content_type":"ostree_ref","additional_metadata":null}
    ]
    ```
    thus, searching for `name = <filename>` doesn't yield anything either.
3. We could consider using the `filename` field for searching, but this is `null` for this upload?

evgeni:
Thanks a a ton @jjeffers!

Minus the problem that the module currently can't find an already uploaded ostree ref, this seems to be working as expected!

While testing this I found a small bug in Katello: https://github.com/Katello/katello/pull/9862
And a bigger issue with the way our installer enables optional content types: https://projects.theforeman.org/issues/34154

plugins/modules/content_upload.py
jjeffers: IIRC the checksum reported in Katello for those units will be the checksum of the OStree commit, not the checksum of the file itself.

I'll see about an alternate search field.

plugins/modules/content_upload.py
```diff
@@ -17,11 +17,13 @@
 # along with this program.  If not, see <http://www.gnu.org/licenses/>.
 
 from __future__ import absolute_import, division, print_function
+
 __metaclass__ = type
 
 
 DOCUMENTATION = '''
 ---
+Note that the OSTree content upload is not idempotent - running mutliple times will attempt to upload the content unit.
```
evgeni: this would need to go in the `notes` section (line 60-ish)

`DOCUMENTATION` is  YAML

jlsherrill:
@evgeni i'm taking this over if it needs any changes, mind kicking the ci?

evgeni:
@jlsherrill ack, when I talked with James the last time, we ended up on "this looks good, and Evgeni will finish the test part", which I obviously didn't yet…

From a Katello perspective, you could have a thought on how can we ensure idempotency. For RPMs, the module asks "search for this NEVRA" and if the search comes up empty, does the upload, otherwise considers it already present. We (James and I) couldn't find a good way to do the same for OStree content, as a single tarball might contain multiple refs and there is no way to ask "do you have all of these" (essentially: the uploaded artifact and the stored artifact have different "forms").

jlsherrill:
yes, that is a tough one :(   I can chat with the pulp team about how we might do this.

evgeni:
Would it be sensible to list all refs from the tarball and ask the api if it has all of them (via one or multiple calls), if any are missing upload the whole tarball, otherwise don't upload?

We'd upload more that needed if only some are missing, but oh well

jlsherrill:
@evgeni i chatted with the pulp team and its possible to get the list of refs and their commit hash.  It is possible by:
1.  extracting the tarball
2.  running `ostree  summary --repo path/to/extracted/repo/REPO_NAME/  --view`
3. parsing this out and getting the hashes
4. querying katello for all refs in the repo and comparing the list

Extracting it out could take a while and use a lot of disk space, but i think its the only way to accomplish this.  Thoughts on this?

jlsherrill:
any thoughts @evgeni ?

evgeni:
That sounds sensible, yeah. But out of scope for this PR

mdellweg:
> @evgeni i chatted with the pulp team and its possible to get the list of refs and their commit hash. It is possible by:
> 
>     1. extracting the tarball
> 
>     2. running `ostree  summary --repo path/to/extracted/repo/REPO_NAME/  --view`
> 
>     3. parsing this out and getting the hashes
> 
>     4. querying katello for all refs in the repo and comparing the list
> 
> 
> Extracting it out could take a while and use a lot of disk space, but i think its the only way to accomplish this. Thoughts on this?

Can you maybe filter the layers in the tar through a hasher without writing anything to disk?

plugins/modules/content_upload.py
```diff
@@ -166,10 +190,15 @@ def main():
...
                 content_upload_payload = {'size': size}
                 content_upload_payload.update(repository_scope)
+
                 content_upload = module.resource_action('content_uploads', 'create', content_upload_payload)
                 content_upload_scope = {'id': content_upload['upload_id']}
                 content_upload_scope.update(repository_scope)
 
+                if module.foreman_params['repository']['content_type'] == 'ostree':
+                    ostree_parameters = {'ostree_repository_name': ostree_repository_name, 'content_type': 'ostree_ref'}
+                    content_upload_scope.update(ostree_parameters)
```
evgeni: @jlsherrill do I read the Katello API correctly, we need to pass `ostree_repository_name` to the `import_uploads` call, but not the individual `content_uploads.update` calls? Aka: this hunk here is unnecessary? Test upload seems to pass fine without on a 4.3 box, but I wanted to be super sure.

evgeni:
I've rebased this and recorded the tests, so it should be good to go.

There is a small inline question and I'd love either @ehelms or @mdellweg have a second look, just because I did touch the code in the past ;-)

mdellweg (approved):
It all looks reasonable to me. But not an OSTree expert here.

plugins/modules/content_upload.py
jlsherrill: correct, the content_uploads api doesn't even know about this parameter

PR #1328, opened 2021-12-13
Warn if the user tries to use a plain HTTP server URL

evgeni:
and fail if the URL is neither HTTPS nor HTTP.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -91,6 +91,7 @@ tests/test_playbooks/vars/server.yml:
 
 dist-test: $(MANIFEST)
 	FOREMAN_SERVER_URL=https://foreman.example.test ansible -m $(NAMESPACE).$(NAME).organization -a "username=admin password=changeme name=collectiontest" localhost | grep -q "Failed to connect to Foreman server.*foreman.example.test"
+	FOREMAN_SERVER_URL=http://foreman.example.test ansible -m $(NAMESPACE).$(NAME).organization -a "username=admin password=changeme name=collectiontest" localhost 2>&1| grep -q "You have configured a plain HTTP server URL."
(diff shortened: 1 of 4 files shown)


Makefile
```diff
@@ -91,6 +91,7 @@ tests/test_playbooks/vars/server.yml:
 
 dist-test: $(MANIFEST)
 	FOREMAN_SERVER_URL=https://foreman.example.test ansible -m $(NAMESPACE).$(NAME).organization -a "username=admin password=changeme name=collectiontest" localhost | grep -q "Failed to connect to Foreman server.*foreman.example.test"
+	FOREMAN_SERVER_URL=http://foreman.example.test ansible -m $(NAMESPACE).$(NAME).organization -a "username=admin password=changeme name=collectiontest" localhost 2>&1| grep -q "You have configured a plain HTTP server URL."
```
mdellweg, line 94: This is to show the warning is happening?

evgeni: correct.

evgeni:
The CodeQL warning is interesting, but unrelated.

mdellweg (approved):

tests/test_crud.py
```diff
@@ -8,6 +8,7 @@
 
 IGNORED_WARNINGS = [
     "Activation Key 'Test Activation Key Copy' already exists.",
+    "You have configured a plain HTTP server URL. All communication will happen unencrypted.",
```
mdellweg, line 11: Now this is an "expected warning"...

evgeni: We use HTTP in the inventory tests, and I am not going to fix that ;)

PR #1329, opened 2021-12-14
Correct collection yum repository

gardar:
Wasn't able to find the collection within the client repository, turns out it's located in the plugins repo.


evgeni (approved):
Indeed, we moved it recently, and forgot to update the docs here. Good catch!

PR #1330, opened 2021-12-16
Add variable in Makefile for python interpreter

bagasse:
RHEL8 don't provides `python` executable, only `python2` or `python3`

This permit to override the default `python` command to the correct local value.


mdellweg (approved):
Sounds good to me.

evgeni (approved):

PR #1333, opened 2022-01-03
Specify the timezone in timestamps

### LINKED ISSUE #1171: The reported facts and reports are ignoring machine timezone

evgeni:
fixes #1171

### DIFF
--- a/plugins/callback/foreman.py
+++ b/plugins/callback/foreman.py
@@ -166,8 +166,10 @@ def get_time():
 def get_now():
     """
     Return the current timestamp as a string to be sent over the network.
+    The time is always in UTC *with* timezone information, so that Ruby
+    DateTime can easily parse it.
     """
-    return datetime.utcnow().isoformat()
+    return datetime.utcnow().strftime("%Y-%m-%d %H:%M:%S+00:00")
 
 
 class CallbackModule(CallbackBase):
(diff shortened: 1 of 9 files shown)


plugins/callback/foreman.py
```diff
@@ -166,8 +166,10 @@ def get_time():
 def get_now():
     """
     Return the current timestamp as a string to be sent over the network.
+    The time is always in UTC *with* timezone information, so that Ruby
+    DateTime can easily parse it.
     """
-    return datetime.utcnow().isoformat()
+    return datetime.utcnow().strftime("%Y-%m-%d %H:%M:%S+00:00")
```
evgeni, line 172: An alternative would be to use `strftime("%Y-%m-%d %H:%M:%SZ")`, where Z means Zulu aka UTC.

Ruby doesn't strictly care, and I must admit I find `+00:00` better to read.

evgeni:
this replaces #1172

mdellweg (approved):

adamruzicka (approved):

evgeni:
:woman_shrugging:

PR #1335, opened 2022-01-14
release 3.1.0

evgeni:


evgeni:
I am not sure why `lint` fails, but not because of the changes in this PR :see_no_evil: 

It fails because it pulls in `rich` 11, which is not compatible with `ansible-lint` before 5.3.2, but I actually don't understand why it just doesn't pull `ansible-lint` 5.3.2, it should just work…

mdellweg (approved):

PR #1337, opened 2022-01-26
Add LDAP auth source role

ephracis:
Add role for managing LDAP authentication sources.

Also extend the documentation of the module with an example of Active Directory integration, in addition to FreeIPA / Red Hat IdM.


ephracis:
Yay! 1337!

ephracis:
I need some assistance on the failing tests. Was that really due to something I changed?

evgeni:
nope

- lint is broken currently (someone *cough* needs to look into that)
- content_rhel sometimes does weird things

roles/auth_source_ldaps/README.md
```diff
@@ -0,0 +1,77 @@
+theforeman.foreman.auth_source_ldaps
```
evgeni: I think the only thing in this PR that prevents me from just merging it is the name of the role…

We have been using the plural for of the module name for most roles, to denote that the role is capable of managing multiple instances of $thing, but `auth_source_ldaps` totally weirds me out.

You can sell me `auth_sources`, or `auth_sources_ldap`, but not `auth_source_ldaps` ;)

ephracis: Haha, yeah I was actually going with `auth_sources_ldap` first at the customer but when I made this PR I noticed that the API endpoint in Foreman is actually called `auth_source_ldaps`.

But I will happily change the name, cause it kinda bugs me too. Just note that it will make the role differ from the API endpoint.

evgeni: To hell with the API endpoint names :)

(you can quote me on that in any Foreman core meeting!)

evgeni:
I think you need to rename a few more files (fixtures, apidoc)

ephracis:
> I think you need to rename a few more files (fixtures, apidoc)

Yeah, I did a new record just to be sure. :)

evgeni (approved):

PR #1340, opened 2022-02-09
properly cleanup locations when we're done with them

evgeni:


tests/test_playbooks/auth_source_ldap.yml
```diff
@@ -73,7 +73,7 @@
       vars:
         location_name: "{{ item }}"
         location_state: "absent"
-      with_items: "{{ domain_locations }}"
+      with_items: "{{ domain_locations | reverse }}"
```
Griffin-Sullivan:
```suggestion
      with_items: "{{ domain_locations[::-1] }}"
```
`reverse` doesn't seem to work correctly. The test was incorrectly passing and still tried to delete the parent location first on the Satellite. Found this syntax online and it seems to be working correctly.

evgeni: Oooh, I know why...

Can you try `| reverse | list`? Reverse probably returns a generator by default 🙄

Griffin-Sullivan: Yeah that works!

Griffin-Sullivan:
Need to reverse these also https://github.com/theforeman/foreman-ansible-modules/blob/develop/tests/test_playbooks/hostgroup.yml#L626

mdellweg (approved):

PR #1342, opened 2022-02-14
Convert2RHEL role - add support for Oracle 8

stejskalleos:


ares:
looks good to me, tests seems to be failing, I may be able to test this for you

stejskalleos:
> looks good to me, tests seems to be failing, I may be able to test this for you

Tests fixed!

tests/test_playbooks/convert2rhel.yml
```diff
@@ -23,8 +23,10 @@
       vars:
         foreman_organization: "Test Organization"
         foreman_validate_certs: false
+        foreman_manifest_path: "~/manifest.zip"
```
evgeni: due to `foreman_convert2rhel_manage_subscription: false` on the next line, this has no effect *plus* the path doesn't exist in CI and would fail then.

evgeni (changes requested):
ACK pending the removal of the one line in the test playbook :)

evgeni (approved):

PR #1343, opened 2022-02-16
don't include _ansible_no_log in the callback fixtures

evgeni:
the value changes in 2.13 from false to null if it's unset


lzap (approved):
Sure, just fixes tests.

PR #1344, opened 2022-02-18
Configure action_groups in /meta/runtime.yml

### LINKED ISSUE #1015: Module defaults groups
##### SUMMARY
As it is now, there are 54 modules that all want
- user
- password
- server-url

Repeating those same informations over and over again with all the modules you use is lots of redundant work.
Creating "module defaults" for each of them is equally redundant.

##### FEATURE REQUEST
 It would be nice to have [Module defaults groups](https://docs.ansible.com/ansible/latest/user_guide/playbooks_module_defaults.html#module-defaults-groups) in order to define that block ONCE for an entire playbook like this:

```
# example_play.yml
- hosts: localhost
  module_defaults:
    group/foreman:
      name: myNameIs
      password: TopS3cr3t
      server_url: foreman.at.home
```
These already exist for aws, azure, gcp, etc..., so I guess most of the needed code should already exist?

miikkajo:
This will configure action_groups in /meta/runtime.yml to allow use of module_defaults
```
  module_defaults:
    group/theforeman.foreman.foreman:
      server_url: https://foreman.local
      username: user
      password: pass
```
reference:
https://docs.ansible.com/ansible/latest/user_guide/playbooks_module_defaults.html


evgeni:
Thank you so much for this!

Can you please add a changelog fragment? Probably something like this:
```yaml
minor_changes:
  - add support for module defaults groups for Ansible core 2.12 (https://github.com/theforeman/foreman-ansible-modules/issues/1015)
```

Also, two questions:
1. Did you intentionally leave out the `*_info` modules?
2. Should we add something to automatically generate this list from the contents of `plugins/modules`? The only module that should not be in the group is IMHO `redhat_manifest`, as it talks to Red Hat portal, not Foreman API.

miikkajo:
@evgeni 

Thanks for commenting,

> 1. Did you intentionally leave out the `*_info` modules?
    should not be in the group is IMHO `redhat_manifest`, as it talks to Red Hat portal, not Foreman API.

I'm actually using Red Hat certified satellite collection and got bit confused about plugin redirects and stuff, 
surely _info modules should also be included.

> 2. Should we add something to automatically generate this list from the contents of `plugins/modules`? The only module that 

I absolutely agree,  best option is to generate list of modules automatically and filter out unwanted modules (redhat_manifest)

mdellweg:
Finally! Thank you so much. And lucky us, older versions of Ansible we try to support seem to ignore this. OTOH, this makes it harder for us to test it in ci.

evgeni:
I have rebased this and added automatic generation of that list \o/

Makefile
```diff
@@ -6,7 +6,8 @@ MANIFEST := build/collections/ansible_collections/$(NAMESPACE)/$(NAME)/MANIFEST.
 
 ROLES := $(wildcard roles/*)
 PLUGIN_TYPES := $(filter-out __%,$(notdir $(wildcard plugins/*)))
-METADATA := galaxy.yml LICENSE README.md meta/runtime.yml requirements.txt changelogs/changelog.yaml CHANGELOG.rst bindep.txt
+RUNTIME_YML := meta/runtime.yml
```
mdellweg, line 9: Is this file ever supposed to live somwhere else?

evgeni: no, I just wanted smth to refer to in multiple places

mdellweg (approved):

evgeni:
I've added tests, let's see that it breaks as expected on non-devel and think how to exclude those there…

evgeni:
@mdellweg would you mind having a quick look whether you find the tests (and their skipping) sane? :)

tests/test_crud.py
```diff
@@ -24,6 +26,8 @@
 
 @pytest.mark.parametrize('module', TEST_PLAYBOOKS)
 def test_crud(tmpdir, module, vcrmode):
+    if module in ['module_defaults'] and not ANSIBLE_SUPPORTS_MODULE_DEFAULTS:
+        pytest.skip("Module defaults only work with Ansible 2.14+")
```
mdellweg: 2.12 or 2.14 now i am confused

tests/test_playbooks/module_defaults.yml
```diff
@@ -0,0 +1,26 @@
...
+    - vars/server.yml
+  module_defaults:
+    group/theforeman.foreman.foreman:
+      username: "{{ foreman_username }}"
+      password: "{{ foreman_password }}"
+      server_url: "{{ foreman_server_url }}"
+  tasks:
+    - name: fetch status
+      status_info:
+        validate_certs: "{{ foreman_validate_certs }}"
```
mdellweg: :screwdriver: Now this works? I thought `collections:` and `module_defaults:` did not work together.

Are you quite sure, the status module _needs_ authentication? Can we also add some other `_info` module?

tests/test_crud.py
evgeni: Thanks, it's 12+!

tests/test_playbooks/module_defaults.yml
mdellweg: https://github.com/ansible/ansible/issues/75279#issuecomment-881946053

evgeni: The tests fail if I don't define the defaults, and pass if I do 🤷‍♀️

And the status module needs auth.

mdellweg (approved):

tests/test_playbooks/module_defaults.yml
evgeni: I did add a FQCN test and another module, just to be like suuuuuuuuper sure ;)

PR #1346, opened 2022-02-24
clarify what src means in content_upload

evgeni:
Fixes: https://bugzilla.redhat.com/show_bug.cgi?id=2055416


evgeni:
the devel/3.10 sanity error is unrelated and is addressed in https://github.com/theforeman/foreman-ansible-modules/pull/1347

ehelms (approved):

PR #1347, opened 2022-02-24
try to use vendored LooseVersion from ansible

evgeni:
this exists since 2.12, given that disutils is deprecated in Python


mdellweg (approved):
I like!

PR #1348, opened 2022-02-24
pin flake8 to < 4

evgeni:
something in the galaxy-importer dependency chain is not compatible with
flake8 v4, which leads to older g-i AND older ansible-lint to be
installed, and that breaks things


evgeni:
LINT IS GREEN :green_apple:

mdellweg (approved):

PR #1353, opened 2022-03-03
vendor distutils.version ourself

### LINKED ISSUE #1352: LooseVersion import doesn't work on ansible < 2.9.20 and < 2.10.8
##### SUMMARY

Originally found by @igramic and reported to me via chat.

When using Ansible 2.9.19 and older, the change from #1347 makes our helper fail to load and thus the whole collection unusable:

```console
% ansible -m theforeman.foreman.domain -a 'server_url=https://localhost username=admin password=changeme' localhost
localhost | FAILED! => {
    "msg": "Could not find imported module support code for foreman_helper.  Looked for either LooseVersion.py or version.py"
}
```

Whereas on 2.9.20 and newer, this correctly loads and executes:
```console
% ansible -m theforeman.foreman.domain -a 'server_url=https://localhost username=admin password=changeme' localhost
localhost | FAILED! => {
    "changed": false,
    "msg": "missing required arguments: name"
}
```

To the best of my knowledge, this was fixed in 2.9.20 via https://github.com/ansible/ansible/commit/f8caa43dc7a7beaf66e719e1e5f7bc6ff70fa7e6

The same applies to the 2.10 branch for [2.10.0,2.10.7], but the error is slightly different:
```console
% ansible -m theforeman.foreman.domain -a 'server_url=https://localhost username=admin password=changeme' localhost
localhost | FAILED! => {
    "msg": "Could not find imported module support code for ansible_collections.theforeman.foreman.plugins.modules.domain.  Looked for (['ansible.module_utils.compat.version.LooseVersion', 'ansible.module_utils.compat.version'])"
}
```

2.10.8 contains https://github.com/ansible/ansible/commit/f51852557dffc2ce5f501c1ab4986e867f55fec07

and then 2.11+ have the "original" fix https://github.com/ansible/ansible/commit/3e1f6484d77f2d7546952cfa22a8534d74ed3dc6

I think the main question is: do we care? And if we do, how should we adjust to not run into this again in future?
Keep in mind, that to "fix" the currently present issue, we'd need to revert the change from #1347 which means we'll see linting errors on Python 3.10+ and with Python 3.12 stuff will stop work until we re-implement the change (thus, again, dropping support for these versions).

For comparison, the various OSes (supported by Foreman) and their Ansible versions:
* Debian 10 (Buster): 2.9.16
* Debian 11 (Bullseye): 2.10.8
* Ubuntu 20.04 (Focal): 2.9.6
* EL7 (EPEL): 2.9.27
* EL7 (cfgmgmt SIG): 2.9.27
* EL8 (cfgmgmt SIG): 2.9.27
* EL8 (Stream): 2.12.2

##### ISSUE TYPE
 - Bug Report

##### ANSIBLE VERSION

```
ansible 2.9.19
  config file = /tmp/aold/lol/ansible.cfg
  configured module search path = ['/home/evgeni/.ansible/plugins/modules', '/usr/share/ansible/plugins/modules']
  ansible python module location = /tmp/aold/lib64/python3.5/site-packages/ansible
  executable location = /tmp/aold/bin/ansible
  python version = 3.5.10 (default, Sep 17 2021, 00:00:00) [GCC 11.2.1 20210728 (Red Hat 11.2.1-1)]
```

##### COLLECTION VERSION

```
3.2.0
```

evgeni:
This reverts commit 5dc0f47d7feeb8b2624d89360635ff40c297ccbf.

### DIFF
--- a/plugins/inventory/foreman.py
+++ b/plugins/inventory/foreman.py
@@ -169,10 +169,7 @@ hostnames:
   - name.split('.')[0]
 '''
 import json
-try:
-    from ansible.module_utils.compat.version import LooseVersion
-except ImportError:
-    from distutils.version import LooseVersion
+from ansible_collections.theforeman.foreman.plugins.module_utils._version import LooseVersion
 from time import sleep
 from ansible.errors import AnsibleError
 from ansible.module_utils._text import to_bytes, to_native, to_text
--- a/plugins/module_utils/_version.py
+++ b/plugins/module_utils/_version.py
@@ -1,0 +1,335 @@
+# Vendored copy of distutils/version.py from CPython 3.9.5
+#
+# Implements multiple version numbering conventions for the
+# Python Module Distribution Utilities.
+#
+# PSF License (see licenses/PSF-license.txt or https://opensource.org/licenses/Python-2.0)
+#
+
+"""Provides classes to represent module version numbers (one class for
+each style of version numbering).  There are currently two such classes
+implemented: StrictVersion and LooseVersion.
+Every version number class implements the following interface:
+  * the 'parse' method takes a string and parses it to some internal
+    representation; if the string is an invalid version number,
+    'parse' raises a ValueError exception
+  * the class constructor takes an optional string argument which,
+    if supplied, is passed to 'parse'
+  * __str__ reconstructs the string that was passed to 'parse' (or
+    an equivalent string -- ie. one that will generate an equivalent
+    version number instance)
+  * __repr__ generates Python code to recreate the version number instance
+  * _cmp compares the current instance with either another instance
+    of the same class or a string (which will be parsed to an instance
+    of the same class, thus must follow the same rules)
+"""
+
+from __future__ import (absolute_import, division, print_function)
+__metaclass__ = type
+
+import re
+
+try:
+    RE_FLAGS = re.VERBOSE | re.ASCII
+except AttributeError:
+    RE_FLAGS = re.VERBOSE
+
+
+class Version:
+    """Abstract base class for version numbering classes.  Just provides
+    constructor (__init__) and reproducer (__repr__), because those
+    seem to be the same for all version numbering classes; and route
+    rich comparisons to _cmp.
+    """
+
+    def __init__(self, vstring=None):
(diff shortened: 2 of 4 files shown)


plugins/inventory/foreman.py
```diff
@@ -169,10 +169,7 @@
   - name.split('.')[0]
 '''
 import json
-try:
-    from ansible.module_utils.compat.version import LooseVersion
-except ImportError:
-    from distutils.version import LooseVersion
```
mdellweg, line 175: Is the vendored `LooseVersion` buggy? Given the careful fallback import i would not expect the new version to be less compatible than the old one.

evgeni: no, but ansible < 2.9.20 and < 2.10.8 is buggy when it comes to imports with try/except: https://github.com/theforeman/foreman-ansible-modules/issues/1352

mdellweg: Oh wow. Are there 2.9.21 and 2.10.9 that fix it? I'd say supporting the latest z in every y would be acceptable.

evgeni: 2.9.20 and 2.10.8 are the fixing versions, yes. versions prior to that flat out go belly up.

I am torn, as not every supported deployment platform of ours has these versions easily available :(

mdellweg: Giving a sad ACK here.

mdellweg (approved):

evgeni:
@mdellweg looking for an opinion…

Other collections ended up vendoring `version.py` themself, until they can drop support for 2.9/2.10 completely.

Example: https://github.com/ansible-collections/community.general/commit/cf7a33356cf746746f5649fe2f3450985e9db16a

What do you think about this approach?

mdellweg:
What a mess. We could depend on community.general and use their `module_utils._version` but i guess that won't make things much better.Vendoring or just carrying our own version compare tools sounds like an option.

evgeni:
We can't depend on `community.general` (well, *we* can, `redhat.satellite` can't, but I kinda have too many hats)

mdellweg (approved):

PR #1360, opened 2022-03-18
declare `hostnames` param of the inventory to be a list of strings

evgeni:


evgeni:
this is something the new `validate-modules` found, which now also scans other plugins:

```
ERROR: plugins/inventory/foreman.py:0:0: parameter-list-no-elements: DOCUMENTATION.options.hostnames: Argument defines type as list but elements is not defined for dictionary value @ data['options']['hostnames']. Got {'description': ['A list of templates in order of precedence to compose inventory_hostname.', 'If the template results in an empty string or None value it is ignored.'], 'type': 'list', 'default': ['name']}
```

mdellweg (approved):

PR #1361, opened 2022-03-18
Adding downstream test to Makefile

Griffin-Sullivan:
These changes will let QE easily test the `test_playbooks` folder in Jenkins. All we are doing is deselecting the `UPSTREAM_ONLY` in `conftest.py`, so that we have accurate results for Satellite.


evgeni:
how about this as an alternative? https://github.com/theforeman/foreman-ansible-modules/pull/1363

this makes it easy for us as it does not enforce any "downstream" notation, nor any definition what "downstream" is (satellite isn't the only ds of foreman).

PR #1363, opened 2022-03-21
allow excluding certain playbooks from test execution

evgeni:
useful if you want to run only a subset of tests, but not call each of
them individually

example:

    FAM_TEST_EXCLUDE_PLAYBOOKS=activation_key.yml,bookmark.yml make test


mdellweg (approved):

evgeni:
Waiting on feedback from @Griffin-Sullivan whether this makes his life easier :)

Griffin-Sullivan:
LGTM. Thanks for the simple implementation!

PR #1367, opened 2022-03-25
ignore fqcn-builtins warnings for now

evgeni:


mdellweg (approved):
Wild guess: We test with ansible versions, where the builtins didn't have a namespace and collection name?

evgeni:
Yeah, I just don't trust it to work properly on 2.9/2.10, and don't want to do the same dance as the last time…

PR #1373, opened 2022-04-06
install ansible from a tarball, not git in CI

evgeni:
in a very unscientific, once-run benchmark, this gives the following
results (all with all dependencies from cache):

* `pip install https://github.com/ansible/ansible/archive/devel.tar.gz` - 8.2s
* `pip install 'git+https://github.com/ansible/ansible.git#egg-name=ansible-core` - 47s

Saving ~40 seconds on every CI run sounds like a good idea.

### DIFF
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -51,7 +51,7 @@ jobs:
       - name: Install libyaml-dev for PyYAML
         run: sudo apt-get install -y libyaml-dev
       - name: Install Ansible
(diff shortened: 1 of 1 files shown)


evgeni:
In reality, the gain will be probably lower, or vary more, as the expensive step is the `git clone` that `pip` performs under the hood, and that is very dependent on which GH frontend it hits (probably faster inside GHA) and tons of other local factors (disk, memory, CPU).

evgeni:
the CI errors are unrelated and fixed in https://github.com/theforeman/foreman-ansible-modules/pull/1374

mdellweg (approved):

PR #1376, opened 2022-04-06
use include_tasks, not include in test playbooks

evgeni:
include is deprecated


mdellweg (approved):
I take the green tests as a proof that we don't support the versions of ansible without `include_tasks` anymore.

evgeni:
> I take the green tests as a proof that we don't support the versions of ansible without `include_tasks` anymore.

We've been using include tasks for a while without issues, just not consistently

PR #1381, opened 2022-04-12
Use correct var for kind in provisioning_templates role

vinzent:
The `bmc_proxy` variable doesn't seem to be correct.


mdellweg (approved):
I cannot argue with that.

evgeni (approved):

PR #1384, opened 2022-04-21
make inventory tests also work with podman

evgeni:
also split the test setup and test execution in own task files, for
better reusability


tests/test_playbooks/tasks/inventory_plugin_setup.yml
```diff
@@ -0,0 +1,107 @@
...
+  docker_container:
+    name: "{{ postgres_host }}"
+    image: docker.io/library/postgres:{{ postgres_version }}
+    env:
+      POSTGRES_USER: foreman
+      POSTGRES_PASSWORD: foreman
+      POSTGRES_DATABASE: foreman
+      PGDATA: /var/lib/postgresql/data/pgdata
+    published_ports:
+      - "0.0.0.0:{{ postgres_port }}:5432"
```
ehelms, line 22: Up above there is a playbook that uses 5433 for the postgresql port. Does that affect this?

evgeni: this line says "expose port 5432 of the container on whatever evgeni set as postgres_port"

so yes, it does affect, but correctly ;)

ehelms (approved):

PR #1397, opened 2022-04-29
Allow to set label for content views

gvde:
This allows to define a specific label when creating a content view.

### DIFF
--- a/changelogs/fragments/1397-content_views-label.yml
+++ b/changelogs/fragments/1397-content_views-label.yml
@@ -1,2 +1,0 @@
- minor_changes: 
-   - content_views - add support to set label during creation.
--- a/changelogs/fragments/1397-content_views-label.yml
+++ b/changelogs/fragments/1397-content_views-label.yml
@@ -1,0 +1,2 @@
+ minor_changes: 
+   - content_views - add support to set label during creation.
--- a/changelogs/fragments/1397-content_view-label.yml
+++ b/changelogs/fragments/1397-content_view-label.yml
@@ -1,0 +1,2 @@
+ minor_changes: 
+   - content_view - add support to set label during creation.
--- a/plugins/modules/content_view.py
+++ b/plugins/modules/content_view.py
@@ -39,7 +39,7 @@ options:
     type: str
   label:
     description:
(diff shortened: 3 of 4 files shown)


evgeni (changes requested):
Thanks!

Would you mind adding a changelog entry for this? A recent example:
https://github.com/theforeman/foreman-ansible-modules/blob/10def2b0ca03d7b779d85a25e1ce3a84b2beda2e/changelogs/fragments/1388-repository-mirroring_policy.yml#L1-L2

Also, I think the description should mention that one can't update a label after the CV was created.

changelogs/fragments/1397-content_views-label.yml
```diff
@@ -0,0 +1,2 @@
+ minor_changes: 
+   - content_views - add support to set label during creation.
```
evgeni: content_view please, to match the module name

gvde:
> Thanks!
> 
> Would you mind adding a changelog entry for this? A recent example:
> 
> https://github.com/theforeman/foreman-ansible-modules/blob/10def2b0ca03d7b779d85a25e1ce3a84b2beda2e/changelogs/fragments/1388-repository-mirroring_policy.yml#L1-L2

O.K. I did.

> Also, I think the description should mention that one can't update a label after the CV was created.

O.K. Updated the description. (Same thing would apply to repository, organization and other places. I think the label can never be changed after creation).

evgeni:
> Same thing would apply to repository, organization and other places. I think the label can never be changed after creation


I think you're right. If you feel like tracking down where we didn't document that properly and open a PR… ;)

evgeni (approved):

gvde:
> > Same thing would apply to repository, organization and other places. I think the label can never be changed after creation
> 
> 
> I think you're right. If you feel like tracking down where we didn't document that properly and open a PR… ;)
> 

I'll check. I'am getting the hang of it. What bothers me more is that I did get an error message when the content view label in the playbook didn't match an existing content view but not in the same case for a repository. Of course in either case a change isn't possible so in either case there should be an error.

But I'll call it a day and look after that next week.

PR #1407, opened 2022-05-13
find operatingsystems by title or full (name,major,minor) tuple

### LINKED ISSUE #1401: operatingsystems ambiguity
##### SUMMARY
Running against a katello 4.3 or 4.4 server my playbook complains about the operating systems to be created/updated:

```
[WARNING]: While constructing a mapping from /home/k/k202081/git/foreman-ansible/first.yaml, line 1834, column 9, found a duplicate dict key (foreman_operatingsystems). Using
last defined value only.
```

It seems the problem is that multiple operating systems can have the same name, e.g. AlmaLinux. I have AlmaLinux with major version 8 and no minor version (AlmaLinux 8) and one with minor version 5 (AlmaLinux 8.5)

When adding and syncing AlmaLinux repositories it automatically generates a distribution and operating system "AlmaLinux 8".

The a client host with AlmaLinux 8.5 is registered it automatically generates an operating system "AlmaLinux 8.5" (from puppet facts or subscription manager?)

So this scenario seems standard.

##### ANSIBLE VERSION

```
$ ansible --version
ansible 2.9.27
  config file = /home/k/k111111/git/foreman-ansible/ansible.cfg
  configured module search path = ['/home/k/k111111/.ansible/plugins/modules', '/usr/share/ansible/plugins/modules']
  ansible python module location = /usr/lib/python3.6/site-packages/ansible
  python version = 3.6.8 (default, Nov 17 2021, 16:10:06) [GCC 8.5.0 20210514 (Red Hat 8.5.0-3)]
```

##### COLLECTION VERSION
Using latest version 3.3.0 from ansible-galaxy with some patches from pull requests.

##### KATELLO/FOREMAN VERSION

```
katello-4.4.0-1.el8.noarch
foreman-3.2.0-1.el8.noarch
```

##### STEPS TO REPRODUCE
Playbook

```
    - role: theforeman.foreman.operatingsystems
      vars:
        foreman_operatingsystems:
          - name: "AlmaLinux"
            major: "8"
            description: "AlmaLinux 8"
            os_family: "Redhat"
            architectures:
              - "x86_64"
            password_hash: "SHA256"
            media:
              - "AlmaLinux 8"
              - "Local AlmaLinux 8"
        foreman_operatingsystems:
          - name: "AlmaLinux"
            major: "8"
            minor: "5"
            description: "AlmaLinux 8.5"
            os_family: "Redhat"
            architectures:
              - "x86_64"
            password_hash: "SHA256"
            media:
              - "AlmaLinux 8"
              - "Local AlmaLinux 8"
           override: disabled
```

##### EXPECTED RESULTS
No error and correct configuration of all operating system entities.

evgeni:
Operatingsystems in Foreman have a rather weird uniqueness construct:
they have a `description` attribute, that is unique, but can be blank,
and they are unique by the tuple (`name`, `major`, `minor`) but here
again `minor` can be blank too.
They also have a generated `title` attribute, that is either the value
of the description attribute (if that is not blank) or a concatenation
of name/major/minor, shortened to 255 characters.

In Rails, this looks like that (roughly quoting
https://github.com/theforeman/foreman/blob/develop/app/models/operatingsystem.rb):

    validates :name, :presence => true, :no_whitespace => true,
              :uniqueness => { :scope => [:major, :minor], :message => N_("Operating system version already exists")}
    validates :description, :uniqueness => true, :allow_blank => true
    before_validation :downcase_release_name, :set_title, :stringify_major_and_minor
    validates :title, :uniqueness => true, :presence => true

    def release
      "#{major}#{('.' + minor.to_s) if minor.present?}"
    end

    def fullname
      "#{name} #{release}"
    end

    # The OS is usually represented as the concatenation of the OS and the revision
    def to_label
      return description if description.present?
      fullname
    end

    def set_title
      self.title = to_label.to_s[0..254]
    end

Overall, this construct makes finding a specific OS rather "interesting":
If you have two OSes, "EL 8" and "EL 8.6" (so both have name:EL, major:8
and one has minor:6), you can't search for "name=EL,major=8" as that
would return *both*, you also can't search for
"name=EL,major=8,not has minor" as the minor could actually be an empty
string instead of NULL.
However, you can search for title="EL 8" and title="EL 8.6" and get the
unique OS that you wanted.
Yet, when we ask the user to provide a description, they can use more
than 255 characters that is allowed for the title -- so we need to look
up in both fields, title AND description as the former is always there
(generated) while the later will have the exact string *if* set.

### DIFF
--- a/plugins/modules/operatingsystem.py
+++ b/plugins/modules/operatingsystem.py
@@ -211,12 +211,12 @@ def main():
         entity = None
         # If we have a description, search for it
         if 'description' in module_params and module_params['description'] != '':
-            search_string = 'description="{0}"'.format(module_params['description'])
+            search_string = 'description="{0}" or title="{0}"'.format(module_params['description'])
             entity = module.find_resource('operatingsystems', search_string, failsafe=True)
         # If we did not yet find a unique OS, search by name & version
         # In case of state == absent, those information might be missing, we assume that we did not find an operatingsytem to delete then
         if entity is None and 'name' in module_params and 'major' in module_params:
-            search_string = ','.join('{0}="{1}"'.format(key, module_params[key]) for key in ('name', 'major', 'minor') if key in module_params)
+            search_string = ','.join('{0}="{1}"'.format(key, module_params.get(key, '')) for key in ('name', 'major', 'minor'))
             entity = module.find_resource('operatingsystems', search_string, failsafe=True)
 
         if not entity and (module.state == 'present' or module.state == 'present_with_defaults'):
--- a/changelogs/fragments/1401-os-search-title.yml
+++ b/changelogs/fragments/1401-os-search-title.yml
@@ -1,0 +1,2 @@
+bugfixes:
+  - operatingsystem - find operatingsystems by title or full (name,major,minor) tuple (https://github.com/theforeman/foreman-ansible-modules/issues/1401)
--- a/tests/test_playbooks/fixtures/operatingsystem-0.yml
+++ b/tests/test_playbooks/fixtures/operatingsystem-0.yml
@@ -14,7 +14,7 @@ interactions:
     uri: https://foreman.example.org/api/status
   response:
     body:
-      string: '{"result":"ok","status":200,"version":"2.1.2","api_version":2}'
+      string: '{"result":"ok","status":200,"version":"3.2.0","api_version":2}'
     headers:
       Cache-Control:
       - max-age=0, private, must-revalidate
@@ -33,7 +33,7 @@ interactions:
       Foreman_current_organization:
       - ; ANY
       Foreman_version:
-      - 2.1.2
+      - 3.2.0
       Keep-Alive:
       - timeout=15, max=100
       Strict-Transport-Security:
@@ -67,12 +67,13 @@ interactions:
       User-Agent:
       - apypie (https://github.com/Apipie/apypie)
     method: GET
-    uri: https://foreman.example.org/api/operatingsystems?search=description%3D%22famos+SP1%22&per_page=4294967296
+    uri: https://foreman.example.org/api/operatingsystems?search=description%3D%22famos+SP1%22+or+title%3D%22famos+SP1%22&per_page=4294967296
   response:
     body:
-      string: "{\n  \"total\": 1,\n  \"subtotal\": 0,\n  \"page\": 1,\n  \"per_page\"\
-        : 4294967296,\n  \"search\": \"description=\\\"famos SP1\\\"\",\n  \"sort\"\
-        : {\n    \"by\": null,\n    \"order\": null\n  },\n  \"results\": []\n}\n"
+      string: "{\n  \"total\": 3,\n  \"subtotal\": 0,\n  \"page\": 1,\n  \"per_page\"\
+        : 4294967296,\n  \"search\": \"description=\\\"famos SP1\\\" or title=\\\"\
+        famos SP1\\\"\",\n  \"sort\": {\n    \"by\": null,\n    \"order\": null\n\
+        \  },\n  \"results\": []\n}\n"
     headers:
       Cache-Control:
       - max-age=0, private, must-revalidate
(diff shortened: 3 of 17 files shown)


plugins/modules/operatingsystem.py
```diff
@@ -211,12 +211,12 @@ def main():
         entity = None
         # If we have a description, search for it
         if 'description' in module_params and module_params['description'] != '':
-            search_string = 'description="{0}"'.format(module_params['description'])
+            search_string = 'description="{0}" or title="{0}"'.format(module_params['description'])
```
mdellweg, line 214: What if two descriptions share the same 255 first characters?
Should we not search for description first and then fallback to title?

edit: :thinking: If you supply the full description, the title will probably never match. So maybe we _are_ OK.

mdellweg (approved):

plugins/modules/operatingsystem.py
evgeni: You mean OS1 has description "longlong255" and OS2 has description "longlong255andmore"? That's not possible:

```irb
irb(main):020:0> os1 = Operatingsystem.new(name: 'test1', major: 8, description: 'AAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAA')
=> #<Operatingsystem id: nil, major: "8", name: "test1", minor: "", nameindicator: nil, created_at: nil, updated_at: nil, release_name: nil, type: nil, description: "AAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAA...", pass...
irb(main):021:0> os1.save
=> true
irb(main):022:0> os2 = Operatingsystem.new(name: 'test2', major: 8, description: 'AAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAandmore')
=> #<Operatingsystem id: nil, major: "8", name: "test2", minor: "", nameindicator: nil, created_at: nil, updated_at: nil, release_name: nil, type: nil, description: "AAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAA...", pass...
irb(main):023:0> os2.save
=> false
```

The second OS fails validation, as the title needs to be unique (but isn't)

mdellweg: So we only need to search by "desc or title" because the user could provide the full _or_ shortened version?

evgeni: In reality we could drop the whole `desc or title` search alltogether, and go by `name,major,minor` only, but that'd be a breaking change and I wanted to avoid that.

And if we keep `desc or title`, it needs to be both for the reasons you stated, yes.

mdellweg: And for the uniqueness you already stated, it should be impossible to create a description that matches the `<name> <major>.<minor>` of a different OS. Also i like the convenience we provide here.

PR #1408, opened 2022-05-13
Update test assertion in module_defaults.yml

Griffin-Sullivan:
The api returns `setting_info['setting']['value'] == null` so in Ansible we need to use None.

### DIFF
--- a/tests/test_playbooks/module_defaults.yml
+++ b/tests/test_playbooks/module_defaults.yml
@@ -33,4 +33,4 @@
       assert:
         that:
           - setting_info['setting']['name'] == "login_text"
-          - setting_info['setting']['value'] == ''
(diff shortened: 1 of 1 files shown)


evgeni:
On which setup do you get `null`?

The fixtures were recorded on 2.5, where it seems to have returned an empty string.

I do have a 3.2 here that says `Version $VERSION` for the `value`, which was set in https://projects.theforeman.org/issues/33417

Edit: ah, I see. On a 3.1 it is indeed `null`. Funny.

I guess we should just use a different setting for this test XD

Griffin-Sullivan:
Haha yes that might be a better idea. I just know this was failing in 6.11 for me. Might also be failing in 6.9 and 6.10.

PR #1412, opened 2022-05-17
Release 3.4.0

evgeni:


mdellweg (approved):

gardar:
Seems like this release hasn't made it in the [yum repo](http://yum.theforeman.org/plugins/latest/) yet, any chance that could be fixed?

evgeni:
> Seems like this release hasn't made it in the [yum repo](http://yum.theforeman.org/plugins/latest/) yet, any chance that could be fixed?

That's not *entirely* correct, it *is* available in the `nightly` repo, but I see your point. I'll do the backports :)

gardar:
Ah, well one would expect the `nightly` repo to be the latest `development` version, and the `latest` repo to contain the latest stable version.
Also the `nightly` repo isn't available on el7, so it's unavailable for those of us that are stuck with redhat satellite on el7.

evgeni:
Packages were built for 3.3 (which is what latest currently points at) and should be available in the repos in the next few hours!

PR #1417, opened 2022-05-25
Allow to promote new version to lifecycle environments automatically.

gvde:
Allow to automatically promote new version to lifecycle environments in role theforeman.foreman.content_view_version.

### DIFF
--- a/changelogs/fragments/1417-content_view_publish-lifecycle-environments.yml
+++ b/changelogs/fragments/1417-content_view_publish-lifecycle-environments.yml
@@ -1,0 +1,2 @@
+minor_changes: 
+   - content_view_publish - add support to promote new version to lifecycle environments.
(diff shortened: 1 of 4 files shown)


gvde:
O.K. Maybe I shouldn't try to add tests and drop that last commit again...

mdellweg:
> O.K. Maybe I shouldn't try to add tests and drop that last commit again...

Quite the opposite. I guess however, you need to update the recoreded fixtures for the test, once you made it pass with `make livetest_<your_test_name>`.

gvde:
> Quite the opposite. I guess however, you need to update the recoreded fixtures for the test, once you made it pass with `make livetest_<your_test_name>`.

Yes. I know. But I would need a complete test environment to record that and a live server to test against. And I don't currently have to time to dive into all that, in particular because for me it's mostly a one-time thing anyway, because I am using ansible only to migrate everything from our old EL7 server to EL8, as far as it's possible fighting all the bugs on the way. So for me, I am good to patch it into my local collection for the time being...

mdellweg:
Ultimately, this is probably @evgeni 's call. Usually, we cannot claim that we support a feature that is not tested. Adding "help wanted" so maybe someone will help recording.

PR #1424, opened 2022-06-01
New download policy options for smart proxy

gvde:
Download policy `background` has been deprecated in Katello 4.0. There is also a new option `inherit`.

### DIFF
--- a/tests/test_playbooks/fixtures/katello_smart_proxy-4.yml
+++ b/tests/test_playbooks/fixtures/katello_smart_proxy-4.yml
@@ -105,7 +105,7 @@ interactions:
       x-xss-protection: [1; mode=block]
     status: {code: 200, message: OK}
 - request:
-    body: !!python/unicode '{"smart_proxy": {"download_policy": "background"}}'
+    body: !!python/unicode '{"smart_proxy": {"download_policy": "inherit"}}'
     headers:
       Accept: [application/json;version=2]
       Accept-Encoding: ['gzip, deflate']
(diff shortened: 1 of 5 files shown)


tests/test_playbooks/fixtures/katello_smart_proxy-4.yml
```diff
@@ -105,7 +105,7 @@ interactions:
       x-xss-protection: [1; mode=block]
     status: {code: 200, message: OK}
 - request:
-    body: !!python/unicode '{"smart_proxy": {"download_policy": "background"}}'
+    body: !!python/unicode '{"smart_proxy": {"download_policy": "inherit"}}'
```
mdellweg, line 108: This is cheating.

gvde: I literally didn't know what I was doing and how all those tests work. After changing it in the code the tests failed. So I cheated.

If you tell me how it's done correctly, I'll change it. Or you create a new pull request the right way. It's a small change anyway...

mdellweg: Sorry for being too brief. These fixtures are actual server responses, and by hand changing them we would loose trust in them. So they should only be recreated by rerecording (see the readme).

But there is also a deeper issue with the change here. I do not think that the modules are supposed to _only_ support Katello >= 4.0. So we cannot just remove the old option. At the very least, you should keep it and tell about the version dependence in the modules inline documentation. Maybe some version dependent validation on execution is also possible.

Last but not least, this change needs a changelog snippet.

gvde: Well, I guess that's beyond my capabilities and available time at the moment. I thought it was a straight-forward change but I didn't see everything else affected. I'll close this and let this to someone more experienced...

evgeni:
This was fixed in 6c919dd5ccb3ad8303c74d5e3836274a9c929642 and https://github.com/theforeman/foreman-ansible-modules/commit/42a0d79706938111c62be8375d692d14d30bab8c

PR #1431, opened 2022-06-15
Added discovery_rule module

Thulium-Drake:
Hi guys!

I took a stab at making a module for managing discovery rules and besides my amazement how easy it was to make (thanks to all the amazing modelling work you've put in ForemanEntityAnsibleModule no doubt ;-) ) it also works!

However, I need a bit more help from you, as I can't find out why it keeps reporting a change:

First the playbook I use:
```
---
- name: 'Test disco'
  hosts: 'sat.rh.lab'
  gather_facts: false
  tasks:
    - name: 'Ensure Disco Rule'
      theforeman.foreman.discovery_rule:
        username: 'admin'
        password: 'banana!!'
        server_url: 'https://sat.rh.lab'
        name: 'ansible-test'
        search: 'mac = bb:bb:bb:bb:bb:bb'
        hostgroup: 'RedHat7-AnsibleController-Infra'
        hostname: 'rhel7-controller'
        enabled: true
        max_count: 1
        priority: 99
        organizations:
          - 'Lab-Inc'
        locations:
          - 'DC1'

    - meta: end_play  # Disable remove task for now, it does work
    - name: 'Ensure Disco Rule'
      theforeman.foreman.discovery_rule:
        username: 'admin'
        password: 'banana!!'
        server_url: 'https://sat.rh.lab'
        name: 'ansible-test'
        state: 'absent'
```

The result of the first pass creates the rule as expected, but the second pass _always_ returns the following diff (ran with ```-vv --diff```)
```
ask path: /root/projects/rhlab/playbooks/sat_disco.yml:6
--- before
+++ after
@@ -3,10 +3,7 @@
   hostgroup_id: 3
   hostname: rhel7-controller
   id: 5
-  location_ids:
-  - 3
+  max_count: 1
   name: ansible-test
-  organization_ids:
-  - 1
   priority: 99
   search: mac = bb:bb:bb:bb:bb:bb

changed: [sat.rh.lab] => changed=true 
  entity:
    discovery_rules:
    - created_at: '2022-06-15T19:04:20.728Z'
      enabled: true
      hostgroup_id: 3
      hostname: rhel7-controller
      id: 5
      max_count: 1
      name: ansible-test
      priority: 99
      search: mac = bb:bb:bb:bb:bb:bb
      updated_at: '2022-06-15T19:12:14.157Z'
```

Unit tests will follow soon, I need to build a test-rig which can record first :)

### DIFF
--- a/plugins/modules/discovery_rule.py
+++ b/plugins/modules/discovery_rule.py
@@ -141,6 +141,13 @@ def main():
     )
 
     with module.api_connection():
+        entity = module.lookup_entity('entity')
+
+        # workround the fact that the API expects `max_count` when modifying the entity
+        # but uses `hosts_limit` when showing one
+        if entity and 'hosts_limit' in entity:
+            entity['max_count'] = entity.pop('hosts_limit')
+
         module.run()
 
 
--- a/plugins/modules/discovery_rule.py
+++ b/plugins/modules/discovery_rule.py
@@ -1,6 +1,5 @@
 #!/usr/bin/python
 # -*- coding: utf-8 -*-
-# (c) 2017, Matthias M Dellweg <dellweg@atix.de> (ATIX AG)
 # (c) 2022, Jeffrey van Pelt <jeff@vanpelt.one>
 #
 # This program is free software: you can redistribute it and/or modify
@@ -23,7 +22,7 @@ __metaclass__ = type
 DOCUMENTATION = '''
 ---
 module: discovery_rule
-version_added: x.x.x
+version_added: 3.5.0
 short_description: Manage Host Discovery Rules
 description:
   - Manage Host Discovery Rules
@@ -64,20 +63,9 @@ options:
       - 0 means no limit
     type: int
 extends_documentation_fragment:
-  locations:
-    description:
-      - List of locations the discovery rule should be assigned to
-    type: list
-    elements: str
-  organizations:
-    description:
-      - List of organizations the discovery rule should be assigned to
-    type: list
-    elements: str
-extends_documentation_fragment:
   - theforeman.foreman.foreman
   - theforeman.foreman.foreman.entity_state
-  - theforeman.foreman.foreman.nested_parameters
+  - theforeman.foreman.foreman.taxonomy
 '''
 
 EXAMPLES = '''
--- a/plugins/modules/discovery_rule.py
+++ b/plugins/modules/discovery_rule.py
@@ -32,15 +32,45 @@ options:
   name:
     description:
       - Name of the Discovery Rule
-    required: true
+    required: True
     type: str
-  parent:
+  search:
     description:
-      - Title of a parent Location for nesting
+      - Expression to match newly discovered hosts with
+    required: True
     type: str
+  hostgroup:
+    description:
+      - Hostgroup to assign hosts to
+    required: True
+    type: str
+  hostname:
+    description:
+      - Hostname to assign to discovered host(s)
+      - When matching multiple hosts, must provide unique hostnames for each of the discovered hosts
+    type: str
+  enabled:
+    description:
+      - Enable or disable the rule
+    type: bool
+  priority:
+    description:
+      - Priority of the rule
+    type: int
+  max_count:
+    description:
+      - Maximum amount of hosts to provision with the rule
+      - 0 means no limit
+    type: int
+extends_documentation_fragment:
+  locations:
+    description:
+      - List of locations the discovery rule should be assigned to
+    type: list
+    elements: str
   organizations:
     description:
-      - List of organizations the location should be assigned to
+      - List of organizations the discovery rule should be assigned to
     type: list
     elements: str
 extends_documentation_fragment:
--- a/plugins/modules/discovery_rule.py
+++ b/plugins/modules/discovery_rule.py
@@ -1,0 +1,119 @@
+#!/usr/bin/python
+# -*- coding: utf-8 -*-
+# (c) 2022, Jeffrey van Pelt <jeff@vanpelt.one>
+#
+# This program is free software: you can redistribute it and/or modify
+# it under the terms of the GNU General Public License as published by
+# the Free Software Foundation, either version 3 of the License, or
+# (at your option) any later version.
+#
+# This program is distributed in the hope that it will be useful,
+# but WITHOUT ANY WARRANTY; without even the implied warranty of
+# MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+# GNU General Public License for more details.
+#
+# You should have received a copy of the GNU General Public License
+# along with this program.  If not, see <http://www.gnu.org/licenses/>.
+
+from __future__ import absolute_import, division, print_function
+__metaclass__ = type
+
+
+DOCUMENTATION = '''
+---
+module: discovery_rule
+version_added: x.x.x
+short_description: Manage Host Discovery Rules
+description:
+  - Manage Host Discovery Rules
+author:
+  - "Jeffrey van Pelt (@Thulium-Drake)"
+options:
+  name:
+    description:
+      - Name of the Discovery Rule
+    required: true
+    type: str
+  parent:
+    description:
+      - Title of a parent Location for nesting
+    type: str
+  organizations:
+    description:
+      - List of organizations the location should be assigned to
+    type: list
+    elements: str
+extends_documentation_fragment:
+  - theforeman.foreman.foreman
+  - theforeman.foreman.foreman.entity_state
+  - theforeman.foreman.foreman.nested_parameters
+'''
+
+EXAMPLES = '''
+- name: 'Ensure Discovery Rule'
+    theforeman.foreman.discovery_rule:
+      username: 'admin'
+      password: 'secret_password'
+      server_url: 'https://foreman.example.com'
+      name: 'my-first-disco'
+      search: 'mac = bb:bb:bb:bb:bb:bb'
+      hostgroup: 'RedHat7-Base'
+      hostname: 'servera'
+      max_count: 1
+      organizations:
+        - 'MyOrg'
+      locations:
+        - 'DC1'
+
+- name: 'Remove Discovery Rule'
+  theforeman.foreman.discovery_rule:
+    username: 'admin'
+    password: 'secret_password'
+    server_url: 'https://foreman.example.com'
+    name: 'my-first-disco'
+    state: 'absent'
+'''
+
+RETURN = '''
+entity:
+  description: Final state of the affected entities grouped by their type.
+  returned: success
+  type: dict
+  contains:
+    discovery_rules:
+      description: List of discovery rules.
+      type: list
+      elements: dict
+'''
+
+from ansible_collections.theforeman.foreman.plugins.module_utils.foreman_helper import ForemanEntityAnsibleModule, NestedParametersMixin
+
+
+class ForemanDiscoveryRuleModule(NestedParametersMixin, ForemanEntityAnsibleModule):
+    pass
+
+
+def main():
+    module = ForemanDiscoveryRuleModule(
+        foreman_spec=dict(
+            name=dict(required=True),
+            search=dict(),
+            hostgroup=dict(type='entity'),
+            hostname=dict(),
+            max_count=dict(type='int'),
+            priority=dict(type='int'),
+            enabled=dict(type='bool'),
+            organizations=dict(type='entity_list'),
+            locations=dict(type='entity_list'),
+        ),
+        required_if=[
+            ['state', 'present', ['hostgroup', 'search']],
+        ],
+    )
+
+    with module.api_connection():
+        module.run()
+
+
+if __name__ == '__main__':
+    main()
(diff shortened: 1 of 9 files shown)


plugins/modules/discovery_rule.py
```diff
@@ -0,0 +1,149 @@
...
+    discovery_rules:
+      description: List of discovery rules.
+      type: list
+      elements: dict
+'''
+
+from ansible_collections.theforeman.foreman.plugins.module_utils.foreman_helper import ForemanEntityAnsibleModule, NestedParametersMixin
+
+
+class ForemanDiscoveryRuleModule(NestedParametersMixin, ForemanEntityAnsibleModule):
```
evgeni: If you use `ForemanTaxonomicEntityAnsibleModule` as a base, you won't need to define orgs/locs as params yourself

evgeni: Also, I don't think you need NestedParameters, disco rules have no params.

evgeni (commented):
I don't see anything obviously wrong why it would want to edit existing stuff. The comments I left are more cosmetic.

Will try to look in detail tomorrow in the office

plugins/modules/discovery_rule.py
Thulium-Drake: I stole this from ```location.py``` and edited it to make it work :-)

Thanks for the feedback! I've updated the module!

evgeni:
So the changed state happens, because while the parameter is called `max_count` in the API, the `show` action renders it as `hosts_limit`:
https://github.com/theforeman/foreman_discovery/blob/develop/app/views/api/v2/discovery_rules/main.json.rabl#L6

Edit, this should fix it:
```patch
diff --git plugins/modules/discovery_rule.py plugins/modules/discovery_rule.py
index 01ffe0ef..9fd42a2d 100644
--- plugins/modules/discovery_rule.py
+++ plugins/modules/discovery_rule.py
@@ -141,6 +141,13 @@ def main():
     )
 
     with module.api_connection():
+        entity = module.lookup_entity('entity')
+
+        # workround the fact that the API expects `max_count` when modifying the entity
+        # but uses `hosts_limit` when showing one
+        if entity and 'hosts_limit' in entity:
+            entity['max_count'] = entity.pop('hosts_limit')
+
         module.run()
 
```

Thulium-Drake:
@evgeni Like a charm!

evgeni:
Nice. Do you need help with the tests?

Thulium-Drake:
It's been a while since I made tests, for now I only have a question.

Can I 'just include' the test tasks for the hostgroup? As that's a dependency for this test :-)

evgeni:
> Can I 'just include' the test tasks for the hostgroup? As that's a dependency for this test :-)

Of course!

Have a look here:
https://github.com/theforeman/foreman-ansible-modules/blob/develop/tests/test_playbooks/architecture.yml

On the `localhost` target, it includes the `operatingsystem` tasks, so that OSes are created (and destroyed) and then later on the `tests` target only the `architecture` stuff is running. That way OS stuff gets executed but not recorded (record only happens for the `tests` host).

Thulium-Drake:
Awesome, thanks! I'll report back when I have some tests :-) Installing a fresh box now (sadly can't use vagrant on Proxmox and my other spare test laptop isn't around right now)

Thulium-Drake:
Something like this? :-)

tests/fixtures/apidoc/discovery_rule.json
```diff
@@ -0,0 +1 @@
+foreman.json
```
evgeni: you'd need to point this at `luna.json`, otherwise the tests won't know what discovery is.

tests/test_playbooks/discovery_rule.yml
```diff
@@ -0,0 +1,128 @@
...
+          - "{{ hostgroup.os.name }} {{ hostgroup.os.major }}.{{ hostgroup.os.minor }}"
+        installation_medium_locations: "{{ hostgroup.locations }}"
+        installation_medium_organizations: "{{ hostgroup.organizations }}"
+        installation_medium_state: present
+    - include_tasks: tasks/hostgroup.yml
+      vars:
+        hostgroup_name: "Discovery hostgroup"
+        hostgroup_description: "Discovery hostgroup"
+        hostgroup_architecture: "{{ hostgroup.arch }}"
+        hostgroup_operatingsystem: "{{ hostgroup.os.name }} {{ hostgroup.os.major }}.{{ hostgroup.os.minor }}"
```
evgeni: I think  you can safely omit these.
```suggestion
```

tests/test_playbooks/discovery_rule.yml
```diff
@@ -0,0 +1,128 @@
...
+        hostgroup_domain: "{{ hostgroup.domains | first }}"
+        hostgroup_subnet: "{{ hostgroup.subnet }}"
+        hostgroup_installation_medium: "{{ hostgroup.os.name }} Mirror"
+        hostgroup_ptable: "{{ hostgroup.ptable }}"
+        hostgroup_pxe_loader: "Grub2 UEFI"
+        hostgroup_parameters:
+          - name: subnet_param1
+            value: value1
+          - name: subnet_param2
+            value: value2
```
evgeni: I think  you can safely omit these.
```suggestion
```

tests/test_playbooks/discovery_rule.yml
```diff
@@ -0,0 +1,128 @@
...
+          - "{{ hostgroup.ptable }}"
+        operatingsystem_state: present
+    - include_tasks: tasks/installation_medium.yml
+      vars:
+        installation_medium_name: "{{ hostgroup.os.name }} Mirror"
+        installation_medium_operatingsystems:
+          - "{{ hostgroup.os.name }} {{ hostgroup.os.major }}.{{ hostgroup.os.minor }}"
+        installation_medium_locations: "{{ hostgroup.locations }}"
+        installation_medium_organizations: "{{ hostgroup.organizations }}"
+        installation_medium_state: present
```
evgeni: I think  you can safely omit these.
```suggestion
```

tests/test_playbooks/discovery_rule.yml
```diff
@@ -0,0 +1,128 @@
...
+        discovery_rule_organizations: "{{ hostgroup.organizations }}"
+        discovery_rule_locations: "{{ hostgroup.locations }}"
+        discovery_rule_state: 'present'
+        expected_change: false
+
+    - include_tasks: tasks/discovery_rule.yml
+      vars:
+        discovery_rule_name: 'New-discovery-rule'
+        discovery_rule_state: 'absent'
+        expected_change: true
```
evgeni, line 76: please test for idempotency of absent, and cleanup after yourself ;-)

```suggestion
        expected_change: true

    - include_tasks: tasks/discovery_rule.yml
      vars:
        discovery_rule_name: 'New-discovery-rule'
        discovery_rule_state: 'absent'
        expected_change: false

- hosts: localhost
  collections:
    - theforeman.foreman
  gather_facts: false
  vars_files:
    - vars/server.yml
    - vars/hostgroup.yml
  tasks:
    - include_tasks: tasks/location.yml
      vars:
        location_name: "{{ item }}"
        location_state: absent
      with_items: "{{ hostgroup.locations | reverse | list }}"
    - include_tasks: tasks/organization.yml
      vars:
        organization_name: "{{ item }}"
        organization_state: absent
      with_items: "{{ hostgroup.organizations }}"
```

evgeni:
> Something like this? :-)

Yes. See inline comments for uh, comments, but generally speaking yes :)

plugins/modules/discovery_rule.py
```diff
@@ -0,0 +1,155 @@
...
+# along with this program.  If not, see <http://www.gnu.org/licenses/>.
+
+from __future__ import absolute_import, division, print_function
+__metaclass__ = type
+
+
+DOCUMENTATION = '''
+---
+module: discovery_rule
+version_added: x.x.x
```
evgeni:
```suggestion
version_added: 3.5.0
```

3.5.0 will be the next version :)

plugins/modules/discovery_rule.py
```diff
@@ -0,0 +1,155 @@
...
+    elements: str
+  organizations:
+    description:
+      - List of organizations the discovery rule should be assigned to
+    type: list
+    elements: str
+extends_documentation_fragment:
+  - theforeman.foreman.foreman
+  - theforeman.foreman.foreman.entity_state
+  - theforeman.foreman.foreman.nested_parameters
```
evgeni:
```suggestion
  - theforeman.foreman.foreman.taxonomy
```

We don't use nested params anymore, but we do have taxonomy! :)

plugins/modules/discovery_rule.py
```diff
@@ -0,0 +1,155 @@
...
+  locations:
+    description:
+      - List of locations the discovery rule should be assigned to
+    type: list
+    elements: str
+  organizations:
+    description:
+      - List of organizations the discovery rule should be assigned to
+    type: list
+    elements: str
```
evgeni:
```suggestion
```

Comes from the `theforeman.foreman.foreman.taxonomy` fragment below.

plugins/modules/discovery_rule.py
Thulium-Drake: Cool! Fixed :)

plugins/modules/discovery_rule.py
```diff
@@ -0,0 +1,155 @@
+#!/usr/bin/python
+# -*- coding: utf-8 -*-
+# (c) 2017, Matthias M Dellweg <dellweg@atix.de> (ATIX AG)
```
evgeni: I think we can safely assume Matthias has not really any more copyright on the result ;)

plugins/modules/discovery_rule.py
Thulium-Drake: Check, done!

plugins/modules/discovery_rule.py
Thulium-Drake: Done!

plugins/modules/discovery_rule.py
Thulium-Drake: Hehe, I'm not always sure how to do those :-) I did steal his ```location.py``` to start working in this :grimacing:

tests/test_playbooks/discovery_rule.yml
Thulium-Drake: Done!

tests/test_playbooks/discovery_rule.yml
Thulium-Drake: Jup!

tests/test_playbooks/discovery_rule.yml
Thulium-Drake: Jup!

tests/test_playbooks/discovery_rule.yml
Thulium-Drake: Jup!

plugins/modules/discovery_rule.py
```diff
@@ -0,0 +1,143 @@
...
+            search=dict(),
+            hostgroup=dict(type='entity'),
+            hostname=dict(),
+            max_count=dict(type='int'),
+            priority=dict(type='int'),
+            enabled=dict(type='bool'),
+        ),
+        required_if=[
+            ['state', 'present', ['hostgroup', 'search']],
+        ],
```
evgeni, line 128: you have both `hostgroup` and `search` defined as `required: True` in the doc string above:
https://github.com/theforeman/foreman-ansible-modules/blob/b90c7de8ecb46fef5f3236d1771c08d4b1a13886/plugins/modules/discovery_rule.py#L37-L46

But not in the spec here, instead you use `required_if`.

I tend to agree that the `required_if` way is slightly more flexible, but then you need to drop the `required: True` from the docs and use a line like `Required if I(state=present)` or similar:
https://github.com/theforeman/foreman-ansible-modules/blob/767eba4de512a04de45cc8bb3167d44e1f32c150/plugins/modules/activation_key.py#L122

Thulium-Drake: Yeah, was just looking at that :-) thanks for pointing me in the right direction!

evgeni:
Ci failures unrelated, seems Ansible devel changed something about callbacks: https://github.com/ansible/ansible/commit/18992b79479848a4bc06ca366a0165eabd48b68e

Thulium-Drake:
Phew! But I guess that's it then :)

Thulium-Drake:
@evgeni Is there still stuff I/we need to do for this? :-)

evgeni:
> @evgeni Is there still stuff I/we need to do for this? :-)

Not at this time, no. Just too many other things on my table right now, sorry.

evgeni:
CI should become green once we get https://github.com/theforeman/foreman-ansible-modules/pull/1437 in :)

evgeni:
@Thulium-Drake mind rebasing on latest `develop`?

Thulium-Drake:
This should do it :)

plugins/modules/discovery_rule.py
```diff
@@ -0,0 +1,143 @@
...
```
evgeni, line 128:
```suggestion
        ],
        required_plugins=[('discovery', ['*'])],
```

Let's make it error nicely if the plugin is not installed ;)

Thulium-Drake: Done! :-)

evgeni (approved):

Thulium-Drake:
:rocket:

PR #1433, opened 2022-06-21
Add content export library to FAM

jeremylenz:
Adds pulpcore content exports to Foreman Ansible Modules.

Covers exports of 
* library
* ~~content view version~~  https://github.com/theforeman/foreman-ansible-modules/pull/1453

* ~~repository~~  https://github.com/theforeman/foreman-ansible-modules/pull/1454

Examples cover creation of `metadata.json` which is required to import the exported content.

Does not cover: 
* imports
* syncable exports

Todo:
- [x] export history info module
- [x] fix tests
- [x] show metadata.json creation in examples

To test:
```sh
sudo make dist-test

ansible-playbook tests/test_playbooks/content_export_info.yml
```


plugins/modules/content_export.py
```diff
@@ -0,0 +1,67 @@
...
+    with module.api_connection():
+        scope = module.scope_for('organization')
+        # payload = {
+        #     'destination_server': module.foreman_params['destination_server'],
+        #     'chunk_size_gb': module.foreman_params['chunk_size_gb'],
+        #     'fail_on_missing_content': module.foreman_params['fail_on_missing_content'],
+        # }
+        payload = {key: module.params[key] for key in module.params.keys() if key in module.foreman_params}
+        payload.update(scope)
+        print(payload)
```
evgeni: this print freaks out ansible and makes it mark the execution as failed -.-

evgeni:
So, I played a bit a round with this, and this patch makes it a bit tidier:

```diff
diff --git plugins/modules/content_export.py plugins/modules/content_export.py
index a5d1b615..3811da38 100644
--- plugins/modules/content_export.py
+++ plugins/modules/content_export.py
@@ -32,7 +32,7 @@ author:
 '''
 
 
-from ansible_collections.theforeman.foreman.plugins.module_utils.foreman_helper import KatelloAnsibleModule
+from ansible_collections.theforeman.foreman.plugins.module_utils.foreman_helper import KatelloAnsibleModule, _flatten_entity
 
 
 class KatelloContentExportModule(KatelloAnsibleModule):
@@ -49,15 +49,8 @@ def main():
     )
 
     with module.api_connection():
-        scope = module.scope_for('organization')
-        # payload = {
-        #     'destination_server': module.foreman_params['destination_server'],
-        #     'chunk_size_gb': module.foreman_params['chunk_size_gb'],
-        #     'fail_on_missing_content': module.foreman_params['fail_on_missing_content'],
-        # }
-        payload = {key: module.params[key] for key in module.params.keys() if key in module.foreman_params}
-        payload.update(scope)
-        print(payload)
+        module.auto_lookup_entities()
+        payload = _flatten_entity(module.foreman_params, module.foreman_spec)
         task = module.resource_action('content_exports', 'library', payload)
 
         module.exit_json(task=task)
```

plugins/modules/content_export.py
```diff
@@ -0,0 +1,67 @@
...
+DOCUMENTATION = '''
+---
+module: content_export
+version_added: 3.5.0
+short_description: Manage pulp3 content exports
+description:
+    - Export content view versions, repositories, or library content to a directory.
+author:
+    - "Jeremy Lenz (@jeremylenz)"
+
```
evgeni:
```suggestion
options:
  destination_server:
    description:
      - Destination Server name
    required: false
    type: str
  chunk_size_gb:
    description:
      - Split the exported content into archives no greater than the specified size in gigabytes.
    required: false
    type: int
  fail_on_missing_content:
    description:
      - Fails if any of the repositories belonging to this organization are unexportable.
    required: false
    type: bool
extends_documentation_fragment:
  - theforeman.foreman.foreman
  - theforeman.foreman.foreman.organization
```

plugins/modules/content_export.py
```diff
@@ -0,0 +1,67 @@
...
+class KatelloContentExportModule(KatelloAnsibleModule):
+    pass
+
+
+def main():
+    module = KatelloContentExportModule(
+        foreman_spec=dict(
+            destination_server=dict(required=False, type='str'),
+            chunk_size_gb=dict(required=False, type='int'),
+            fail_on_missing_content=dict(required=False, type='bool', default=False),
```
evgeni:
```suggestion
            fail_on_missing_content=dict(required=False, type='bool'),
```

I wouldn't set a default here, the API should (and does already).

jeremylenz:
@evgeni here's the link to the docs: https://docs.theforeman.org/nightly/Content_Management_Guide/index-katello.html#Synchronizing_Content_Between_Servers_content-management

jeremylenz:
As far as how to divide it up, what I was thinking is

3 modules in this PR, each covering both full and incremental:
* content_export_library
* content_export_repository
* content_export_version

and a corresponding follow-up card for Import. @parthaa does that sound good?

plugins/modules/content_export_library.py
```diff
@@ -0,0 +1,99 @@
...
+        incremental = module.params['incremental']
+        endpoint = 'content_export_incrementals' if incremental else 'content_exports'
+
+        if module.params.get('from_history_id') and incremental is not True:
+            module.fail_json(msg='from_history_id is only valid for incremental exports')
+
+        payload = _flatten_entity(module.foreman_params, module.foreman_spec)
+        task = module.resource_action(endpoint, 'library', payload)
+
+        module.exit_json(task=task)
```
parthaa, line 141: @jeremylenz might need to add stuff for generate-metadata endpoint.

parthaa:
> and a corresponding follow-up card for Import. @parthaa does that sound good?
Sounds good to me

wbclark (commented):
Examples are really helpful for users to understand how to work with these modules. When they are first getting started it is often easier to copy and paste some example code and then edit it, rather than writing it all from scratch. Could you add maybe 2 examples beneath the `DOCUMENTATION` here, showing the full and incremental export?

A basic _example_ (har) of what this looks like, can be found in the `lifecycle_environment` module: https://github.com/theforeman/foreman-ansible-modules/blob/767eba4de512a04de45cc8bb3167d44e1f32c150/plugins/modules/lifecycle_environment.py#L57-L69

plugins/modules/content_export_library.py
```diff
@@ -0,0 +1,99 @@
...
+  incremental:
+    description:
+      - Export only the content that has changed since the last export.
+    required: false
+    type: bool
+  from_history_id:
+    description:
+      - Export history identifier used for incremental export. If not provided the most recent export history will be used.
+    required: false
+    type: str
```
wbclark, line 56: Does the successful export return this ID? It might be valuable to show in the examples how to register that ID and save it for a future incremental export task. Since users would not want to manually enter this data each time.

wbclark:
> As far as how to divide it up, what I was thinking is
> 
> 3 modules in this PR, each covering both full and incremental:
> 
>     * content_export_library
> 
>     * content_export_repository
> 
>     * content_export_version
> 
> 
> and a corresponding follow-up card for Import. @parthaa does that sound good?

I _think_ it should be doable to have all this within a single module. The parameters could include a Boolean for Library. If that is not set to True then it would look either for a Repository name, or a CV name and version ID, or a CV name and lifecycle environment.

For an example of what conditionally required parameters look like, see: https://github.com/theforeman/foreman-ansible-modules/blob/d77a0f1162142022e69110ad4cdd692c8d2c15c5/plugins/modules/bookmark.py#L141-L143

You can also make parameters mutually exclusive, as in: https://github.com/theforeman/foreman-ansible-modules/blob/a1330d3a6d1867eb7b7112d0d030ad0439babd6b/plugins/module_utils/foreman_helper.py#L1345

Per https://docs.ansible.com/ansible/latest/dev_guide/developing_program_flow_modules.html there is also `required_one_of`, `required_together`, etc. So there are powerful tools to specify what combinations of data the module will accept, and then you can write the module assuming that a "correct" input was received and specifying the endpoints appropriately.

plugins/modules/content_export_library.py
evgeni: The module currently returns the whole foreman-tasks structure, which includes:

```
            "output": {
                "export_history_id": 6
            },
```

But that feels a bit cumbersome. We could make the module return this one *specifically*.

However, people won't find that useful -- they don't need the history id the moment they did the export, they need it the moment they want to do an incremental one. And for that I think we should have a `*_info` module, that can look up the "last" (or any) export id for any given export?

plugins/modules/content_export_library.py
```diff
@@ -0,0 +1,99 @@
...
+    )
+
+    with module.api_connection():
+        module.auto_lookup_entities()
+
+        incremental = module.params['incremental']
+        endpoint = 'content_export_incrementals' if incremental else 'content_exports'
+
+        if module.params.get('from_history_id') and incremental is not True:
+            module.fail_json(msg='from_history_id is only valid for incremental exports')
```
evgeni, line 136: instead, you could add the following in the module creation above:
```python
        foreman_spec=dict(
…
        ),
        argument_spec=dict(
…
        ),
        required_together=[
              ['from_history_id', 'incremental'],
        ],
    )
```

On the other hand, should we just assume the user wanted `incremental=True` if they passed a history id? Feels more natural to me?

evgeni:
> I _think_ it should be doable to have all this within a single module.

We *could*. But from talking to Jeremy and reading the docs, the three workflows (library, version and repo exports) seem to be solving completely different user scenarios and the only thing they have in common is the fact that the end result is a "content export". Given the workflows are different (and so are the `hammer` commands: `hammer content-export complete version` - not `hammer content-export complete --type version`), it feels more natural (to me) to have different modules, even if they will end up sharing a lot of code together.

plugins/modules/content_export_library.py
wbclark: My intuition about the design is that it would be nice for the module not to make these assumptions, and rather expect the user to be explicit in order to limit any potential disparities between what the module will do vs. what the user understood or intended it to do. And then to have a couple very small roles which are more workflow driven and are allowed to make nice assumptions about what the user intended from minimal inputs.

Let me give an example of why this might be beneficial. Let's say I'm modifying the data I provide to an existing role or playbook, with the intent to change an incremental export to a full export, and I currently have some data like:

```yaml
  incremental: true
  some_other: "{{ stuff }}"
  some_more: "{{ other_stuff }}"
  from_history_id: "{{ saved_export_history_id }}"
```
And then I take a quick look and make the seemingly reasonable assumption that I can flip `true` to `false` and I will get a full backup.

But Ansible assumes, without telling me so, that actually I really meant `true`, and I'm disappointed _much later on_ when I discover that I'm relying on what I thought was a full export that was in fact only incremental.

OTOH if the module catches this and tells me that I'm trying to do something that doesn't make any sense, I'll realize my error right away before any damage is done (like deleting my prior content backups because I think I have generated a new full backup).

Another way of thinking about this is that I am looking at the modules as being sort of an API or library for role developers. So this viewpoint is aligned with having a single module for everything related to content exports, eliminating lots of duplicated code there, and then having a couple very small and easy to use roles corresponding to the different user scenarios.

evgeni: > And then I take a quick look and make the seemingly reasonable assumption that I can flip `true` to `false` and I will get a full backup.
> 
> But Ansible assumes, without telling me so, that actually I really meant `true`, and I'm disappointed _much later on_ when I discover that I'm relying on what I thought was a full export that was in fact only incremental.

Well, yeah, I did mean to not to have an (exposed) `incremental` parameter at all, so there would be no assumption that flipping it to `false` would create a full export.

> Another way of thinking about this is that I am looking at the modules as being sort of an API or library for role developers.

I strongly disagree, if a module requires a wrapping role to function, the module is bad. Modules are not APIs, they are user facing interfaces.

jeremylenz:
> the three workflows (library, version and repo exports) seem to be solving completely different user scenarios and the only thing they have in common is the fact that the end result is a "content export". Given the workflows are different (and so are the `hammer` commands: `hammer content-export complete version` - not `hammer content-export complete --type version`), it feels more natural (to me) to have different modules, even if they will end up sharing a lot of code together.

> if the module catches this and tells me that I'm trying to do something that doesn't make any sense, I'll realize my error right away before any damage is done 

Thanks for the thoughtful discussion everyone!

As a FAM newb I really like the idea of everything being explicit. So you each win one -- I like @evgeni's reasoning for having different modules for everything, and I like @wbclark's idea of passing explicit values instead of relying on hidden logic that may not be immediately apparent.

I'll move forward with this unless y'all feel super strongly and want to keep debating here.

plugins/modules/content_export_library.py
wbclark: > Well, yeah, I did mean to not to have an (exposed) `incremental` parameter at all, so there would be no assumption that flipping it to `false` would create a full export.

Thank you, that makes a lot more sense than how I interpreted it.

wbclark:
> Thanks for the thoughtful discussion everyone!
> 
> As a FAM newb I really like the idea of everything being explicit. So you each win one -- I like @evgeni's reasoning for having different modules for everything, and I like @wbclark's idea of passing explicit values instead of relying on hidden logic that may not be immediately apparent.
> 
> I'll move forward with this unless y'all feel super strongly and want to keep debating here.

Actually, I will cede my point. I have been totally convinced by @evgeni 's perspective.

It makes the most sense not to have any `incremental` parameter at all, and just infer whether this is a complete or incremental export based on the presence of `from_history_id`.

jeremylenz:
```
from_history_id:
    description:
      - Export history identifier used for incremental export. If not provided the most recent export history will be used.
```

Just realized that setting `incremental` based on the presence of `from_history_id` will remove the convenience described here - the user will have to look up a history ID for every incremental export.  I think this tips the balance back in favor of explicit.. thoughts? @wbclark @parthaa @evgeni

evgeni:
Oooh, if it can look up the recent automatically, then perfect, leave it!

plugins/modules/content_export_library.py
parthaa: Sounds like generate-metadata is not an end point. I 'd suggest trying to do a get for export history id or task and pull it from there.

jeremylenz:
>  It might be valuable to show in the examples how to register that ID and save it for a future incremental export task

@wbclark  I would love to, but I need help with how to do it :confused: 
(right now my IDs and content view names are hard-coded.)

I've tried using `debugger: always` but I didn't get far.

jeremylenz:
okay, Added CV version and repository exports.

still need help with
* what's going on with the tests
* ~~how not to hard-code values in the test playbook~~

evgeni:
Shall we sit down on Monday or so together and talk tests?

jeremylenz:
> Shall we sit down on Monday or so together and talk tests?

sounds good to me!

plugins/modules/content_export_library.py
jeremylenz: :thinking:

jeremylenz:
> Could you add maybe 2 examples beneath the `DOCUMENTATION` here, showing the full and incremental export?

@wbclark Done! :+1:

plugins/modules/content_export_library.py
jeremylenz: @evgeni Added the `info` module.

This will list all your content exports, but unfortunately the Katello API doesn't allow searching, ordering or sorting by
* created_at
* updated_at
* content_view_name (only id)
* content_view_version name (only id)
* whether the export was of library, repository, or version

@parthaa I think we should add these to Katello at some point.

plugins/modules/content_export_library.py
jeremylenz: Still not sure what to do about this. Generating metadata is a special task performed by Hammer, due to the fact that hammer is run as root and it needs filesystem permissions in `/var/lib/pulp` (if I understand correctly). It's not something the Katello API does. In light of that, what should we do here?

jeremylenz:
Updated tests. Not sure what the failure means..

evgeni:
Was the repository export added recently to Katello? The error says that there is no "repository" action on the exports controller.

As the tests run against a recorded api, it uses a static apidoc.json, which might be missing that (it's 4.3 right now, we're sometimes lazy updating)?

jeremylenz:
> Was the repository export added recently to Katello?

Yes, it's pretty recent..

jeremylenz:
okay, if February is recent? https://github.com/Katello/katello/pull/9925

tests/fixtures/apidoc/content_export_info.json
```diff
@@ -0,0 +1 @@
+katello.json
```
jeremylenz, line 1: I guess this is what Github looks like when you symlink a file.. :shrug:

evgeni: Yeah, the red is annoying but the link is correct.

evgeni:
https://github.com/Katello/katello/commit/4c1e72119659be50dd60682b1ef10e13c3ab682a seems to be in 4.5+, so yeah, "recent".

Feel free to replace `katello.json` with a fresh 4.5 one (but vanilla Katello with ostree enabled please, no other plugins)

jeremylenz:
https://projects.theforeman.org/issues/34374 landed in Katello 4.5, so I guess that's the issue.

jeremylenz:
@wbclark any pointers on how to fix the "checkmode" tests?

jeremylenz:
:green_apple: :green_apple: :green_apple: :green_apple: :green_apple: :green_apple: :apple: :green_apple: :green_apple: :green_apple: :green_apple: :green_apple: :apple: :green_apple: :green_apple: :green_apple: :green_apple: 

I believe the remaining test failures are unrelated!

plugins/modules/content_export_repository.py
```diff
@@ -0,0 +1,130 @@
...
+description:
+    - Export repository content to a directory.
+author:
+    - "Jeremy Lenz (@jeremylenz)"
+options:
+  id:
+    description:
+      - Repository identifier.
+    required: true
+    type: int
```
evgeni: I don't think the user should have to pass the repo-id, but more the product and repo name, and then the module should figure this out on its own.

plugins/modules/content_export_version.py
```diff
@@ -0,0 +1,156 @@
...
+description:
+    - Export a content view version to a directory.
+author:
+    - "Jeremy Lenz (@jeremylenz)"
+options:
+  id:
+    description:
+      - Content view version identifier.
+    required: true
+    type: int
```
evgeni: I don't think the user should have to pass the cv-id, but more the cv name, and then the module should figure this out on its own.

plugins/modules/content_export_info.py
```diff
@@ -0,0 +1,165 @@
...
+  content_view_version_id:
+    description:
+      - Content view version identifier.
+    required: false
+    type: int
+  content_view_id:
+    description:
+      - Content view identifier.
+    required: false
+    type: int
```
evgeni: same, we should not need to address CVs by id

plugins/modules/content_export_info.py
```diff
@@ -0,0 +1,165 @@
...
+- name: "List all full exports in the organization"
+  content_export_info:
+    organization: "Default Organization"
+    type: complete
+    username: "admin"
+    password: "changeme"
+    server_url: "https://foreman.example.com"
+- name: "Get a specific export history and register the result for the next task"
+  vars:
+    organization_name: "Export Org"
```
evgeni, line 76:
```suggestion
```

this var is unused

plugins/modules/content_export_info.py
```diff
@@ -0,0 +1,165 @@
...
+
+    with module.api_connection():
+        module.auto_lookup_entities()
+
+        endpoint = 'content_exports'
+
+        payload = _flatten_entity(module.foreman_params, module.foreman_spec)
+        task = module.resource_action(endpoint, 'index', payload)
+
+        module.exit_json(task=task)
```
evgeni: We have a shortcut for `resource_action('index')`: `list_resource`, it also takes care of stripping the `results` "outer-array".

```suggestion
        resource = 'content_exports'

        payload = _flatten_entity(module.foreman_params, module.foreman_spec)
        info = {resource: module.list_resource(resource, params=payload)}

        module.exit_json(**info)
```

We could completely avoid this code, if we would base the module off `ForemanInfoAnsibleModule` and mark `name` as `invisible`. Yes!

```python
from ansible_collections.theforeman.foreman.plugins.module_utils.foreman_helper import KatelloInfoAnsibleModule


class KatelloContentExportInfo(KatelloInfoAnsibleModule):
    pass


def main():
    module = KatelloContentExportInfo(
        foreman_spec=dict(
            id=dict(required=False, type='int'),
            content_view_version=dict(type='entity', scope=['content_view'], required=False),
            content_view=dict(type='entity', scope=['organization'], required=False),
            type=dict(required=False, type='str', choices=['complete', 'incremental']),
            search=dict(required=False, type='str'),
            name=dict(invisible=True),
        ),
    )

    with module.api_connection():
        module.run()


if __name__ == '__main__':
    main()
```

that works :)

evgeni:
Here is the full diff I applied to the info module:

```diff
diff --git plugins/modules/content_export_info.py plugins/modules/content_export_info.py
index d9097595..6a17826a 100644
--- plugins/modules/content_export_info.py
+++ plugins/modules/content_export_info.py
@@ -34,26 +34,21 @@ options:
       - Export history identifier.
     required: false
     type: int
-  content_view_version_id:
+  content_view_version:
     description:
-      - Content view version identifier.
+      - Content view version.
     required: false
-    type: int
-  content_view_id:
+    type: str
+  content_view:
     description:
-      - Content view identifier.
+      - Content view name.
     required: false
-    type: int
+    type: str
   destination_server:
     description:
       - Destination server name
     required: false
     type: str
-  organization_id:
-    description:
-      - Organization identifier.
-    required: false
-    type: int
   type:
     description:
       - Specify complete or incremental exports.
@@ -62,14 +57,10 @@ options:
     choices:
     - complete
     - incremental
-  search:
-    description:
-      - Search string.
-    required: false
-    type: str
 extends_documentation_fragment:
   - theforeman.foreman.foreman
-  - theforeman.foreman.foreman.organization
+  - theforeman.foreman.foreman.katelloinfomodule
+  - theforeman.foreman.foreman.infomodulewithoutname
 '''
 
 EXAMPLES = '''
@@ -81,8 +72,6 @@ EXAMPLES = '''
     password: "changeme"
     server_url: "https://foreman.example.com"
 - name: "Get a specific export history and register the result for the next task"
-  vars:
-    organization_name: "Export Org"
   content_export_info:
     id: 29
     username: "admin"
@@ -92,13 +81,14 @@ EXAMPLES = '''
   register: result
 - name: "Write metadata.json to disk using data from the previous task"
   vars:
-    metadata: "{{ result['task']['results'][0]['metadata'] }}"
+    metadata: "{{ result['content_exports'][0]['metadata'] }}"
   ansible.builtin.copy:
     content: "{{ metadata }}"
     dest: ./metadata.json
 - name: "List all exports of a specific content view version"
   content_export_info:
-    content_view_version_id: 379
+    content_view: RHEL8
+    content_view_version: '1.0'
     username: "admin"
     password: "changeme"
     server_url: "https://foreman.example.com"
@@ -112,7 +102,7 @@ EXAMPLES = '''
     organization: "Default Organization"
 - name: "List incremental exports of a specific content view version marked for a specific destination server"
   content_export_info:
-    content_view_id: 1
+    content_view: RHEL8
     destination_server: "airgapped.example.com"
     type: incremental
     username: "admin"
@@ -121,7 +111,7 @@ EXAMPLES = '''
     organization: "Default Organization"
 - name: "List all exports of a specific content view marked for a specific destination server"
   content_export_info:
-    content_view_id: 1
+    content_view: RHEL8
     destination_server: "airgapped.example.com"
     username: "admin"
     password: "changeme"
@@ -130,35 +120,28 @@ EXAMPLES = '''
 
 '''
 
-from ansible_collections.theforeman.foreman.plugins.module_utils.foreman_helper import KatelloAnsibleModule, _flatten_entity
+from ansible_collections.theforeman.foreman.plugins.module_utils.foreman_helper import KatelloInfoAnsibleModule
 
 
-class KatelloContentExportInfoModule(KatelloAnsibleModule):
+class KatelloContentExportInfo(KatelloInfoAnsibleModule):
     pass
 
 
 def main():
-    module = KatelloContentExportInfoModule(
+    module = KatelloContentExportInfo(
         foreman_spec=dict(
             id=dict(required=False, type='int'),
-            content_view_version_id=dict(required=False, type='int'),
-            content_view_id=dict(required=False, type='int'),
+            content_view_version=dict(type='entity', scope=['content_view'], required=False),
+            content_view=dict(type='entity', scope=['organization'], required=False),
             destination_server=dict(required=False, type='str'),
-            organization_id=dict(required=False, type='int'),
             type=dict(required=False, type='str', choices=['complete', 'incremental']),
             search=dict(required=False, type='str'),
+            name=dict(invisible=True),
         ),
     )
 
     with module.api_connection():
-        module.auto_lookup_entities()
-
-        endpoint = 'content_exports'
-
-        payload = _flatten_entity(module.foreman_params, module.foreman_spec)
-        task = module.resource_action(endpoint, 'index', payload)
-
-        module.exit_json(task=task)
+        module.run()
 
 
 if __name__ == '__main__':
```

plugins/modules/content_export_library.py
```diff
@@ -0,0 +1,145 @@
...
+
+from __future__ import absolute_import, division, print_function
+__metaclass__ = type
+
+
+DOCUMENTATION = '''
+---
+module: content_export_library
+version_added: 3.5.0
+short_description: Manage pulp3 content exports
```
evgeni: Is it really important (to the user) that those are Pulp3 exports?
IMHO the user shouldn't have to know what Pulp is when running Katello/Satellite.

evgeni:
I am happy with the `content_export_library` module, but I think the others need a bit more work to eliminate the `cv_id` etc references.

Shall we split this PR into four, one per module? Then we can get the library one in instantly and work on the others?

jeremylenz:
@evgeni I need to keep `info` and `library` together because the tests use `info`..

jeremylenz:
See  https://github.com/theforeman/foreman-ansible-modules/pull/1453 and https://github.com/theforeman/foreman-ansible-modules/pull/1454

I'll rebase them once this is in.

plugins/modules/content_export_info.py
```diff
@@ -130,35 +122,28 @@
...
             type=dict(required=False, type='str', choices=['complete', 'incremental']),
             search=dict(required=False, type='str'),
+            name=dict(invisible=True),
         ),
     )
 
     with module.api_connection():
-        module.auto_lookup_entities()
-
-        endpoint = 'content_exports'
```
jeremylenz: I still don't understand how it knows which endpoint to call if we delete this. :open_mouth:

jeremylenz:
also I don't know how to make lint happy

plugins/modules/content_export_info.py
evgeni: The class is called `KatelloContentExportInfo`, it strips "Katello" and "Info" and then translates the rest from "CamelCaseSingular" to "snake_case_plural".

https://github.com/theforeman/foreman-ansible-modules/blob/e27433353a67f65e5c0f845fd1d6e60eade77651/plugins/module_utils/foreman_helper.py#L1349-L1369

meta/runtime.yml
```diff
@@ -10,6 +10,10 @@ action_groups:
   - compute_resource
   - config_group
   - content_credential
+  - content_export_info
+  - content_export_library
+  - content_export_repository
+  - content_export_version
```
evgeni:
```suggestion
```

These modules are not part of this PR, so they should not be in the metadata either ;)

evgeni:
the remaining sanity errors will be fixed in https://github.com/theforeman/foreman-ansible-modules/pull/1455 and are unrelated to this PR!

evgeni (approved):

PR #1444, opened 2022-07-24
Add option to the inventory plugin to retrieve a list of ansible roles enabled within foreman

stuwilkins:
This addition to the inventory plugin pulls the defined ansible roles from foreman and sets the variable `foreman_ansible_roles`

The option is set via the `wants_ansible_roles` setting.


mdellweg:
This repository contains so much more than just the inventory plugin. Please drop the name of the component somewhere in the title of this issue and the commit messages.

stuwilkins:
@mdellweg Thanks for the prompt reply, and thanks for the work on this collection, we use much more than just the plugin but I added this to help with using Tower and Foreman together. 

I changed the commit messages and also the PR. Please let me know if you need anything else.

mdellweg:
That is much better now. And i hope people more familiar with the inventory than me will be triggered by this. ;)

PR #1446, opened 2022-07-29
Fixes missing memory issue when creating snapshot

Manisha15:
When creating VMWare snapshot using ansible-playbook, it throws error if memory value is not provided. 

This PR fixes the issue by adding `include_ram` parameter in the snapshot module.

### DIFF
--- a/plugins/modules/snapshot.py
+++ b/plugins/modules/snapshot.py
@@ -47,6 +47,11 @@ options:
       - Name of related Host
     required: true
     type: str
+  include_ram:
+    description:
+      - Option to add RAM (only available for VMWare compute-resource)
+    required: false
+    type: bool
   state:
     description:
       - State of Snapshot
@@ -125,6 +130,7 @@ def main():
             host=dict(type='entity', required=True, ensure=False),
             name=dict(required=True),
             description=dict(),
+            include_ram=dict(type='bool'),
         ),
         required_plugins=[('snapshot_management', ['*'])],
         entity_opts={'scope': ['host']},


evgeni:
Does this option *have* to be set when using VMware? That's how I read your commit message, but not the option description.

Manisha15:
> Does this option _have_ to be set when using VMware? That's how I read your commit message, but not the option description.

Yes, it needs to be set for VMWare otherwise it will throw `missing required memory parameter`.

evgeni:
> > Does this option _have_ to be set when using VMware? That's how I read your commit message, but not the option description.
> 
> Yes, it needs to be set for VMWare otherwise it will throw `missing required memory parameter`.

That sounds wrong to me then? The API (https://github.com/ATIX-AG/foreman_snapshot_management/blob/master/app/controllers/api/v2/snapshots_controller.rb#L48) says it's optional (and doesn't even mention VMware only).

If that is always required, I think `foreman_snapshot_management` should just always set it (for VMware).
If it's optional, `foreman_snapshot_management` should not error out :)

cc @mdellweg who contributed to that feature in the past I think ;-)

Manisha15:
> > > Does this option _have_ to be set when using VMware? That's how I read your commit message, but not the option description.
> > 
> > 
> > Yes, it needs to be set for VMWare otherwise it will throw `missing required memory parameter`.
> 
> That sounds wrong to me then? The API (https://github.com/ATIX-AG/foreman_snapshot_management/blob/master/app/controllers/api/v2/snapshots_controller.rb#L48) says it's optional (and doesn't even mention VMware only).
> 
> If that is always required, I think `foreman_snapshot_management` should just always set it (for VMware). If it's optional, `foreman_snapshot_management` should not error out :)
> 
> cc @mdellweg who contributed to that feature in the past I think ;-)

Yeah I agree, it is not a required parameter in snapshot plugin but for some reason it throws error in FAM when creating snapshot that the memory field is required. It used to work earlier but something changed in either FAM modules or snapshot plugin. I will again have a look.

evgeni:
Could it be, that when we don't pass `include_ram` in FAM (and thus it's not part of the API request), this ends up setting `include_ram: nil` in https://github.com/ATIX-AG/foreman_snapshot_management/blob/328071c1418143c1105143a5a7b716b2234808ca/app/controllers/api/v2/snapshots_controller.rb#L52

And RBVmomi/fog/whatever doesn't like it to be `nil`?

Could using smth like 
```ruby
@snapshot = resource_class.new(snapshot_params.to_h.merge(host: @host).merge(include_ram: Foreman::Cast.to_bool(params[:include_ram]))) 
```
help?

Manisha15:
> Could it be, that when we don't pass `include_ram` in FAM (and thus it's not part of the API request), this ends up setting `include_ram: nil` in https://github.com/ATIX-AG/foreman_snapshot_management/blob/328071c1418143c1105143a5a7b716b2234808ca/app/controllers/api/v2/snapshots_controller.rb#L52
> 
> And RBVmomi/fog/whatever doesn't like it to be `nil`?
> 
> Could using smth like
> 
> ```ruby
> @snapshot = resource_class.new(snapshot_params.to_h.merge(host: @host).merge(include_ram: Foreman::Cast.to_bool(params[:include_ram]))) 
> ```
> 
> help?

Yes, when we don't pass include_ram in FAM it sets `include_ram` as nil or empty string. The solution you provided on snapshot side works, thanks.
I will close this PR.

PR #1451, opened 2022-08-07
fix: allow value_type to be resource in module job_template

hairmare:
To use the already provided `resource_type` option we also need to specify inputs as being `value_type=resource`, not just plain, search, and date.

### DIFF
--- a/plugins/modules/job_template.py
+++ b/plugins/modules/job_template.py
@@ -137,6 +137,7 @@ options:
           - plain
           - search
           - date
+          - resource
         type: str
       resource_type:
         description:
@@ -322,6 +323,7 @@ template_input_foreman_spec = {
         'plain',
         'search',
         'date',
+        'resource',
     ]),
     'resource_type': dict(),
 }


evgeni:
Introduced in 3.1 via https://github.com/theforeman/foreman/commit/989e0aad01df80dea8fd4734a78e794770b39f2d

Thanks!

evgeni (approved):

hairmare:
heyyy,,,, @evgeni 

thanks for being so proactive wrt merging my shit over the last few weeks

i'm not done with my tests/poc/prod yet, but i already want to let you know that i very much appreciate how productive and rewarding getting changes to FAM merged has been! :purple_heart:

evgeni:
Glad you're enjoying the contributor experience @hairmare!

FAM is a lot about scratching own itches (it started as "we need to get this done, let's use ansible for that"), so if something is missing, it's probably because noone found that itch yet. Now you did and fixed it! On to the next one (and I am sure you'll find more ;))

PR #1453, opened 2022-08-11
Add content export version

jeremylenz:
Split off from https://github.com/theforeman/foreman-ansible-modules/pull/1433


evgeni:
please do the rebase dance! :woman_dancing:

tests/test_playbooks/content_export_info.yml
```diff
@@ -21,13 +21,6 @@
         incremental: false
         destination_server: "airgapped.example.com"
 
-- hosts: tests
-  collections:
-    - theforeman.foreman
-  gather_facts: false
-  vars_files:
-    - vars/server.yml
-  tasks:
```
evgeni: why are you deleting these? the tasks need to run against the `tests` "host", so they are properly recorded.

tests/test_playbooks/content_export_version.yml
```diff
@@ -0,0 +1,77 @@
...
+        id: "{{ cv_info['content_view_versions'][0]['id'] }}"
+        organization_name: "Export Org"
+        foreman_organization: "Export Org"
+        destination_server: "airgapped.example.com"
+    - include_tasks: tasks/content_export_show.yml
+      vars:
+        id: "{{ result['task']['output']['export_history_id'] | default(12345) }}"
+        organization_name: "Export Org"
+        foreman_organization: "Export Org"
+        destination_server: "airgapped.example.com"
```
evgeni, line 56: this needs a `when: not ansible_check_mode` as in the library export, as in check mode, no export happened so you can't `info` it.

tests/test_playbooks/content_export_info.yml
jeremylenz: hmm, maybe this was a rebase error? I'll look again

plugins/modules/content_export_version.py
```diff
@@ -0,0 +1,167 @@
...
+
+
+class KatelloContentExportModule(KatelloAnsibleModule):
+    pass
+
+
+def main():
+    module = KatelloContentExportModule(
+        foreman_spec=dict(
+            content_view_version=dict(type='entity', scope=['content_view'], search_by='version', required=False),
```
evgeni: this stores the cv version as `content_view_version`, but the controller expects it as `id`, so let's rename it:

```suggestion
            content_view_version=dict(type='entity', scope=['content_view'], search_by='version', flat_name='id', required=False),
```

evgeni:
Here is the full diff that I applied locally, and afterwards I could record test fixtures just fine without any errors.

```patch
diff --git plugins/modules/content_export_version.py plugins/modules/content_export_version.py
index f5792e8a..b3b5002b 100644
--- plugins/modules/content_export_version.py
+++ plugins/modules/content_export_version.py
@@ -136,7 +136,7 @@ class KatelloContentExportModule(KatelloAnsibleModule):
 def main():
         foreman_spec=dict(
-            content_view_version=dict(type='entity', scope=['content_view'], search_by='version', required=False),
+            content_view_version=dict(type='entity', scope=['content_view'], search_by='version', flat_name='id', required=False),
             content_view=dict(type='entity', scope=['organization'], required=False),
             chunk_size_gb=dict(required=False, type='int'),
diff --git tests/test_playbooks/content_export_version.yml tests/test_playbooks/content_export_version.yml
index 310d3923..d8058a6a 100644
--- tests/test_playbooks/content_export_version.yml
+++ tests/test_playbooks/content_export_version.yml
@@ -35,21 +35,11 @@
   vars_files:
     - vars/server.yml
   tasks:
-    - name: Fetch content view version info
-      content_view_version_info:
-        username: "{{ foreman_username }}"
-        password: "{{ foreman_password }}"
-        server_url: "{{ foreman_server_url }}"
-        organization: "Test Organization"
-        validate_certs: "{{ foreman_validate_certs }}"
-        content_view: "Test Content View"
-        search: 'version = "1.0"'
-      register: cv_info
     - include_tasks: tasks/content_export_version.yml # Full export
       vars:
         content_view: "Test Content View"
-        content_view_version: "{{ cv_info['content_view_versions'][0]['version'] }}"
+        content_view_version: "1.0"
         incremental: false
       register: result
 
@@ -58,10 +48,12 @@
         expected_change: true
         incremental: true
         content_view: "Test Content View"
-        content_view_version: "{{ cv_info['content_view_versions'][0]['version'] }}"
+        content_view_version: "1.0"
     - include_tasks: tasks/content_export_show.yml
       vars:
         id: "{{ result['task']['output']['export_history_id'] | default(12345) }}"
+      when:
+        - not ansible_check_mode
 ...
```

plugins/modules/content_export_version.py
evgeni: (and this is the reason you've been seeing "Error while performing version on content_exports: The following required parameters are missing: id" the other day: the apidoc says the controller expects the "id" parameter, but nothing was providing that)

jeremylenz:
@evgeni updated! I _think_ the lint failures are unrelated..

evgeni:
> @evgeni updated! I _think_ the lint failures are unrelated..

And you're right! https://github.com/theforeman/foreman-ansible-modules/pull/1464 should hide that lint issue.

evgeni:
rebased and applied two mini fixes

evgeni:
meh, of course, I removed a call, now VCR freaks out -.-

plugins/modules/content_export_version.py
```diff
@@ -0,0 +1,169 @@
...
+
+class KatelloContentExportModule(KatelloAnsibleModule):
+    pass
+
+
+def main():
+    module = KatelloContentExportModule(
+        foreman_spec=dict(
+            content_view_version=dict(type='entity', scope=['content_view'], search_by='version', flat_name='id', required=False),
+            content_view=dict(type='entity', scope=['organization'], required=False),
```
evgeni: @jeremylenz under which circumstances are those not required?

https://github.com/Katello/katello/blob/master/app/controllers/katello/api/v2/content_exports_controller.rb#L31 lists the `id` as required, and we can only get the `id` if we know the name of the CV and the "name/version" of the version, so I'd mark both as required, right?

(when changing here, don't forget to update the docs above, a shame it can't be autogenerated…)

jeremylenz:
@evgeni thanks, updated!

evgeni (approved):

PR #1454, opened 2022-08-11
Add content export repository

jeremylenz:
Split off from https://github.com/theforeman/foreman-ansible-modules/pull/1433


evgeni:
please do the rebase dance! :woman_dancing:

evgeni:
looking at https://github.com/Katello/katello/blob/fe5c7a80683e5376105826b85dfb15c2e3f8bb60/app/controllers/katello/api/v2/content_exports_controller.rb#L74-L89 it does not support the `destination_server` param like the others. any reason not  to have that?

jeremylenz:
@parthaa 👀

evgeni:
And there is also no `fail_on_missing_content`?

jeremylenz:
> it does not support the `destination_server` param like the others. any reason not to have that?

@parthaa do you recall why `content_export_repository` doesn't support `destination_server` or `fail_on_missing_content`?

parthaa:
> @parthaa do you recall why `content_export_repository` doesn't support `destination_server` or `fail_on_missing_content`?

`fail_on_missing_content` is automatically true for a repository export. If the only repository you are trying to export is not `immediate` we will just fail.

`destination_server` - When we were designing the repository, destination_server added a lot of complexity to the code. More importantly the typical the repository export user
- Exports each iteration repository incrementally or otherwise.
- Stores it at some webserver 
- Interested downstream  satellite will download individual increments and import.
`destination_server` was deemed an overkill for this scenario. But if we get push back we can look into implementing that.

evgeni:
Thanks for the explanation! No pushback from my side. It just looked weird without the background information. ;)1

jeremylenz:
Any more changes needed here?

evgeni:
> Any more changes needed here?

Yes. Instead of `id`, we need to pass a `repository` (with `flat_name=id` and `scope=['product']`) and a `product` (with `scope=['organization']`) so that the user can pass names instead of IDs.

jeremylenz:
Sanity tests are failing but I don't know what that means

jeremylenz:
```
FATAL: The 1 sanity test(s) listed below (out of 32) failed. See error output above for details.
validate-modules
```

plugins/modules/content_export_repository.py
```diff
@@ -0,0 +1,136 @@
...
+# along with this program.  If not, see <http://www.gnu.org/licenses/>.
+
+from __future__ import absolute_import, division, print_function
+__metaclass__ = type
+
+
+DOCUMENTATION = '''
+---
+module: content_export_repository
+version_added: 3.5.0
```
evgeni:
```suggestion
version_added: 3.6.0
```

plugins/modules/content_export_repository.py
```diff
@@ -0,0 +1,136 @@
...
+
+class KatelloContentExportModule(KatelloAnsibleModule):
+    pass
+
+
+def main():
+    module = KatelloContentExportModule(
+        foreman_spec=dict(
+            repository=dict(type='entity', flat_name='id', scope=['product']),
+            product=dict(type='entity', scope=['organization']),
```
evgeni:
```suggestion
            repository=dict(type='entity', flat_name='id', scope=['product'], required=True),
            product=dict(type='entity', scope=['organization'], required=True),
```

This is what sanity was complaining about!

```
ERROR: Found 2 validate-modules issue(s) which need to be resolved:
ERROR: plugins/modules/content_export_repository.py:0:0: doc-required-mismatch: Argument 'product' in argument_spec is not required, but is documented as being required
ERROR: plugins/modules/content_export_repository.py:0:0: doc-required-mismatch: Argument 'repository' in argument_spec is not required, but is documented as being required
```

plugins/modules/content_export_repository.py
```diff
@@ -0,0 +1,136 @@
...
+            chunk_size_gb=dict(required=False, type='int'),
+            from_history_id=dict(required=False, type='str'),
```
evgeni, line 113: I'd argue this should be an int and not a string?
(And then also fix it in the other modules where we've overlooked that :see_no_evil: )

tests/test_playbooks/content_export_repository.yml
```diff
@@ -0,0 +1,70 @@
...
+        foreman_organization: "Export Org"
+    - include_tasks: tasks/content_export_show.yml
+      vars:
+        id: "{{ result['task']['output']['export_history_id'] | default(12345) }}"
+        organization_name: "Export Org"
+        foreman_organization: "Export Org"
+        destination_server: "airgapped.example.com"
+      when:
+        - not ansible_check_mode
+...
```
evgeni, line 65: can you add another play, that goes against `localhost` again and tears down the created repo/product/org?

tests/test_playbooks/content_export_repository.yml
```diff
@@ -0,0 +1,70 @@
...
+        organization_name: "Export Org"
+      register: result
+- hosts: tests
+  collections:
+    - theforeman.foreman
+  gather_facts: false
+  vars_files:
+    - vars/server.yml
+  tasks:
+    - include_tasks: tasks/repository.yml
```
evgeni: this looks unnecessary, you've already created the repo in the `localhost` play above

tests/test_playbooks/content_export_repository.yml
```diff
@@ -0,0 +1,70 @@
...
+      register: result
+    - include_tasks: tasks/content_export_repository.yml # Full export
+      vars:
+        product: "Export Test Product"
+        repository: "TestRepo"
+        expected_change: true
+        incremental: false
+        organization_name: "Export Org"
+        foreman_organization: "Export Org"
+    - include_tasks: tasks/repository.yml # i dont know why but it loses result in the next step if I don't include repository.yml here
```
evgeni: same, this shouldn't be needed?

tests/test_playbooks/content_export_repository.yml
```diff
@@ -0,0 +1,70 @@
...
+        expected_change: true
+        incremental: true
+        organization_name: "Export Org"
+        foreman_organization: "Export Org"
+    - include_tasks: tasks/content_export_show.yml
+      vars:
+        id: "{{ result['task']['output']['export_history_id'] | default(12345) }}"
+        organization_name: "Export Org"
+        foreman_organization: "Export Org"
+        destination_server: "airgapped.example.com"
```
evgeni: given repository exports don't support a destination server, this seems odd?

evgeni (changes requested):

plugins/modules/content_export_repository.py
jeremylenz: Oh! That's the error message I would've preferred to stumble upon..

plugins/modules/content_export_repository.py
jeremylenz: yeah, int makes a lot more sense! I see [your other PR](https://github.com/theforeman/foreman-ansible-modules/pull/1468) now so I guess I'll leave that change over there.

tests/test_playbooks/content_export_repository.yml
jeremylenz: see the comment.. I was having trouble with it forgetting the registered variable. Will test again, maybe I was just insane :laughing:

tests/test_playbooks/content_export_repository.yml
jeremylenz: we love copy-paste around here :confounded:

plugins/modules/content_export_repository.py
evgeni: It was in the output of the GitHub Action, maybe a tad hidden

tests/test_playbooks/content_export_repository.yml
jeremylenz: Whaddaya know.. Works fine. Removing.

jeremylenz:
:green_apple: :partying_face:

tests/test_playbooks/tasks/repository.yml
```diff
@@ -34,7 +34,7 @@
     ssl_client_key: "{{ repository_ssl_client_key | default(omit) }}"
   register: result
 - assert:
-    fail_msg: "Ensuring repositroy is {{ repository_state }} failed! (expected_change: {{ expected_change | default('unknown') }})"
+    fail_msg: "Ensuring repository is {{ repository_state }} failed! (expected_change: {{ expected_change | default('unknown') }})"
```
evgeni, line 37: :laughing: 

How are you finding all those typos?!

evgeni (approved):

PR #1456, opened 2022-08-16
Add include_ram option when creating VMWare snapshot

Manisha15:
At present, `include_ram` option when creating VMWare snapshot is set to False by default in foreman_snapshot_management plugin which is utilized by FAM modules. With this PR, user can now change the default value of `include_ram` and can set to it `True` as well.


evgeni (approved):

Manisha15:
Thanks @evgeni for merging the PR.

PR #1457, opened 2022-08-16
deprecate `foreman_setting`, users should use entity

evgeni:


evgeni:
Mhh, shouldn't this happen automatically anyways? *thinking*

mdellweg (approved):
So this is not even a code change...

evgeni:
Correct, as the underlying code already did that for us.

PR #1466, opened 2022-08-26
fix relative path of deps files so ansible-builder sanitize works

thedoubl3j:
ansible-builder sanitize cannot resolve relative paths, while the build command is able to. For this particular file and its requirement to placed be here, the paths will need to be absolute for sanitize to work as intended. sanitize is ran during EE builds that have multiple collections inside them so if a user is building with this collection included in it, it will fail. 

see this [awx](https://github.com/ansible/awx-ee/issues/132#issue-1351360059) issue for context.

### DIFF
--- a/meta/execution-environment.yml
+++ b/meta/execution-environment.yml
@@ -2,8 +2,8 @@
 version: 1
 
 build_arg_defaults:
-  EE_BASE_IMAGE: 'quay.io/ansible/ansible-runner:latest'
+  EE_BASE_IMAGE: "quay.io/ansible/ansible-runner:latest"
 
 dependencies:
-  python: ../requirements.txt
-  system: ../bindep.txt
+  python: requirements.txt
+  system: bindep.txt
--- a/meta/execution-environment.yml
+++ b/meta/execution-environment.yml
@@ -2,7 +2,7 @@
 version: 1
 
 build_arg_defaults:
-  EE_BASE_IMAGE: "quay.io/ansible/ansible-runner:latest"
+  EE_BASE_IMAGE: 'quay.io/ansible/ansible-runner:latest'
 
 dependencies:
   python: requirements.txt


evgeni:
CI seems to say the build command cannot figure out the path :(

thedoubl3j:
@evgeni yeah...we are seeing the problems that we didn't see coming when we first introduced the change for the file. context: https://github.com/ansible/ansible-builder/issues/393 

the problem here is that if relative, it breaks anyone else building with it, if absolute, fails the build for you. maybe have a second EE file that CI runs the builds from?

evgeni:
I don't think having a separate file for CI (and thus, let's be honest, always forget to keep the one the CI is not using in sync) is a workable solution long term.

thedoubl3j:
I am not sure what would be a long term solution then as currently, the CI is not using the file as builder intended (see issue linked). If the file stays the same currently, any user that is wants to build the foreman collection into an EE will not be able to build the environment outside of the current EE. This would also throw an error in the importer as well.

evgeni:
Let's turn the question around: how am I, as a collection maintainer, supposed to verify that the file I am shipping inside my collection is doing what it's supposed to? When we initially introduced this, my reading of the docs (and talking to Ansible people) was: call `ansible-builder build` on it, and try to use it in `ansible-navigator` and friends, and that's exactly what we implemented. If that's wrong, cool, we can change that, but we need a way to verify we're no shipping bad stuff (as we do right now, it seems).

evgeni:
Okay, I think I have a conceptual problem understanding how EEs should work :(

Looking at https://ansible-builder.readthedocs.io/en/stable/collection_metadata/#collection-level-metadata, I get the impression that `requirements.txt` and `bindep.txt` is *always* looked at, regardless what `meta/execution-environment.yml` is pointing at.

This also seems to match what I see here:

```console
% ansible-builder --version               
1.1.0

% ansible-builder introspect --sanitize build/collections
# Sanitized dependencies for build/collections
---
python:
- 'requests>=2.4.2  # from collection theforeman.foreman'
- 'ipaddress  # from collection theforeman.foreman'
system:
- 'python3-rpm [(platform:redhat platform:base-py3)]  # from collection theforeman.foreman'
- 'rpm-python [(platform:redhat platform:base-py2)]  # from collection theforeman.foreman'
- 'python38-requests [platform:centos-8 platform:rhel-8]  # from collection theforeman.foreman'

% rm build/collections/ansible_collections/theforeman/foreman/meta/execution-environment.yml
% ansible-builder introspect --sanitize build/collections                                                                                         
# Sanitized dependencies for build/collections
---
python:
- 'requests>=2.4.2  # from collection theforeman.foreman'
- 'ipaddress  # from collection theforeman.foreman'
system:
- 'python3-rpm [(platform:redhat platform:base-py3)]  # from collection theforeman.foreman'
- 'rpm-python [(platform:redhat platform:base-py2)]  # from collection theforeman.foreman'
- 'python38-requests [platform:centos-8 platform:rhel-8]  # from collection theforeman.foreman'
```

In reality, the contents of `meta/execution-environment.yml` *override* the paths. So if I set only `dependencies.python` there, there will be no system deps, and if it has no `dependencies` key at all, I get no deps listed. But if the file is missing completely, stuff is loaded properly?

So, what is this file really for?
It's not for building an EE for the collection (as that would require "wrong" paths, as we see in this discussion).
It's not *really* required if the collection deps are in the standard paths.

evgeni:
cc @alisonlhart because I think it was you who I originally spoke about the contents of this file and the "weird" looking relative paths :)

PR #1467, opened 2022-08-30
BZ#2118790 - Convert2RHEL - Content views for activation keys

stejskalleos:
Add content views for Convert2RHEL activation keys
with enabled repositories only for the specific OS

**Steps to test**
- install centos7
- update host
- reboot host so that new kernel is loaded
- register host to satellite with "CentOS 7 converting" host group
- Run `yum install convert2rhel`

For more details see https://bugzilla.redhat.com/show_bug.cgi?id=2118790

### DIFF
--- a/roles/convert2rhel/README.md
+++ b/roles/convert2rhel/README.md
@@ -36,7 +36,7 @@ Convert2RHEL
         foreman_manifest_path: "~/manifest.zip"
         foreman_content_rhel_enable_rhel7: true
         foreman_content_rhel_enable_rhel8: true
-        foreman_content_rhel_rhel8_releasever: 8.4
+        foreman_content_rhel_rhel8_releasever: 8.5
         foreman_content_rhel_wait_for_syncs: false
         foreman_convert2rhel_lifecycle_env: "Library"
         foreman_convert2rhel_content_view: "Default Organization View"
--- a/roles/convert2rhel/tasks/activation_keys.yml
+++ b/roles/convert2rhel/tasks/activation_keys.yml
@@ -31,8 +31,7 @@
     organization: "{{ foreman_organization }}"
     name: "{{ foreman_convert2rhel_key_centos7 }}"
     lifecycle_environment: "{{ foreman_convert2rhel_lifecycle_env }}"
-    content_view: "{{ foreman_convert2rhel_content_view }}"
-    subscriptions: "{{ foreman_convert2rhel_org_info['organization']['simple_content_access'] | ternary(omit, foreman_convert2rhel_centos7_subs) }}"
+    content_view: "{{ foreman_convert2rhel_key_centos7 }}"
   when: foreman_content_rhel_enable_rhel7
 
 - name: "Create '{{ foreman_convert2rhel_key_centos8 }}' activation key"
@@ -44,9 +43,7 @@
     organization: "{{ foreman_organization }}"
     name: "{{ foreman_convert2rhel_key_centos8 }}"
     lifecycle_environment: "{{ foreman_convert2rhel_lifecycle_env }}"
-    content_view: "{{ foreman_convert2rhel_content_view }}"
-    auto_attach: false
-    subscriptions: "{{ foreman_convert2rhel_org_info['organization']['simple_content_access'] | ternary(omit, foreman_convert2rhel_centos8_subs) }}"
+    content_view: "{{ foreman_convert2rhel_key_centos8 }}"
   when: foreman_content_rhel_enable_rhel8
 
 - name: "Create '{{ foreman_convert2rhel_key_oracle7 }}' activation key"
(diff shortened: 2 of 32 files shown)


roles/convert2rhel/tasks/activation_keys.yml
```diff
@@ -31,8 +31,7 @@
     organization: "{{ foreman_organization }}"
     name: "{{ foreman_convert2rhel_key_centos7 }}"
     lifecycle_environment: "{{ foreman_convert2rhel_lifecycle_env }}"
-    content_view: "{{ foreman_convert2rhel_content_view }}"
-    subscriptions: "{{ foreman_convert2rhel_org_info['organization']['simple_content_access'] | ternary(omit, foreman_convert2rhel_centos7_subs) }}"
```
evgeni, line 35: Are subscriptions unnecessary now? (I'd guess in the non-SCA case, they are still needed?)

roles/convert2rhel/tasks/activation_keys.yml
```diff
@@ -31,8 +31,7 @@
     organization: "{{ foreman_organization }}"
     name: "{{ foreman_convert2rhel_key_centos7 }}"
     lifecycle_environment: "{{ foreman_convert2rhel_lifecycle_env }}"
-    content_view: "{{ foreman_convert2rhel_content_view }}"
```
evgeni, line 34: This CV is still used in the `host_groups` playbook, is that correct?!

roles/convert2rhel/tasks/content_views.yml
```diff
@@ -0,0 +1,104 @@
...
+  theforeman.foreman.content_view:
+    username: "{{ foreman_username }}"
+    password: "{{ foreman_password }}"
+    server_url: "{{ foreman_server_url }}"
+    validate_certs: "{{ foreman_validate_certs | default(omit) }}"
+    organization: "{{ foreman_organization }}"
+    name: "{{ foreman_convert2rhel_key_oracle7 }}"
+    repositories:
+      - name: '{{ foreman_convert2rhel_oracle7_repo }}'
+        product: '{{ foreman_convert2rhel_oracle7_product }}'
```
evgeni: don't you need both, `foreman_convert2rhel_rhel7_repo` AND `foreman_convert2rhel_oracle7_repo`, in the Oracle case?

roles/convert2rhel/tasks/content_views.yml
```diff
@@ -0,0 +1,104 @@
...
+  theforeman.foreman.content_view:
+    username: "{{ foreman_username }}"
+    password: "{{ foreman_password }}"
+    server_url: "{{ foreman_server_url }}"
+    validate_certs: "{{ foreman_validate_certs | default(omit) }}"
+    organization: "{{ foreman_organization }}"
+    name: "{{ foreman_convert2rhel_key_oracle8 }}"
+    repositories:
+      - name: '{{ foreman_convert2rhel_oracle8_repo }}'
+        product: '{{ foreman_convert2rhel_oracle8_product }}'
```
evgeni: don't you need both, `foreman_convert2rhel_rhel8_repo` AND `foreman_convert2rhel_oracle8_repo`, in the Oracle case?

evgeni:
The tests seem to pass and all, but I am not sure about the actual changes and their influence on the c2r workflow.

Paging @ares as I don't have an environment to test this fully.

roles/convert2rhel/tasks/activation_keys.yml
stejskalleos: Yes it's correct. Now if I'm thinking about that, we should probably also change the CVs for the host groups, kinda forgot about it when was doing this PR

stejskalleos:
@evgeni updated

* Changes only for CentOS 7 and CentOS 8
* There is still probably issue with Oracle but  I am unable to verify it due to various issues. Since the issue is reported for CentOS let's focus on CentOS only
* Removed content views and environments from the host groups. Registered host will get them from activation keys anyway

roles/convert2rhel/README.md
```diff
@@ -36,7 +36,7 @@ Convert2RHEL
         foreman_manifest_path: "~/manifest.zip"
         foreman_content_rhel_enable_rhel7: true
         foreman_content_rhel_enable_rhel8: true
-        foreman_content_rhel_rhel8_releasever: 8.4
+        foreman_content_rhel_rhel8_releasever: 8.5
```
evgeni, line 39: Why not 8.6? ;)

stejskalleos: Because it doesn't work with `8.6`, see https://access.redhat.com/discussions/5676681

evgeni: wow.

evgeni:
The changes look good to me from an Ansible PoV, but I don't have an environment to test the full workflow.

@stejskalleos @ares if you're confident this works, we can merge.

evgeni (approved):

stejskalleos:
@ares are you ok with the changes? I'm confident it works, tested workflow with CentOS 7 & 8 and it worked.

ares (approved):
I think this is fine and can be merged.

PR #1468, opened 2022-08-30
make from_history_id an integer

evgeni:


evgeni:
:eyes: @jeremylenz

plugins/modules/content_export_library.py
```diff
@@ -103,7 +103,7 @@
     organization: "Default Organization"
     destination_server: "airgapped.example.com"
     incremental: true
-    from_history_id: "12345"
+    from_history_id: 12345
```
evgeni, line 106: funnily, this was the only example using a string, the others were using ints ;)

mdellweg (approved):
Ansible-lint does not look at the examples section? tztztz

evgeni:
> Ansible-lint does not look at the examples section? tztztz

Neither does `ansible-test sanity`, it's essentially the wild west down there.

PR #1471, opened 2022-09-01
Fix paths inside meta/execution-environment.yml

ssbarnea:
Related: https://github.com/ansible/ansible-builder/issues/406#issuecomment-1195584188
Related: https://github.com/ansible/creator-ee/issues/92

### DIFF
--- a/meta/bindep.txt
+++ b/meta/bindep.txt
@@ -1,0 +1,1 @@
+../bindep.txt
--- a/meta/execution-environment.yml
+++ b/meta/execution-environment.yml
@@ -5,5 +5,8 @@ build_arg_defaults:
   EE_BASE_IMAGE: 'quay.io/ansible/ansible-runner:latest'
 
 dependencies:
-  python: ../requirements.txt
-  system: ../bindep.txt
+  # File paths are relative to collection root folder when collection is
+  # installed but relative to meta/ when it is built, so we need symlink magic
+  # See https://github.com/ansible/ansible-builder/issues/406#issuecomment-1195584188
+  python: requirements.txt
+  system: bindep.txt
--- a/meta/requirements.txt
+++ b/meta/requirements.txt
@@ -1,0 +1,1 @@
+../requirements.txt


evgeni:
Thanks!

Are the symlinks supposed to end up in the tarball release? I guess so. Then they need to be added to https://github.com/theforeman/foreman-ansible-modules/blob/85f978f76b840c060cab80c0b209b3516c93fd85/Makefile#L10

Could you (or anyone else who understands EE and builder) look at my last comments in https://github.com/theforeman/foreman-ansible-modules/pull/1466? As I still don't exactly understand what this file is doing for us.

ssbarnea:
To be honest, I do not know if these are needed inside the tarfile, as builder will use the one  from repository root in that case. I think that the symlinks were needed only to make your execution work.

Yes, I totally share the confusion with you here. I am just someone else that got hit by the builder buss. The good part is that I raised the issue on the internal builder channel, so we should be able to get some better feedback.

nitzmahone:
I've recently inherited architectural ownership of ansible-builder- looking at how `meta/execution-environment.yml`  is used by builder (at least in its current form), it looks like more of a liability to you than a help. As long as your `requirements.txt` and `bindep.txt` remain in the root of the collection (ie both in git and included in the built tarball), builder will pick them up from there if the meta file is absent. If you want an execution env def for your own CI purposes or whatever that can refer to those files, it's currently easier to just put it at the root of the collection as well (and probably don't include it in the collection). That should still provide you with a buildable EE def as well as allowing the metadata to be available to other EEs that want to include this collection.

This is clearly a design problem with using the EE def for metadata exchange (since doing so results in it not being usable for its original purpose). We're working on a bunch of builder enhancements right now, so I'll add this one to the pile and see if there's any decent solution to it, but for now, the easiest thing is probably just to delete it or move it to the root.

evgeni:
Thanks @nitzmahone for chiming in!

Well, we have this file, because back in May we got told that this file is a *requirement* for being a Certified Collection, and the `redhat.satellite` one is such a collection and is built from the very same source as `theforeman.foreman` (minus branding changes).

So did that requirement change? If it did, I'd gladly get rid of that file :)

nitzmahone:
That's another whole group of folks that make those requirements- there may be other tools that are using that file today, but I kinda suspect not (maybe it's aspirational). Regardless, we've clearly got a problem here- if you want/need to keep the file *and* make it buildable as an EE in its own right, the symlinks are probably the best option for the moment, but we might be able to get someone to waive that requirement for the moment or permanently, depending on the reason it was added in the first place...

cc @thedoubl3j

thedoubl3j:
reason for addition was for python dep resolution in galaxy-importer which uses sanitize to confirm that the requirements.txt and or bindep.txt files are present in certified collections because all certified collections (of which downstream foreman is) MUST contain these files to be built into EEs. I can raise it up but can't make promises for its removal. 

tl;dr certified collections must be able to be built into EEs, this was the path taken to confirm that the files were present and usable by builder if they were needed for the EE.

nitzmahone:
After a quick back-channel with @thedoubl3j it sounds like they're cool temporarily waiving the EE def requirement for now until we get the right way to solve this sorted out- it's already effectively just a warning in the Automation Hub import anyway.

So basically do whatever works for you- if you want to have the EE def there and buildable, add the symlinks in `meta/` for the bindep and requirements files, otherwise just nuke it altogether for now (but leave the bindep and requirements files in the root of the repo and built collection tarball) and do your release without it.

evgeni:
I'm cool with the current workaround for now (and will merge this PR after this comment).

It's just that I don't see the point of that file, as things *seem* to be working fine without and it seems to just complicate things more than it should?

evgeni (approved):

PR #1475, opened 2022-09-06
subscription_manifest - increase the import timeout to 10 minutes

### LINKED ISSUE #1474: Occasional timeout with subscription_manifest
I have an older test manifest which can be uploaded to Satellite and refreshed manually, however when refreshing it using Ansible the operation times out ("Timout waiting for Task e7b9fc02-2a76-4c4c-87b7-cc7c147d4da9"). The typo in the message reveals that it is coming from `plugins/module_utils/foreman_helper.py` but there doesn't seem a way to increase timeout for the task.

The timeout occurs after 5 minutes. The operation seems to continue in the background and finishes perhaps in 7 minutes. Given that this is an old test manifest it might be ok to stick with the 5 minute timeout but OTOH increasing the timeout would avoid the issue. So I'll leave it for your consideration what would be the most appropriate course of action here, if any. Thanks.

evgeni:
Fixes: #1474

### DIFF
--- a/plugins/modules/subscription_manifest.py
+++ b/plugins/modules/subscription_manifest.py
@@ -84,7 +84,7 @@ def main():
         supports_check_mode=False,
     )
 
-    module.task_timeout = 5 * 60
+    module.task_timeout = 10 * 60
 
     with module.api_connection():
         organization = module.lookup_entity('organization')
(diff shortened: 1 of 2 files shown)


mdellweg (approved):

plugins/modules/subscription_manifest.py
```diff
@@ -84,7 +84,7 @@ def main():
         supports_check_mode=False,
     )
 
-    module.task_timeout = 5 * 60
+    module.task_timeout = 10 * 60
```
mdellweg, line 87: Do we want to expose that as a parameter or are we rather safe now with 10 minutes?

evgeni: I prefer to not have a param for that.

10 should be rather safe (in most cases even 5 is fine)

PR #1489, opened 2022-09-14
Build docs in release workflow

maximiliankolb:
Required to package FAM docs in https://github.com/theforeman/foreman-packaging/pull/8267.


evgeni:
Right now, the docs folder is not part of the distributed tarball, so even if you'd build it like you do here, it won't be in the artifact as you'd like.

PR #1495, opened 2022-10-07
redhat_manifest - return the UUID of the manifest so it can be reused

evgeni:


evgeni:
lint fixes in https://github.com/theforeman/foreman-ansible-modules/pull/1496

mdellweg (approved):

Issue #1494, opened 2022-10-06
switch redhat_manifest to org mode by default


evgeni:
##### SUMMARY
RH defaults to SCA these days, and so should we.

approved by @sideangleside himself

##### ISSUE TYPE
 - Bug Report

##### ANSIBLE VERSION

```

```

##### COLLECTION VERSION

```

```

##### KATELLO/FOREMAN VERSION

```

```

##### STEPS TO REPRODUCE

```yaml

```

##### ACTUAL RESULTS

```

```

PR #1499, opened 2022-10-12
redhat_manifest - set default ``quantity`` to 1

evgeni:
The code decides whether subs need to be attached or not by the presence of the `pool_id` parameter. However, to be able to actually attach a pool, one also needs to pass the quantity.

We could document that and set a `required_together` property, or do the sensible thing and just default the `quantity` to 1, which is a sane default in the days of SCA where we need to attach a sub to make Katello aware of the fact that we have access to said content, but otherwise don't need the *correct amount* of subs attached for providing them to clients.


ikke-t:
I second the default of one. I was thinking to ask you this, but you already did it.

sideangleside:
+1 from me as well, especially given that #1494 will make the default manifest mode `org_environment`

ehelms (approved):

PR #1502, opened 2022-10-14
drop apidoc patches that are not required since Foreman 2.0

evgeni:


evgeni:
This is *technically* a breaking change, but at the same time we document we only support current stable releases, which pre-2.0 certainly are not.

mdellweg (approved):

Issue #1436, opened 2022-07-01
content_view and content_view_publish roles variable isse


maciej-markowski:
##### SUMMARY
content_view and content_view_publish are using the same named variable but first one expects a list of dictionaries while second list of strings.

##### ISSUE TYPE
 - Bug Report

##### ANSIBLE VERSION

```
ansible --version
ansible [core 2.12.4]

```

##### COLLECTION VERSION

```
theforeman.foreman   3.4.0

```

##### KATELLO/FOREMAN VERSION

```
foreman-3.3.0-1.el8.noarch
```

##### STEPS TO REPRODUCE

Try to use both roles from the same playbook

```yaml

```

##### EXPECTED RESULTS

Unified variable type
##### ACTUAL RESULTS

```

```

gvde:
It's unclear to me what you mean exactly.

There are two roles:

- `theforeman.foreman.content_view_publish`
- `theforeman.foreman.content_views`

There is no role with "content_view".

There is only a module `theforeman.foreman.content_view`.

The role content_views creates/updates a content view. It uses the content_view module to do so. So obviously the role content_views needs a list of dictionaries to define all parameters needed to set up a content_view.

The role content_view_publish publishes a new version of one or more content views. For this, it only needs to identify the content view to be published, thus it only needs a list of string.

I don't know how else it should be...

maciej-markowski:
Sorry for the typo, I meant content_views role.
Both roles are using the same named variable, which are of different type so it is confusing and eg.  if you want to use both roles from single playbook, you cannot just define the variable in eg. in host vars, but you need to define it separately for each role eg.:

```
    - role: theforeman.foreman.content_views
      vars:
        foreman_content_views: "{{ create_foreman_content_views }}"
    - role: theforeman.foreman.content_view_publish
      vars:
        foreman_content_views: "{{ publish_foreman_content_views }}"
```

Basically, if role content_view_publish would use `content_view.name` inside the loop instead of `content_view`, you wouldn't need to define the variables separately.

PR #1507, opened 2022-10-24
Content Views Publish Role Variable Conflict

willtome:
This PR fixes Issue #1436

**Summary of Issue**
The `content_views` role and `content_view_publish` rule used the same variable name `foreman_content_views` but accepted different data structures.

Please expedite merge and release. This is a blocking issue for [RedHatGov/product-demos](https://github.com/RedHatGov/product-demos/blob/2f123b6359542f8de5eaad96ee99f330285d8d9d/satellite/setup.yml#L195).

### DIFF
--- a/roles/content_view_publish/README.md
+++ b/roles/content_view_publish/README.md
@@ -10,7 +10,7 @@ This role supports the [Common Role Variables](https://github.com/theforeman/for
 
 ### Required
 
-- `foreman_content_views`: List of content views to publish
+- `foreman_content_views_publish`: List of content views to publish
 
 Example Playbook
 ----------------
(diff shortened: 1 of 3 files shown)


wbclark (commented):
Thanks @willtome 

Should we use a convention of `$collectionname_$rolename_$varname` to be consistent and unambiguous?

For example, `foreman_content_view_publish_content_views` or append whatever alternate $varname works best.

knumskull:
I've provided another PR #1571 - This allows to use the same variable without any change or double definition. Just define the variable as required for `content_views` and it will work with `content_view_publish` too.

It's tested with https://github.com/knumskull/ansible-satellite

evgeni:
closing in favor of #1571, which is indeed better UX

PR #1509, opened 2022-10-25
add ``ignore_types`` to locations and organizations modules

evgeni:


evgeni:
needs new fixtures, only here to test for @sbernhard

Manisha15:
Result of `ansible-playbook --diff <playbook.yml>` with `ignore_types: ["Subnet"]`:
```
--- before
+++ after
@@ -1,5 +1,11 @@
 {
     "locations": [
-        {}
+        {
+            "id": 6,
+            "name": "Test Location",
+            "organization_ids": [
+                1
+            ]
+        }
     ]
 }
```
Result of `ansible-playbook --diff --check <playbook.yml>` with `ignore_types: ["Subnet"]`:
```--- before
+++ after
@@ -3,9 +3,7 @@
         {
             "id": 6,
             "ignore_types": [
-                "Subnet",
-                "ProvisioningTemplate",
-                "Hostgroup"
+                "Subnet"
             ],
             "name": "Test Location",
             "organization_ids": [
```

The `all host groups` is still checked even though it is not in the ignore_types list.

evgeni:
Are you editing an Organization or a Location?

Manisha15:
> Are you editing an Organization or a Location?

I created new location and then edited it.

evgeni:
for locations, the code seems to work for me, but the reporting in ansible is broken (I know why, but not sure how to fix it)

evgeni:
okay, it seems to work fine for me now for locations

Manisha15:
> okay, it seems to work fine for me now for locations

Okay, I will re-test it. It was not unchecking the hostgroups unless the ignore_types is empty list, in all other cases it checked all host groups.

evgeni:
yeah, and after patching https://projects.theforeman.org/issues/35687 locally, orgs also work fine

evgeni:
the lint issues are unrelated (see #1515)

evgeni:
And @Manisha15 and I filed https://projects.theforeman.org/issues/35732 for the weird behavior for locations, but there is not much we can do here at the moment :(

wbclark (commented):
It seems like it was once desired behavior ( https://projects.theforeman.org/issues/5619 ) for hostgroups and provisioning templates to be in ignored_types for a Location by default.

If the user wants to create a Location which does not have hostgroups or provisioning templates as ignored_types, could we add some documentation instructing them to have a duplicate task in their playbook, so that the Location is first created, and then gets updated to have only the desired ignored_types?

What do you think @evgeni @Manisha15 ?

evgeni:
No, because while that would untick the "all" checkbox, all *existing* HostGroups and ProvisioningTemplates got the location assigned and would require manual unassignment.

ehelms (approved):

evgeni:
(added changelog)

PR #1513, opened 2022-11-02
Use FQCNs in roles/

wbclark:
I noticed on another PR that the linter was complaining about this


evgeni:
I guess we should then drop `fqcn-builtins` from the ignore list?
https://github.com/theforeman/foreman-ansible-modules/blob/4a17931409fa9979028a74db41af2f5348ad1c97/.ansible-lint#L3

(I am still torn on that particular lint, but oh well)

evgeni (approved):

PR #1515, opened 2022-11-08
ensure redhat cert is available during lint too

evgeni:


evgeni:
I tend to think, this is a bug in `ansible-lint` and filed https://github.com/ansible/ansible-lint/issues/2661 for that.

Edit: probably more `ansible` bug: https://github.com/ansible/ansible/issues/79291

evgeni:
this got fixed in ansible

PR #1516, opened 2022-11-09
Add module to list all snapshots for a host

Manisha15:


plugins/modules/snapshot_info.py
```diff
@@ -0,0 +1,82 @@
...
+# along with this program.  If not, see <http://www.gnu.org/licenses/>.
+
+from __future__ import absolute_import, division, print_function
+__metaclass__ = type
+
+
+DOCUMENTATION = '''
+---
+module: snapshot_info
+version_added: 1.0.0
```
evgeni:
```suggestion
version_added: 3.8.0
```

;)

plugins/module_utils/foreman_helper.py
```diff
@@ -657,7 +657,8 @@ def list_resource(self, resource, search=None, params=None):
 
         if search is not None:
             params['search'] = search
-        params['per_page'] = PER_PAGE
+        if not resource == 'snapshots':
+            params['per_page'] = PER_PAGE
```
evgeni: Why is this needed? The API reply below in the fixtures does contain the `page` and `per_page` attributes, so I would expect the API to support paging?

https://github.com/ATIX-AG/foreman_snapshot_management/blob/master/app/controllers/api/v2/snapshots_controller.rb#L15 also mentions Pagination.

Manisha15: When recording and testing the fixtures, it was not able to match the query if there was pagination. It recorded the fixtures but threw error when testing the same fixtures with mismatch query due to pagination.

evgeni: :thinking: that's weird!

Would you mind dropping that change, re-record, and push that as a separate commit so that I can look at the diff?

evgeni: Hah!

I think I know what's up. Could it be that the apidoc snapshot (lol) we're using is older than https://github.com/ATIX-AG/foreman_snapshot_management/commit/7c91a1e9e7db3c34249d37543246e1ebc4acae2d?

```
% git log tests/fixtures/apidoc/snapshot.json
commit 241d49fd27e7f9926373a442768f12cbc2273e4e
Author: Manisha Singhal <manisha1595@gmail.com>
Date:   Thu Jun 27 14:57:05 2019 +0200

    Add Snapshot module
```

but the commit above is Dec 2019.

Please update the apidoc file and we're good! :)

Manisha15: Thanks, changing the apidoc file did the work.

plugins/modules/snapshot_info.py
```diff
@@ -0,0 +1,95 @@
...
+    required: true
+    type: str
+extends_documentation_fragment:
+  - theforeman.foreman.foreman
+  - theforeman.foreman.foreman.infomodule
+'''
+
+EXAMPLES = '''
+- name: "Show all snapshots for a host"
+  theforeman.foreman.subscription_info:
```
evgeni:
```suggestion
  theforeman.foreman.snapshot_info:
```

plugins/modules/snapshot_info.py
```diff
@@ -0,0 +1,95 @@
...
+    username: "admin"
+    password: "changeme"
+    server_url: "https://foreman.example.com"
+    host: "server.example.com"
+
+- name: "Show a snapshot"
+  theforeman.foreman.subscription_info:
```
evgeni:
```suggestion
  theforeman.foreman.snapshot_info:
```

evgeni (approved):

PR #1519, opened 2022-11-13
feat(job_template): add default param for template_inputs

hairmare:
Add support for setting a default value in job template inputs.

### DIFF
--- a/plugins/modules/job_template.py
+++ b/plugins/modules/job_template.py
@@ -147,6 +147,10 @@ options:
         description:
           - The value contains sensitive information and should't be normally visible, useful e.g. for passwords
         type: bool
+      default:
+        description:
+          - Default value for user input
+        type: str
 extends_documentation_fragment:
   - theforeman.foreman.foreman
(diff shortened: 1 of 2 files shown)


plugins/modules/job_template.py
```diff
@@ -147,6 +147,10 @@
         description:
           - The value contains sensitive information and should't be normally visible, useful e.g. for passwords
         type: bool
+      default:
+        description:
+          - Default value for user input
+        type: str
```
evgeni, line 153:
```suggestion
        type: str
        version_added: 3.8.0
```

evgeni (changes requested):
:+1: on the change!

Could you please add the `version_added` tag as in the suggestion and add a changelog snippet in the `changelogs/fragments` directory?

Something like `changelogs/fragments/job_template-template_inputs-default.yml` reading:
```yaml
minor_changes:
  - job_template - add ``default`` option to the ``template_inputs`` parameter

evgeni (approved):

PR #1520, opened 2022-11-15
Extend snapshot module to manage multiple snapshots with same name

Manisha15:
The [foreman_snapshot_management](https://github.com/ATIX-AG/foreman_snapshot_management) plugin provides option to create multiple snapshots with same name. This PR extends the existing snapshot module to provide this feature.

The snapshot module is extended to provide following features:
1. A boolean param, `new_entity` (default = False), if set true then user can create new snapshot entity with same name.
2. If `new_entity` is set to false, then user can update, revert, delete the snapshot by providing the snapshot id. The snapshot id can be fetched utilizing this [PR](https://github.com/theforeman/foreman-ansible-modules/pull/1516)
3. If new_entity and id params are not provided, the module will modify the oldest snapshot from the ones with same name.


Manisha15:
Failed test is unrelated.

plugins/modules/snapshot.py
```diff
@@ -58,6 +58,16 @@
     default: present
     choices: ["present", "reverted", "absent"]
     type: str
+  new_entity:
+    description:
+      - Flag to identity new snaoshot entity
```
m-bucher: Typo :grin:

plugins/modules/snapshot.py
```diff
@@ -125,19 +157,29 @@ def main():
     module = ForemanSnapshotModule(
         argument_spec=dict(
             state=dict(default='present', choices=['present', 'absent', 'reverted']),
+            new_entity=dict(required=False, type='bool', default=False)
```
m-bucher: Would it be possible to use a new `state` choice like `new` or `created` instead of a new parameter?

m-bucher (changes requested):

plugins/modules/snapshot.py
evgeni: Technically yes, but I am not sure what the better UX is, need to think about that a bit (and have time for that 😞)

sbernhard: how can we proceed with this?

plugins/modules/snapshot.py
Manisha15: removed the parameter.

plugins/modules/snapshot.py
Manisha15: Added a new state called `new_state`.

Manisha15:
docs tests are failing with unrelated reasons.

plugins/modules/snapshot.py
```diff
@@ -124,20 +149,29 @@ class ForemanSnapshotModule(ForemanEntityAnsibleModule):
...
             include_ram=dict(type='bool'),
+            id=dict(),
         ),
         required_plugins=[('snapshot_management', ['*'])],
         entity_opts={'scope': ['host']},
     )
 
     with module.api_connection():
-        module.run()
+        host_val = module.find_resource_by_name('hosts', name=module.foreman_params['host'])
```
evgeni:
```suggestion
        host_val = module.lookup_entity('host')
```

we have a helper for that and this saves the entity in a way that later `module.run()` doesn't have to look it up again.

evgeni:
I am still mulling over the nicest way to present this to the users, but let me recap the intent here.
The "name" of a snapshot is not unique, aka a single host can have two distinct snapshots with the name "snap".

Thus, the current behavior of the module is:
* called with `host=host.example.com` and `name=snap` and such a snapshot doesn't exist yet → create a snapshot
* called with `host=host.example.com` and `name=snap` and such a snapshot already exists → update existing snapshot

This is "correct", but now you want to extend this to allow a second snapshot to be created with the same name. Doing so also breaks the "update" case, as now the snapshot is not unique anymore and the module can't find the correct one to update.

So properly achieve your goal you need
* a new way to uniquely identify the snapshot (→ `id`)
* a way for the user to say "yes, create a NEW snapshot with the name 'snap'"

Is that correct?

Manisha15:
> I am still mulling over the nicest way to present this to the users, but let me recap the intent here. The "name" of a snapshot is not unique, aka a single host can have two distinct snapshots with the name "snap".
> 
> Thus, the current behavior of the module is:
> 
>     * called with `host=host.example.com` and `name=snap` and such a snapshot doesn't exist yet → create a snapshot
> 
>     * called with `host=host.example.com` and `name=snap` and such a snapshot already exists → update existing snapshot
> 
> 
> This is "correct", but now you want to extend this to allow a second snapshot to be created with the same name. Doing so also breaks the "update" case, as now the snapshot is not unique anymore and the module can't find the correct one to update.
> 
> So properly achieve your goal you need
> 
>     * a new way to uniquely identify the snapshot (→ `id`)
> 
>     * a way for the user to say "yes, create a NEW snapshot with the name 'snap'"
> 
> 
> Is that correct?

Yes, this is exactly what this PR is doing. And to get `snapshot-ids` user can utilize the functionality of `list_snapshots` module.

evgeni:
So I still don't really *like* this (non-unique names), but I guess if the backend allows it, we should also allow the users to manage it.

One thing that I'd change is the state name. We already have a "special" state `reverted`, so instead of `new_entity` I'd probably use `new_snapshot`?

Manisha15:
> So I still don't really _like_ this (non-unique names), but I guess if the backend allows it, we should also allow the users to manage it.
> 
> One thing that I'd change is the state name. We already have a "special" state `reverted`, so instead of `new_entity` I'd probably use `new_snapshot`?

Changed the state name to `new_snapshot`.

evgeni (approved):

Manisha15:
Thank you @evgeni

PR #1521, opened 2022-11-15
redhat_manifest: search by UUID on the server side if UUID is known

evgeni:
This is faster, as the Server has an index on the UUID in the database. This also allows to fetch manifests for accounts that have more than 1000 allocations and otherwise would require pagination (which we currently do not implement).

Co-Authored-By: Pavel Moravec <pmoravec@redhat.com>


pmoravec (approved):
Works great, thanks for the PR!

mdellweg (approved):

PR #1525, opened 2022-11-22
fix last spacing lint issue

evgeni:


evgeni:
fixed for python 2.7/3.5/3.6 in #1526 and unrelated to the change in this PR :)

mdellweg (approved):

PR #1531, opened 2022-12-01
don't sync the whole product if the repository could not be found

evgeni:
Fixes: dab1c937eb379aa7e77fd24b00ee34f6ec84b9ea

### DIFF
--- a/plugins/modules/repository_sync.py
+++ b/plugins/modules/repository_sync.py
@@ -66,7 +66,7 @@ def main():
     module = KatelloAnsibleModule(
         foreman_spec=dict(
             product=dict(type='entity', scope=['organization'], required=True),
-            repository=dict(type='entity', scope=['product'], failsafe=True),
+            repository=dict(type='entity', scope=['product']),
             # This should be scoped more explicit for better serch performance, but needs rerecording
             # repository=dict(type='entity', scope=['organization', 'product'], failsafe=True),
         ),
--- a/plugins/modules/repository_sync.py
+++ b/plugins/modules/repository_sync.py
@@ -67,8 +67,6 @@ def main():
         foreman_spec=dict(
             product=dict(type='entity', scope=['organization'], required=True),
             repository=dict(type='entity', scope=['product']),
-            # This should be scoped more explicit for better serch performance, but needs rerecording
-            # repository=dict(type='entity', scope=['organization', 'product'], failsafe=True),
         ),
     )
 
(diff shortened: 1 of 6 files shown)


pmoravec (approved):
Works great in all three scenarios:

1) just product name supplied (whole product synced)
2) product name and repo name correctly supplied (exactly the repo synced)
3) valid product name and invalid repo name supplied (originally, whole product synced, newly `"Found no results while searching for repositories with name=\"invalid repo name\""` reported as a failure)

tests/test_playbooks/fixtures/repository_ostree-1.yml
```diff
@@ -253,13 +248,20 @@ interactions:
       User-Agent:
       - apypie (https://github.com/Apipie/apypie)
     method: POST
-    uri: https://foreman.example.org/katello/api/products/4/sync
+    uri: https://foreman.example.org/katello/api/repositories/229/sync
```
evgeni, line 251: This is exactly the bug that this PR is fixing ;)

plugins/modules/repository_sync.py
```diff
@@ -66,9 +66,7 @@ def main():
     module = KatelloAnsibleModule(
         foreman_spec=dict(
             product=dict(type='entity', scope=['organization'], required=True),
-            repository=dict(type='entity', scope=['product'], failsafe=True),
-            # This should be scoped more explicit for better serch performance, but needs rerecording
-            # repository=dict(type='entity', scope=['organization', 'product'], failsafe=True),
+            repository=dict(type='entity', scope=['product']),
```
mdellweg, line 69: I get, the `failsafe` was the issue. What about the "better scoping" was that real?

evgeni: No, the commit that dropped the comment says:
> the product is already scoped by org, so there is no benefit of scoping
> the repository by both product *and* org.

tests/test_playbooks/repository_ostree.yml
```diff
@@ -40,7 +40,7 @@
     - name: sync repository
       include_tasks: tasks/katello_sync.yml
       vars:
-        repository: "Test Ostree Repository"
+        repository: "Test OStree Repository"
```
mdellweg, line 43: So this typo triggered the bug and we never noticed?
Should we keep a version of this expecting to properly fail now?

mdellweg (approved):
Just a suggestion for a test. Approving anyway.

tests/test_playbooks/repository_ostree.yml
evgeni: Correct, the typo triggered the wrong codepath being executed.

I'll merge as-is and add a test in a separate PR.

evgeni: https://github.com/theforeman/foreman-ansible-modules/pull/1532

PR #1533, opened 2022-12-05
Fixing content_view test

Griffin-Sullivan:
This part of the test was causing a failure in CI. I'm not sure why we need to do a negative assertion for a content view version for a content view that does not exist? If we want to negatively assert for CVV we at least would need a present CV to work with. Removing the task for now, but we can always add the content view to do this check then remove everything before we start the main part of the test.

### DIFF
--- a/tests/test_playbooks/content_view.yml
+++ b/tests/test_playbooks/content_view.yml
@@ -27,7 +27,7 @@
       vars:
         content_view_name: "Test Composite Content View"
         content_view_state: absent
-    - include_tasks: tasks/content_view_version.yml
+    - import_tasks: tasks/content_view_version.yml
       vars:
         state: absent
         version: "1.0"


evgeni:
What was the error you were getting?

Griffin-Sullivan:
```
TASK [Create katello content view version] *************************************
18:48:00  [0;31mfatal: [localhost]: FAILED! => {[0m
18:48:00  [0;31m    "changed": false,[0m
18:48:00  [0;31m    "invocation": {[0m
18:48:00  [0;31m        "module_args": {[0m
18:48:00  [0;31m            "content_view": "Test Content View",[0m
18:48:00  [0;31m            "current_lifecycle_environment": null,[0m
18:48:00  [0;31m            "description": null,[0m
18:48:00  [0;31m            "force_promote": false,[0m
18:48:00  [0;31m            "force_yum_metadata_regeneration": false,[0m
18:48:00  [0;31m            "lifecycle_environments": [[0m
18:48:00  [0;31m                "Library"[0m
18:48:00  [0;31m            ],[0m
18:48:00  [0;31m            "organization": "Test Organization",[0m
18:48:00  [0;31m            "password": "VALUE_SPECIFIED_IN_NO_LOG_PARAMETER",[0m
18:48:00  [0;31m            "server_url": "<server_url>",[0m
18:48:00  [0;31m            "state": "absent",[0m
18:48:00  [0;31m            "username": "admin",[0m
18:48:00  [0;31m            "validate_certs": false,[0m
18:48:00  [0;31m            "version": "1.0"[0m
18:48:00  [0;31m        }[0m
18:48:00  [0;31m    },[0m
18:48:00  [0;31m    "msg": "Found no results while searching for content_views with name=\"Test Content View\""[0m
```

Griffin-Sullivan:
Side note: the task before this makes a CV named "Test Composite Content View" and even if you specify that as the name for the CVV task it doesn't work.

evgeni:
Can you instead try the following please?

```patch
diff --git tests/test_playbooks/content_view.yml tests/test_playbooks/content_view.yml
index 4e93594e..c687ca63 100644
--- tests/test_playbooks/content_view.yml
+++ tests/test_playbooks/content_view.yml
@@ -27,7 +27,7 @@
       vars:
         content_view_name: "Test Composite Content View"
         content_view_state: absent
-    - include_tasks: tasks/content_view_version.yml
+    - import_tasks: tasks/content_view_version.yml
       vars:
         state: absent
         version: "1.0"
```

evgeni:
Just some context:
This is not a test, but setup for later tests. And it's not an assertion per se, we just need to make sure that the content view (that is deleted in the lines *after* that) has no version anymore, as otherwise the deletion fails, but we need the CV to be absolutely absent before we can run the tests.

This was broken by https://github.com/theforeman/foreman-ansible-modules/commit/7d1c9401b19f30a2433030d9c3a3aba53d0bc06d which assumed `include_tasks` would inherit the `ignore_errors` the same way as `include`, alas it did not.

Griffin-Sullivan:
That fixed it! But just to clarify, you are saying we are doing these absent tests to make sure there are no existing content views from other tests?

evgeni:
Yeah, as said, technically those are setup/teardown, not actual tests (even tho they use the same codebase).

The tests are destructive, as in, they change the system under test, but we want to perform more than one test before destroying the whole system, so we clean up things after a test has run (so kind of a teardown). Now the problem is that when you just run the "main" tests (the ones concerning the module in question and contain actual asserts) and they fail, they leave the system still in an unclean state, so our setup *also* does a bit of cleanup -- and that fails here as the system is actually clean but we broke the "ignore errors" part.

Also, see https://github.com/theforeman/foreman-ansible-modules/blob/develop/docs/testing.md#writing-tests

evgeni (approved):

PR #1540, opened 2022-12-17
Generate changelog for develop doc builds

evgeni:


.github/workflows/docs.yml
```diff
@@ -24,6 +24,9 @@ jobs:
           python-version: "3.9"
       - name: Install dependencies
         run: make doc-setup
+      - name: Generate changelog
+        run: antsibull-changelog release
+        if: github.ref == 'refs/heads/develop'
```
evgeni: Mhh, this will break committing the docs later:

```console
% git checkout gh-pages
error: Your local changes to the following files would be overwritten by checkout:
	CHANGELOG.rst
	changelogs/changelog.yaml
Please commit your changes or stash them before you switch branches.
Aborting
```

ehelms (approved):

Issue #854, opened 2020-06-24
Add module stream filter_type to katello_content_view_filter


ich199:
##### SUMMARY
From both the Katello UI and via hammer, you are able to add a content view filter rule to filter module streams from Yum content. This is not currently possible using the katello_content_view_filter module.

![screenshot_2020-06-24-10:13:08](https://user-images.githubusercontent.com/17448821/85528466-d43ac580-b603-11ea-9ef4-9d0065751c53.png)

##### ISSUE TYPE
 - Feature Idea

nesanton:
I might take it in a week if no one does before.

evgeni:
So I had a brief look at the API for that and I fear it's non-trivial to add right now.

The CV Filter Rule API expects a ModuleStream ID, which is just the numerical ID in the DB, so if we add a `module_streams` parameter, we'll have to do a search on the modulestream API endpoint to find the right stream id, and the API allows searching for the following terms: name, uuid, stream, version, context, arch, host, module_spec. the spec is the most verbose (it's NAME:STREAM:VERSION:CONTEXT:ARCH/PROFILE essentially) but we should allow the user to do any of these (well, maybe not host, that's IMHO not too useful in this context)

I vote we implement that when we have split the module into `content_view_filter` and `content_view_filter_rule`.

Blackmoresknight:
It would also be nice to be able to filter out all module streams (exclude all) if possible.

tinsjourney:
any update on the split of the module ?
Would be nice to be able to use ansible module to create a filter like "Include all Module Streams with no errata."

RJBJongejan:
Any update on this?

parmstro:
Please split the module into content_view_filter and content_view_filter_rule. Soonest. And implement modulemd. I need this functionality badly. You can't create content views with filters on anything with modules without this. My tests in Satellite show that you need both All RPMS no errata + All Streams no errata + your date based erratum filter or your content view is missing components and installs and patching fails. If we can at least support an include filter of type modulemd with original_packages=true that gets many content views over the hump.

Fabian1976:
Any update on this?

parmstro:
Not yet, my day job is getting in my way. :-)
I have started working on it.


Paul Armstrong
Principal Technical Specialist
Red Hat Technical Sales Organization
647-285-0044

RJBJongejan:
Can you give an update on this? I would very much like to automate my rhel8/rhel9 content views with this.

[...]

Issue #803, opened 2020-04-28
Feature requests for katello_content_view_filter


gabriel-mh:
##### SUMMARY
A several feature requests for katello_content_view_filter

##### ISSUE TYPE
Hello folks,

I've written a role for handling the content views yum filters & yum filter rules in Satellite and would like to request the following features:

> Split the model in 2:
i) Filter creation
ii) Rule creation

> Allow specifying the package groups by UUID (just a paramater, API takes it) or a product ID and implement an UUID look-up for the valid packages. (Call + pattern matching) Currently, if several repositories of the same name are exposed by different products, there's no way to specify a package group and trying to do so fails with "multiple results returned" error.

> Expose an explicit `filter_type` parameter, instead of changing the module's behaviour based on which variables are defined.

This is especially important for bulk processing, the only way to currently achieve this is conditionally importing a playbook with these definitions.

evgeni:
This is roughly what we want to do, see https://github.com/theforeman/foreman-ansible-modules/issues/695#issuecomment-582794505
At least the first step is identical, the rest might need to be split up, but sounds valid!

Would you be willing to work on this? :)

gabriel-mh:
Thanks for the reply!

Yes, I've started today :)

jeffwarnica:
I'm running into this identical problem, and more or less had modeled out the top proposal as a path forward. Even down to missing the duplicate group name problem.

Whatever the anti-denvercoder feeling is, I'm feeling it!

parmstro:
This issue should be resolved with the release of v3.9.0

Issue #695, opened 2020-02-04
I should be able to add multiple rules to a content view filter


jwbernin:
##### SUMMARY

The current model in use in katello_content_view_filter only allows the addition of a single rule per filter item. If the same call is reissued with the same filter name and a different rule, the rule is replaced rather than appended to. I would like to be able to either append to a rule with a follow-on play, or I would like to be able to specify multiple rules within a given filter.

##### ISSUE TYPE
 - Feature Idea

sean797:
I think if you supply the same name, but a different rule_name this should
work?

evgeni:
Yepp, setting different `rule_name`s should work. It just defaults to the same value as the filter name itself.

jwbernin:
`rule_name` is an alias of `package_name` so when I set a `rule_name`, I get `[WARNING]: Both option rule_name and its alias package_name are set.` and the rule is replaced, not appended to.

evgeni:
ah, can you post your whole playbook then to reproduce please?

jwbernin:
```
---
- name: Add filters
  hosts: localhost

  tasks:
    - name: Add filter
      katello_content_view_filter:
        username: admin
        password: password
        server_url: "https://satellite.example.com"
        validate_certs: false
        name: "pkg filter 1"
        organization: MyOrg
        content_view: rhel7
        filter_type: "rpm"
        package_name: "389-ds-base"
        version: "0:1.3.1.6-26.el7_0"
        inclusion: true

    - name: Add filter
      katello_content_view_filter:
        username: admin
        password: password
        server_url: "https://satellite.example.com"
        validate_certs: false
        name: "pkg filter 1"
        organization: MyOrg
        content_view: rhel7
        filter_type: "rpm"
        package_name: "389-ds-base"
        version: "0:1.3.3.1-13.el7"
        inclusion: true
```

jwbernin:
That is the smallest reproducer snippet - it will result in one filter with one rule, when it should have one filter with two rules.

evgeni:
Right, so after looking (long enough) at your example, the Katello UI around this, and the code of our module, I think that we should follow the workflow Katello UI does here, and split the module into two: `cv_filter` and `cv_filter_rule`, that way you first have to create a filter, and then, can add/remove as many rules to/from it as you like.

Another way would be to allow a `rules` list in the current module, and then compare the rules with the existing ones, but I think I prefer separate modules.

[...]

PR #1545, opened 2022-12-22
Updated Filter and Filter Rule model

parmstro:
This currently meets requirements of the issues #854,  #803, and #695, the outstanding issue is removing a modulemd rule because it is hard to identify. Working on it, but having to get the whole list of rules for the modulemd filter and then find the one that matches the spec for removal. This is related to the challenge that the katello API doesn't allow a search of the module stream filters by the module_stream_id or a NSVCA module spec. Looking for a quick hand if someone has a suggestion.


parmstro:
Updating with latest changes and closing this one. Outstanding issues resolved.

PR #1552, opened 2023-01-20
increase task timeout to 12h as export tasks can be time intensive

evgeni:
Fixes: https://bugzilla.redhat.com/show_bug.cgi?id=2162678


jeremylenz (approved):
LGTM :+1:

mdellweg (approved):

PR #1555, opened 2023-02-01
hostgroup_info

Greek-Fire:
I added this module because I needed a list of all the hostgroups foreman/satellite. Instead of doing it using the uri module, I created this so that others may use this modules. This is easier than using the URI module and creating your own http request and I hope it helps.


evgeni:
Thanks!

It's not exactly clear to me why CI fails, as your stuff looks correct. I'll have a deeper look.

evgeni (approved):

evgeni:
Thank you @TheRedGreek!

FTR, there is a `resource_info` module, so you don't need to use `uri` for things that are not yet present.

Greek-Fire:
Wow, that module is awesome. No more apidoc for me! If I notice a need for a different *_info, would it help if I submit a another PR?

evgeni:
Absolutely! Dedicated modules are always better.

Greek-Fire:
Sounds good!

Thank you.

Greek-Fire:
Also, while I have your attention. I would like to contribute to the OVAL scanning in https://github.com/theforeman/foreman_openscap. Is there a channel working out there that I can join to stay up-to-date on foreman_openscap?

evgeni:
There are links to our forum and chat at https://theforeman.org/support.html, both with development sections. But nothing openscap specific.

Greek-Fire:
Thank you!

PR #1558, opened 2023-02-10
content_view_filter_rule module

evgeni:
split out of #1547

### DIFF
--- a/plugins/modules/content_view_filter_rule.py
+++ b/plugins/modules/content_view_filter_rule.py
@@ -277,22 +277,15 @@ def main():
         search_scope = cvf_scope
         content_view_filter_rule = None
 
-        # there are really 2 erratum filter types by_date and by_id
-        # however the table backing them is denormalized to support both, as is the api
-
-        if 'errata_id' in module.foreman_params:
-            # this filter type supports many rules
-            # we need to search by errata_id, because it really doesn't have a name field.
-            rule_spec = content_filter_rule_erratum_id_spec
-            search_scope = {'errata_id': module.foreman_params['errata_id']}
-            search_scope.update(cvf_scope)
-            search = None
-            content_view_filter_rule = module.find_resource('content_view_filter_rules', search, params=search_scope, failsafe=True)
-
         if filter_type == 'erratum':
+            # this filter type supports many rules
+            # there are really 2 erratum filter types by_date and by_id
+            # however the table backing them is denormalized to support both, as is the api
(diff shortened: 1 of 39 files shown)


plugins/modules/content_view_filter_rule.py
```diff
@@ -0,0 +1,340 @@
...
+
+            if not module.desired_absent:
+                # if the state is present and the module_id is NOT in the exising list, add module_stream_id.
+                if not existing_rule:
+                    module.foreman_params['module_stream_ids'].append(module_stream['id'])
+
+                # if the state is present and the module_id IS in the list,
+                # make sure that the current and desired state are identical
+                else:
+                    content_view_filter_rule = module.foreman_params
```
evgeni: This *shouldn't* be necessary (but it is), so worth investigating further.

plugins/modules/content_view_filter_rule.py
```diff
@@ -0,0 +1,335 @@
...
+
+        elif filter_type == 'package_group':
+            # this filter type support many rules
+            # the name is the key to finding the proper one and is required for these types
+            # uuid is also a required value creating, but is implementation specific and not easily knowable to the end user - we find it for them
+            content_view_filter_rule = module.find_resource_by_name('content_view_filter_rules', module.foreman_params['name'],
+                                                                    params=search_scope, failsafe=True)
+
+            package_group = module.find_resource_by_name('package_groups', module.foreman_params['name'], params=cv_scope)
+            module.foreman_params['uuid'] = package_group['uuid']
```
evgeni: I'm wondering if simplifying this to the following will make it better readable or not?

```suggestion
        elif filter_type in ('rpm', 'docker', 'package_group'):
            # these filter types support many rules
            # the name is the key to finding the proper one and is required for these types
            content_view_filter_rule = module.find_resource_by_name('content_view_filter_rules', module.foreman_params['name'],
                                                                    params=search_scope, failsafe=True)

            if filter_type == 'package_group':
                # uuid is also a required value creating, but is implementation specific and not easily knowable to the end user - we find it for them
                package_group = module.find_resource_by_name('package_groups', module.foreman_params['name'], params=cv_scope)
                module.foreman_params['uuid'] = package_group['uuid']
```

parmstro: Yes. That works and means less maintenance.

parmstro (approved):
Yes, Evgeni. I think that the change is good. Clearer and less maintenance.

PR #1560, opened 2023-02-14
move ansible-format-automatic-specification to a version specific ignore

evgeni:
it's not a thing anymore in devel and thus pylint complains

    unknown-option-value: Unknown option value for 'disable', expected a valid pylint message and got 'ansible-format-automatic-specification'


mdellweg (approved):

evgeni:
and why does it not like my ignores?!

evgeni:
fixed in ansible devel via https://github.com/ansible/ansible/pull/79997 \o/

PR #1565, opened 2023-03-06
Workaround for Foreman without loc/org enabled

evgeni:


evgeni:
none of the failures are related, merging.

PR #1566, opened 2023-03-06
use pre-2.0 runner on all pre-3.8 pythons

evgeni:
see https://github.com/ansible/ansible-runner/issues/1205 for details


mdellweg (approved):
:facepalm:

PR #1570, opened 2023-03-13
Module additions

kent-hydra:
Adds the following modules:
- architecture_info
- location_info
- operatingsystem_info
- role_info
- user_info
- usergroup_info

Adds `parent` parameter to organization module

Adds `disabled` and `mail_enabled` parameter to user module


kent-hydra:
Could i get any guidance on the error messages here? Im not really sure whats going on with it. My best guess is the tests do not like the additional parameters being added, but Im not real sure why. I have been using these new/changed modules without any issues.

plugins/modules/user.py
```diff
@@ -515,9 +530,11 @@ def main():
     module = ForemanUserModule(
         foreman_spec=dict(
             login=dict(required=True, aliases=['name']),
+            disabled=dict(required=False, type='bool', default=False),
```
evgeni:
```suggestion
            disabled=dict(required=False, type='bool'),
```

Please don't set defaults unless really necessary. Ideally we depend on the API to provide those defaults to us.

plugins/modules/user.py
```diff
@@ -515,9 +530,11 @@ def main():
...
             firstname=dict(required=False),
             lastname=dict(required=False),
             mail=dict(required=False),
+            mail_enabled=dict(required=False, type='bool', default=False),
```
evgeni:
```suggestion
            mail_enabled=dict(required=False, type='bool'),
```

Please don't set defaults unless really necessary. Ideally we depend on the API to provide those defaults to us.

evgeni (changes requested):
> Could i get any guidance on the error messages here? Im not really sure whats going on with it. My best guess is the tests do not like the additional parameters being added, but Im not real sure why. I have been using these new/changed modules without any issues.

The tests don't like that you added the parameters with default values, but the pre-recorded API interactions that we use for the tests don't have those parameters (yet). If you drop the `default`s from the spec, this particular test will succeed.

Overall the tests will still fail tho, as you've added new modules, but did not add any tests for those. I can show you how (the TL;DR is in https://github.com/theforeman/foreman-ansible-modules/blob/develop/docs/testing.md#writing-tests but it's not strictly trivial when doing the first time) or do it for you if you prefer?

That said, would you mind splitting up this PR in three? one for the user module, one for the organization module, and one for the new info modules? That would make merging them easier/faster :)

PR #1573, opened 2023-03-20
add support for creating modulemd filters to cv_filter

evgeni:


evgeni:
merging as-is, as that's Pauls work :)

PR #1576, opened 2023-03-23
allow setting the format for an export, thus allowing syncable exports

evgeni:


evgeni:
note for future me: we probably should factor out most of the params (and docs) into a common thing, given they are, common.

ehelms (approved):

PR #1578, opened 2023-03-24
refactor export modules to have more common base

evgeni:


plugins/modules/content_export_library.py
```diff
@@ -116,42 +92,24 @@
...
 
-class KatelloContentExportModule(KatelloAnsibleModule):
+class KatelloContentExportModule(KatelloContentExportBaseModule):
     pass
 
 
 def main():
     module = KatelloContentExportModule(
         foreman_spec=dict(
             destination_server=dict(required=False, type='str'),
```
evgeni, line 105: I've opted into not moving `destination_server` and `fail_on_missing_content` because:
1. they don't exist for `repository` exports, so we'd need another layer of sub-classing
2. the doc-strings differ for `fail_on_missing_content`, so we'd not save much on the doc-fragments

ehelms (approved):

PR #1579, opened 2023-03-24
document how to obtain valid setting names for the setting module

evgeni:


lessfoobar:
when you hover over the config names in the web ui you are getting the var name to set. also there is this documentation if you prefer copy pasta:

[3.5.2 Configuration Options](https://theforeman.org/manuals/3.5/index.html#3.5Configuration)

[Appendix A: Administration Settings](https://docs.theforeman.org/3.5/Administering_Project/index-katello.html#administer_settings_information_admin)

ehelms (approved):

PR #1580, opened 2023-03-30
include information about the published version in the return value

evgeni:


plugins/modules/content_view_version.py
```diff
@@ -230,6 +227,12 @@ def main():
             module.ensure_entity('content_view_versions', None, content_view_version, params=scope)
         else:
             module.auto_lookup_entities()
+
+            # only need to record if not desired_absent, as ensure_entity already does all the proper recording for us
+            content_view_version_record = _flatten_entity(content_view_version, module.foreman_spec)
+            content_view_version_record['environments'] = [environment['name'] for environment in content_view_version_record.get('environments', [])]
```
evgeni, line 233: a tad unsure about this `environments` flattening here. it slightly changes the behaviour from what we had before, but the data here is supposed to be "flat" anyway.

jeremylenz: LGTM :shrug: 

I don't see why Ansible users would need any other lifecycle environment metadata here, other than the names.

evgeni: Yeah, plus, they're still getting the raw data in the "entities" return block, the flattening here is only for the diff output where the name-only notation makes most sense as that's also the input the user made

ehelms (approved):

PR #1586, opened 2023-04-07
stop using requests

evgeni:
this is using https://github.com/Apipie/apypie/pull/129 and some dirty tricks to make it work

don't try this at home, kids

### DIFF
--- a/plugins/module_utils/foreman_helper.py
+++ b/plugins/module_utils/foreman_helper.py
@@ -36,7 +36,6 @@ try:
     except ImportError:
         from plugins.module_utils import _apypie as apypie
         from plugins.module_utils.ansible_requests import RequestSession
-    import requests.exceptions
     HAS_APYPIE = True
     APYPIE_IMP_ERR = None
     inflector = apypie.Inflector()
@@ -1187,7 +1186,7 @@ class ForemanAnsibleModule(AnsibleModule):
 
     def fail_from_exception(self, exc, msg):
         fail = {'msg': msg}
-        if isinstance(exc, requests.exceptions.HTTPError):
+        if hasattr(exc, 'response'):
             try:
                 response = exc.response.json()
                 if 'error' in response:
--- a/plugins/module_utils/foreman_helper.py
+++ b/plugins/module_utils/foreman_helper.py
@@ -21,7 +21,7 @@ from collections import defaultdict
 from functools import wraps
 
 from ansible.module_utils.basic import AnsibleModule, missing_required_lib, env_fallback
-from ansible.module_utils._text import to_bytes, to_native
+from ansible.module_utils._text import to_native
 from ansible.module_utils import six
 from ansible.module_utils.urls import Request
 try:
@@ -690,8 +690,8 @@ class ForemanAnsibleModule(AnsibleModule):
         verify_ssl = self._foremanapi_ca_path if (self._foremanapi_validate_certs and self._foremanapi_ca_path) else self._foremanapi_validate_certs
         self.foremanapi = apypie.ForemanApi(
             uri=self._foremanapi_server_url,
-            username=to_bytes(self._foremanapi_username),
-            password=to_bytes(self._foremanapi_password),
+            username=self._foremanapi_username,
+            password=self._foremanapi_password,
             verify_ssl=verify_ssl,
             kerberos=self._foremanapi_use_gssapi,
             task_timeout=self.task_timeout,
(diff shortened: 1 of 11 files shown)


mdellweg:
Looks like a castling to me...
I'm about to take the exact opposite turn here: https://github.com/pulp/squeezer/pull/106

evgeni:
"heh"

the ever-returning problem is the python-package-dependency-handling of ansible-collection (which is effectively non-existant)

plugins/module_utils/foreman_helper.py
```diff
@@ -592,10 +664,11 @@ def connect(self):
 
         self.foremanapi = apypie.Api(
             uri=self._foremanapi_server_url,
-            username=to_bytes(self._foremanapi_username),
-            password=to_bytes(self._foremanapi_password),
```
evgeni, line 619: this was done as part of https://github.com/theforeman/foreman-ansible-modules/pull/609, but it seems ansible's `Request` does that for us instead.

plugins/module_utils/foreman_helper.py
```diff
@@ -89,6 +95,75 @@ class NoEntity(object):
...
+        return status
+
+    @property
+    def headers(self):
+        return self._resp.headers
+
+    def raise_for_status(self):
+        http_error_msg = ""
+        if self.status_code > 400:
+            http_error_msg = "you failed"
```
ekohl: If it's > 500 the server failed

plugins/module_utils/foreman_helper.py
```diff
@@ -89,6 +95,75 @@ class NoEntity(object):
...
+
+    @verify.setter
+    def verify(self, value):
+        self.validate_certs = value
+
+    def request(self, method, url, **kwargs):
+        validate_certs = kwargs.pop('verify', None)
+        params = kwargs.pop('params', None)
+        if params:
+            url += '?' + six.moves.urllib.parse.urlencode(params)
```
ekohl: Doesn't Ansible's HTTP client code itself implement this?

evgeni: So far, nope

plugins/module_utils/foreman_helper.py
```diff
@@ -1193,7 +1195,7 @@ def wait_for_task(self, task, ignore_errors=False):
 
     def fail_from_exception(self, exc, msg):
         fail = {'msg': msg}
-        if isinstance(exc, requests.exceptions.HTTPError):
+        if hasattr(exc, 'response'):
```
evgeni, line 1190: right now, the exceptions we'd raise wouldn't have that, and thus we never execute that code :see_no_evil:

evgeni:
So this works (mostly?) and is a good ground for the discussion.

Currently open issues:
* the `url += '?' + six.moves.urllib.parse.urlencode(params)` is ugly (as pointed out by Ewoud)
* the callback is totally untested
* it might have broken a shit ton unicode things
* I am using a method only available in Ansible 2.10+ -- time to say good bye to 2.9? Or make that use `requests`?

ekohl:
> the `url += '?' + six.moves.urllib.parse.urlencode(params)` is ugly (as pointed out by Ewoud)

Is there at least an upstream issue to support it? I think your approach to pop the parameter is safe, even if it's added, but not having to think about it is even nicer.

> I am using a method only available in Ansible 2.10+ -- time to say good bye to 2.9? Or make that use `requests`?

That also means dropping the ability to run on EL7, right? I don't know how much that's used, but I'd consider keeping it for a bit longer.

evgeni:
> > I am using a method only available in Ansible 2.10+ -- time to say good bye to 2.9? Or make that use `requests`?
> 
> That also means dropping the ability to run on EL7, right? I don't know how much that's used, but I'd consider keeping it for a bit longer.

"On EL7 by using only official packages", yes. Nothing forbids to run `pip` installed Ansible there.

evgeni:
> > the `url += '?' + six.moves.urllib.parse.urlencode(params)` is ugly (as pointed out by Ewoud)
> 
> Is there at least an upstream issue to support it? I think your approach to pop the parameter is safe, even if it's added, but not having to think about it is even nicer.

I don't think upstream cares, but I can ask.
Their wrapper is not trying to be `requests`-compatible and just expects the `url` parameter to contain all parameters already. Which is totally fine.

ekohl:
If it's an issue closed as WONTFIX that at least is a good code comment why it's done.

evgeni:
> If it's an issue closed as WONTFIX that at least is a good code comment why it's done.

https://github.com/ansible/ansible/issues/84274

PR #1589, opened 2023-04-12
don't require `content_type` and `content` parameters when removing

### LINKED ISSUE #1588: theforeman.foreman.content_credential shouldn't need to specify content parameter when state=absent
Is it necessary to specify content: parameter when deleting (state: absent) a content credential?  IMHO, you shouldn't have to [specify the content of the credential itself](https://docs.ansible.com/ansible/latest/collections/theforeman/foreman/content_credential_module.html#parameter-content) in order to delete it.  You should just be able to delete it.

evgeni:


evgeni:
lint fixes in https://github.com/theforeman/foreman-ansible-modules/pull/1590

mdellweg (approved):
So we can still have a `state: present` call without these and it is supposed to succeed if it is not creating a new object? Do we need to introduce a `required_if_new` thing for nice error messages?

evgeni:
`required_if_new` would be cool (we have weird logic to mimic this in a few places), but IMHO overkill for now

PR #1590, opened 2023-04-12
pin ansible-lint to less than 6.14.5

evgeni:
6.14.5 has a bug that breaks role validation: https://github.com/ansible/ansible-lint/issues/3277


evgeni:
6.14.5 was yanked

PR #1592, opened 2023-04-20
try using the rpm-shim instead of the rpm-py-installer hack

evgeni:


evgeni:
well, that only works for the system python (3.8 on ubuntu 20.04 which we use as a base), but not for all the other pythons. *sad*

PR #1593, opened 2023-04-20
pin pip to <23.1

evgeni:
23.1 dropped non-wheel installation and this breaks rpm-py-installer

also use `--force-reinstall` as sometimes pip is installed in a way
where there is a `pip` binary, but `python -m pip` doesn't work (no, I
have no idea *how* such an environment can be created) and becausse the
rpm-py-installer install script relies on that to work it explodes with
no good reason

PR #1594, opened 2023-04-21
don't duplicate apidoc jsons

evgeni:


evgeni:
contains https://github.com/theforeman/foreman-ansible-modules/pull/1593 as otherwise tests are just broken

mdellweg (approved):

PR #1595, opened 2023-04-24
don't add dependencies when building the source tree in build/src/

evgeni:
when there is a dependency that has a similar name as a target (say, uh, `tests/test_crud.py` and `test_%`), trying to make `build/src/tests/test_crud.py` will try to call the `test_crud` target and explode

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -105,8 +105,8 @@ endif
 $(MANIFEST): $(NAMESPACE)-$(NAME)-$(VERSION).tar.gz
 	ansible-galaxy collection install -p build/collections $< --force
 
-build/src/%: %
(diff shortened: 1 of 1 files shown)


evgeni:
@mdellweg this line is originally your magic, would you mind looking at this change?

(my intent is to ship tests in the built collection, but that currently prevents it)

mdellweg:
I can see collision with our special test role. But taking away the dependency from the copy role will make make fail to copy the file again had it been changed in the meantime. This may work in ci because all files are essentially new, but not properly in the development case.

evgeni:
Right... file refreshes. of fucking course.

PR #1596, opened 2023-04-24
include test data in collection

evgeni:
the main wish here is to have all the data in place to be able to run the tests *live*

before:

    -rw-r--r--. 1 evgeni evgeni 177K Apr 24 11:34 theforeman-foreman-3.11.0-dev.tar.gz

after:

    -rw-r--r--. 1 evgeni evgeni 787K Apr 24 11:34 theforeman-foreman-3.11.0-dev.tar.gz

edit: after removing apidoc fixtures:

    -rw-r--r--. 1 evgeni evgeni 440K Apr 25 14:24 theforeman-foreman-3.11.0-dev.tar.gz

440K is OK, IMHO -- there are collections with a single module being 1.4MB big (and the whole tarball is 2.4MB, compressed)

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -8,8 +8,9 @@ ROLES := $(wildcard roles/*)
 PLUGIN_TYPES := $(filter-out __%,$(notdir $(wildcard plugins/*)))
 RUNTIME_YML := meta/runtime.yml
 METADATA := galaxy.yml LICENSE README.md $(RUNTIME_YML) requirements.txt changelogs/changelog.yaml CHANGELOG.rst bindep.txt PSF-license.txt meta/execution-environment.yml
+TESTDATA := Makefile pytest.ini $(shell find tests/ ! -type d ! -path '*/__pycache__/*' ! -path '*/test_playbooks/fixtures/*' ! -path '*/fixtures/apidoc/*')
 $(foreach PLUGIN_TYPE,$(PLUGIN_TYPES),$(eval _$(PLUGIN_TYPE) := $(filter-out %__init__.py,$(wildcard plugins/$(PLUGIN_TYPE)/*.py)) $(wildcard plugins/$(PLUGIN_TYPE)/*.yml)))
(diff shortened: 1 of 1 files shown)


Makefile
```diff
@@ -8,8 +8,9 @@ ROLES := $(wildcard roles/*)
 PLUGIN_TYPES := $(filter-out __%,$(notdir $(wildcard plugins/*)))
 RUNTIME_YML := meta/runtime.yml
 METADATA := galaxy.yml LICENSE README.md $(RUNTIME_YML) requirements.txt changelogs/changelog.yaml CHANGELOG.rst bindep.txt PSF-license.txt meta/execution-environment.yml
+TESTDATA := Makefile pytest.ini $(shell find tests/ ! -type d ! -path '*/__pycache__/*' ! -path '*/test_playbooks/fixtures/*')
```
evgeni: the file blows up to 3.5MB if I do not exclude these fixtures

mdellweg: Sounds a lot like `wildcard`, `notdir` and `filterout` to me... ;)

evgeni: yeah, but at what price? :)

mdellweg:
Is this a the ask to "ship the tests with the product"? With live-testing, I guess the recorded fixtures are not important...

Are people aware that it is a bad idea running these tests on a production system?

evgeni:
> Is this a the ask to "ship the tests with the product"? With live-testing, I guess the recorded fixtures are not important...

The ask/idea is to be able to say "here is the collection and you can call `make test` (against a throw away) system to ensure the collection is compatible with your system"

> Are people aware that it is a bad idea running these tests on a production system?

Depends on the "people".
The ones asking? Yes, absolutely.
Everyone else, maybe, maybe not. Given just installing the collection won't give you an environment that you can immediately destroy (there is no `vars/server.yml` with the right credentials, and several dependencies are still missing), I think it's safe tho

Griffin-Sullivan (approved):

ehelms (approved):

PR #1600, opened 2023-04-25
assume modern ansible in tests if detection failed

evgeni:
the way we detect the ansible version fails if we use the system-installed ansible but pytest is running in a venv, so let's just assume we have a modern (2.14 at the time of writing) ansible in this case

### DIFF
--- a/tests/conftest.py
+++ b/tests/conftest.py
@@ -103,7 +103,7 @@ def run_playbook_vcr(tmpdir, module, extra_vars=None, limit=None, inventory=None
 
 
 def get_ansible_version():
-    ansible_version = None
+    ansible_version = '2.14.0'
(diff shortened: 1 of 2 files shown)


evgeni:
the alternative is to parse the output of `ansible --version`, but that can be tricky and I don't see a good reason for doing so?

```console
% ansible --version
ansible 2.9.27

% ansible --version            
ansible 2.10.17

% ansible --version                
ansible [core 2.11.12] 

% ansible --version
ansible [core 2.12.10]

% ansible --version                
ansible [core 2.13.9]

% ansible --version                  
ansible [core 2.14.5]

% ansible --version 2>/dev/null
ansible [core 2.16.0.dev0]
```

ehelms (approved):

PR #1601, opened 2023-04-25
explicitly set login_text to '' in setting info test

evgeni:
we assert the value, but recent Foreman versions changed the default

### DIFF
--- a/tests/test_playbooks/setting_info.yml
+++ b/tests/test_playbooks/setting_info.yml
@@ -7,6 +7,9 @@
     - vars/server.yml
   tasks:
     - include_tasks: tasks/setting.yml
+      vars:
+        setting_name: 'login_text'
+        setting_value: ''
 
 - hosts: tests
   collections:


tests/test_playbooks/setting_info.yml
```diff
@@ -7,6 +7,8 @@
     - vars/server.yml
   tasks:
     - include_tasks: tasks/setting.yml
+      vars:
+        setting_value: ''
```
ekohl, line 12: I think it's confusing that we set the setting name in [tasks/setting.yml](https://github.com/theforeman/foreman-ansible-modules/blob/85698bc686e4a8301a889c54a976788423459774/tests/test_playbooks/tasks/setting.yml#L4) but the value here. Is that just me?

evgeni: No, but I guess you're right and we could be explicit here too.

ekohl (approved):

PR #1610, opened 2023-05-24
Add support for deb content view filters

lumarel:
Currently it is not possible to create deb filters with the content_view_filter module,
so this changes that.
And also makes content_view_filter_rule correctly understand the deb filters.

Just to make sure, I dug myself through the code to find out if I make the correct, and also all changes here, so it could possibly missing something, which I just didn't see 🙂 
As well, should I create test cases for this? I didn't see a content view test where a deb repo is used, or if there is I just overlooked it.

I saw that there is another PR open that might change the behavior of the whole module as a whole, if this is a unnecessary change here then, I'm sorry for the noise 🙂 

Thanks Lukas

### DIFF
--- a/plugins/modules/content_view_filter.py
+++ b/plugins/modules/content_view_filter.py
@@ -79,6 +79,7 @@ options:
       - erratum
       - docker
       - modulemd
+      - deb
     type: str
   rule_name:
     description:
@@ -249,7 +250,7 @@ def main():
             inclusion=dict(type='bool', default=False),
             original_packages=dict(type='bool'),
             content_view=dict(type='entity', scope=['organization'], required=True),
-            filter_type=dict(required=True, choices=['rpm', 'package_group', 'erratum', 'docker', 'modulemd']),
+            filter_type=dict(required=True, choices=['rpm', 'package_group', 'erratum', 'docker', 'modulemd', 'deb']),
             filter_state=dict(default='present', choices=['present', 'absent']),
             rule_state=dict(default='present', choices=['present', 'absent']),
             rule_name=dict(aliases=['package_name', 'package_group', 'tag']),
@@ -297,7 +298,7 @@ def main():
             foreman_spec=content_filter_spec,
         )
 
-        if content_view_filter is not None and module.foreman_params['filter_type'] not in ['modulemd']:
+        if content_view_filter is not None and module.foreman_params['filter_type'] not in ['modulemd', 'deb']:
             cv_filter_scope = {'content_view_filter_id': content_view_filter['id']}
             if 'errata_id' in module.foreman_params:
                 # should we try to find the errata the user is asking for? or just pass it blindly?
--- a/tests/test_playbooks/content_view_filter.yml
+++ b/tests/test_playbooks/content_view_filter.yml
@@ -6,36 +6,69 @@
   vars_files:
     - vars/server.yml
   tasks:
-    - include_tasks: tasks/organization.yml
+    - name: "Create Test Organization"
+      ansible.builtin.include_tasks: tasks/organization.yml
       vars:
         organization_state: present
-    - include_tasks: tasks/product.yml
+
+    - name: "Create Test Product"
+      ansible.builtin.include_tasks: tasks/product.yml
       vars:
         product_state: present
-    - include_tasks: tasks/repository.yml
+
+    - name: "Create Test Repository"
+      ansible.builtin.include_tasks: tasks/repository.yml
       vars:
         repository_state: present
-    - include_tasks: tasks/repository.yml
+
+    - name: "Create Addititional Repository"
+      ansible.builtin.include_tasks: tasks/repository.yml
       vars:
         repository_state: present
         repository_url: "https://repos.fedorapeople.org/pulp/pulp/demo_repos/zoo/"
-    - include_tasks: tasks/repository.yml
+
+    - name: "Create Container repository"
+      ansible.builtin.include_tasks: tasks/repository.yml
       vars:
         repository_state: present
         repository_name: Test Docker Repository
         repository_content_type: docker
-        repository_url: https://registry.hub.docker.com
-        repository_docker_upstream_name: library/busybox
-    - include_tasks: tasks/katello_sync.yml
-    - include_tasks: tasks/katello_sync.yml
+        repository_url: https://quay.io
+        repository_docker_upstream_name: quay/busybox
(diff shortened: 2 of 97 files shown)


plugins/modules/content_view_filter.py
```diff
@@ -235,6 +236,12 @@
     'rule_name': {'flat_name': 'name'},
 }
 
+content_filter_rule_deb_spec = {
+    'id': {},
+    'rule_name': {'flat_name': 'name'},
+    'architecture': {},
+}
+
```
evgeni: this is not necessary if you add `deb` to the list in https://github.com/theforeman/foreman-ansible-modules/blob/1727958add1324dd8ee8a0c796806946aed7ae69/plugins/modules/content_view_filter.py#L300

this will enforce the creation of rules via the new `content_view_filter_rule` module, which is the right way ™

evgeni:
if you rebase this on latest `develop`, CI should pass again.

shall we add tests for that? ;-)

lumarel:
Sure, here is the rebase 🙂 
Tbh I assumed it's necessary, because the `tests/test_playbooks/content_view_filter.*\.yml` tests looked like they have tests for all other filter types. But I didn't go down the rabbit hole to understand how to extend these tests correctly.

plugins/modules/content_view_filter.py
lumarel: Okay understood 🙂

evgeni:
Do you *want to* understand? ;-)
(feel free to say no, and I'll merge this as is, and add tests myself later)

lumarel:
Yes I do want to understand, I just normally take the time to play around with that stuff to learn, so don't really hold of PRs because of learning time :)

lumarel:
@evgeni I think I got it working 🙂 

Only to confirm, I setup a Foreman testsystem on Rocky 8 (I chose the 3.7RC2, so yay bleeding edge, if I should use 3.6, also fine just need to scrap and need to rerun the installer, no harm done, `foreman-installer --scenario katello --enable-foreman-cli-ansible --enable-foreman-cli-puppet --enable-foreman-compute-vmware --enable-foreman-plugin-ansible --enable-foreman-plugin-puppet --enable-foreman-plugin-remote-execution --enable-foreman-plugin-scc-manager --enable-foreman-proxy-plugin-ansible --enable-foreman-proxy-plugin-remote-execution-script --enable-puppet --foreman-initial-organization "Test Organization" --foreman-initial-location "Test Location" --foreman-initial-admin-username admin --foreman-initial-admin-password changeme --foreman-proxy-dhcp false --foreman-proxy-dns false --foreman-proxy-tftp false`),
and a seperate box for the ansible runs on Rocky 9 (`dnf install python3-pip git docker-ce make sshpass; python3 -m pip install ansible-core pytest pytest-vcr; git clone <repo>; cd <repo>; make setup`).
Aaand created the correct server.yml for my environment!

And then I ran the record with: `make record_content_view_filter`
and test with: `make test_content_view_filter`

I also tried to fix the _info, _rules, _rules_info tests, but I can't figure out how the `ansible.builtin.include_tasks: tasks/content_view_filter.yml` works, because it doesn't for me. (or better said, it does not exist, I know what it should create, so could create it, but hmm maybe I'm understanding something wrong)

tests/test_playbooks/tasks/repository.yml
```diff
@@ -23,7 +23,7 @@
     upstream_username: "{{ repository_upstream_username | default(omit) }}"
     upstream_password: "{{ repository_upstream_password | default(omit) }}"
     deb_releases: "{{ repository_deb_releases | default(omit) }}"
-    deb_components: "{{ repository_deb_componentes | default(omit) }}"
+    deb_components: "{{ repository_deb_components | default(omit) }}"
```
lumarel, line 26: I noticed that this variable had a typo
Had to update the repository_deb fixtures due to this 🙂

tests/test_playbooks/content_view_filter.yml
```diff
@@ -6,36 +6,69 @@
...
+      ansible.builtin.include_tasks: tasks/repository.yml
       vars:
         repository_state: present
         repository_name: Test Docker Repository
         repository_content_type: docker
-        repository_url: https://registry.hub.docker.com
-        repository_docker_upstream_name: library/busybox
-    - include_tasks: tasks/katello_sync.yml
-    - include_tasks: tasks/katello_sync.yml
+        repository_url: https://quay.io
```
lumarel, line 36: Switched to quay here because Docker Hub limited their api calls, and that breaks it for me 🙂

tests/test_playbooks/content_view_filter.yml
```diff
@@ -6,36 +6,69 @@
...
+
+    - name: "Create Deb Repository"
+      ansible.builtin.include_tasks: tasks/repository.yml
+      vars:
+        repository_state: present
+        repository_name: Test Debian Repository
+        repository_content_type: deb
+        repository_url: https://ftp.debian.org/debian
+        repository_deb_releases: buster
+        repository_deb_components: contrib
```
lumarel, line 47: I saw that in other tests main is used, but I didn't want to use it here because it's so many packages, takes ages to download.

lumarel:
Rebuilt everything, all 4 tests are working now 👍🏻

tests/test_playbooks/content_view_filter_info.yml
```diff
@@ -68,9 +84,13 @@
             product: "Test Product"
           - name: "Test Modular Repository"
             product: "Test Product"
+          - name: "Test Debian Repository"
+            product: "Test Product"
 
     - name: "Create content view filters"
       ansible.builtin.include_tasks: tasks/content_view_filter.yml
+      vars:
+        expected_change: true
```
evgeni:
```suggestion
```
no need to assert change in the setup step

evgeni (approved):
Tiny comment, will just apply the suggested change myself

ACK

tests/test_playbooks/content_view_filter.yml
evgeni: you could try explicitly setting `repository_download_policy: on_demand`, as I think Katello defaults to full downloads for non-RH repos?

tests/test_playbooks/content_view_filter.yml
evgeni: yeah... I've heard people complaining about that

PR #1620, opened 2023-06-02
don't log repositories content when creating products

evgeni:


evgeni:
before:
```
TASK [theforeman.foreman.repositories : Create Products] ******************************************************************************************************************************************************************************
changed: [localhost] => (item={'name': 'Ansible', 'repositories': [{'name': 'Red Hat Collections', 'content_type': 'ansible_collection', 'url': 'https://console.redhat.com/api/automation-hub/content/540155-synclist/', 'ansible_collection_requirements': '---\ncollections:\n- redhat.satellite\n', 'upstream_username': 'lol', 'upstream_password': 'satellite'}]})
```

after:
```
TASK [theforeman.foreman.repositories : Create Products] ******************************************************************************************************************************************************************************
changed: [localhost] => (item={'name': 'Ansible', 'repositories': '[FILTERED]'})
```

roles/repositories/tasks/main.yml
```diff
@@ -43,7 +43,7 @@
     gpg_key: "{{ item.gpg_key | default(omit) }}"
     state: present
   with_items:
-    - "{{ foreman_products | selectattr('repositories', 'defined') | list }}"
+    - "{{ foreman_products | selectattr('repositories', 'defined') | map('combine', {'repositories': '[FILTERED]'})| list }}"
```
evgeni: the alternative is to unconditionally set `no_log: true`, as "finding out whether any of the defined repos has a password set" is way to complicated in this step (it's easy in the next one, where we actually iterate over the repos).

ehelms (approved):

PR #1621, opened 2023-06-07
Add new parameters to job_invocation

lumarel:
This adds the recurrance.purpose and description_format parameters, as they are not exposed in the module up to now, cross checked with the api docs, how they need to be added here 🙂 

I don't see any need for test changes here, if wanted I can of course extend the job_invocation tests!

(I saw that the `make info` command didn't treat the line breaks for me, so I also added this here, feel free to let me remove these if not wanted)


evgeni (approved):

evgeni:
Danke!

PR #1624, opened 2023-06-08
Convert2RHEL - Sync repos before CV publish

stejskalleos:


stejskalleos:
Can anybody take a look at the review? I have no idea how to fix the tests

stejskalleos:
@evgeni

evgeni:
The tests are failing, because you've recorded them on a 4.9 Katello, but we still have an older Katello apidoc stored in the repo, so that it generates slightly different requests than what you recorded. Let me see if I can fix that by updating the katello apidoc :)

evgeni:
okay, the last bit is that it generates a warning as you set mirror_on_sync... lets see

evgeni (approved):

PR #1625, opened 2023-06-08
Plugin content_view_version timeout increase

jjansky1:
In specific content views which are using filters publish or promote may exceed task_timeout. Proposing update from 60 * 60 to 180 * 60.

### DIFF
--- a/plugins/modules/content_view_version.py
+++ b/plugins/modules/content_view_version.py
@@ -200,7 +200,7 @@ def main():
         mutually_exclusive=[['current_lifecycle_environment', 'version']],
     )
 
-    module.task_timeout = 60 * 60
+    module.task_timeout = 180 * 60
 
     if 'version' in module.foreman_params and not re.match(r'^\d+\.\d+$', module.foreman_params['version']):
         try:


evgeni:
What on earth makes the publish action take 3h?! I mean, we can surely change the timeout, but that feels like papering over an actual issue.

evgeni:
ping @jjansky1

ShitovAndrey:
@evgeni 
Hello.
ContentView with nine Oracle Linux 8 repositories, about 70,000 packages and filters published in more than 1 hour.

evgeni:
> @evgeni Hello. ContentView with nine Oracle Linux 8 repositories, about 70,000 packages and filters published in more than 1 hour.

My Oracle CV only has 54000 packages, but the publish takes 2 minutes (on a VM with 16 cores and 32G RAM).
It really shouldn't take >1h with a few more packages.

Can you start a thread on https://community.theforeman.org describing your setup (hw and sw) and tagging me (`@evgeni`) so someone can look into this?

evgeni (approved):
Going to merge this, but I still think it's a band aid

PR #1627, opened 2023-06-14
run ansible-lint in offline mode

evgeni:
avoids things like https://github.com/ansible/ansible-lint/issues/3560


evgeni:
lint passed :see_no_evil:

PR #1629, opened 2023-06-21
update default Python versions for CI

evgeni:
- run tests by default with 3.10, as core is dropping 3.9 support soon
- add 3.9 as an additional test with core 2.11
- run all other steps with 3.11, as that's the latest supported one


evgeni:
of course, GHA dropped Python 2.7 now…

PR #1630, opened 2023-06-21
run Python 2.7 in a container

evgeni:


evgeni:
Right. Starting a container from inside a container is a bit funny. :disappointed:

evgeni:
going to pretend this was reviewed :see_no_evil:

PR #1632, opened 2023-06-22
Job template

pipopopo:
- The with_fileglob wasn't correctly indented so it wasn't looping over the files
- The other example didn't include the name parameter which will result in ` "msg": "No name specified and no filename to infer it."}`

### DIFF
--- a/plugins/modules/job_template.py
+++ b/plugins/modules/job_template.py
@@ -254,7 +254,7 @@ EXAMPLES = '''
     password: "changeme"
     state: present
     template: '{{ lookup("file", item.src) }}'
-    name: '{{ item }}'
+    name: '{{ item.path }}'
   with_filetree: '/path/to/job/templates'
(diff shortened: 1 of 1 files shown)


plugins/modules/job_template.py
```diff
@@ -254,6 +254,7 @@
     password: "changeme"
     state: present
     template: '{{ lookup("file", item.src) }}'
+    name: '{{ item }}'
```
evgeni: Shouldn't this be `item.path` or something? `item` would be the whole dict as returned by `with_fileglob`

pipopopo: Yeah, if the path is too long you'll get a super long name and mostly you'll only be interested in using the filename so item.path would be cleaner.

pipopopo: Modified it

evgeni (approved):

PR #1647, opened 2023-07-20
feat(wait_task): Add a wait_task module.

### LINKED ISSUE #1609: When publishing a ContentViewVersion to a LifeCycleEnvironment it would be helpful to have a way to wait for SmartProxy/Capsule synchronization to finish
##### SUMMARY
In a use-case where we use a load-balanced setup of SmartProxies to distribute content, we want to use a single playbook to manage the ContentViewVersion promotion and the update of the machines to have a consistent and reproducible way to perform updates on clusters.
This way we can perform all preparations for update, safety checks, and reboots in the playbook, and not need manual validation that the SmartProxies are synchronized.

However, since the Capsule Synchronization task is asynchronous to ContentViewVersion Promotions.

##### ISSUE TYPE
Currently there is no "native" way to chain ContentViewVersion Promotion and updating of the machines automatically when SmartProxies are involved.

A helper action to wait for completion of SmartProxy synchronization (or a generic way to wait for Foreman Task(s) completion) would help greatly in that case. It is involved to wait this in pure ansible, so I personnally think that it makes sense to have it in this repository.

JGodin-C2C:
# What : 
And a new module that simply expose the wait_for_task function
This allows to wait for a asynchronous task to finish before continuing to apply the playbook. 

# How 
It simply expose the function "wait_for_task" that was already available as a helper. 

# Related

fixes #1609 

# tests 

Have been tested with a custom/minimal playbook on an internal satellite server.

### DIFF
--- a/meta/runtime.yml
+++ b/meta/runtime.yml
@@ -79,6 +79,7 @@ action_groups:
   - templates_import
   - user
   - usergroup
+  - wait_for_task
 plugin_routing:
   modules:
     foreman_architecture:
--- a/plugins/modules/wait_for_task.py
+++ b/plugins/modules/wait_for_task.py
@@ -1,0 +1,94 @@
+#!/usr/bin/python
+# -*- coding: utf-8 -*-
+# (c) 2023, Julien Godin <julien.godin@camptocamp.com>
+#
+# This program is free software: you can redistribute it and/or modify
+# it under the terms of the GNU General Public License as published by
+# the Free Software Foundation, either version 3 of the License, or
+# (at your option) any later version.
+#
+# This program is distributed in the hope that it will be useful,
+# but WITHOUT ANY WARRANTY; without even the implied warranty of
+# MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+# GNU General Public License for more details.
+#
+# You should have received a copy of the GNU General Public License
+# along with this program.  If not, see <http://www.gnu.org/licenses/>.
+
+from __future__ import absolute_import, division, print_function
+__metaclass__ = type
+
+DOCUMENTATION = '''
+---
+module: wait_for_task
+version_added: 3.13.0
+short_description: Wait for a task
+description:
+  - Wait for a task to finish
+author:
+  - "Julien Godin (@JGodin-C2C)"
+options:
+  task:
+    description:
+      - Task id to wait for.
+    required: true
+    type: str
+  timeout:
+    description:
+      - How much time the task should take to be finished
+    required: false
+    type: int
+    default: 60
+extends_documentation_fragment:
+  - theforeman.foreman.foreman
+'''
+
+
+EXAMPLES = '''
+- name: wait for a task to finish in at least 10 seconds
+  theforeman.foreman.wait_for_task:
+    server_url:  "https://foreman.example.com"
+    password: changeme
+    username: admin
+    task: "{{ item }}"
+    timeout: 60
+  loop: "{{ tasks.resources }}"
+
+'''
+
+RETURN = '''
+entity:
+  description: Designated task is finished
+  returned: success
+  type: dict
+  contains:
+    task:
+        description: The finished task
+        type: dict
+'''
+
+
+from ansible_collections.theforeman.foreman.plugins.module_utils.foreman_helper import ForemanAnsibleModule
+
+
+class ForemanWaitForTask(ForemanAnsibleModule):
+    pass
+
+
+def main():
+    module = ForemanWaitForTask(
+        argument_spec=dict(
+            task=dict(type="str", required=True),
+            timeout=dict(type="int", required=False, default=60),
+        )
+    )
+    module.task_timeout = module.foreman_params["timeout"]
+    with module.api_connection():
+
+        task = module.wait_for_task(module.show_resource(
+            'foreman_tasks', module.foreman_params["task"]))
+        module.exit_json(task=task, task_id=task['id'])
+
+
+if __name__ == "__main__":
+    main()
--- a/galaxy.yml
+++ b/galaxy.yml
@@ -93,6 +93,7 @@ authors:
   - "metalcated <mike.gomon@gmail.com>"
   - "russianguppie <46544650+russianguppie@users.noreply.github.com>"
   - "willtome <willtome@gmail.com>"
+  - "Julien Godin <julien.godin@camptocamp.com>"
 version: "3.13.0-dev"
 license:
   - "GPL-3.0-or-later"
--- a/tests/fixtures/apidoc/wait_for_task.json
+++ b/tests/fixtures/apidoc/wait_for_task.json
@@ -1,0 +1,1 @@
+foreman.json
--- a/tests/test_playbooks/tasks/wait_for_task.yml
+++ b/tests/test_playbooks/tasks/wait_for_task.yml
@@ -1,0 +1,59 @@
+---
+- name: "Create katello content view version"
+  vars:
+    - content_view_name: "Test Content View"
+    - organization_name: "Test Organization"
+    - lifecycle_environments: Library
+  content_view_version:
+    username: "{{ foreman_username }}"
+    password: "{{ foreman_password }}"
+    server_url: "{{ foreman_server_url }}"
+    validate_certs: "{{ foreman_validate_certs }}"
+    content_view: "{{ content_view_name }}"
+    organization: "{{ organization_name }}"
+    description: "{{ description | default(omit)}}"
+    version: "{{ version | default(omit)}}"
+    lifecycle_environments: "{{ lifecycle_environments }}"
+    force_promote: "{{ force_promote | default(omit) }}"
+    force_yum_metadata_regeneration: "{{ force_yum_metadata_regeneration | default(omit) }}"
+    current_lifecycle_environment: "{{ current_lifecycle_environment | default(omit) }}"
+    state: "{{ state | default(omit) }}"
+  async: 9999
+
+- name: Wait for API to react
+  ansible.builtin.pause:
+    seconds: 10
+
+- name: "Search for previously created task"
+  resource_info:
+    username: "{{ foreman_username }}"
+    password: "{{ foreman_password }}"
+    server_url: "{{ foreman_server_url }}"
+    validate_certs: "{{ foreman_validate_certs }}"
+    resource: foreman_tasks
+    search: "(label = Actions::Katello::ContentView::Publish and state = running)"
+
+- name: wait for the task to finish
+  wait_for_task:
+    username: "{{ foreman_username }}"
+    password: "{{ foreman_password }}"
+    server_url: "{{ foreman_server_url }}"
+    validate_certs: "{{ foreman_validate_certs }}"
+    task: "{{ item }}"
+    timeout: 900
+  loop: "{{ tasks.resources }}"
+
+
+- name: "Search for previously created task"
+  resource_info:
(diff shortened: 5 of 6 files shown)


JGodin-C2C:
@evgeni Failing tests does not seems to be part of my PR.

evgeni:
They are, or at least partially.

Have a look at the 3.10/2.12 sanity failures here: https://github.com/theforeman/foreman-ansible-modules/actions/runs/5612522667/job/15206459221#step:13:1

(The newer ansible versions fail for some reason, but *that* is certainly unrelated)

I won't have time to properly review this PR before next week, but wanted to ask if you considered naming the module "wait_for_task" which (IMHO) would better fit with other wait_for_… modules that exist in Ansible core.

JGodin-C2C:
A ! i see, the problem is WAY above the actual failure. I will fix. 
Also, i can totally rename it "wait_for_task"

JGodin-C2C:
Seems much better.

CHANGELOG.rst
```diff
@@ -6,6 +6,14 @@ theforeman.foreman Release Notes
 
 This changelog describes changes after version 0.8.1.
 
+v3.13.0
+=======
+
+Minor Changes
+-------------
+
+- Add a wait for task module
```
evgeni: This (and the change to `changelogs/changelog.yaml`) are not necessary, they will be automatically generated when we do the release.

meta/runtime.yml
```diff
@@ -156,7 +157,7 @@ plugin_routing:
     foreman_user:
       redirect: theforeman.foreman.user
     foreman_usergroup:
-      redirect: theforeman.foreman.usergroup
+      redirect: theforeman.foreman.wait_for_task
```
evgeni: that seems wrong :)

plugins/modules/wait_for_task.py
```diff
@@ -0,0 +1,92 @@
...
+description:
+  - Wait for a task to finish
+author:
+  - "Julien Godin (@JGodin-C2C)"
+options:
+  task:
+    description:
+      - Task to wait for. Typical output of the resource_info module
+    required: true
+    type: dict
```
evgeni: I wonder if this should really be a dict like returned by `resource_info` vs a plain string with the ID, which we then feed like 
```python
module.wait_for_task({'id': module.foreman_params['task']})
```

@mdellweg @ehelms opinions?

plugins/modules/wait_for_task.py
```diff
@@ -0,0 +1,92 @@
...
+def main():
+    module = ForemanWaitForTask(
+        argument_spec=dict(
+            task=dict(type="dict", required=True),
+            timeout=dict(type="int", required=False, default=60),
+        )
+    )
+    module.task_timeout = module.foreman_params["timeout"]
+    with module.api_connection():
+        module.wait_for_task(module.foreman_params["task"])
```
evgeni:
```suggestion
        task = module.wait_for_task(module.foreman_params["task"])
        module.exit_json(task=task)
```

that way we return the task details so users can take further actions based on the finished task result

JGodin-C2C: Ok, but i think it is smarter to return the same type as the one passed to this module. Waiting for the other comment to be resolved.

evgeni: We can also do `module.exit_json(task=task, task_id=task['id'])` to have both directly available?

My trail of thought was that:
- users might want to have the additional data here, because later steps might depend on details of the finished task
- users might want to start waiting based on id only (as that's really the only data we need) because they get the ID by some other means than `resource_info`

plugins/modules/wait_for_task.py
evgeni: technically, I think we could even accept both via `type: raw` and then decide in code how to call `wait_for_task` 

(personally, I wouldn't, and would do just string, but wanted to highlight possibilities)

ehelms: Do we expect this module to mostly be used in conjunction with output from our other modules or do we expect a 50/50 of it being used "independently" and in conjunction?  If mostly in conjunction, I would err towards a dict.

JGodin-C2C: I thought about this (leaving the decision to the code how to call `wait_for_task`), but not sure if that is a good idea. any additional input/opinion  would be great IMHO.

plugins/modules/wait_for_task.py
JGodin-C2C: Is returning both task AND task["id"] useful thought ? I mean, the ID is already in the task that just finished. 

I think the best return value would be the task, as you mentionned, users may need additional info.

meta/runtime.yml
JGodin-C2C: Corrected

CHANGELOG.rst
JGodin-C2C: Removed

plugins/modules/wait_for_task.py
JGodin-C2C: Hey @ehelms sorry, didnt see your comment. 
Honestly , i do not think this module will be used a lot independently. It could cover some corner cases. But i think it will mostly be used in conjunction with other modules.

meta/runtime.yml
mdellweg: I don't think we even need this, because we never had a foreman* version of this module we now need to alias for.

plugins/modules/wait_for_task.py
mdellweg: What would it be? The task id? Its url? I tend towards a short identifier. If you come along with the output of a resource_info call, you can still use something like "result.task.href", no?

plugins/modules/wait_for_task.py
```diff
@@ -0,0 +1,93 @@
...
+def main():
+    module = ForemanWaitForTask(
+        argument_spec=dict(
+            task=dict(type="dict", required=True),
+            timeout=dict(type="int", required=False, default=60),
+        )
+    )
+    module.task_timeout = module.foreman_params["timeout"]
+    with module.api_connection():
+        task = module.wait_for_task(module.foreman_params["task"])
```
mdellweg: If the task we wait on failed, this module will fail too, right?
(I think that is the right way to handle it.)

plugins/modules/wait_for_task.py
JGodin-C2C: No, it will simply be a task object.

mdellweg: So let's say, all i have in my playbook is the task href, do i need to fake a dict to make this module work, while all the module will ever look at is the id/href inside the dict? That sounds like a strange interface.

mdellweg: Even more so, we need to add a weird state attribute in that dict, and `wait_for_tasks` will try to interpret that. This sounds all wrong to me. All we should pass to the module is the id. And the module should first query the task state from the server.

JGodin-C2C: Ok. Thats two peoples in favor of the id as a parameter. I will implement.

plugins/modules/wait_for_task.py
mdellweg: So you could use `result.task.id` for another call to this module, right?
In a way, this makes this module like a `task_info` module that also waits on the task first. (Not saying this is bad.)

plugins/modules/wait_for_task.py
evgeni: Yeah, `module.wait_for_task` fails in that case:
https://github.com/theforeman/foreman-ansible-modules/blob/845116fdefcc4e9dac8fc34ed35782eee8598ef7/plugins/module_utils/foreman_helper.py#L1198-L1199

meta/runtime.yml
evgeni: technically, that is correct, a new module doesn't need a mapping entry for the old name.

plugins/modules/wait_for_task.py
JGodin-C2C: Okay ! 
It now uses the ID of the task and output the id + the task itself.

plugins/modules/wait_for_task.py
JGodin-C2C: Done, it now return both values.

plugins/modules/wait_for_task.py
JGodin-C2C: thanks @evgeni .

meta/runtime.yml
JGodin-C2C: @evgeni Yes, it has been totally removed in my latest push :smiley:

plugins/modules/wait_for_task.py
```diff
@@ -0,0 +1,95 @@
...
+    task:
+        description: The finished task
+        type: dict
+'''
+
+
+from ansible_collections.theforeman.foreman.plugins.module_utils.foreman_helper import ForemanAnsibleModule
+
+
+
```
evgeni: `E303: too many blank lines (3)` -- linters can be a tad too picky at times :crying_cat_face: 
```suggestion
```

JGodin-C2C: na, my bad, i didnt run the linters after my update.

JGodin-C2C:
All green ! :muscle:

plugins/modules/wait_for_task.py
```diff
@@ -0,0 +1,94 @@
...
+class ForemanWaitForTask(ForemanAnsibleModule):
+    pass
+
+
+def main():
+    module = ForemanWaitForTask(
+        argument_spec=dict(
+            task=dict(type="str", required=True),
```
mdellweg, line 81: I'm a bit out of touch, but a faint memory nags me that we may move this into `foreman_spec` and the lookup will be done for us. Is this correct?

evgeni: it's not an entity, so nope

(I think the code is fine as is, but I can try to play around with it, but wouldn't bother Julien with that right now)

mdellweg: I'm pretty sure, `lookup_entity` works with the plain `ForemanAnsibleModule`. Or is the task not an entity in foreman is what you were trying to tell me?

https://github.com/theforeman/foreman-ansible-modules/blob/develop/plugins/module_utils/foreman_helper.py#L831

evgeni: Yeah. It's sort-of-an-entity-but-not-quite.

For example, it has no `name` (but `label` and `action`) attribute. And the only *unique* attribute is the `id` which we get passed here, but `lookup_entity` will, by default, try a search via name.
But given we already have the id, we don't need to perform a search at all (all it would give us is the same id anyway) and can just call `show_resource` on the ID directly (which we do in the code below).

mdellweg: Right! Thank you.

JGodin-C2C:
Ok, what would be the next step to see this merged ?

evgeni:
Me finding the time to finalize the review ;-)
Which I did this morning!

The only "real" thing I wanted/had to change were the tests -- I didn't like to have to do a full publish to get them to run and made them delete a product, which is also an async action we can wait on.

While at it, I also extended the examples a bit ;-)

As I couldn't push to your fork, I opened a second PR with your commit as the first one, and then my changes in mostly logical chunks: https://github.com/theforeman/foreman-ansible-modules/pull/1656

Once CI passes on that one I'll merge it (or you can pull the changes into this PR if you prefer to have the original one to be "closed as merged" -- I don't mind either way)

evgeni:
merged as https://github.com/theforeman/foreman-ansible-modules/pull/1656

PR #1649, opened 2023-07-25
don't fail when removing a content credential from a repository

evgeni:


evgeni:
This now of course fails the "no warnings" test, as it raises "The following parameters are not supported by your server when performing create on repositories: {'organization_id'}. They were ignored." :exploding_head:

evgeni:
On the other hand, why does it even fail for the content credential, but not the product :confused:

evgeni:
probably should add tests, right? :see_no_evil:

evgeni:
~~and the tests say I am stupid…~~

and the tests helped me to write the fix correctly

evgeni:
Mhh. The existing apidoc doesn't have https://projects.theforeman.org/issues/36497 yet. Makes sense.

evgeni:
@ehelms @mdellweg @Griffin-Sullivan does anyone want to look at this one? :pleading_face:

Griffin-Sullivan (approved):
LGTM. Is the `/fixtures` directory what we use for "recorded" testing?

evgeni:
> LGTM. Is the `/fixtures` directory what we use for "recorded" testing?

Yes, exactly.

plugins/module_utils/foreman_helper.py
```diff
@@ -575,6 +575,21 @@ def _patch_organization_ignore_types_api(self):
             _organization_create['params'].append(_ignore_types_param)
             _organization_update['params'].append(_ignore_types_param)
 
+    @_check_patch_needed(fixed_version='3.8.0', plugins=['katello'])
+    def _patch_products_repositories_allow_nil_credential(self):
+        """
+        This is a workaround for the missing allow_nil: true in the Products and Repositories controllers
```
ehelms, line 581: This is a in two-three releases of Katello this work around gets dropped?

ehelms (approved):

plugins/module_utils/foreman_helper.py
evgeni: We've been rather conservative in dropping those apidoc patches, but the code is a noop once it detects a new enough version.

https://github.com/theforeman/foreman-ansible-modules/blob/87e1b51886ae75b60e35c3d7f40eaca287b8febf/plugins/module_utils/foreman_helper.py#L109-L135

PR #1655, opened 2023-08-09
allow disabling/removing of repositories by setting the state parameter

evgeni:


roles/repositories/tasks/main.yml
```diff
@@ -9,7 +9,7 @@
     product: "{{ item.0.name }}"
     name: "{{ item.1.name }}"
     all_repositories: false
-    state: enabled
+    state: "{{ item.1.state | default(omit) }}"
```
evgeni, line 12: I decided to `default(omit)` here (and later) as `enabled` and `present` are the defaults anyway.

Griffin-Sullivan (approved):
Looks good. Could we add a section to the end of https://github.com/theforeman/foreman-ansible-modules/blob/develop/tests/test_playbooks/repositories_role.yml to remove the repositories.

evgeni:
We could, but I so far avoided that, as the re-recording of that role requires a manifest and I am lazy :D

PR #1656, opened 2023-08-11
feat(wait_task): Add a wait_task module

evgeni:


tests/test_playbooks/tasks/wait_for_task.yml
```diff
@@ -41,8 +39,7 @@
     validate_certs: "{{ foreman_validate_certs }}"
     task: "{{ item }}"
     timeout: 900
-  loop: "{{ tasks.resources }}"
-
+  loop: "{{ tasks.resources | map(attribute='id') | list }}"
```
mdellweg: Just to confirm, we could also have used `{{ item.id }}`?

tests/test_crud.py
```diff
@@ -42,6 +42,8 @@ def test_crud(tmpdir, module, vcrmode):
 def test_check_mode(tmpdir, module):
     if module in ['subscription_manifest', 'templates_import', 'puppetclasses_import', 'content_rhel_role']:
         pytest.skip("This module does not support check_mode.")
+    if module in ['wait_for_task']:
+        pytest.skip("The tests for this module do not support check_mode.")
```
mdellweg: This is because we can never create a task to wait on in check_mode? Is it worth supporting check_mode by this module at all?
What if we moved the "deleting a product" task out of the test section and have the running task imprinted in the fixtures already?

mdellweg (approved):

tests/test_crud.py
mdellweg: Maybe this even speeds up the tests altogether.

evgeni: it's because we can't trigger the delete "async" in check mode. (async and check mode are not allowed together in ansible)

so moving the delete to the setup phase? that could work. yeah.

tests/test_playbooks/tasks/wait_for_task.yml
evgeni: sure, but where would be the fun.

my brainfart, sorry. :)

evgeni: on the other hand, this notation makes the ansible output nicer, as it only shows the ids and not ALL the things we got from `resource_info`. (which is irrelevant in tests, but makes a nicer example in the docs?)

mdellweg: ^ That's what i thought when i was about to make this a suggestion.

mdellweg: And there is also
```yaml
loop_control:
  label: "{{ item.name }}:{{ item.id }}"
```
If you want _nice_.

evgeni: modern shit :D

tests/test_crud.py
evgeni: nice, that worked and now we can have check mode tests

PR #1659, opened 2023-08-15
add smart_class_parameter_override_value module

evgeni:


plugins/modules/smart_class_parameter_override_value.py
```diff
@@ -0,0 +1,115 @@
...
+
+class ForemanOverrideValueModule(ForemanEntityAnsibleModule):
+    pass
+
+
+def main():
+    module = ForemanOverrideValueModule(
+        foreman_spec=dict(
+            puppetclass=dict(required=True, type='entity', ensure=False),
+            smart_class_parameter=dict(required=True, type='entity', scope=['puppetclass'], search_by='parameter'),
```
evgeni: I am a bit torn on the naming of those two parameters.

We call them `puppetclass_name` and `parameter` in the `smart_class_parameter` module:
https://github.com/theforeman/foreman-ansible-modules/blob/25815553557f99010f9fe66843cbec817a7aadd5/plugins/modules/smart_class_parameter.py#L208-L209

But that seems somehow wrong? We could add aliases for those names, so copy/paste is easier?

evgeni:
the two failing checks are unrelated and should be fixed in https://github.com/theforeman/foreman-ansible-modules/pull/1662

plugins/modules/smart_class_parameter_override_value.py
Griffin-Sullivan: What seems wrong with it? I feel like the expectation for the parameters that are in both modules would be 1 to 1 naming.

Griffin-Sullivan (approved):
Everything looks great to me! I would consider changing the parameter names that you pointed out. I will follow the community thread to see how everything goes.

plugins/modules/smart_class_parameter_override_value.py
evgeni: I find the `_name` suffix odd, as we use the "name" of an entity everywhere, but this is the only place where it's part of the parameter name.

and `parameter` is - while shorter - not the name the API expects or the UI shows.

Griffin-Sullivan: I would like them to be the same or at least work interchangeably. Will aliasing help? From a UX perspective there's an expectation that they should be the same IMO.

evgeni: Yeah, aliasing would help.

I'd make both modules accept both ways to write it.

Griffin-Sullivan: Great lets do that

evgeni: done

tests/test_playbooks/fixtures/global_parameter-0.yml
```diff
@@ -133,13 +129,15 @@ interactions:
     uri: https://foreman.example.org/api/common_parameters
   response:
     body:
-      string: '{"created_at":"2020-10-16 13:53:30 UTC","updated_at":"2020-10-16 13:53:30
-        UTC","hidden_value?":false,"hidden_value":"*****","id":11,"name":"TheAnswer","parameter_type":"string","value":42}'
+      string: '{"created_at":"2023-08-24 07:13:28 UTC","updated_at":"2023-08-24 07:13:28
+        UTC","hidden_value?":false,"hidden_value":"*****","id":9,"name":"TheAnswer","parameter_type":"string","value":"42"}'
```
evgeni: funnily, it seems that Foreman actually did accept an *integer* here, while the type was explicitly set to string :upside_down_face:

evgeni:
Thanks to @areyus for having "bad" data in their Foreman and spotting the string casting issue :)

plugins/module_utils/foreman_helper.py
```diff
@@ -1755,9 +1755,7 @@ def _is_resolved(spec, what):
 # Helper for (global, operatingsystem, ...) parameters
 def parameter_value_to_str(value, parameter_type):
     """Helper to convert the value of parameters to string according to their parameter_type."""
-    if parameter_type in ['real', 'integer']:
-        parameter_string = str(value)
-    elif parameter_type in ['array', 'hash', 'yaml', 'json']:
+    if parameter_type in ['real', 'integer', 'string', 'array', 'hash', 'yaml', 'json'] and not isinstance(value, six.string_types):
```
evgeni: @mdellweg would you mind double-checking my refactoring here?

evgeni: okay, that broke at least OS tests, so it's not the same -.-

evgeni:
I'm going to back out the conversion part of this PR and merge the working "new module" one :)

PR #1660, opened 2023-08-15
document that some inventory params only work for report api inventories

evgeni:


plugins/inventory/foreman.py
```diff
@@ -108,39 +108,57 @@
         type: int
         default: 600
       want_organization:
-        description: Toggle, if true the inventory will fetch organization the host belongs to and create groupings for the same.
+        description:
+          - Toggle, if true the inventory will fetch organization the host belongs to and create groupings for the same.
+          - Only for Reporting API based inventories.
```
evgeni: If a native english speaker can come up with a better wording, I'll gladly take it ;-)

omenos: > Only applies to inventories using the Reporting API - attribute is ignored otherwise.

> Exclusive to inventories generated by the Reporting API.

Some combination of the above? Though "Only for Reporting API based inventories" is fairly understandable, the only confusion I'd see would come from not being aware of what that is. Maybe a statement in the Synopsis that explains the different APIs the plugin will use when gathering host information?

evgeni: have a look now, is the description more clear?

Griffin-Sullivan (approved):
Everything makes sense to me. I see you changed the word `Reporting API` to `Reports API`, so you could also change this variable if you wanted to `MINIMUM_FOREMAN_VERSION_FOR_REPORTING_API`

PR #1661, opened 2023-08-16
Deprecate REX time span option

adamruzicka:
A companion to https://github.com/theforeman/foreman_remote_execution/pull/834

Opening as a draft until the REX PR goes in.

### DIFF
--- a/plugins/modules/job_invocation.py
+++ b/plugins/modules/job_invocation.py
@@ -117,6 +117,7 @@ options:
       time_span:
         description:
           - Distribute tasks over given number of seconds
+          - This is removed since foreman_remote_execution-11.0.0
         type: int
       concurrency_level:
         description:


adamruzicka:
Should I also regenerate all the apidoc data and fixtures?

evgeni:
NACK-ish.
We try to support older versions too.
Can you document that the option is removed in REX version something and we'll remove it at some point in the future?
(When the apidoc doesn't expose a param that the user has set, the param won't be sent and a warning is displayed that the param is not supported)

adamruzicka:
Ah, I didn't know that, thank you. Will do

plugins/modules/job_invocation.py
```diff
@@ -117,6 +117,7 @@
       time_span:
         description:
           - Distribute tasks over given number of seconds
+          - This is deprecated with foreman_remote_execution-11.0.0
```
evgeni: deprecated or removed? the PR you linked fully removes that parameter from the API

evgeni (approved):

adamruzicka:
We got the green light, time span is going to go away

PR #1662, opened 2023-08-22
ignore `retries` keyword in callback tests

evgeni:
its value is sometimes present and sometimes not, based on ansible version


evgeni:
it was dropped in 2.16+ (see https://github.com/ansible/ansible/commit/a48feb4cfc0feb24bebc680aaec679bb13d2483b) and thus doesn't show up in the report anymore

Griffin-Sullivan (approved):

PR #1664, opened 2023-08-29
add import_only to content_view module

evgeni:


plugins/modules/content_view.py
```diff
@@ -68,6 +68,11 @@
     description:
       - Solve RPM dependencies by default on Content View publish
     type: bool
+  import_only:
+    description:
+      - Designate this Content View for importing from upstream servers only.
+    type: bool
+    version_added: 3.14.0
```
Griffin-Sullivan, line 75: What's the point of `version_added`?

Griffin-Sullivan (approved):
I've never imported a content view so the precise language is lost on me, but the code looks good.

plugins/modules/content_view.py
evgeni: It gets rendered in the docs, so the reader can know since which version they are able to use this specific parameter without checking all the docs/changelig versions

evgeni: See https://theforeman.github.io/foreman-ansible-modules/develop/plugins/content_view_module.html#parameter-import_only

PR #1668, opened 2023-09-07
ensure we test and document the lowest ansible possible

evgeni:


.github/workflows/main.yml
```diff
@@ -61,13 +61,16 @@ jobs:
         run: sudo apt-get install -y libyaml-dev rpm
       - name: Install Ansible
         run: pip install --upgrade https://github.com/ansible/ansible/archive/${{ matrix.ansible }}.tar.gz
+      - name: Ensure Jinja version is old enough for Ansible 2.9
+        run: pip install --upgrade 'Jinja2<3.1'
+        if: matrix.ansible == 'v2.9.17'
```
evgeni, line 66: only later versions on 2.9 are compatible with Jinja 3.1+, and I kinda think it's up to the user to have a working environment

meta/runtime.yml
```diff
@@ -187,4 +187,4 @@ plugin_routing:
       redirect: theforeman.foreman.sync_plan
     katello_upload:
       redirect: theforeman.foreman.content_upload
-requires_ansible: '>=2.9'
+requires_ansible: '>=2.9.17'
```
evgeni, line 190: technically, support for collections is there since 2.9.10, but collection roles are somewhat broken until .17 and this seemed like the safer boundary.

ehelms (approved):

PR #1672, opened 2023-09-26
use convert2rhel from cdn-public.redhat.com

evgeni:
thus not needing to create a dedeicated CA


evgeni:
See https://issues.redhat.com/browse/RHELC-1166

ehelms:
Looks sane to me -- is the draft status just to fix tests or is there some unknown?

evgeni:
Just the tests

ehelms (approved):

PR #1673, opened 2023-10-06
add Python 3.12 to test matrix

evgeni:


evgeni:
includes #1674 as we need a newer `pip` than we pinned to, but we can't unpin w/o getting rid of `rpm-py-installer`

evgeni:
seems vcrpy needs a fix (https://github.com/kevin1024/vcrpy/commit/69621c67fb29dedd9ece4a7bdbf50380fbe4c5ee) released for that to work

ehelms (approved):

PR #1682, opened 2023-11-16
do not set the default url for the inventory to localhost:3000

evgeni:


Griffin-Sullivan (approved):

changelogs/fragments/inventory-no-default-localhost.yml
```diff
@@ -0,0 +1,2 @@
+breaking_changes:
+  - inventory plugin - do not default to ``http://localhost:3000`` as the Foreman URL, provividing an URL is now mandatory
```
Griffin-Sullivan:
```suggestion
  - inventory plugin - do not default to ``http://localhost:3000`` as the Foreman URL, providing a URL is now mandatory
```

PR #1683, opened 2023-11-16
add registration_command module

### LINKED ISSUE #1284: ansible native way to use the global registration feature
##### SUMMARY
#1283 proposed an ansible module and basic role for generating registration commands with the global registration feature. After discussion, the author decided to withraw the role as it was not needed for their immediate goal.

I do think the role could be useful as a higher level / more user friendly interface to the module, particularly if it registers the generated commands as an ansible variable and provides some basic features like booleans `output_to_terminal`, `output_to_file` and a configurable path `registration_command_filepath` to work with that last one.

So one workflow this enables would be to have a task running on a single host (the ansible control node) to generate the registration commands for a particular type of server and write them to a file on that machine, and a follow up task running on a group of servers in the inventory to copy the contents to each one as a script and execute.

Copying a script and executing it is probably not the most "Ansible native" way of doing things (from another point of view, it's exactly the Ansible way of doing things, only at a lower level) but with registering the generated commands in an ansible variable there is maybe some possibility of parsing that and utilizing standard ansible modules to perform equivalents of each piece... at that point, maybe we are better off giving the registration feature itself the option to generate native, well written ansible, or at least an ansible data structure to use with a pre-written role, and that could be the preferred way for ansible users to interact with this feature.

An aim of this collection is that anything that most anything that can be done with Katello WebUI, hammer, or API, should be also doable with ansible in a way that conforms to sane ansible practices. What that might look like for this particular feature is imo an interesting topic that deserves further discussion.

##### ISSUE TYPE
 - Feature Idea

evgeni:
Fixes: #1284

### DIFF
--- a/plugins/modules/registration_command.py
+++ b/plugins/modules/registration_command.py
@@ -1,0 +1,205 @@
+#!/usr/bin/python
+# -*- coding: utf-8 -*-
+# (c) Evgeni Golov
+#
+# This program is free software: you can redistribute it and/or modify
+# it under the terms of the GNU General Public License as published by
+# the Free Software Foundation, either version 3 of the License, or
+# (at your option) any later version.
+#
+# This program is distributed in the hope that it will be useful,
+# but WITHOUT ANY WARRANTY; without even the implied warranty of
+# MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+# GNU General Public License for more details.
+#
+# You should have received a copy of the GNU General Public License
+# along with this program.  If not, see <http://www.gnu.org/licenses/>.
+
+from __future__ import absolute_import, division, print_function
+__metaclass__ = type
+
+
+DOCUMENTATION = '''
+---
+module: registration_command
+version_added: 4.0.0
+short_description: Manage Registration Command
+description:
+  - Manage Registration Command
+author:
+  - "Evgeni Golov (@evgeni)"
+options:
+  activation_keys:
+    description:
+    - Activation keys for subscription-manager client, required for CentOS and Red Hat
+      Enterprise Linux.
+    - Required only if host group has no activation keys.
+    required: false
+    type: list
+    elements: str
+  force:
+    description:
+    - "Clear any previous registration and run C(subscription-manager) with C(--force)."
+    required: false
+    type: bool
+  hostgroup:
+    description:
+    - Host group to register the host in.
+    required: false
+    type: str
+  ignore_subman_errors:
+    description:
+    - Ignore C(subscription-manager) errors for C(subscription-manager register) command.
+    required: false
+    type: bool
+  insecure:
+    description:
+    - Enable insecure argument for the initial C(curl).
+    required: false
+    type: bool
+  jwt_expiration:
+    description:
+    - Expiration of the authorization token (in hours).
+    required: false
+    type: int
+  lifecycle_environment:
+    description:
+    - Lifecycle environment for the host.
+    required: false
+    type: str
+  operatingsystem:
+    description:
+    - Operating System to register the host in.
+    - Operating system must have a C(host_init_config) template assigned.
+    required: false
+    type: str
+  packages:
+    description:
+    - Packages to install on the host when registered.
+    - Multiple packages are to be given as a space delimited string.
+    required: false
+    type: str
+  remote_execution_interface:
+    description:
+    - Identifier of the Host interface for Remote execution.
+    required: false
+    type: str
+  repo:
+    description:
+    - Repository URL (yum/dnf) or full sources.list entry (apt).
+    required: false
+    type: str
+  repo_gpg_key_url:
+    description:
+    - URL of the GPG key for the repository.
+    required: false
+    type: str
+  setup_insights:
+    description:
+    - If this is set to C(true), C(insights-client) will be installed
+      and registered on Red Hat family operating systems.
+    required: false
+    type: bool
+  setup_remote_execution:
+    description:
+    - If this is set to true, SSH keys will be installed on the host.
+    required: false
+    type: bool
+  setup_remote_execution_pull:
+    description:
+    - If this is set to true, pull provider client will be deployed on the host.
+    required: false
+    type: bool
+  smart_proxy:
+    description:
+    - Name of Smart Proxy.
+    - This Proxy must have both the C(Templates) and C(Registration) features enabled.
+    required: false
+    type: str
+  update_packages:
+    description:
+    - Update all packages on the host.
+    required: false
+    type: bool
+  organization:
+    description:
+      - Organization to register the host in.
+    required: false
+    type: str
+  location:
+    description:
+      - Location to register the host in.
+    required: false
+    type: str
+
+extends_documentation_fragment:
+  - theforeman.foreman.foreman
+'''
+
+EXAMPLES = '''
+- name: "Generate registration command"
+  theforeman.foreman.registration_command:
+    username: "admin"
+    password: "changeme"
+    server_url: "https://foreman.example.com"
+  register: command
+
+- name: "Perform registration"
+  ansible.builtin.shell:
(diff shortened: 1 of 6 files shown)


plugins/modules/registration_command.py
```diff
@@ -0,0 +1,194 @@
...
+      - Location to register the host in
+    required: false
+    type: str
+
+extends_documentation_fragment:
+  - theforeman.foreman.foreman
+'''
+
+EXAMPLES = '''
+'''
```
evgeni, line 150: Well… ;)

plugins/modules/registration_command.py
```diff
@@ -0,0 +1,204 @@
...
+  - Manage Registration Command
+author:
+  - "Evgeni Golov (@evgeni)"
+options:
+  activation_keys:
+    description:
+    - Activation keys for subscription-manager client, required for CentOS and Red Hat
+      Enterprise Linux.
+    - Required only if host group has no activation keys.
+    required: false
```
Griffin-Sullivan, line 37: Is `required: false` the default when writing these?

plugins/modules/registration_command.py
```diff
@@ -0,0 +1,204 @@
...
+    description:
+    - Operating System to register the host in.
+    - Operating system must have a C(host_init_config) template assigned.
+    required: false
+    type: str
+  packages:
+    description:
+    - Packages to install on the host when registered.
+    required: false
+    type: str
```
Griffin-Sullivan, line 81: Are you supposed to write these as a list in a string?

plugins/modules/registration_command.py
```diff
@@ -0,0 +1,204 @@
...
+  repo_gpg_key_url:
+    description:
+    - URL of the GPG key for the repository
+    required: false
+    type: str
+  setup_insights:
+    description:
+    - If this is set to C(true), insights client will be installed
```
Griffin-Sullivan:
```suggestion
    - If this is set to C(true), insights-client will be installed
```

plugins/modules/registration_command.py
```diff
@@ -0,0 +1,204 @@
...
+            operatingsystem=dict(type='entity'),
+            smart_proxy=dict(type='entity'),
+            setup_insights=dict(type='bool'),
+            setup_remote_execution=dict(type='bool'),
+            jwt_expiration=dict(type='int'),
+            insecure=dict(type='bool'),
+            packages=dict(type='str'),
+            update_packages=dict(type='bool'),
+            repo=dict(type='str'),
+            repo_gpg_key_url=dict(type='str', no_log=False),
```
Griffin-Sullivan, line 179: What does `no_log=False` do here? Isn't that default?

Griffin-Sullivan (approved):
Nitpicking a little here but some of your descriptions have periods at the end and some don't. I'm not sure which is better but it would look best if they were all the same.

plugins/modules/registration_command.py
evgeni: It is, but Ansible has a check for parameters that *look* like secrets (e.g. it matches on "key"), and explicitly saying no log False here tells the check "yes, the param is a key but not a secret"

plugins/modules/registration_command.py
evgeni: yeah

plugins/modules/registration_command.py
evgeni: yes, updated the docs to reflect this.

plugins/modules/registration_command.py
evgeni: danke

evgeni:
@Griffin-Sullivan thanks for the review! I've updated the doc strings a bit, mind having another look?

plugins/modules/registration_command.py
```diff
@@ -0,0 +1,205 @@
...
+    required: false
+    type: bool
+  insecure:
+    description:
+    - Enable insecure argument for the initial C(curl).
+    required: false
+    type: bool
+  jwt_expiration:
+    description:
+    - Expiration of the authorization token (in hours).
```
sayan3296, line 62: @evgeni @Griffin-Sullivan  , Is it worth mentioning that , the default jwt_expiration period is of 4 hours , just for clarification ?

evgeni: We usually try not to document server side defaults, as those can change and then our docs get out of sync.

sayan3296: I was thinking from the point of view of a user who might wonder what is the default expiration period of the token after generating it. But I understand your point as well.  Thnx

evgeni: What is the usual user flow for that function? I mean nobody will install Foreman, fire up an editor and start writing an Ansible playbook to register a host. They will visit the UI, read the docs, etc. So these places should already tell the defaults?

sayan3296: True. They should do that only but there are users who need spoonfeeding as well.

The only place from where they can guess the default is `Hosts --> Register Hosts --> Advanced tab` and we pre-populate the `Token life time` field with value 4.  Other than that whether it's API or hammer ,or docs , nowhere we explicitly mention that 4 hours would be the default lifetime. The only thing we mention is `Expiration of the authorization token (in hours)`.

evgeni: Huh, interesting.
Maybe only that UI enforces the 4h and otherwise you get an unlimited token? That's something to investigate, but not on a Sunday morning ;)

sayan3296: But as i mentioned earlier, I understand your point of view as well. The default can change and we may not be able to keep the new change in sync with every other place where it needs to be.

sayan3296: It's 4 for sure. https://github.com/theforeman/foreman/blob/develop/app/controllers/concerns/foreman/controller/registration_commands.rb#L35-L39

evgeni: Oh huh, it can be the string "unlimited"?! How? The api says the value is a number:
https://github.com/theforeman/foreman/blob/74933559fafc3f52494845bf82a5a67d83b2d584/app/controllers/api/v2/registration_commands_controller.rb#L18

(And we translate this to "int" here in Ansible)

sayan3296: That is `kind of` reported in https://bugzilla.redhat.com/show_bug.cgi?id=2180568

evgeni: Interesting, thanks

sayan3296: if assuming correctly, We would not be able to pass `unlimited` as a value here for `jwt_expiration` right as it only accepts `int`.

It would have been great if `0` could be treated as `unlimited` somehow then ( even in the main API itself ).

evgeni: I'd prefer not to diverge from the API, so if we make that accept "unlimited", we should also accept strings here.

But you're right, at the moment non-integers would be rejected

evgeni: https://github.com/theforeman/foreman/pull/9951 - the api is moving to a "string" with stricter validation afterwards, so we should follow.

sayan3296: Is it not the opposite i.e. Leos suggested that we move to numbers instead i.e. 

* x < 0 raise an error, invalid value
* x = 0 unlimited expiration
* x > 0 expiration in hours.

and for both UI as well as API.

evgeni: it seems the general consensus in that PR is that the field becomes "number only", with either `0` or `-1` meaning "unlimited".
I think we can go ahead and merge *this* PR as-is (at it already accepts "numbers only"), and once the other PR is in we can adjust the doc string to mention what needs to be done for unlimited?

sayan3296: Agreed.. No problems with that I guess.. LGTM ..

PR #1684, opened 2023-11-20
Change typo resouce_type to resource_type

ehelms:


plugins/module_utils/foreman_helper.py
```diff
@@ -1316,7 +1316,7 @@ def __init__(self, **kwargs):
         ))[0])
 
         if 'parent' in self.foreman_spec and self.foreman_spec['parent'].get('type') == 'entity':
-            if 'resouce_type' not in self.foreman_spec['parent']:
+            if 'resource_type' not in self.foreman_spec['parent']:
```
evgeni, line 1319: heh, this uncovers an interesting bug. :see_no_evil: 

the `resource_type` attribute of `parent` is *always* set via 
https://github.com/theforeman/foreman-ansible-modules/blob/2864f87f96fe4d347c93287c88374ba60c17453d/plugins/module_utils/foreman_helper.py#L1655
but it's *always* `parents`.

What we need to check here is not the _current_ `foreman_spec` (as that has been enriched already), but the original one passed to `__init__`. But then again, I see absolutely no reason for setting this, as I can't think of a situation where a parent would be of a different type than the child?!

evgeni: cc @mdellweg :)

ehelms: I think you are saying this chunk of code might be useless?

evgeni: that's what I am saying.

ehelms: How do we figure out if it's useless?

evgeni: The line below this reads:

```python
self.foreman_spec['parent']['resource_type'] = self.foreman_spec['entity']['resource_type']
```

That means we're using the type of the main entity (the one we're currently executing). As I can't think of a situation where the parent would be a different type, this works just fine

mdellweg: So we can safely assume that the condition always evaluated true and we always set the parent type to match the current ones type. I guess that is at the very least better than "parents".

So if I understand your comment correctly, that means "fixing" this condition will make it false always, now leaving the parent's type to be "parents"? Is this breaking anything? What do we use that `resource_type` for? Do we need it to access apypie apis properly?

evgeni: https://github.com/theforeman/foreman-ansible-modules/pull/1725

ehelms: Does https://github.com/theforeman/foreman-ansible-modules/pull/1725 replace this PR?

evgeni: It does, and that's why it had a "closes" on this one ;)

PR #1690, opened 2023-12-08
do not use jinja in assert, we can use filters directly

evgeni:


tests/test_playbooks/host_info.yml
```diff
@@ -54,7 +54,7 @@
     - name: check host details
       assert:
         that:
-          - host_info['host']['name'] == "test-host.{{ host.domain }}"
+          - host_info['host']['name'] == "test-host.{}".format(host.domain)
```
mdellweg, line 57: As if that were easier...
So it looks like the whole condition is put through jinja2. Is that correct?

mdellweg (approved):

tests/test_playbooks/host_info.yml
evgeni: there is no jinja here anymore, just plain python :D

(no, I have not the slightes idea why this should be any better)

evgeni: OTOH, you can use `|filter`, so maybe there is still some jinja... the hell I know

mdellweg: That's what it looks to me. Whatever you assert here seems to be processed by jinja.

evgeni: instead of melting my brain over this, I'll go get some sushi :)

PR #1692, opened 2023-12-14
don't set entity_name in product module

evgeni:
it can be derived from the class name just fine

### DIFF
--- a/plugins/modules/product.py
+++ b/plugins/modules/product.py
@@ -120,7 +120,6 @@ class KatelloProductModule(KatelloEntityAnsibleModule):
 
 def main():
     module = KatelloProductModule(
-        entity_name='product',
         foreman_spec=dict(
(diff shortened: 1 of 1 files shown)


evgeni:
things you find while looking for examples for a demo :)

Griffin-Sullivan (approved):

Griffin-Sullivan:
does it use the class name or the name of the file?

evgeni:
Class name

Griffin-Sullivan:
is the class name not KatelloProductModule?

evgeni:
It is. And the code strips `Katello` and `Module` and then transforms the rest:

https://github.com/theforeman/foreman-ansible-modules/blob/b526b1ac7a68bd763a7fad481c8e1aadc25c27d5/plugins/module_utils/foreman_helper.py#L1336-L1356

PR #1693, opened 2023-12-14
virt_who_config - add a new module to support virt_who_config

### LINKED ISSUE #1483: Support virt-who
##### SUMMARY
Support the creation and loading from virt-who configurations.

dgoetz:
Fixes: https://github.com/theforeman/foreman-ansible-modules/issues/1483


dgoetz:
Wanted to try my hands on this after today's deep dive.

Some questions:
- something required to install the plugin for testing?
- filtering_mode is using 0, 1, 2 should we translate this to "none", "whitelist", "blacklist"?
- some options are only used with a specific hypervisor, API will take and store them even if not used, should we ensure they are not set at all?

PR #1695, opened 2023-12-20
exclude tests and changelog.yml from linting

evgeni:
while our CI doesn't lint those anyway, ansible-lint as invoked by galaxy does and produces a ton of rather useless warnings. lets just ignore them :)


Griffin-Sullivan (approved):
So we don't lint our test_playbooks?

evgeni:
> So we don't lint our test_playbooks?

Yeah, they are _wild_ :)

PR #1696, opened 2024-01-04
Add webhooks module

### LINKED ISSUE #1452: Support for webhook
##### SUMMARY
We are using foreman ansible modules to configure our RedHat Satellite. We find them very good and would like to have a module/role which create a webhook. 

Thanks

Griffin-Sullivan:
Ok @evgeni give it all you got.

This should partially solve https://github.com/theforeman/foreman-ansible-modules/issues/1452.


plugins/modules/webhook.py
```diff
@@ -0,0 +1,161 @@
...
...succeeded, actions.remote_execution.run_host_job_katello_packages_update_by_search_succeeded, actions.remote_execution.run_host_job_katello_service_restart_succeeded, actions.remote_execution.run_host_job_leapp_preupgrade_succeeded, actions.remote_execution.run_host_job_leapp_remediation_plan_succeeded, actions.remote_execution.run_host_job_leapp_upgrade_succeeded, actions.remote_execution.run_host_job_puppet_run_host_succeeded, actions.remote_execution.run_host_job_rh_cloud_connector_run_playbook_succeeded, actions.remote_execution.run_host_job_rh_cloud_remediate_hosts_succeeded, actions.remote_execution.run_host_job_run_script_succeeded, actions.remote_execution.run_host_job_succeeded, actions.remote_execution.run_hosts_job_running, actions.remote_execution.run_hosts_job_succeeded, build_entered, build_exited, content_view_created, content_view_destroyed, content_view_updated, domain_created, domain_destroyed, domain_updated, host_created, host_destroyed, host_updated, hostgroup_created, hostgroup_destroyed, hostgroup_updated, model_created, model_destroyed, model_updated, status_changed, subnet_created, subnet_destroyed, subnet_updated, user_created, user_destroyed, user_updated
```
evgeni: I wouldn't list all of the possible ones here.
* The list is specific to the installed plugins, and thus will vary on every installation
* The `:` after `one of` makes YAML think it's an array, confusing the Ansible doc linter.

plugins/modules/webhook.py
```diff
@@ -0,0 +1,161 @@
...
+      - Required when creating a new webhook
+    type: str
+  webhook_template_id:
```
evgeni: this would require the user to know the DB id to pass it.
better we make it `webhook_template` and let the API look it up?

plugins/modules/webhook.py
```diff
@@ -0,0 +1,161 @@
...
+
+def main():
+    module = ForemanWebhookModule(
+        foreman_spec=dict(
+            name=dict(required=True),
+            target_url=dict(),
+            http_method=dict(),
+            http_content_type=dict(),
+            event=dict(),
+            webhook_template_id=dict(),
```
evgeni:
```suggestion
            webhook_template=dict(type='entity'),
```

untested, this *should* make it look up templates by name

plugins/modules/webhook.py
```diff
@@ -0,0 +1,161 @@
...
+class ForemanWebhookModule(ForemanTaxonomicEntityAnsibleModule):
+    pass
+
+
+def main():
+    module = ForemanWebhookModule(
+        foreman_spec=dict(
+            name=dict(required=True),
+            target_url=dict(),
+            http_method=dict(),
```
evgeni:
```suggestion
            http_method=dict(choices=['POST', 'GET', 'PUT', 'DELETE', 'PATCH']),
```

(and the same above in the doc)

this will force the user to use a proper http verb :)

tests/fixtures/apidoc/webhook.json
```diff
@@ -0,0 +1 @@
+luna.json
```
evgeni, line 1: I changed this to `luna.json`, as `foreman.json` has no plugins and thus did make the tests fail

evgeni (changes requested):

plugins/modules/webhook.py
Griffin-Sullivan: What should I put for the description here then?

evgeni: "Name of the event that shall trigger the webhook"?

plugins/modules/webhook.py
```diff
@@ -0,0 +1,167 @@
...
+    description:
+      - Password for the webhook, if required
+    type: str
+  http_headers:
+    description:
+      - HTTP headers for the webhook
+    type: str
+  proxy_authorization:
+    description:
+      - Authorize with Satellite client certificate and validate capsule CA from Settings
```
evgeni:
```suggestion
      - Authorize with client certificate and validate CA from Settings
```

damn, missed that Satellite branding here :)

evgeni (approved):

PR #1708, opened 2024-02-11
Adding missing dns_proxy and description arguments to domain role

sscheib:
Despite the role's description stating that `description` and `dns_proxy` are being passed to the module call of `theforeman.foreman.domain`, these attributes are not passed to it.

This PR fixes that by passing both `description` and `dns_proxy` to the module call if specified. They will be omitted by default.

### DIFF
--- a/roles/domains/tasks/main.yml
+++ b/roles/domains/tasks/main.yml
@@ -6,6 +6,8 @@
     server_url: "{{ foreman_server_url | default(omit) }}"
     validate_certs: "{{ foreman_validate_certs | default(omit) }}"
     name: "{{ item.name }}"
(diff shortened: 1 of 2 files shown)


sscheib:
Domain tests seem to be fixed, but I don't understand the issue that is present currently. Could sombody review this and give me a hint where to look? Thanks!

evgeni:
> Domain tests seem to be fixed, but I don't understand the issue that is present currently. Could sombody review this and give me a hint where to look? Thanks!

Ignore that, it's fixed in https://github.com/theforeman/foreman-ansible-modules/pull/1707

evgeni (approved):

PR #1713, opened 2024-02-23
Fixes #1712: Fix content_view_version_cleanup slice error

### LINKED ISSUE #1712: content_view_version_cleanup role fails with error 'slice indices must be integers or None or have an __index__ method'
##### SUMMARY
content_view_version_cleanup role fails with error 'slice indices must be integers or None or have an __index__ method'

##### ISSUE TYPE
 - Bug Report

##### ANSIBLE VERSION
```
ansible [core 2.15.9]
  config file = /home/local/COMMON/sme8401/ansible/tme-satellite/ansible.cfg
  configured module search path = ['/home/local/COMMON/sme8401/.ansible/plugins/modules', '/usr/share/ansible/plugins/modules']
  ansible python module location = /home/local/COMMON/sme8401/bin/ansible-venv-py39/lib64/python3.9/site-packages/ansible
  ansible collection location = /home/local/COMMON/sme8401/.ansible/collections:/usr/share/ansible/collections
  executable location = /home/local/COMMON/sme8401/bin/ansible
  python version = 3.9.18 (main, Nov 16 2023, 02:13:35) [GCC [8.5.0 20210514](callto:8.5.0 20210514) (Red Hat [8.5.0-20.0.1](callto:8.5.0-20.0.1))] (/home/smeeus/bin/ansible-venv-py39/bin/python3.9)
  jinja version = 3.1.2
  libyaml = True
```

##### COLLECTION VERSION
```
[smeeus@watson satellite]$ cat /home/smeeus/.ansible/collections/ansible_collections/theforeman/foreman/MANIFEST.json | grep version
  "version": "4.0.0",
```

##### KATELLO/FOREMAN VERSION

```
rubygem-katello-4.8.1-1.el8.noarch
foreman-3.6.1-1.el8.noarch
```

##### STEPS TO REPRODUCE

```yaml
- name: "Cleanup Satellite content views"
  hosts: satellite
  tasks:
    - name: "Content Cleanup | Include theforeman.foreman.content_credentials role"
      ansible.builtin.import_role:
        name: theforeman.foreman.content_view_version_cleanup
      tags: [ "katello_cleanup" ]
```

##### EXPECTED RESULTS

Cleanup of unused content views

##### ACTUAL RESULTS

```
ansible-playbook -i inventories playbook-satellite-content.yml --diff --tags=katello_cleanup -e "foreman_cleanup_content=true" -e "foreman_content_view_version_cleanup_keep=0"
...
TASK [theforeman.foreman.content_view_version_cleanup : Delete unused composite content view versions] *******************************************************************************************************************************************************************
Friday 23 February 2024  11:27:52 +0100 (0:00:02.368)       0:00:10.594 *******
fatal: [satellite]: FAILED! => {"msg": "The conditional check '(__foreman_ccv.versions | rejectattr('environment_ids') | map(attribute='version') | reverse | list)[foreman_content_view_version_cleanup_keep:]' failed. The error was: Unexpected templating type error occurred on ({% if (__foreman_ccv.versions | rejectattr('environment_ids') | map(attribute='version') | reverse | list)[foreman_content_view_version_cleanup_keep:] %} True {% else %} False {% endif %}): slice indices must be integers or None or have an __index__ method. slice indices must be integers or None or have an __index__ method\n\nThe error appears to be in '/home/smeeus/.ansible/collections/ansible_collections/theforeman/foreman/roles/content_view_version_cleanup/tasks/main.yml': line 19, column 3, but may\nbe elsewhere in the file depending on the exact syntax problem.\n\nThe offending line appears to be:\n\n\n- name: \"Delete unused composite content view versions\"\n  ^ here\n"}
```

smeeus:
An explicit conversion to int of the `foreman_content_view_version_cleanup_keep` variable is required to fix the `slice indices must be integers or None or have an __index__ method` error.

### DIFF
--- a/roles/content_view_version_cleanup/tasks/find_and_delete_unused_cv_versions.yml
+++ b/roles/content_view_version_cleanup/tasks/find_and_delete_unused_cv_versions.yml
@@ -17,4 +17,4 @@
     foreman_cv_name: "{{ foreman_cv.name }}"
     foreman_cv_versions: "{{ (__foreman_versions.resources | rejectattr('environments') | rejectattr('composite_content_view_ids') |
       rejectattr('published_in_composite_content_view_ids') | map(attribute='version') | map('float') | sort |
-      map('string') | reverse | list)[foreman_content_view_version_cleanup_keep:] }}"
+      map('string') | reverse | list)[(foreman_content_view_version_cleanup_keep | int):] }}"
--- a/roles/content_view_version_cleanup/tasks/main.yml
+++ b/roles/content_view_version_cleanup/tasks/main.yml
@@ -21,12 +21,12 @@
   vars:
     foreman_cv_name: "{{ __foreman_ccv.name }}"
     foreman_cv_versions: "{{ (__foreman_ccv.versions | rejectattr('environment_ids') | map(attribute='version') | map('float') | sort
-      | map('string') | reverse | list)[foreman_content_view_version_cleanup_keep:] }}"
+      | map('string') | reverse | list)[(foreman_content_view_version_cleanup_keep | int):] }}"
   loop: "{{ __foreman_all_cvs.resources | selectattr('composite') | list }}"
   loop_control:
     label: "{{ __foreman_ccv.label }}"
     loop_var: "__foreman_ccv"
-  when: (__foreman_ccv.versions | rejectattr('environment_ids') | map(attribute='version') | reverse | list)[foreman_content_view_version_cleanup_keep:]
+  when: (__foreman_ccv.versions | rejectattr('environment_ids') | map(attribute='version') | reverse | list)[(foreman_content_view_version_cleanup_keep | int):]
 
 - name: "Find and delete unused content view versions"
   ansible.builtin.include_tasks: find_and_delete_unused_cv_versions.yml
@@ -34,4 +34,4 @@
   loop_control:
     label: "{{ foreman_cv.label }}"
     loop_var: "foreman_cv"
-  when: (foreman_cv.versions | rejectattr('environment_ids') | map(attribute='version') | reverse | list)[foreman_content_view_version_cleanup_keep:]
+  when: (foreman_cv.versions | rejectattr('environment_ids') | map(attribute='version') | reverse | list)[(foreman_content_view_version_cleanup_keep | int):]


smeeus:
@evgeni Could you try and prioritize this fix in one of the next releases please?

evgeni:
How are you setting it so it doesn't end up being an int?

smeeus:
Very good question, and I don't have an answer there. I have had this error for a long while now and actually never considered that it was due to that variable being an issue.
It's only today by really debugging all the selectattr, rejectattr and map filters that I figured out it was related to this variable.
When I debug the variable, it does show as an integer, but it seems it doesn't when you use it to slice the lists.
There might be moregoing on, but the implicit conversion does work.
In the issue, I've used `0` as a value, but I've also tested the fix with `1` and `3` against a satellite 6.14 server.

evgeni:
This feels like papering over an actual issue.

In our CI this works just fine:
https://github.com/theforeman/foreman-ansible-modules/blob/28ea226a5610aab7d892b28b3e758080fcfa44a3/tests/test_playbooks/content_view_version_cleanup_role.yml#L58

And so does my prod environment where I use this role regularly.

Do your variables come from some external source, not a vars file?

smeeus:
I always set it from command line with `-e "foreman_content_view_version_cleanup_keep=0"`
In the production environments I work with, it passes through another collection I am working on: https://github.com/kangaroot/ansible-collection-kangaroot-foreman

But even with the example playbook in the issue, I have the same error.

evgeni:
out of curiousity, can you try with `-e '{"foreman_content_view_version_cleanup_keep":5}'` (= use JSON and not a string to pass the var)

smeeus:
I don't have the error if I use the JSON style parameter.
I've also tried some variations with single and double quotes on the non-JSON style and they all give the same error.

This is one of those python variable type conversion things again I think...

Something I have issues with on a regular basis is `ansible_distribution_major_version >= 8` for example in `when` clauses, they also need an implicit conversion to `int` everywhere I check for a version these days. I use different ansible versions in some cases (customer dependent). Something tells me this is somewhat related and python not having proper knowledge of the type of the variables.

smeeus:
Now, regardless of the JSON style working or not, I can imagine that other people might have faced the same issue.
Even though it is pampering something that shouldn't be fixed, it might be a good idea to just do it regardless, always be sure that it is an int instead of having to rely on a certain method of passing command line variables etc.

evgeni:
Yeah, I absolutely agree we should fix this. I just wanted to better understand why it needs fixing.

The cli `-e` string/json difference is a good one.

evgeni (approved):

smeeus:
Thanks @evgeni

PR #1720, opened 2024-03-19
content import modules (info, library, repository, version)

parthaa:
Added bindings to content import  versions, repositories and library via foreman ansible modules. 
```yaml
--- 
- name: import to foreman
  hosts: localhost
  gather_facts: false
  tasks:
  - name: "Import repository"
    theforeman.foreman.content_import_repository:
        path: "/var/lib/pulp/exports/export-10380/Export-katello-1/1.0/2024-03-19T15-26-35-00-00"
        metadata: "{{ lookup('file', '/tmp/metadata.json') | from_json }}"
        username: "admin"
        password: "changeme"
        server_url: "http://darwin.lagrange.example.com"
        organization: "foo org"

```
you can also optionally use `metadata_file` instead of `metadata`


plugins/module_utils/foreman_helper.py
```diff
@@ -1592,6 +1592,39 @@ def run(self, **kwargs):
...
+        metadata_file = self.params.get('metadata_file')
+        self.auto_lookup_entities()
+        payload = _flatten_entity(self.foreman_params, self.foreman_spec)
+
+        if payload.get("metadata") is None and metadata_file:
+            payload["metadata"] = json.load(open(metadata_file))
+            payload.pop("metadata_file")
+
+        if payload.get("metadata") is None:
+            self.fail_json(msg="Metadata not specified.")
```
evgeni: if you pass
```
required_one_of=[['metadata', 'metadata_file']]
```
to the constructor in line 1609, you will get the "metadata not specified" error "for free"

plugins/modules/content_import_library.py
```diff
@@ -0,0 +1,87 @@
...
+    description:
+      - Directory containing the exported library.
+    required: true
+    type: str
+  metadata:
+    description:
+      - Contents of the metadata.json file. This is not required if the metadata_file location is provided.
+  metadata_file:
+    description:
+      - Location of the metadata.json file. Not required if the metadata has been already provided via the other parameter
```
evgeni: I think all three options could live in a documentation fragment, so we don't have to repeat ourself.

plugins/modules/content_import_library.py
```diff
@@ -0,0 +1,87 @@
...
+# You should have received a copy of the GNU General Public License
+# along with this program.  If not, see <http://www.gnu.org/licenses/>.
+
+from __future__ import absolute_import, division, print_function
+__metaclass__ = type
+
+DOCUMENTATION = '''
+---
+module: content_import_library
+version_added: 4.0.0
```
evgeni:
```suggestion
version_added: 4.1.0
```

4.0.0 is already released :)

evgeni:
Not a full review, as I know you're still working on this, but looks pretty nice already!

plugins/modules/content_import_info.py
```diff
@@ -0,0 +1,120 @@
...
+version_added: 4.0.0
+short_description: List content imports
+description:
+    - List information about content imports.
+author:
+    - "Partha Aji (@parthaa)"
+options:
+  id:
+    description:
+      - Export history identifier.
```
jeremylenz: Is this an export history or an import history? Are they the same?

plugins/modules/content_import_info.py
```diff
@@ -0,0 +1,120 @@
...
+- name: "Get a specific import history and register the result for the next task"
+  vars:
+    organization_name: "Export Org"
+  theforeman.foreman.content_import_info:
+    id: 29
+    username: "admin"
+    password: "changeme"
+    server_url: "https://foreman.example.com"
+    organization: "Default Organization"
+  register: result
```
jeremylenz: In content_export_info this was there so you could save your metadata. But the "register result for the next task" part is maybe not needed here.

plugins/modules/content_import_info.py
```diff
@@ -0,0 +1,120 @@
...
+  theforeman.foreman.content_import_info:
+    content_view: RHEL8
+    content_view_version: '1.0'
+    username: "admin"
+    password: "changeme"
+    server_url: "https://foreman.example.com"
+    organization: "Default Organization"
+- name: "List incremental imports of a specific content view version"
+  theforeman.foreman.content_import_info:
+    content_view: RHEL8
```
jeremylenz: This doesn't actually specify a version.
```suggestion
- name: "List incremental imports of a specific content view"
  theforeman.foreman.content_import_info:
    content_view: RHEL8
```
(I think this bug is also in the one you copied from ;)

plugins/modules/content_import_library.py
parthaa: do you have examples of document fragments?

evgeni: https://github.com/theforeman/foreman-ansible-modules/blob/d195da2d1e6c4c766739e5fdf4156bc585e91807/plugins/doc_fragments/foreman.py#L390-L417
used e.g. here:
https://github.com/theforeman/foreman-ansible-modules/blob/d195da2d1e6c4c766739e5fdf4156bc585e91807/plugins/modules/content_export_library.py#L45

plugins/modules/content_import_info.py
parthaa: It should be import

tests/test_playbooks/content_import.yml
evgeni, line 1: if you split this file into three files, `content_import_repository.yml`, `content_import_version.yml` and `content_view_library.yml`, the "module is tested" test will recognize that (it simply matches `something.py` against `something.yml`)

(this will also rename your fictures to `content_import_THING-0.yml`, you probably can just `git mv` those to save re-recording time)

plugins/modules/content_import_info.py
jeremylenz: Still think this example can be deleted.

plugins/modules/content_import_info.py
```diff
@@ -0,0 +1,120 @@
...
+# along with this program.  If not, see <http://www.gnu.org/licenses/>.
+
+from __future__ import absolute_import, division, print_function
+__metaclass__ = type
+
+
+DOCUMENTATION = '''
+---
+module: content_import_info
+version_added: 4.0.0
```
evgeni:
```suggestion
version_added: 4.1.0
```

plugins/modules/content_import_info.py
```diff
@@ -0,0 +1,120 @@
...
+- name: "Get a specific import history and register the result for the next task"
+  vars:
+    organization_name: "Export Org"
+  theforeman.foreman.content_import_info:
+    id: 29
+    username: "admin"
+    password: "changeme"
+    server_url: "https://foreman.example.com"
+    organization: "Default Organization"
+  register: result
```
evgeni: I see no reason to register the result here.

tests/test_playbooks/content_import_info.yml
```diff
@@ -0,0 +1,66 @@
...
+  tasks:
+    - name: Setup Org
+      include_tasks: tasks/organization.yml
+      vars:
+        organization_state: present
+    - name: Test Import Repository
+      include_tasks: tasks/content_import.yml # Full export
+      vars:
+        import_type: "repository"
+        expected_change: true
```
evgeni: we don't expect a change here (it will happen, but we shall not expect it, it's the setup phase)

tests/test_playbooks/tasks/content_import.yml
```diff
@@ -0,0 +1,19 @@
...
+    path: "/var/lib/pulp/imports/content-import"
+    metadata: "{{ lookup('file', 'data/metadata/metadata-' + import_type + '.json') | from_json }}"
+    username: "{{ foreman_username }}"
+    password: "{{ foreman_password }}"
+    server_url: "{{ foreman_server_url }}"
+    organization: "{{ organization_name }}"
+  register: result
+
+- assert:
+    fail_msg: "Repository export failed!"
```
evgeni:
```suggestion
    fail_msg: "Repository import failed!"
```

;-)

evgeni (commented):
couple of nitpicks, but overall this looks good!

evgeni:
You dropped the wrong register, that one is actually required

tests/test_playbooks/tasks/content_import.yml
```diff
@@ -0,0 +1,19 @@
+---
+- name: "{{ 'Import' + import_type }}"
+  vars:
+    - organization_name: "Test Organization"
```
evgeni:
```suggestion
    organization_name: "Test Organization"
```

plugins/modules/content_import_info.py
```diff
@@ -0,0 +1,119 @@
...
+- name: "List all imports in the organization"
+  theforeman.foreman.content_import_info:
+    organization: "Default Organization"
+    type: complete
+    username: "admin"
+    password: "changeme"
+    server_url: "https://foreman.example.com"
+- name: "Get a specific import history and register the result for the next task"
+  vars:
+    organization_name: "Import Org"
```
evgeni:
```suggestion
- name: "Get a specific import history entry"
```

plugins/modules/content_import_library.py
```diff
@@ -0,0 +1,75 @@
...
+from __future__ import absolute_import, division, print_function
+__metaclass__ = type
+
+DOCUMENTATION = '''
+---
+module: content_import_library
+version_added: 4.1.0
+short_description: Manage library content imports
+description:
+    - Import library content to foreman.
```
evgeni:
```suggestion
    - Import library content.
```

plugins/modules/content_import_repository.py
```diff
@@ -0,0 +1,74 @@
...
+__metaclass__ = type
+
+
+DOCUMENTATION = '''
+---
+module: content_import_repository
+version_added: 4.1.0
+short_description: Manage repository content imports
+description:
+    - Import repository content to foreman.
```
evgeni:
```suggestion
    - Import repository content.
```

plugins/modules/content_import_version.py
```diff
@@ -0,0 +1,74 @@
...
+__metaclass__ = type
+
+
+DOCUMENTATION = '''
+---
+module: content_import_version
+version_added: 4.1.0
+short_description: Manage content view version content imports
+description:
+    - Import a content view version to a foreman.
```
evgeni:
```suggestion
    - Import a content view version.
```

evgeni (approved):

evgeni:
thanks @parthaa!

PR #1721, opened 2024-03-21
user module fixes

evgeni:


evgeni:
so glad we have tests :see_no_evil:

ehelms (approved):

PR #1724, opened 2024-04-03
Feature: Add description to location module

PfurtschellerP:
Currently - even though the REST API and python utils would allow it - the location module does not implement the parameter `description`.
However, this can be helpful especially if a company uses e.g. computed site codes as locations in Foreman or Satellite that are not easily understandable at first glance (but help with automation).

### DIFF
--- a/changelogs/fragments/1724-location_description.yml
+++ b/changelogs/fragments/1724-location_description.yml
@@ -1,0 +1,2 @@
+minor_changes:
+  - location, locations role - add ``description`` parameter to set the description
--- a/plugins/modules/location.py
+++ b/plugins/modules/location.py
@@ -34,6 +34,12 @@ options:
       - Name of the Location
     required: true
     type: str
+  description:
+    description:
+      - Description of the Location
+    required: false
+    type: str
+    version_added: 4.2.0
   parent:
     description:
       - Title of a parent Location for nesting
@@ -123,6 +129,7 @@ def main():
     module = ForemanLocationModule(
         foreman_spec=dict(
             name=dict(required=True),
+            description=dict(required=False),
             parent=dict(type='entity'),
             organizations=dict(type='entity_list'),
             ignore_types=dict(type='list', elements='str', required=False, aliases=['select_all_types']),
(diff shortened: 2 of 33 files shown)


evgeni:
On a first glance, your changes look good (and you have a picky editor 😉).
You've added tests, but did not re-record the fixtures, which is why things are now failing in CI. Do you need any help with that step?

PfurtschellerP:
> On a first glance, your changes look good (and you have a picky editor 😉). You've added tests, but did not re-record the fixtures, which is why things are now failing in CI. Do you need any help with that step?

Hi,
yes I would definetly need help with that. I tried to replicate it like it was done for the other tests, but I'm not familiar with creating fixtures.

evgeni:
I'll try to ping you with some pointers next week when back in the office.

evgeni:
Okay, here I am. Let's see if we can get this going for you.

There is some basic documentation at https://github.com/theforeman/foreman-ansible-modules/blob/develop/docs/testing.md#writing-tests but I'll try to re-phrase it (and you tell me which version was better understandable ;) )

You'll need:
- a Foreman installation you have admin privileges on and that you can "destroy" (the tests create and delete a lot of things, so this should ideally be a throw-away VM or something, I use https://github.com/theforeman/forklift for that, but it's not a requirement)
- a checkout of your branch
- a python environment with our [development dependencies](https://github.com/theforeman/foreman-ansible-modules/blob/develop/requirements-dev.txt) installed (the easiest way for that is IMHO a Python venv and then calling `make test-setup`)
- url/user/password of your Foreman installation set in `tests/test_playbooks/vars/server.yml` (this file gets created by `make test-setup`, if you don't use that just copy the example manually from https://github.com/theforeman/foreman-ansible-modules/blob/develop/tests/test_playbooks/vars/server.yml.example )

Once you have that, you *should* be ready to go. You can verify the overall setup by calling an existing test like `make test_bookmark`, this won't touch your installation yet, but run the same tests that we run on GitHub here.

As a counter-test, you can run `make test_location` and that will fail, as the playbooks do not (yet) match the recorded fixtures.

You can now run `make record_location` which will delete all existing fixtures and re-record them. As you've added new tests, you'll see new files in `tests/test_playbooks/fixtures/location-*.yml` (and the old ones change a bit, as you re-record everything).
If `make record_location` passes, you can add the new (and changed) files to git, and commit them. Once pushed here I'd expect the CI to turn green.
If it doesn't pass, it might be a bug in the changes you made, or something else. Ideally just paste the whole output here and we'll look over it together.

evgeni (approved):

PR #1726, opened 2024-04-08
always declare vars as dicts

evgeni:
previously ansible also accepted lists, but not anymore


Griffin-Sullivan (approved):
Not sure about the warnings but everything looks good to me

PR #1730, opened 2024-04-12
Update AK tests using entitlement subscriptions

Griffin-Sullivan:
Everything is SCA now so we don't need to specify the specific subs for an AK. This should fix 2 tests in our live automation runs.

### DIFF
--- a/plugins/modules/activation_key.py
+++ b/plugins/modules/activation_key.py
@@ -49,6 +49,7 @@ options:
     description:
       - List of subscriptions that include either Name, Pool ID, or Upstream Pool ID.
       - Pool IDs are preferred since Names and Upstream Pool IDs are not guaranteed to be unique. The module will fail if it finds more than one match.
+      - This parameter is not supported in SCA mode.
     type: list
     elements: dict
     suboptions:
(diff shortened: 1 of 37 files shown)


evgeni:
@jeremylenz reminded me that we also should look at the examples in our docs and update these.

plugins/modules/activation_key.py
```diff
@@ -49,6 +49,7 @@
     description:
       - List of subscriptions that include either Name, Pool ID, or Upstream Pool ID.
       - Pool IDs are preferred since Names and Upstream Pool IDs are not guaranteed to be unique. The module will fail if it finds more than one match.
+      - NOTE: New versions no longer support this parameter
```
evgeni: The YAML gods don't like that. This is now an entry that is itself a dict `{'NOTE': 'New versions…'}`.
You need to either quote this as a whole string:
```suggestion
      - "NOTE: New versions no longer support this parameter"
```
Or rephrase to not have a colon:
```suggestion
      - This parameter is not supported in SCA mode.
```
(I prefer that, as it also removes the ambiguity what "new versions" are.)

Griffin-Sullivan: I wasn't sure what the proper upstream subscription language was so I avoided adding SCA.

evgeni (approved):

PR #1731, opened 2024-04-23
Make registration_command run against live Satellite

Griffin-Sullivan:
Enabling this in robottelo in a separate PR: https://github.com/SatelliteQE/robottelo/pull/14866

While verifying a BZ I tried to run the test playbook for RC and found it needed an activation key.

### DIFF
--- a/tests/fixtures/apidoc/registration_command.json
+++ b/tests/fixtures/apidoc/registration_command.json
@@ -1,1 +1,1 @@
-foreman.json
+katello.json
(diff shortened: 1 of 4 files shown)


evgeni:
Heh, that's an interesting one.
In plain foreman, there are no AKs and they are not needed for registration. Once you add Katello to the mix things change.

I'm fine to only test the katello way (in other places we have multiple test playbooks, one for plain foreman, one for katello/luna).


You gotta change the symlink of the apidoc json to katello, otherwise tests will be failing as AK is an unknown parameter in the foreman scenario.

evgeni (approved):

PR #1732, opened 2024-05-03
Include external_usergroup vars in server.yml

Griffin-Sullivan:
This is failing on a live test because you need a valid LDAP set up. Adding these vars to the server.yml file so that we can perform live testing correctly.


evgeni:
*technically* the re-recording was not necessary, as you moved the vars and kept the values, right?

evgeni (approved):

PR #1733, opened 2024-05-08
Remove deb CV filters from tests and make special deb test

Griffin-Sullivan:
The content view filter tests were using deb repositories which was breaking on live Satellites. I decided to remove the deb filters and rules from the `content_view_filter_*` test playbooks and give it it's own simple playbook. 

I also simplified the `content_view_filter_info` test since it was outdated after creating the `content_view_filter_rule` module. The test now just creates a filter and makes sure the info module works on it. No need to run an info on every type of filter, since the other tests are much more comprehensive.


Griffin-Sullivan:
@evgeni can't get forklift to work for me so I still need to record the fixtures for the deb playbook

evgeni:
I've started more splitting, so let's do https://github.com/theforeman/foreman-ansible-modules/pull/1761

evgeni:
Closing in favor of #1761

PR #1735, opened 2024-05-15
Add wait for repo metadata generation in export playbooks

Griffin-Sullivan:
Surely there's a better way to do this, but to summarize here's the issue:

1. We make repo and then right after do an export in both these tests
2. When the export is started the task for generating the repo metadata is still running (although it's pretty fast)
3. Rerunning after a failure will work since the task is done by then. Adding a wait for 2 seconds achieves the same result. 

Is there a more synchronous way to do this? Maybe wait for the task? I'm not sure if the metadata task is reported back when the repository create task is finished.

### DIFF
--- a/tests/test_playbooks/content_export_info.yml
+++ b/tests/test_playbooks/content_export_info.yml
@@ -16,6 +16,10 @@
       vars:
         repository_state: present
 
+    - name: Wait for Repository metadata generation
+      ansible.builtin.pause:
+        seconds: 2
+
     - include_tasks: tasks/content_export_library.yml # Full export
       vars:
         incremental: false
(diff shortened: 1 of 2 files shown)


evgeni:
Well, our modules already do wait for task completion if there is a "direct" task:
https://github.com/theforeman/foreman-ansible-modules/blob/a7077cc1cf6fc2d6ac907072b82462b1368d3a95/plugins/module_utils/foreman_helper.py#L1193-L1196

So whatever does the generation is not properly reported as part of the task (maybe intentionally?)

I also see no task-related info in https://github.com/theforeman/foreman-ansible-modules/blob/develop/tests/test_playbooks/fixtures/repository-0.yml, so I'd just assume this is how it is today and merge your "sleep 2".

evgeni (approved):
No idea why 3.5 is failing, but certainly not you.

PR #1751, opened 2024-07-03
Fix provisioning_template removal

Griffin-Sullivan:
Trying to remove all provisioning templates does not work on a Satellite. Changing the test to only remove what it creates.


Griffin-Sullivan:
Looks like my recording might not having worked properly?

Griffin-Sullivan:
ok tests should pass now

evgeni (approved):

PR #1753, opened 2024-07-05
Update role test_playbook to not use removed filter perm

Griffin-Sullivan:
It seems like the `puppetrun_hosts` permission no longer exists or is not on Satellite by default. Either way we can use another permission to test adding additional filters to roles that already have one filter.


evgeni (approved):
Yeah, that one was moved out to the puppet plugin and that's not installed by default.

PR #1754, opened 2024-07-09
Fix content_view_version_cleanup_role version listing in test

Griffin-Sullivan:
`remaining_cvs.resources[1].versions|map(attribute='version')|list` was returning `['5.0', '4.0', '3.0']` which is != to `['3.0', '4.0', '5.0']`. Let's use `difference()` and compare them like we would with a set.

### DIFF
--- a/tests/test_playbooks/content_view_version_cleanup_role.yml
+++ b/tests/test_playbooks/content_view_version_cleanup_role.yml
@@ -70,8 +70,8 @@
     - name: check remaining content view versions
       assert:
         that:
-          - remaining_cvs.resources[0].versions|map(attribute='version')|list == ['3.0', '4.0', '5.0']
-          - remaining_cvs.resources[1].versions|map(attribute='version')|list == ['3.0', '4.0', '5.0']
+          - remaining_cvs.resources[0].versions|map(attribute='version')|list | difference(['3.0', '4.0', '5.0'])|length == 0
+          - remaining_cvs.resources[1].versions|map(attribute='version')|list | difference(['3.0', '4.0', '5.0'])|length == 0
     - name: clean up all unused CVs
       include_role:
         name: content_view_version_cleanup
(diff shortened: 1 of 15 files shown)


tests/test_playbooks/content_view_version_cleanup_role.yml
```diff
@@ -70,8 +70,8 @@
     - name: check remaining content view versions
       assert:
         that:
-          - remaining_cvs.resources[0].versions|map(attribute='version')|list == ['3.0', '4.0', '5.0']
-          - remaining_cvs.resources[1].versions|map(attribute='version')|list == ['3.0', '4.0', '5.0']
+          - remaining_cvs.resources[0].versions|map(attribute='version')|list | difference(['3.0', '4.0', '5.0'])|length == 0
+          - remaining_cvs.resources[1].versions|map(attribute='version')|list | difference(['3.0', '4.0', '5.0'])|length == 0
```
evgeni, line 74: Those two are not equal. When the `remaining_cvs` list is a *subset* of 3,4,5, you still get `true`:

```console
% ansible -m debug -a "msg='{{ [1] | difference([1,2,3]) }}'" localhost
localhost | SUCCESS => {
    "msg": []
}
```

evgeni: In https://github.com/theforeman/foreman-ansible-modules/blob/1fe33bfd214cba3515f78a8706e7da6b5790bdf3/roles/content_view_version_cleanup/tasks/find_and_delete_unused_cv_versions.yml#L19 we use `map('float') | sort` to make things stable. Should we do the same here? Given we only have <10 versions, we can even avoid the floating conversion.

evgeni: https://github.com/theforeman/foreman-ansible-modules/pull/1758

PR #1758, opened 2024-07-12
sort versions in content_view_version_cleanup_role test

evgeni:


evgeni:
@Griffin-Sullivan this passes with both the existing fixtures and the ones you uploaded in #1754

Griffin-Sullivan:
ok let's just merge yours and close mine then. Tested against live Sat and it works.

Griffin-Sullivan (approved):

PR #1761, opened 2024-07-24
Split out cv_filter* Debian testing into own playbooks

evgeni:
This allows running the main playbook on setups that do not support Debian content


evgeni:
I've verified that all `content_view*.yml` test playbooks without `_deb` now record correctly against a Satellite \o/

ehelms (approved):

PR #1762, opened 2024-07-25
add packit

evgeni:


.packit.yaml
```diff
@@ -0,0 +1,46 @@
...
+  - job: copr_build
+    trigger: pull_request
+    targets:
+      rhel-8:
+        additional_modules: "foreman:el8"
+        additional_repos:
+          - https://yum.theforeman.org/releases/nightly/el8/x86_64/
+      rhel-9:
+        additional_repos:
+          - https://yum.theforeman.org/releases/nightly/el9/x86_64/
```
evgeni: technically we don't need those -- keep or drop?

ehelms: drop

ehelms (approved):

PR #1764, opened 2024-07-26
activation_keys role -- allow deleting of keys

cojmckee:
Added Feature to Activation_keys role to delete activation keys

### DIFF
--- a/roles/activation_keys/tasks/main.yml
+++ b/roles/activation_keys/tasks/main.yml
@@ -1,5 +1,5 @@
 ---
-- name: 'Create 0r Delete Activation Keys'  # noqa: args[module]
+- name: 'Create or Delete Activation Keys'  # noqa: args[module]
   theforeman.foreman.activation_key:
     username: "{{ foreman_username | default(omit) }}"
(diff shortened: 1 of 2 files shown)


roles/activation_keys/tasks/main.yml
```diff
@@ -21,6 +21,6 @@
     purpose_usage: "{{ item.purpose_usage | default(omit) }}"
     purpose_role: "{{ item.purpose_role | default(omit) }}"
     purpose_addons: "{{ item.purpose_addons | default(omit) }}"
-    state: present
-  with_items:
+    state: "{{ item.state | default('present') }}"
+  loop:
```
evgeni: If you keep "with_items" things will work again.
"loop" is stricter about the syntax

evgeni (changes requested):

roles/activation_keys/tasks/main.yml
cojmckee: Replaced loop with with_items

evgeni (approved):

PR #43, opened 2017-09-01
Foreman operatingsystem

mdellweg:
This adds yet another module. Hurray!
This time for Operating systems.

PR #1779, opened 2024-09-12
don't require password_hash when creating an OS

evgeni:
this has a working default value, no idea why it was set to required
when new.


mdellweg (approved):

mdellweg:
> this has a working default value, no idea why it was set to required when new.

I would guess on a lack of knowledge.

evgeni:
> > this has a working default value, no idea why it was set to required when new.
> 
> I would guess on a lack of knowledge.

Yeah, the author of https://github.com/theforeman/foreman-ansible-modules/pull/43 is known for that :P

PR #1785, opened 2024-10-04
support hidden values when creating params

evgeni:


evgeni:
contrary to what CI makes us believe, this actually breaks idempotency because of https://github.com/theforeman/foreman/commit/eaeeaef6325cd0a38c6a8cc0d83a8961bd10f029

plugins/module_utils/foreman_helper.py
```diff
@@ -1094,6 +1097,15 @@ def _update_entity(self, resource, desired_entity, current_entity, params, forem
                     sort_key = 'name'
                 else:
                     sort_key = list(value[0].keys())[0]
+                # special handling for parameters created by ParametersMixin
```
evgeni, line 1096: @mdellweg how much of a decontamination shower do you need after reading this?

evgeni: also should probably move that up by 4 lines so it's not in the middle of the sorting logic

mdellweg: image

Oh dear. Part of me is wondering if this should be handled by the mixin somehow. But i guess that would be a huge architectural change now.
Also I'm a bit confused:
1. The use of `value` and `new_value` seems mixed up here.
2. Should we not drop all the list entries that are the same in `new_value` and `old_value` at this point?

evgeni: > Part of me is wondering if this should be handled by the mixin somehow. But i guess that would be a huge architectural change now.

I tried, but couldn't come up with a solution that would actually work.
Can certainly try again, now that most nightmares of the first implementation round are over.

>   1. The use of `value` and `new_value` seems mixed up here.

How so?

The idea was:
1. take the `new_value` (aka: the parameters the user has set in their playbook) and iterate over it (thus, looking at each parameter individually)
2. find the same parameter in the `old_value` (aka: the one known to the server)
3. when found, remove all parameter-attributes that are set in the old but not the new version of the parameter
4. let the normal flow of the `update_entity` compare the values.

>   2. Should we not drop all the list entries that are the same in `new_value` and `old_value` at this point?

You mean `old_param` and `new_param`?
We could, but I don't think there is a big benefit. Actually, I think this might break things when we don't pass at least `name` and `value` -- no idea how the API behaves then.

mdellweg: > > Part of me is wondering if this should be handled by the mixin somehow. But i guess that would be a huge architectural change now.
> 
> I tried, but couldn't come up with a solution that would actually work. Can certainly try again, now that most nightmares of the first implementation round are over.

Not for me anyway...

> > 1. The use of `value` and `new_value` seems mixed up here.
> 
> How so?
> 
> The idea was:
> 
>     1. take the `new_value` (aka: the parameters the user has set in their playbook) and iterate over it (thus, looking at each parameter individually)
> 
>     2. find the same parameter in the `old_value` (aka: the one known to the server)
> 
>     3. when found, remove all parameter-attributes that are set in the old but not the new version of the parameter
> 
>     4. let the normal flow of the `update_entity` compare the values.

We create a (shallow) copy `new_value` of `value` in line 1086 in case we want to overwrite it. But we still use `value` in line 1095 for example.
The shallow here might be an actual issue when we pop anything from within `new_value`, for `value` being the thing ansible will report in the details about the modules use later.

> > 2. Should we not drop all the list entries that are the same in `new_value` and `old_value` at this point?
> 
> You mean `old_param` and `new_param`? We could, but I don't think there is a big benefit. Actually, I think this might break things when we don't pass at least `name` and `value` -- no idea how the API behaves then.

What if the user specified one changed and one unchanged parameter? Will we send both to the update call? Should we?

For better understanding, the old_value looks like this?
```
"parameters":[{"priority":60,"created_at":"2024-10-08
        11:08:07 UTC","updated_at":"2024-10-08 11:08:21 UTC","id":30,"name":"subnet_param1","parameter_type":"string","associated_type":"host
        group","hidden_value?":false,"value":"new_value1"},{"priority":60,"created_at":"2024-10-08
        11:08:21 UTC","updated_at":"2024-10-08 11:08:21 UTC","id":38,"name":"subnet_param3","parameter_type":"string","associated_type":"host
        group","hidden_value?":false,"value":"value3"}]
```
(Found it in the host-group. It is named "parameters" there...)

evgeni: > We create a (shallow) copy `new_value` of `value` in line 1086 in case we want to overwrite it. But we still use `value` in line 1095 for example.
> The shallow here might be an actual issue when we pop anything from within `new_value`, for `value` being the thing ansible will report in the details about the modules use later.

Yeah. I think the idea was (but the shallow copy slightly breaks that): we use `new_value` for all the comparison magic (including partially converting things, because, Ansible), but *iff* we decided that the new value is actually new and needs to be submitted, we take the *raw* `value` that the user provided.

> What if the user specified one changed and one unchanged parameter? Will we send both to the update call? Should we?

We have to. Otherwise the API would delete the unsent parameter. Or so I expect it to.



> For better understanding, the old_value looks like this?

Yes

mdellweg (approved):
If you want, you can make the comment even more scary.
And maybe discuss how we need to identify if any of the parameters need attention, at which point we need to "touch them all".

PR #1788, opened 2024-10-17
allow overriding pytest command

evgeni:


Makefile
```diff
@@ -18,7 +18,8 @@ COLLECTION_COMMAND ?= ansible-galaxy
 SANITY_OPTS = --venv
 TEST =
 FLAGS =
-PYTEST = pytest -n 4 --forked -vv
+PYTEST_COMMAND ?= pytest -vv
```
ekohl: Does it make sense to have one variable just for `pytest` and another for the options?

Note pytest natively already has `PYTEST_ADDOPTS` so you can use
```suggestion
PYTEST_COMMAND ?= pytest
PYTEST_ADDOPTS ?= -vv
```

evgeni: I can never remember, are these then natively available in shells spawned by make, or do they need some kind of exporting?

ekohl: That's a good question.

ekohl: Submitted too early, but the answer is no:
```console
$ cat Makefile 
PYTEST_ADDOPTS ?= -vv

default:
	env
$ make | grep PYTEST
$ 
```

evgeni: Then I don't think it gives us much benefit to split those up

ekohl: To me the biggest benefit is that you don't need to think about `-v` or `-vv` when overriding it.

evgeni: k, found a rather clean way to get what we both wanted :)

ekohl (approved):

PR #1789, opened 2024-10-28
use ForemanApi from apypie

evgeni:
includes #1797

### DIFF
--- a/plugins/module_utils/foreman_helper.py
+++ b/plugins/module_utils/foreman_helper.py
@@ -13,7 +13,6 @@ import json
 import os
 import operator
 import re
-import time
 import traceback
 
 from contextlib import contextmanager
@@ -397,7 +396,6 @@ class ForemanAnsibleModule(AnsibleModule):
             self.fail_json(msg="The server URL needs to be either HTTPS or HTTP!")
 
         self.task_timeout = 60
-        self.task_poll = 4
 
         self._thin_default = False
         self.state = 'undefined'
@@ -608,12 +606,12 @@ class ForemanAnsibleModule(AnsibleModule):
         that are required by the module.
         """
 
-        self.foremanapi = apypie.Api(
+        self.foremanapi = apypie.ForemanApi(
             uri=self._foremanapi_server_url,
             username=to_bytes(self._foremanapi_username),
             password=to_bytes(self._foremanapi_password),
-            api_version=2,
             verify_ssl=self._foremanapi_validate_certs,
+            task_timeout=self.task_timeout,
         )
 
         _status = self.status()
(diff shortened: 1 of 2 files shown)


plugins/module_utils/_apypie.py
```diff
@@ -921,3 +921,188 @@ def path_with_params(self, params=None):
...
+
+        :param resource: Plural name of the api resource to show
+        :type resource: str
+        :param search: Search string as accepted by the API to limit the results
+        :type search: str, optional
+        :param params: Lookup parameters (i.e. parent_id for nested entities)
+        :type params: Union[dict,None], optional
+        """
+        PER_PAGE = 2 << 31
+        payload = {'per_page': PER_PAGE}
```
ekohl: Foreman supports `per_page=all`, but Katello has its own mechanism using `full_result=true`. I started https://github.com/Katello/katello/pull/11126 to unify this, but I don't think that's ready to be merged any time soon.

plugins/module_utils/foreman_helper.py
```diff
@@ -700,18 +677,7 @@ def list_resource(self, resource, search=None, params=None):
...
         """
 
-        if params is None:
-            params = {}
-        else:
-            params = params.copy()
-
-        if search is not None:
-            params['search'] = search
-        params['per_page'] = PER_PAGE
```
ekohl, line 710: The constant definition is now unused.

plugins/module_utils/_apypie.py
```diff
@@ -921,3 +921,188 @@ def path_with_params(self, params=None):
...
+    def show(self, resource, resource_id, params=None):
+        """
+        Execute the ``show`` action on an entity.
+
+        :param resource: Plural name of the api resource to show
+        :type resource: str
+        :param resource_id: The ID of the entity to show
+        :type resource_id: int
+        :param params: Lookup parameters (i.e. parent_id for nested entities)
+        :type params: Union[dict,None], optional
```
ekohl: There's also `Optional[dict]` according to https://docs.python.org/3/library/typing.html#typing.Optional so is this also sufficient?
```suggestion
        :type params: Optional[dict]
```

plugins/module_utils/foreman_helper.py
```diff
@@ -1024,7 +990,7 @@ def _validate_supported_payload(self, resource, action, payload):
         :return: The payload as it can be submitted to the API
         :rtype: dict
         """
-        filtered_payload = self._resource_prepare_params(resource, action, payload)
+        filtered_payload = self.foremanapi._resource_prepare_params(resource, action, payload)
```
ekohl: Calling a private API isn't very nice.

ekohl (commented):
Consider this some of the notes I made while reading this. I realize it's very much a work in progress and should result in a real PR to apypie itself before updating the vendored copy.

plugins/module_utils/foreman_helper.py
evgeni: That's not correct ;)
```
plugins/modules/activation_key.py:from ansible_collections.theforeman.foreman.plugins.module_utils.foreman_helper import KatelloEntityAnsibleModule, PER_PAGE
```

plugins/module_utils/_apypie.py
evgeni: Yeah, that's the same.

plugins/module_utils/foreman_helper.py
evgeni: so let's not do that! :)
(fixed)

ekohl (approved):

PR #1792, opened 2024-10-30
set collections_path in ansible.cfg for tests

evgeni:
devel (2.19+) dropped support for the old, deprecated collections_paths


ansible.cfg
```diff
@@ -1,5 +1,6 @@
 [defaults]
 collections_paths = build/collections
```
evgeni, line 3: this is still needed for 2.9 -- can be dropped once that is gone

PR #1793, opened 2024-11-07
Drop support for Ansible 2.9

evgeni:


ekohl (approved):
Any particular reason you chose 2.10 as the new lower bound?

evgeni:
> Any particular reason you chose 2.10 as the new lower bound?

The fact that this is the version shipped in Ubuntu 22.04 and Debian 11: https://canidepend.com/#ansible

(Yes, we're getting rid of D11 soon, and you *can* have 2.12 in U22, but small steps…)

PR #1796, opened 2024-11-19
support bare "from typing" imports that are not guarded

evgeni:


evgeni:
This produces no changes when used with the currently vendored `0.4.0` of `apypie` and is just here to support https://github.com/Apipie/apypie/pull/182

ekohl (approved):

PR #1802, opened 2024-11-25
use the official Ansible sanity action

evgeni:


.github/workflows/main.yml
```diff
@@ -12,6 +12,31 @@ concurrency:
...
+  sanity:
+    name: Sanity (Ⓐ${{ matrix.ansible }})
+    strategy:
+      matrix:
+        ansible:
+          - stable-2.15
+          - stable-2.16
+          - stable-2.17
+          - stable-2.18
+          - devel
```
evgeni, line 24: Before we ran sanity on 2.12+, but the way the action executes it it doesn't pass on older 2.14 and older, and I really don't care about sanity for such old versions.

(Mind you: we still call CI for them, just don't adhere to outdated development best practices anymore, which sounds sane (lol) to me)

ehelms (approved):

PR #1808, opened 2024-12-05
drop fallback to Host API when Reports API fails

evgeni:
The results the Host API returns are different from the Reports API, so
when the user asks for Reports API (via `use_reports_api`), they should
get exactly that, as otherwise their Playbooks/Roles might miss-behave
given different data is fed into them.

### DIFF
--- a/changelogs/fragments/drop-host-api-fallback.yml
+++ b/changelogs/fragments/drop-host-api-fallback.yml
@@ -1,0 +1,2 @@
+bugfixes:
+  - inventory - Drop fallback to Host API when Reports API fails, as this leads to possibly wrong data being used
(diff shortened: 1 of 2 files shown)


mdellweg (approved):

changelogs/fragments/drop-host-api-fallback.yml
```diff
@@ -0,0 +1,2 @@
+bugfixes:
```
evgeni, line 1: I wasn't sure about this. Conceptually it's a bugfix (you really do not want wrong data when running stuff), but it's also technically a breaking change :see_no_evil: 

Realistically, this is only a breaking change for users who
- do not have `foreman_ansible` installed
- have `use_reports_api: true`
- have ignored the "Failed to use Reports API, falling back to Hosts API" warning ever since
:woman_shrugging:

mdellweg: And additionally if that thing ever triggered, there inventory ran into the wrong data (format) and presumably failed anyway down the line? IMHO It definitely is a bugfix. It breaks bug to bug compatibility, right?

PR #1813, opened 2024-12-23
Use --no-proxy in favor of -Y off for wget

maximiliankolb:
I created this PR while looking at the commit in foreman and then checking all my local git repositories. I do not know how/when the YAML files that I've edited are used. I might need some guidance if GHA fail and my PR needs more work.

* Refs [38067](https://projects.theforeman.org/issues/38067) in redmine
* Refs [87652e585755381bb4601461a5b2811b93c6d191](https://github.com/theforeman/foreman/commit/87652e585755381bb4601461a5b2811b93c6d191) in foreman


evgeni:
I don't think we need this. Those files are recorded fixtures from Foreman. Whenever we record them new, they will be updated. Until then the old reply is fine for CI and nothing else uses that file.

PR #1816, opened 2025-01-20
Add 'report_diff' key in result data in case of available diff data

rpluem-vf:
Enrich the data for config reports with unified diffs in case the Ansible tasks provides diff data. This is part of a series of PR's to enable the Ansible diff mode on Foreman.

### DIFF
--- a/plugins/callback/foreman.py
+++ b/plugins/callback/foreman.py
@@ -110,6 +110,7 @@ from datetime import datetime
 from collections import defaultdict
 import json
 import time
+import re
 
 try:
     import requests
@@ -121,36 +122,6 @@ from ansible.module_utils.common.json import AnsibleJSONEncoder
 from ansible.module_utils.common.text.converters import to_text
 from ansible.module_utils.parsing.convert_bool import boolean as to_bool
 from ansible.plugins.callback import CallbackBase
-
-
-def build_log_foreman(data_list):
-    """
-    Transform the internal log structure to one accepted by Foreman's
-    config_report API.
-    """
-    for data in data_list:
-        result = data.pop('result')
-        task = data.pop('task')
-        result['failed'] = data.get('failed')
-        result['module'] = task.get('action')
-        if data.get('failed'):
-            level = 'err'
-        elif result.get('changed'):
-            level = 'notice'
-        else:
-            level = 'info'
-
-        yield {
-            "log": {
-                'sources': {
-                    'source': task.get('name'),
-                },
-                'messages': {
-                    'message': json.dumps(result, sort_keys=True, cls=AnsibleNoVaultJSONEncoder),
-                },
-                'level': level,
-            }
-        }
 
 
 def get_time():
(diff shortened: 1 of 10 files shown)


evgeni:
Why would you alter the data here instead of passing before/after verbatim to Foreman and let the UI render it in whichever way it likes?

rpluem-vf:
> Why would you alter the data here instead of passing before/after verbatim to Foreman and let the UI render it in whichever way it likes?

This saves space. If you have a 100 KB file and change only one line you have to save 200 KB in the report (100 KB for the file before the change, 100 KB for the file after the change). If you have this for thousands of hosts and a number of larger configuration files this sums up.

rpluem-vf:
> > Why would you alter the data here instead of passing before/after verbatim to Foreman and let the UI render it in whichever way it likes?
> 
> This saves space. If you have a 100 KB file and change only one line you have to save 200 KB in the report (100 KB for the file before the change, 100 KB for the file after the change). If you have this for thousands of hosts and a number of larger configuration files this sums up.

UI wise this uses what is used for the Puppet data to render which only delivers the unified diffs. Plus `difflib` is part of core Python whereas in Ruby a 3rd party GEM (https://rubygems.org/gems/diff-lcs/) would be required to generate the needed diff. But you are correct that providing the unified diff leaves less render options for the UI. I guess it is a kind of preference which way to go. Just let men know.

evgeni:
The Ansible callback here is just "the messenger", we (well, you ;) ) can adapt to whatever foreman/foreman_ansible (I don't recall which one carries the report parsing these days) expects. If it's easier (and cheaper) to follow the Puppet path, so be it.

BTW, I always thought that Ansible already provided a unified diff in the `diff` attribute. But my memory might be missguided (I know we only do before/after in our own modules, and the normal Ansible cli shows a unified diff fine, but I have no idea who actually generates it)

rpluem-vf:
> The Ansible callback here is just "the messenger", we (well, you ;) ) can adapt to whatever foreman/foreman_ansible (I don't recall which one carries the report parsing these days) expects. If it's easier (and cheaper) to follow the Puppet path, so be it.
> 
> BTW, I always thought that Ansible already provided a unified diff in the `diff` attribute. But my memory might be missguided (I know we only do before/after in our own modules, and the normal Ansible cli shows a unified diff fine, but I have no idea who actually generates it)

I was surprised as well. I will have a look at my code again because there seems to be a [function](https://github.com/ansible/ansible/blob/8c5e33cd3aa604124234d407044d317a028e84ed/lib/ansible/plugins/callback/__init__.py#L349) already for callbacks:  and it is used [here](https://github.com/ansible/ansible/blob/devel/lib/ansible/plugins/callback/minimal.py#L78) and [here](https://github.com/ansible/ansible/blob/devel/lib/ansible/plugins/callback/default.py#L225).

rpluem-vf:
I have two further PR's that are unrelated to this new feature, but fix issues that cause tests to fail:

https://github.com/theforeman/foreman-ansible-modules/pull/1819
https://github.com/theforeman/foreman-ansible-modules/pull/1820

@evgeni Can you have a look?

Once they are in I would rebase this PR. Furthermore I rewrote the code to use the [function](https://github.com/ansible/ansible/blob/8c5e33cd3aa604124234d407044d317a028e84ed/lib/ansible/plugins/callback/__init__.py#L349) I mentioned above to generate the diff instead of my own. It is more generic and covers cases that I did not consider so far.

rpluem-vf:
PR's for this feature in other repositories:

https://github.com/theforeman/smart_proxy_ansible/pull/98
https://github.com/theforeman/foreman_ansible/pull/749
https://github.com/theforeman/foreman-documentation/pull/3613

evgeni:
I have merged the test cleanup PRs, thanks for that!

The diff here looks quite reasonable, but I guess we should first get movement on the receiving end (foreman_ansible) for proceeding further?

rpluem-vf:
> I have merged the test cleanup PRs, thanks for that!
> 

As promised I just pushed a rebased commit with tests that makes use of the  existing [function](https://github.com/ansible/ansible/blob/8c5e33cd3aa604124234d407044d317a028e84ed/lib/ansible/plugins/callback/__init__.py#L349) 

> The diff here looks quite reasonable, but I guess we should first get movement on the receiving end (foreman_ansible) for proceeding further?

I think there is no hard dependency. If the updated callback runs against an older Foreman there are just no diffs in the data coming from Ansible. And even if there are as I added a new key `report_diff` the existing reporting UI would just ignore it.

rpluem-vf:
I see that there are failures currently. I will take care of them and then you can have a look. Stay tuned.

rpluem-vf:
@evgeni : Now all is green :-)

rpluem-vf:
Any update?

evgeni:
I think the diff looks fine, but I'd prefer https://github.com/theforeman/smart_proxy_ansible/pull/98 and https://github.com/theforeman/foreman_ansible/pull/749 being merged first.

rpluem-vf:
> I think the diff looks fine, but I'd prefer [theforeman/smart_proxy_ansible#98](https://github.com/theforeman/smart_proxy_ansible/pull/98) and [theforeman/foreman_ansible#749](https://github.com/theforeman/foreman_ansible/pull/749) being merged first.

Understood. Any idea how we can push these to get merged?

rpluem-vf:
The last push was just a rebase that solved a conflict.

rpluem-vf:
The last push was just a rebase that solved a conflict in [tests/test_callback.py](https://github.com/theforeman/foreman-ansible-modules/blob/4c611fa7df036df4f6c62c6182e4ad454ac1bca3/tests/test_callback.py#L45).

PR #1819, opened 2025-01-21
Fix exception when serializing results to json

rpluem-vf:
Fix the 'Object of type AnsibleVaultEncryptedUnicode is not JSON serializable' exception when serializing the results to json.

This happens when executing make test TEST="tests/test_callback.py"

### DIFF
--- a/plugins/callback/foreman.py
+++ b/plugins/callback/foreman.py
@@ -146,7 +146,7 @@ def build_log_foreman(data_list):
                     'source': task.get('name'),
                 },
                 'messages': {
-                    'message': json.dumps(result, sort_keys=True),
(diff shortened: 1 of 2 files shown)


evgeni:
The diff makes total sense, but I can't reproduce the Exception without the patch.

Do you have any other Ansible config present that would make it report those facts?

evgeni:
ooooh, I can! it just doesn't fail the run :exploding_head:

evgeni (approved):

evgeni:
and https://github.com/theforeman/foreman-ansible-modules/pull/1822 should prevent such mistakes from happening again

thanks!

PR #1822, opened 2025-01-29
also assert that there are no warnings in callback tests

evgeni:


tests/conftest.py
```diff
@@ -110,3 +115,13 @@ def get_ansible_version():
...
+
+
+def assert_no_warnings(run):
+    for event in run.events:
+        # check for play level warnings
+        assert not event.get('event_data', {}).get('warning', False)
+
+        # check for task level warnings
+        event_warnings = [warning for warning in event.get('event_data', {}).get('res', {}).get('warnings', []) if warning not in IGNORED_WARNINGS]
+        assert [] == event_warnings, str(event_warnings)
```
ekohl: I see you're just moving it around, but any reason you're not just trusting:
```suggestion
        assert not event_warnings
```

evgeni: I see no reason why this was done, no :)

ekohl (approved):

PR #1825, opened 2025-02-17
implement Kerberos auth

evgeni:


plugins/module_utils/_apypie.py
```diff
@@ -248,6 +248,14 @@ def _prepare_route_params(self, input_dict):
 except ImportError:
     pass
 
+try:
```
evgeni, line 251: Those changes come from https://github.com/Apipie/apypie/pull/204

ehelms (approved):
Untested

PR #1833, opened 2025-03-18
[WIP] Remove oVirt

stejskalleos:
Main tracker: https://projects.theforeman.org/issues/38266
Target version: Foreman 3.16

### DIFF
--- a/plugins/module_utils/foreman_helper.py
+++ b/plugins/module_utils/foreman_helper.py
@@ -764,7 +764,7 @@ class ForemanAnsibleModule(AnsibleModule):
         return [self.find_puppetclass(name, **kwargs) for name in names]
 
     def find_cluster(self, name, compute_resource):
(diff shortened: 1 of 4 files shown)


evgeni:
We support older versions of Foreman, so no, we won't remove oVirt support just yet.

stejskalleos:
> We support older versions of Foreman, so no, we won't remove oVirt support just yet.

I didn't know that. Do we have information about supported Foreman versions somewhere?

evgeni:
Officially: https://github.com/theforeman/foreman-ansible-modules?tab=readme-ov-file#supported-foreman-and-plugins-versions

Unofficially, we've been trying to keep compatibility as long as possible (= no cost incurred/no improvement blocked)

evgeni:
And to come back to this: oVirt was extracted to a plugin (see https://community.theforeman.org/t/rfc-deprecation-and-removal-of-ovirt-compute-resource/42527) and thus we can keep ovirt support in FAM

PR #1836, opened 2025-03-21
don't mangle basestring

evgeni:
apypie dropped that code, so we don't need to mangle it during vendoring


evgeni:
https://github.com/Apipie/apypie/pull/183 dropped it

ekohl (approved):

PR #1839, opened 2025-03-21
simplify inventory example to make ansible-lint pass

evgeni:


evgeni:
replaces https://github.com/theforeman/foreman-ansible-modules/pull/1742 as that is not compatible with all ansible versions

PR #1847, opened 2025-04-24
repository - add `rhel-10` to os version filter choices

evgeni:


evgeni:
failures are unrelated due to changes in ansible 2.19/devel

ehelms (approved):

PR #1849, opened 2025-04-25
Add support for setting metadata_expire via the repository module

DaTurr3t:
This PR adds support for setting the metadata_expire option for yum repositories.
The version_added attribute is still missing because I am unsure which version I'd have to set it to. Should that be `version_added: 5.4.0` or already `5.5.0`?


evgeni:
5.4.0 is correct, that will be the next release

evgeni (approved):

evgeni:
Thanks!

PR #1851, opened 2025-05-02
Document the LCE param to registration_command ignored since Katello 4.12

evgeni:


plugins/modules/registration_command.py
```diff
@@ -65,6 +65,7 @@
   lifecycle_environment:
     description:
     - Lifecycle environment for the host.
+    - Ignored since Katello 4.12.
```
ekohl: Is this because a host can now be in multiple LCEs? Is there a replacement we can point to?

evgeni: The AK drives the LCE(s)

ekohl: Can we include that in the message?

evgeni: Sure

ekohl (approved):

plugins/modules/registration_command.py
```diff
@@ -65,6 +65,7 @@
   lifecycle_environment:
     description:
     - Lifecycle environment for the host.
+    - Ignored since Katello 4.12, where the Lifecycle Environment is derived from the Activation Key.
```
ekohl: According to @jeremylenz users could always set the LCE via AK so we should steer users away even stronger, also on older releases.

evgeni: Have any better wording suggestions?
Maybe something like
- The LCE.
- normally unused, because it gets set by the ak
- ignored since 4.12

ekohl: How about something like

* The LCE
* Deprecated: prefer the AK
* Removed since Katello 4.12

evgeni: Okay, I ended up with:

- Deprecated: The Lifecycle Environment of a Host should be set by the Activation Key.
- Removed from the API since Katello 4.12.

ekohl (approved):

PR #1854, opened 2025-05-12
update documentation of the hostgroup module

### LINKED ISSUE #1853: Hostgroup module, parent parameter
##### SUMMARY
Can we provide more information to the parent parameter in the hostgroup module ? 
https://docs.ansible.com/ansible/latest/collections/theforeman/foreman/hostgroup_module.html

it says: "Hostgroup parent name"

Can we provide a full path hierarchy like this : `group_a/group_a2/group_a23` ? 

##### ISSUE TYPE
 - Feature Idea

evgeni:
- document nested parents
- update AK to be an own module parameter

Fixes: #1853

### DIFF
--- a/plugins/modules/hostgroup.py
+++ b/plugins/modules/hostgroup.py
@@ -42,7 +42,9 @@ options:
     required: false
     type: str
   parent:
-    description: Hostgroup parent name
+    description:
+      - Title of the parent Hostgroup.
+      - Can be nested in the form "Main Group/Sub Group".
     required: false
     type: str
   organization:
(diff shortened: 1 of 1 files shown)


plugins/modules/hostgroup.py
```diff
@@ -125,9 +127,7 @@
     content_source: capsule.example.com
     lifecycle_environment: "Production"
```
ekohl, line 128: If an AK is given, is the LCE still used?

evgeni: for registration, no, to be able to select the right kickstart repo, yes.

ehelms (approved):

PR #1866, opened 2025-05-26
implement content view environment support for hosts

evgeni:


evgeni:
The way facets are done totally confuses our flattening logic, so I am not sure how to easily fix that :/

PR #1870, opened 2025-06-04
Add content view type rolling

m-bucher:
Add new Katello content view type `rolling`: https://github.com/Katello/katello/pull/11240

## Summary by CodeRabbit

- **New Features**
  - Added support for a new "rolling" option when managing content views, allowing users to create or update rolling content views.
- **Documentation**
  - Updated module documentation and examples to include the new "rolling" parameter.
- **Tests**
  - Introduced new test cases and fixtures to validate rolling content view functionality, including creation and deletion scenarios.

### DIFF
--- a/changelogs/fragments/content_view-add_rolling.yml
+++ b/changelogs/fragments/content_view-add_rolling.yml
@@ -1,0 +1,4 @@
+---
+minor_changes:
+  - content_view - add ``rolling``-flag to create a Rolling Content View
+...
--- a/plugins/modules/content_view.py
+++ b/plugins/modules/content_view.py
@@ -73,6 +73,11 @@ options:
       - Designate this Content View for importing from upstream servers only.
     type: bool
     version_added: 3.14.0
+  rolling:
+    description:
+      - A rolling content view contains latest synced state of repositories.
+    type: bool
+    version_added: 5.5.0
   composite:
     description:
       - A composite view contains other content views.
@@ -133,6 +138,18 @@ EXAMPLES = '''
         content_view_version: 1.0
       - content_view: Internal CV
         latest: true
+
+- name: "Create or update rolling content view"
+  theforeman.foreman.content_view:
+    username: "admin"
+    password: "changeme"
+    server_url: "https://foreman.example.com"
+    name: "Fedora RCV"
+    organization: "My Cool new Organization"
+    rolling: true
+    repositories:
+      - name: 'Fedora 26'
+        product: 'Fedora'
 '''
 
 RETURN = '''
@@ -170,6 +187,7 @@ def main():
             description=dict(),
             label=dict(),
             composite=dict(type='bool', default=False),
+            rolling=dict(type='bool'),
             auto_publish=dict(type='bool', default=False),
             solve_dependencies=dict(type='bool'),
             import_only=dict(type='bool'),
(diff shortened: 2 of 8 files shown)


plugins/modules/content_view.py
```diff
@@ -170,6 +188,7 @@ def main():
             description=dict(),
             label=dict(),
             composite=dict(type='bool', default=False),
+            rolling=dict(type='bool', default=False),
```
evgeni: I guess the API defaults to false if nothing is sent, right? Then you can omit it here which will allow us to skip re-recording VCR and won't trigger "empty" updates when the users run this the first time.

evgeni (changes requested):

plugins/modules/content_view.py
m-bucher: I guess, yes.
However, I would like to add a test for the option. However, it looks like I not only have to re-record the VCR-cassette but also update /tests/fixtures/apidoc/katello.json :-/
Is there a command to this like with the VCRs?

evgeni: Just spin up a katello box and download the apidoc, there is no command for it (but if you contribute one, there will be ;) )

plugins/modules/content_view.py
```diff
@@ -73,6 +73,12 @@
       - Designate this Content View for importing from upstream servers only.
     type: bool
     version_added: 3.14.0
+  rolling:
+    description:
+      - A rolling content view contains latest synced state of repositories.
+    default: false
```
evgeni:
```suggestion
```

evgeni:
updated katello.json in https://github.com/theforeman/foreman-ansible-modules/pull/1877, please rebase

plugins/modules/content_view.py
evgeni, line 1: sanity complains that you changed this file to be executable, please don't :)

plugins/modules/content_view.py
```diff
@@ -73,6 +73,11 @@
       - Designate this Content View for importing from upstream servers only.
     type: bool
     version_added: 3.14.0
+  rolling:
+    description:
+      - A rolling content view contains latest synced state of repositories.
+    type: bool
+    version_added: 3.16.0
```
evgeni:
```suggestion
    version_added: 5.5.0
```

that's the collection version, not the katello version :)

plugins/modules/content_view.py
m-bucher: Sorry, my `vim`-config automagically makes python-scripts executable :see_no_evil:

evgeni:
LGTM!

Can you please add a changelog snippet like https://github.com/theforeman/foreman-ansible-modules/blob/effd9a3c0d2b7d6bcc7e4324c942526903f7b4a2/changelogs/fragments/repository-retain_package_versions_count.yml?

@coderabbitai review

evgeni:
@coderabbitai review

evgeni:
Sorry for the CodeRabbit spam @m-bucher, could not not try it on a "foreign" PR, and was sufficiently disappointed :)

changelogs/fragments/content_view-add_rolling.yml
```diff
@@ -0,0 +1,4 @@
+---
+minor_change:
```
evgeni:
```suggestion
minor_changes:
```

evgeni (changes requested):

evgeni (approved):

PR #1879, opened 2025-07-03
Harden release flow by using the gh release tool

ekohl:
<del>Ansible has the GALAXY_TOKEN configuration option (since https://github.com/ansible/ansible/commit/576335e53dc7721811a9dfe02b29a36ec7f27e56), which is exposed via the ANSIBLE_GALAXY_TOKEN environment variable. This means we don't need to pass it on the command line and can instead pass it in securely.</del> This was dropped in https://github.com/ansible/ansible/commit/83909bfa22573777e3db5688773bda59721962ad.

It avoids the use of an external action to create the release and instead uses the `gh` command line tool. This reduces the attack surface in case the action gets compromised.

Disclaimer: untested.

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -28,7 +28,6 @@ jobs:
       - name: Deploy Ansible Collection
         run: make publish GALAXY_API_KEY=${{ secrets.GALAXY_API_KEY }}
       - name: Create GitHub Release
-        uses: softprops/action-gh-release@v2
-        with:
-          files: theforeman-foreman-*.tar.gz
-          body_path: release-changelog.txt
+        env:
+          GH_TOKEN: ${{ github.token }}
+        run: gh release create ${{ github.ref_name }} -F release-changelog.txt theforeman-foreman-*.tar.gz


.github/workflows/release.yml
```diff
@@ -26,9 +26,10 @@ jobs:
       - name: Build Changelog
         run: antsibull-changelog generate --output release-changelog.txt --only-latest
       - name: Deploy Ansible Collection
-        run: make publish GALAXY_API_KEY=${{ secrets.GALAXY_API_KEY }}
+        run: make publish
+        env:
+          ANSIBLE_GALAXY_TOKEN: ${{ secrets.GALAXY_API_KEY }}
```
evgeni: I don't think `ANSIBLE_GALAXY_TOKEN` is a thing :/

ekohl: Reading https://github.com/ansible/ansible/commit/576335e53dc7721811a9dfe02b29a36ec7f27e56 I see in the commit message:
> This change lets user store token in configuration file or environment to prevent exposing the secret on the command line.

In `base.yml` I see `ANSIBLE_GALAXY_TOKEN` as an `env` var so am I reading it wrong?

evgeni: That patch? No, you're reading it all right.
What you're missing is https://github.com/ansible/ansible/commit/83909bfa22573777e3db5688773bda59721962ad which removed that again.

https://docs.ansible.com/projects/ansible/latest/reference_appendices/config.html doesn't mention `ANSIBLE_GALAXY_TOKEN`

ekohl: Of course, it can't be easy.

ekohl:
What should we do with this PR? Want me to trim it down to only using `gh` instead of `softprops/action-gh-release@v2` or close it altogether?

evgeni:
I'd take the `gh` change, yeah

ekohl:
I've dropped the commit for Ansible. I'm very happy with past me who already created 2 separate commits.

evgeni (approved):

PR #1880, opened 2025-07-04
Use proper composite property when creating a CCV

pablomh:


evgeni:
Great find!
This needs the fixtures re-recorded (see https://github.com/theforeman/foreman-ansible-modules/blob/develop/docs/testing.md)
If you want, I can help you setting you up for recording, or just do it for you :)

pablomh:
Thinking about it later, I thought that maybe the intention was to use the existence of `components` as a boolean value as well (as in "if we're setting `components`, this needs to be a composite one).

But my use case was to create initially an empty CCV, so I think it should be covered as well :)

roles/content_views/tasks/_create_content_view.yml
```diff
@@ -9,7 +9,7 @@
     name: "{{ content_view.name }}"
     auto_publish: "{{ content_view.auto_publish | default(omit) }}"
     components: "{{ content_view.components | default(omit) }}"
-    composite: "{{ content_view.components | default(false) | ternary(true, false) }}"
+    composite: "{{ content_view.composite | default(false) | ternary(true, false) }}"
```
evgeni:
```suggestion
    composite: "{{ content_view.components | default(content_view.composite) | default(false) | ternary(true, false) }}"
```

does that work for your "empty CCV" case?

pablomh: Give me like 4h and I'll be able to tell you. Thanks!!

pablomh: It's the same code I've ended up using after my re-thought of the issue, so big +1 from me :)

pablomh:
Hi @evgeni, is anything else missing?

pablomh:
I see that the fixtures re-recording (mentioned in https://github.com/theforeman/foreman-ansible-modules/pull/1880#issuecomment-3043794922) is still missing. I'll read about it and will try to do it myself.

evgeni:
No, after our changes no re-recording is needed, this just needs a merge now

evgeni (approved):

PR #1882, opened 2025-07-09
Add Wake on LAN module

ares:


ares:
Eh, how do I set the right author?

dgoetz:
Ansible defines the author as `Name of the module author in the form First Last (@GitHubID)` so it would be `Marek Hulan (@ares)`.

PR #1885, opened 2025-07-11
Add environments for rolling content views

m-bucher:
Add additional parameter to specify lifecycle-environments for rolling Content Views.

Requires https://github.com/Katello/katello/pull/11407

Draft until the Katello PR is merged.

### DIFF
--- a/plugins/modules/content_view.py
+++ b/plugins/modules/content_view.py
@@ -78,6 +78,14 @@ options:
       - A rolling content view contains latest synced state of repositories.
     type: bool
     version_added: 5.5.0
+  lifecycle_environments:
+    description:
+      - List of lifecycle environments that content of this rolling content view will be promoted to.
+      - Only valid for I(rolling=True).
+    required: False
+    type: list
+    elements: str
+    version_added: 5.9.0
   composite:
     description:
       - A composite view contains other content views.
@@ -192,6 +200,7 @@ def main():
             solve_dependencies=dict(type='bool'),
             import_only=dict(type='bool'),
             components=dict(type='nested_list', foreman_spec=cvc_foreman_spec, resolve=False),
+            lifecycle_environments=dict(type='entity_list', flat_name='environment_ids'),
             repositories=dict(type='entity_list', elements='dict', resolve=False, options=dict(
                 name=dict(required=True),
                 product=dict(required=True),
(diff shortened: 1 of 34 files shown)


mhjacks:
I hit the issue that this would fix; and was happy to see there's a PR that addresses it. Now that the Katello PR is merged, when would be a good time to refresh and merge it?

plugins/modules/content_view.py
```diff
@@ -78,6 +78,14 @@
...
     type: bool
     version_added: 5.5.0
+  lifecycle_environments:
+    description:
+      - List of lifecycle environments that content of this rolling content view will be promoted to.
+      - Ignored if I(rolling=False).
+    required: False
+    type: list
+    elements: str
+    version_added: 5.5.0
```
evgeni:
```suggestion
    version_added: 5.9.0
```

tests/fixtures/apidoc/katello.json
evgeni, line 1: `develop` now contains Katello 4.20-rc1 apidoc, so you can drop that file from the PR

evgeni:
I've asked @ianballou and team to review this

sjha4:
Had claude set me up a nice playbook https://paste.centos.org/view/07c95362 to test the workflows..Worked like a charm with the PR here. Functionality wise this looks good. 👍🏼 

```
"========== TEST SUMMARY ==========",
"✓ Test 1: Rolling CV without LCE - PASSED",
"✓ Test 2: Rolling CV with LCE - PASSED",
"✓ Test 3: Non-rolling CV - PASSED",
"✓ Test 4: Non-rolling CV with LCE (expected fail) - PASSED",
"✓ Test 5: Non-rolling CV without LCE - PASSED",
"=================================="
```

plugins/modules/content_view.py
```diff
@@ -78,6 +78,14 @@
       - A rolling content view contains latest synced state of repositories.
     type: bool
     version_added: 5.5.0
+  lifecycle_environments:
+    description:
+      - List of lifecycle environments that content of this rolling content view will be promoted to.
+      - Ignored if I(rolling=False).
```
sjha4: This is not ignored, rather the failure below "Lifecycle Environments can only be specified for Rolling Content View." is expected?

evgeni: Good catch, how about:

```suggestion
      - Only valid for I(rolling=True).
```

evgeni:
Oh, and we should have a changelog entry in `changelogs/fragments`, something like:

```yaml
minor_changes:
  - content_view - add support for lifecycle environments in rolling content views
```

evgeni (approved):

PR #1887, opened 2025-07-15
activation_key: workaround to ensure LCE and CV are always sent together

evgeni:
With the introduction of Content View Environments in Katello, you need to either set a CVE or both LCE and CV to update where a host is consuming content from. As we currently don't have a way to correctly set the CVE, this workaround ensures LCE and CV are always sent so that Katello can correctly calculate the CVE for us in the background.


evgeni:
cc @jeremylenz

pablomh:
Yes, please :)

jeremylenz (approved):

PR #1888, opened 2025-07-18
Content export repository

parmstro:
add destination_server option to content_export_repository to match other content_export_* modules.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -15,3 +15,5 @@ docs/plugins/
 docs/_build/
 changelogs/.plugin-cache.yaml
 importer_result.json
+.vscode
+*.code-workspace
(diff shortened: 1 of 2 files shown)


evgeni:
But Katello doesn't support that?

https://github.com/katello/katello/blob/master/app/controllers/katello/api/v2/content_exports_controller.rb

parmstro:
Yes, I see that now and did not test. My apologies, I thought it was merely an oversight as it is listed as simply a description field. 
Aside: If it is simply a description field, it begs the question as to why katello doesn't support it.

evgeni:
That's a fair question, which *I* have no answer to.

parmstro:
Hahaha! After looking at the code, me neither

PR #1889, opened 2025-07-21
Fix idempotency of hostgroup module for child hostgroups.

### LINKED ISSUE #1865: foreman-ansible hostgroup module returns changed when there is no change
##### SUMMARY

When creating a new hostgroup that inherits ansible roles from the parent hostgroup, when a new role is defined for the childgroup (that is not listed in the parent roles), ansible returns "changed" for every run of the playbook. This is a bug, for the second run on the same system ansible should return "ok".

##### ISSUE TYPE
 - Bug Report

##### ANSIBLE VERSION

```
ansible [core 2.16.3]
  configured module search path = ['/root/.ansible/plugins/modules', '/usr/share/ansible/plugins/modules']
  ansible python module location = /usr/lib/python3.12/site-packages/ansible
  ansible collection location = /root/.ansible/collections:/usr/share/ansible/collections
  python version = 3.12.8 (main, Dec 12 2024, 09:41:14) [GCC 8.5.0 20210514 (Red Hat 8.5.0-22)] (/usr/bin/python3.12)
  jinja version = 3.1.2
  libyaml = True
```

##### COLLECTION VERSION

```
# /root/.ansible/collections/ansible_collections
Collection            Version
--------------------- -------
ansible.utils         5.1.2  

# /usr/share/ansible/collections/ansible_collections
Collection            Version
--------------------- -------
ansible.posix         1.5.4  
community.general     8.3.0  
theforeman.foreman    4.0.0  
theforeman.operations 3.0.0  
```

##### KATELLO/FOREMAN VERSION

```
rubygem-katello-4.13.1-152.10.0.atix.el8.noarch
foreman-3.11.2-105.1.1.atix.el8.noarch
```

##### STEPS TO REPRODUCE

Wehn running the second task which creates the nested hostgroup "test" that assigns a new ansible role that is not listed in the parent hostgroup, ansible returns always 'changed' for every run, no matter if there is a change on the endsystem or not. This is a bug, for the second run it should return 'ok'.

```yaml

- hosts: localhost
  vars:
    connection: local
  tasks:
    - name: "Create base Hostgroups"
      theforeman.foreman.hostgroup:
        name: "base"
        server_url: "{{ foreman }}"
        username: "{{ user }}"
        password: "{{ passwd }}"
        organization: atix
        ansible_roles:
          - testrole

- hosts: localhost
  vars:
    connection: local
  tasks:
    - name: "Create nested Hostgroups"
      theforeman.foreman.hostgroup:
        name: "test"
        server_url: "{{ foreman }}"
        username: "{{ user }}"
        password: "{{ passwd }}"
        organization: atix
        ansible_roles:
          - foreman.foreman_open_scap
        parent: "base"
```

##### EXPECTED RESULTS

Expect ansible to return in the second run 'ok'.
##### ACTUAL RESULTS

Ansible always returns 'changed'.

```

```

OttaviaB:
Closes #1865 .

### Description

`theforeman.foreman.hostgroup` was always returning `changed` when using the `ansible_roles` parameter on a child hostgroup. (See the linked issue for a reproducer)

The root cause was that to determine the changed status we were comparing `desired_ansible_role_ids`, which contains only the roles explicitly assigned to the child hostgroup and not the ones inherited from the parent, with `current_ansible_role_ids`, which contains the child-specific as well as the inherited roles.

This led to the status always being `changed`.

### Changes

I have added an `if` condition which expands `desired_ansible_role_ids` to include the parent's roles when necessary.

### DIFF
--- a/plugins/modules/hostgroup.py
+++ b/plugins/modules/hostgroup.py
@@ -198,15 +198,31 @@ def main():
             ensure_puppetclasses(module, 'hostgroup', entity, expected_puppetclasses)
 
         ansible_roles = module_params.get('ansible_roles')
+
+        parent_ansible_role_ids = []
+
         if not module.desired_absent and ansible_roles is not None:
             desired_ansible_role_ids = [item['id'] for item in ansible_roles]
+
+            if entity.get('parent_id') is not None:
+                parent_ansible_role_ids = [
+                    item['id']
+                    for item in module.resource_action(
+                        'hostgroups',
+                        'ansible_roles',
+                        {'id': entity['parent_id']},
+                        ignore_check_mode=True,
+                        record_change=False,
+                    )
+                ]
+
             current_ansible_role_ids = [
                 item['id'] for item in module.resource_action(
                     'hostgroups', 'ansible_roles', {'id': entity['id']},
                     ignore_check_mode=True, record_change=False,
                 )
             ] if old_entity else []
-            if set(current_ansible_role_ids) != set(desired_ansible_role_ids):
+            if set(current_ansible_role_ids) != set(desired_ansible_role_ids + parent_ansible_role_ids):
                 module.resource_action(
                     'hostgroups', 'assign_ansible_roles',
                     {'id': entity['id'], 'ansible_role_ids': desired_ansible_role_ids},
--- a/changelogs/fragments/hostgroup_idempotency.yml
+++ b/changelogs/fragments/hostgroup_idempotency.yml
@@ -1,0 +1,4 @@
+---
+bugfixes:
+  - hostgroup - fix idempotency of hostgroup module when assigning Ansible roles to a hostgroup with a parent hostgroup (https://github.com/theforeman/foreman-ansible-modules/issues/1865)
+...


plugins/modules/hostgroup.py
```diff
@@ -200,6 +200,20 @@ def main():
         ansible_roles = module_params.get('ansible_roles')
         if not module.desired_absent and ansible_roles is not None:
             desired_ansible_role_ids = [item['id'] for item in ansible_roles]
+
+            if entity['ancestry'] is not None:
```
evgeni:
```suggestion
            if entity.get('ancestry') is not None:
```

(for some reason it's sometimes not defined at all)

plugins/modules/hostgroup.py
```diff
@@ -200,6 +200,20 @@ def main():
...
+                    item['id']
+                    for item in module.resource_action(
+                        'hostgroups',
+                        'ansible_roles',
+                        {'id': entity['parent_id']},
+                        ignore_check_mode=True,
+                        record_change=False,
+                    )
+                ]
+                desired_ansible_role_ids = desired_ansible_role_ids + parent_ansible_role_ids
```
evgeni: I think this is wrong, as you'd be calling `assign_ansible_roles` with the parent ones if the lists still don't match.

Instead you should check `if set(current_ansible_role_ids) != set(desired_ansible_role_ids + parent_ansible_role_ids)` below

plugins/modules/hostgroup.py
OttaviaB: Yes, you are absolutely right.

I also noticed that `parent_id` is defined more consistently (even in check mode).
If you don't see any problems with that, I would use `entity.get('parent_id')` instead of `entity.get('ancestry')`.

plugins/modules/hostgroup.py
OttaviaB: You are right here as well.

For what it's worth, in my tests it did not seem to make a difference: it runs `assign_ansible_roles` with the _wrong_ (= too large) `desired_ansible_roles`, but does not throw an error or get the wrong "changed" status due to the extra inherited roles.

Your solution is definitely more elegant, so I changed it to that. (I also initialized `parent_ansible_role_ids` to an empty list.)

plugins/modules/hostgroup.py
evgeni: I never looked much into the whole ancestry stuff, so I'll just go and trust you here :)

evgeni (approved):

PR #1901, opened 2025-09-01
Use ansible.module_utils.common.text.converters

evgeni:
Replace use of old `ansible.module_utils._text` which is deprecated in 2.20+ and will be removed in 2.24
ansible.module_utils.common.text.converters exists since 2.10, so it's safe to just replace our usage with it.


stejskalleos (approved):
:green_apple: **LGTM**

Works as expected, the CentOS Stream 9 host was created from the imported facts.

PR #1904, opened 2025-09-02
fall-back to rpm binary when library can't be imported

evgeni:


plugins/modules/content_upload.py
```diff
@@ -144,6 +153,11 @@ def get_rpm_info(path):
     return (name, epoch, version, release, arch)
 
 
+def get_rpm_info_bin(path):
+    return subprocess.check_output(['rpm', '--nosignature', '--queryformat', '%{NAME} %{EPOCHNUM} %{VERSION} %{RELEASE} %{ARCH}',
+                                    '-qp', path]).decode('ascii').split()
```
evgeni: The following needs to be ported:
```
    arch = to_native(rpmhdr[rpm.RPMTAG_ARCH])
    if arch == 'noarch' and rpmhdr[rpm.RPMTAG_SOURCEPACKAGE] == 1:
        arch = 'src'
```

requirements-dev.txt
```diff
@@ -8,7 +8,7 @@ ansible_runner<2.0; python_version < '3.8'
 ansible_runner; python_version >= '3.8'
 python-debian<0.1.40; python_version < '3.7'
 python-debian; python_version >= '3.7'
-rpm
+rpm; python_version >= '3.12'
```
evgeni, line 11: This ensures both codepaths are tested in CI.

evgeni:
The devel failures are unrelated, the fix is in https://github.com/theforeman/foreman-ansible-modules/pull/1901

plugins/modules/content_upload.py
```diff
@@ -120,6 +123,12 @@ def get_deb_info(path):
 
 
 def get_rpm_info(path):
+    if HAS_RPM_LIB:
+        return get_rpm_info_lib(path)
+    return get_rpm_info_bin(path)
```
stejskalleos, line 128: Should be here the `if HAS_RPM_BIN`?

evgeni: No, because if both are `False`, so will be `HAS_RPM` and then it errors out when running the module at
https://github.com/theforeman/foreman-ansible-modules/blob/eab0fbb769f5b8c11897586cd1d1293bec0fa389/plugins/modules/content_upload.py#L174-L175

stejskalleos (approved):
Haven't tested, but the changes LGTM

PR #1905, opened 2025-09-03
Update CR and image tests

evgeni:


evgeni:
Ansible devel issues unrelated, fix in https://github.com/theforeman/foreman-ansible-modules/pull/1901

shubhamsg199 (approved):
Looks good, Ack.
Non-blocking, we can also update Test Organization and Test Location the same way

stejskalleos (commented):
The changes look good to me, but what was the reason for updating the tests? Were they failing?

evgeni:
>  The changes look good to me, but what was the reason for updating the tests? Were they failing?

In the recorded state they were not failing, but when we run them live (or trying to re-record) we had no control about the CR setup "from the outside" (the vars files we use), so e.g. downstream tests were failing.

>  Non-blocking, we can also update Test Organization and Test Location the same way

Technically, yes, but that name is hard coded in so many other places, that right now it wouldn't be helping much.

stejskalleos (approved):

stejskalleos:
Thanks @evgeni

PR #1907, opened 2025-09-11
don't call task.serialize() on Ansible 2.20+

evgeni:
This was removed in favor of dump_attrs() in https://github.com/ansible/ansible/pull/85724


shubhamsg199 (approved):
ACK

PR #1915, opened 2025-09-24
use instance_hosts not http_proxies in resource_info test

evgeni:
we want to test an unscoped list of entities, but depending on the setup there might be http proxies defined, but we're super sure instance_hosts is always one


shubhamsg199 (approved):
Looks good

PR #1918, opened 2025-10-06
don't record the authorization token in the registration command

evgeni:
while the token won't work for anyone, this makes security scanners not stumble upon this as a false positive leak


stejskalleos (approved):
🍏 CI is green
🍏 No token in the fixtures

LGTM

PR #1923, opened 2025-10-16
Support the `feature` parameter in `job_invocation`

pablomh:


plugins/modules/job_invocation.py
```diff
@@ -127,6 +126,11 @@
     description:
       - Override the description format from the template for this invocation only
     type: str
+  feature:
+    description:
+      - Feature label that should be triggered
+      - A job template assigned to this feature will be used
+    type: str
```
evgeni, line 133:
```suggestion
    type: str
    version_added: 5.8.0
```

evgeni (approved):

PR #1925, opened 2025-10-27
Don't expect changes during content_view_info setup

evgeni:
This was accidentally set to true here, but during setup we do not want to assert any changes, as it might have already happened in a previous run.


stejskalleos (approved):
🍏 LGTM

PR #1934, opened 2025-11-26
Deprecate auto_attach to match Katello

Gauravtalreja1:
Ref https://github.com/Katello/katello/pull/11526


plugins/modules/activation_key.py
```diff
@@ -103,10 +103,6 @@
           - default
         type: str
         required: true
-  auto_attach:
-    description:
-      - Set Auto-Attach on or off
-    type: bool
```
evgeni, line 116: FAM supports older Katello releases, so I'd prefer that instead of removing the parameter we document that it's been removed in Katello 4.20 (or was it 4.19?) and drop it from the examples and tests.

Gauravtalreja1: Fair enough, added this param back with removal note for Katello 4.19

plugins/modules/activation_key.py
```diff
@@ -105,7 +105,7 @@
         required: true
   auto_attach:
     description:
-      - Set Auto-Attach on or off
+      - Set Auto-Attach on or off; auto_attach option has been removed in Katello 4.19
```
```suggestion
      - Set Auto-Attach on or off.
      - The C(auto_attach) parameter is deprecated since Katello 4.13 and not available since Katello 4.19.
```

plugins/modules/activation_key.py
```diff
@@ -228,7 +227,6 @@ def main():
             lifecycle_environment=dict(type='entity', flat_name='environment_id', scope=['organization']),
             content_view=dict(type='entity', scope=['organization']),
             host_collections=dict(type='entity_list', scope=['organization']),
-            auto_attach=dict(type='bool'),
```
evgeni, line 257: It still needs to be defined here.

roles/activation_keys/tasks/main.yml
```diff
@@ -13,7 +13,6 @@
     host_collections: "{{ item.host_collections | default(omit) }}"
     subscriptions: "{{ item.subscriptions | default(omit) }}"
     content_overrides: "{{ item.content_overrides | default(omit) }}"
-    auto_attach: "{{ item.auto_attach | default(omit) }}"
```
evgeni, line 17: please keep for now

evgeni (changes requested):

evgeni:
Why did you re-record user and organization? These expect plain Foreman (not Katello) APIs in our tests.

evgeni (approved):

PR #1935, opened 2025-12-04
activation key: add support for multi cv

martin-schlossarek:
This pull request adds support for multiple content views in activation keys. The tests have been updated accordingly, and the fixtures have been regenerated.


plugins/modules/activation_key.py
```diff
@@ -45,6 +45,12 @@
     description:
       - Name of the content view
     type: str
+  content_view_environments:
+    description:
+      - List of content view environments to add to activation key
+      - Ignored if content_view_id and lifecycle_environment_id are specified
```
```suggestion
      - Ignored if C(content_view) and C(lifecycle_environment) are specified
```
`C()` makes nice formatting, and the params are called without the `_id` here.

But this made me wonder: we can enforce mutual exclusivity in Ansible, so the user gets an error when they pass `content_view` *and* `content_view_environments`. Should we do that?

Then the description here would become something like
```suggestion
      - Mutually exclusive with C(content_view) and C(lifecycle_environment).
```

And around line 280 we'd need to add something like
```python
       mutually_exclusive=[
           ['content_view', 'content_view_environments'],
           ['lifecycle_environment', 'content_view_environments'],
       ],
```

Which would result in an error like:
```
fatal: [localhost]: FAILED! => {"changed": false, "msg": "parameters are mutually exclusive: content_view|content_view_environments"}
```

plugins/modules/activation_key.py
```diff
@@ -227,6 +251,7 @@ def main():
             description=dict(),
             lifecycle_environment=dict(type='entity', flat_name='environment_id', scope=['organization']),
             content_view=dict(type='entity', scope=['organization']),
+            content_view_environments=dict(type='list', elements='str'),
```
evgeni, line 255: When I tried implementing CV Environments for hosts in https://github.com/theforeman/foreman-ansible-modules/pull/1866, I used an "entity list", not a string list here (as CV Environments are entities that you can search for etc).

However, now that you did it differently, I wonder what's the most "correct" way is.

- The user needs to pass in the environment labels in both cases, so the user interface doesn't change.
- Using a "real" entity has the benefit of having a more direct error message if the CVE label can't be found, at the cost of an additional API request (see below)

Error with your code:
```
fatal: [localhost]: FAILED! => {"changed": false, "msg": "Failed to ensure entity state: ForemanApiException: Error while performing create on activation_keys: 422 Client Error: Unprocessable Content for url: https://centos9-stream-katello-nightly.juuni.example.com/katello/api/activation_keys - {'displayMessage': 'No content view environments found with names: RHEL9/blafasel', 'errors': ['No content view environments found with names: RHEL9/blafasel']}"}
```

Error with my code:
```
fatal: [localhost]: FAILED! => {"changed": false, "msg": "Found no results while searching for content_view_environments with label=\"RHEL9/blafasel\""}
```

I think paying an API request just to validate the label is too much, so I'd go with your solution, but wanted to highlight the possibility nevertheless

plugins/modules/activation_key.py
```diff
@@ -45,6 +45,12 @@
     description:
       - Name of the content view
     type: str
+  content_view_environments:
+    description:
+      - List of content view environments to add to activation key
+      - Ignored if content_view_id and lifecycle_environment_id are specified
+    type: list
+    elements: str
```
evgeni, line 53:
```suggestion
    elements: str
    version_added: 5.8.0
```

That way docs render this as a "new" parameter and people don't try to use it on older versions

evgeni:
Thanks a lot Martin! This looks very good!

I left a few inline comments, but they are pretty much cosmetic.

Could you please add a small YAML snippet with a changelog entry in `changelogs/fragments/`? Something like `changelogs/fragments/activation_key-cv_environments.yml`:
```yaml
minor_changes:
  - activation_key - add ``content_view_environments`` parameter to support multi CV (https://github.com/theforeman/foreman-ansible-modules/pull/1935)
```

plugins/modules/activation_key.py
martin-schlossarek: > But this made me wonder: we can enforce mutual exclusivity in Ansible, so the user gets an error when they pass content_view and content_view_environments. Should we do that?

TBH, I just copied the behavior of the API/hammer one-to-one. content-view-envionments will be ignored if you specify a single content-view/lifecycle-environment. No errors. I don't like that too....

The question is: should the ansible collection behave like the API or should it be better? :)

plugins/modules/activation_key.py
martin-schlossarek: I should also say that I based my work on how the API currently does things.

Right now, the API lets you pass in a list of content-view-environment labels or a list of environment IDs (entities). I implemented the first one. The second one would push things more toward an entity-list approach, like you mentioned.

TBH, I just went with the more pragmatic option because I want to use this feature soon and needed something that works. I played around with an entity-list version for a bit, but since I didn’t have the time or the deeper knowledge of the code, I decided to drop it for now.

plugins/modules/activation_key.py
evgeni: I think we're allowed to be better ;-)

(Although I will poke @jeremylenz and @parthaa, as I think API and hammer could/should be stricter here too)

martin-schlossarek:
Thanks for the hints! I updated the code and adjusted the tests.

Regarding labels vs entities: I also would prefer an entity based implementation, but I really struggled with that flatten logic for content view environments. Maybe it would be an option to rename the parameter to _content_view_environment_labels_? This leaves open the option for an entity based implementation in the future...?

evgeni (approved):
Given there is no user difference whether we use entities or not: ACK!

I'll wait with the merge until @jeremylenz or @parthaa had a chance to look at this

jeremylenz (approved):
> Maybe it would be an option to rename the parameter to _content_view_environment_labels_? This leaves open the option for an entity based implementation in the future...?

I wouldn't want to rename it, since that would affect the `--content-view-environments` parameter in Hammer. But aliasing them seems fine, so I opened https://projects.theforeman.org/issues/38937 in Katello.

I didn't test this but LGTM from a Katello / multiCV perspective! Thanks again @martin-schlossarek :heart:

PR #1939, opened 2025-12-11
add support for repo_data parameter in registration_command module

### LINKED ISSUE #1938: Add support for the 'repo_data' parameter in the 'registration_command' module
##### SUMMARY
The Foreman API has introduced the 'repo_data' parameter for [registration commands](https://apidocs.theforeman.org/foreman/latest/apidoc/v2/registration_commands/create.html). This parameter replaces the deprecated 'repo' and 'repo_gpg_key_url' parameters, and allows for multiple repository configurations in a single registration. This provides more flexibility and convenience.

However, the Ansible module `theforeman.foreman.registration_command` currently does not support the 'repo_data' parameter.

##### ISSUE TYPE
 - Feature Idea

Andrewich:
Add repo_data parameter support to registration_command module

- Added repo_data parameter as a list of dictionaries
- Updated documentation with deprecation notices

Fixes #1938

### DIFF
--- a/plugins/modules/registration_command.py
+++ b/plugins/modules/registration_command.py
@@ -204,7 +204,7 @@ def main():
             repo_gpg_key_url=dict(type='str', no_log=False),
             repo_data=dict(type='list', elements='dict', options=dict(
                 repo=dict(type='str', required=True),
-                repo_gpg_key_url=dict(type='str'),
+                repo_gpg_key_url=dict(type='str', no_log=False),
             )),
             remote_execution_interface=dict(type='str'),
             setup_remote_execution_pull=dict(type='bool'),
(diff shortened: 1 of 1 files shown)


plugins/modules/registration_command.py
```diff
@@ -89,13 +89,33 @@
   repo:
     description:
     - Repository URL (yum/dnf) or full sources.list entry (apt).
+    - "Deprecated: Use the I(repo_data) option instead."
```
evgeni: Should we note that it's deprecated since 3.11?

evgeni (approved):

plugins/modules/registration_command.py
Andrewich: I can just add a note to the description of the option: "This option was deprecated in version 3.11.0."
Is that okay?

evgeni: I'd do "Forman 3.11" so people are not confused with the version of the Ansible collection, but otherwise yeah that's what I meant!

Andrewich:
Done. Added a note that the 'repo' and 'repo_gg_key_url' parameters have been deprecated since Foreman version 3.11

PR #1941, opened 2025-12-12
Make verbosity configurable

shubhamsg199:
Add support for making the ansible verbosity configurable through environment variable.

### DIFF
--- a/tests/conftest.py
+++ b/tests/conftest.py
@@ -63,7 +63,7 @@ def run_playbook(module, extra_vars=None, limit=None, inventory=None, check_mode
     if inventory is None:
         inventory = os.path.join(os.getcwd(), 'tests', 'inventory', 'hosts')
     kwargs['inventory'] = inventory
-    kwargs['verbosity'] = 4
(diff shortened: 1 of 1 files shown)


tests/conftest.py
```diff
@@ -63,7 +63,7 @@ def run_playbook(module, extra_vars=None, limit=None, inventory=None, check_mode
     if inventory is None:
         inventory = os.path.join(os.getcwd(), 'tests', 'inventory', 'hosts')
     kwargs['inventory'] = inventory
-    kwargs['verbosity'] = 4
+    kwargs['verbosity'] = int(os.environ.get('FAM_ANSIBLE_VERBOSITY', 4))
```
evgeni:
```suggestion
    kwargs['verbosity'] = int(os.environ.get('FAM_TEST_ANSIBLE_VERBOSITY', 4))
```

evgeni (changes requested):
ACK pending that one variable name change

tests/conftest.py
shubhamsg199: Updated

evgeni (approved):

PR #1943, opened 2025-12-14
Replace pkg_resources with importlib.metadata

Gauravtalreja1:
While running tests, we're seeing a following warning for pkg_resources deprecation, which will be removed soon, so replacing pkg_resources with importlib.metadata 
```
=========================================================== warnings summary ===========================================================
tests/conftest.py:5
  /Users/gtalreja/my_workspace/foreman-ansible-modules/tests/conftest.py:5: UserWarning: pkg_resources is deprecated as an API. See https://setuptools.pypa.io/en/latest/pkg_resources.html. The pkg_resources package is slated for removal as early as 2025-11-30. Refrain from using this package or pin to Setuptools<81.
    import pkg_resources

-- Docs: https://docs.pytest.org/en/stable/how-to/capture-warnings.html
========================================================== 1 warning in 0.02s ==========================================================
```

### DIFF
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -113,6 +113,9 @@ jobs:
       - name: Install required collections for ansible 2.10
         run: ansible-galaxy collection install https://galaxy.ansible.com/api/v3/plugin/ansible/content/published/collections/artifacts/community-docker-2.7.12.tar.gz
         if: matrix.ansible == 'v2.10.4'
+      - name: Install importlib-metadata for python_version < '3.8'
+        run: pip install importlib-metadata
+        if: matrix.python in ['3.7', '3.8']
       - name: Run crud tests
         run: make test-crud
       - name: Run other tests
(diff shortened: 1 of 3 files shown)


tests/conftest.py
```diff
@@ -2,7 +2,7 @@
 import os
 
 import ansible_runner
-import pkg_resources
+from importlib.metadata import version, PackageNotFoundError
```
evgeni: When you use `importlib-metadata` backport package, you need to import from `importlib_metadata` instead, so:
```suggestion
try:
    from importlib.metadata import version, PackageNotFoundError
except ModuleNotFoundError:
    from importlib_metadata import version, PackageNotFoundError
```

.github/workflows/main.yml
```diff
@@ -113,6 +113,9 @@ jobs:
       - name: Install required collections for ansible 2.10
         run: ansible-galaxy collection install https://galaxy.ansible.com/api/v3/plugin/ansible/content/published/collections/artifacts/community-docker-2.7.12.tar.gz
         if: matrix.ansible == 'v2.10.4'
+      - name: Install importlib-metadata for python_version < '3.8'
+        run: pip install importlib-metadata
+        if: matrix.python in ['3.7', '3.8']
```
evgeni: you added it to `requirements-dev.txt`, so this is not necessary (and also `in` doesn't work in GHA)
```suggestion
```

evgeni (changes requested):

evgeni (approved):

PR #1944, opened 2025-12-16
automatically use system certs via truststore

omenos:
`requests` defaults to using the certificate trust provided by `certifi`, which precludes the inclusion of custom trusted certificates. Rather than needing to set the `REQUESTS_CA_BUNDLE` environment variable in every shell/config, use [`truststore`](https://github.com/sethmlarson/truststore) to override the way certs are read and automatically use the system trust store.

This will only happen transparently in environments using [Python >= 3.10](https://github.com/sethmlarson/truststore/blob/main/src/truststore/__init__.py) and is effectively a noop in any other scenario.

### DIFF
--- a/plugins/inventory/foreman.py
+++ b/plugins/inventory/foreman.py
@@ -197,6 +197,12 @@ from ansible.plugins.inventory import BaseInventoryPlugin, Cacheable, to_safe_gr
 
 # 3rd party imports
 try:
+    import truststore
+    truststore.inject_into_ssl()
+except ImportError:
+    pass
+
+try:
     import requests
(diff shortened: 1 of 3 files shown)


omenos:
Copying over my comment from ansible-collections/vmware.vmware#295 after Mike noted the behavioral change:

I just did a quick check (on RHEL) and if `SSL_CERT_FILE` is specified, it will be used rather than the system store.

So it's a bit of a behavior reversal, rather than:

* Default: `certifi`
* `REQUESTS_CA_BUNDLE`: selected bundle

it becomes:

* Default: system store via `truststore`
* `SSL_CERT_FILE`: selected bundle

---

Whether or not this project views that as a breaking change is not something I can define.

PR #1946, opened 2026-01-20
use run_command instead of subprocess.check_output

evgeni:
Ansible develop fails sanity if you use subprocess directly


shubhamsg199 (approved):
Ack

PR #1958, opened 2026-03-05
Fix content_view_filter_rule to consider versions for deb filters

OttaviaB:
# Description
When creating a filter rule for a filter of type `deb`, like in the following example, the result did not take into account the `min_version`, `version` and `max_version` parameters and always included all versions instead (see screenshot below).

```yaml
    - name: "Ensure presence of CV filter rules"
      theforeman.foreman.content_view_filter_rule:
        content_view: "Salt deb - example"
        content_view_filter: "Salt 3006"
        min_version: "3007"
        name: "*"
        organization: "default"
```
before_fix_1

The issue was that  the dictionary `content_filter_rule_deb_spec` did not include these entries and thus they were being ignored.

### DIFF
--- a/changelogs/fragments/deb_filter_rule.yml
+++ b/changelogs/fragments/deb_filter_rule.yml
@@ -1,0 +1,3 @@
+---
+bugfixes:
+  - content_view_filter_rule - fix content_filter_rule_deb_spec to take into account desired versions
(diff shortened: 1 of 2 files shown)


evgeni (approved):

theft6142:
:heart: I've been beating my head against this bug for a few days now trying to isolate it.  Thank you.  Just tested develop and it works for my Debian content views.

PR #1959, opened 2026-03-10
Remove subscriptions from convert2rhel role

shubhamsg199:
As per https://github.com/Katello/katello/pull/11616 , You can no longer add/remove subscription pools to/from activation keys. The simple_content_access field is removed from the organization API response, and the SCA enable/disable/eligible/status endpoints and controller are deleted


roles/convert2rhel/tasks/activation_keys.yml
```diff
@@ -44,7 +20,6 @@
     name: "{{ foreman_convert2rhel_key_centos8 }}"
     lifecycle_environment: "{{ foreman_convert2rhel_lifecycle_env }}"
     content_view: "{{ foreman_convert2rhel_key_centos8 }}"
-  when: foreman_content_rhel_enable_rhel8
```
Gauravtalreja1, line 47: why removing this conditions, is this intended?

shubhamsg199: It defaults to true [here](https://github.com/theforeman/foreman-ansible-modules/blob/71e7640a33d9c41d36ecafd420d376a2a7bce69e/tests/test_playbooks/convert2rhel.yml#L28-L29) so no point in keeping it here

Gauravtalreja1: yes, but this is a playbook where we've both set to true, while it still makes sense to keep this in role, so user is able to enable repos/create AK as per defined variable for each major version

evgeni: yes, please keep the `enable` vars

evgeni (changes requested):

evgeni (approved):

Gauravtalreja1 (approved):

PR #1868, opened 2025-05-28
LDAP RFC4519 group membership

adamruzicka:
Companion to https://github.com/theforeman/foreman/pull/10527 . Draft until the main PR settles

PR #1960, opened 2026-03-12
Bugfixes

ivarmu:
- Fix the lifecycle_environments role so the loop is correctly defined
- The auth_sources_ldap role is sending incompatible parameters when the server_type is active_directory

### DIFF
--- a/roles/auth_sources_ldap/tasks/main.yml
+++ b/roles/auth_sources_ldap/tasks/main.yml
@@ -24,7 +24,7 @@
     port: "{{ item.port | default(omit) }}"
     server_type: "{{ item.server_type | default(omit) }}"
     ldap_filter: "{{ item.ldap_filter | default(omit) }}"
-    use_netgroups: "{{ item.use_netgroups | default(omit) }}"
+    use_netgroups: "{{ omit if item.server_type is match('active_directory') else item.use_netgroups | default(omit) }}"
     state: "{{ item.state | default('present') }}"
   loop: "{{ foreman_auth_sources_ldap }}"
   no_log: "{{ item.account_password is defined }}"
--- a/roles/auth_sources_ldap/tasks/main.yml
+++ b/roles/auth_sources_ldap/tasks/main.yml
@@ -25,6 +25,7 @@
     server_type: "{{ item.server_type | default(omit) }}"
     ldap_filter: "{{ item.ldap_filter | default(omit) }}"
     use_netgroups: "{{ omit if item.server_type is match('active_directory') else item.use_netgroups | default(omit) }}"
+    ldap_group_membership: "{{ omit if item.server_type is not match('posix') else item.ldap_group_membership | default(omit) }}"
     state: "{{ item.state | default('present') }}"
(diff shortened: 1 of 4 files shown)


roles/lifecycle_environments/tasks/main.yml
```diff
@@ -11,5 +11,4 @@
     prior: "{{ item.prior }}"
     label: "{{ item.label | default(omit) }}"
     state: "{{ item.state | default('present') }}"
-  with_items:
-    - "{{ foreman_lifecycle_environments }}"
+  loop: "{{ foreman_lifecycle_environments }}"
```
evgeni, line 14: Aren't those identical in the result?

roles/auth_sources_ldap/tasks/main.yml
```diff
@@ -24,7 +24,7 @@
     port: "{{ item.port | default(omit) }}"
     server_type: "{{ item.server_type | default(omit) }}"
     ldap_filter: "{{ item.ldap_filter | default(omit) }}"
-    use_netgroups: "{{ item.use_netgroups | default(omit) }}"
+    use_netgroups: "{{ omit if item.server_type is match('active_directory') else item.use_netgroups | default(omit) }}"
```
evgeni, line 27: In https://github.com/theforeman/foreman-ansible-modules/pull/1868 we introduced a new `ldap_group_membership` parameter, but do not expose those in the role.

Should we add support for that too here?

cc @adamruzicka as the original author of `ldap_group_membership` code

adamruzicka: That would be nice

ivarmu: Nice!, I'm adding this ;-)

roles/lifecycle_environments/tasks/main.yml
ivarmu: `with_items` is replaced by `loop` and should be updated. It's equivalent, but here was badly written (although it was working nice, curiously). The correct form would be:

```yaml
with_items: "{{ foreman_lifecycle_environments }}"
```

In the original code, it was looping over the list contained into the first element of the provided list... like making an auto-flatten of the input list...

I prefer to have that fixed :-)

evgeni: Yes, `with_items` does an implicit flatten: https://docs.ansible.com/projects/ansible/latest/playbook_guide/playbooks_loops.html#comparing-loops

I was just looking whether this fixes a bug (as in: bad behavior) or just cleans up things :)

roles/auth_sources_ldap/tasks/main.yml
```diff
@@ -24,7 +24,8 @@
...
+    ldap_group_membership: "{{ omit if item.server_type is not match('posix') else item.ldap_group_membership | default(omit) }}"
```
evgeni: server_type free_ipa also supports this

ivarmu: Did you reffer only to `ldap_group_membership` or also to `use_netgroups`?

PR #1961, opened 2026-03-13
Release 5.10.0

shubhamsg199:


Gauravtalreja1 (approved):

CHANGELOG.rst
```diff
@@ -6,6 +6,19 @@ theforeman.foreman Release Notes
 
 This changelog describes changes after version 0.8.1.
 
+v5.10.0
+=======
+
+Breaking Changes / Porting Guide
+--------------------------------
+
+- convert2rhel role - removed subscription management from activation keys as subscription pools can no longer be added/removed from activation keys
```
evgeni: If that's a breaking change the version should be 6.0.0

However, I don't think it is, given that it was a noop since a long time (non-SCA Katello is not a thing since Katello 4.12)

shubhamsg199: I thought so too, `antsibull-changelog`  gave me this so kept it as is. So should  `Minor Changes`  be suitable here ?

evgeni: It did so, because you added it to `breaking_changes` list, but at the same time it doesn't check versioning at all.

I'd downgrade this to a `minor_change` yes, and probably reword it to something like: `removed subscription support as it's been unused since Katello 4.12 (05/2024)`

evgeni:
You ammended the wrong commit :)

evgeni (approved):

shubhamsg199:
> You ammended the wrong commit :)

Updated

PR #1962, opened 2026-03-20
Do not add content view filters rules on specific filters

### LINKED ISSUE #1896: [create_content_view] Do not add content view filters rules when using original_packages or original_module_streams filters
##### SUMMARY
When using original_module_streams or original_packages filter, role content_views fails

##### ISSUE TYPE
 - Bug Report

##### ANSIBLE VERSION

```
ansible [core 2.18.7]
  config file = /sat_example_playbook/ansible.cfg
  configured module search path = ['/.ansible/plugins/modules', '/usr/share/ansible/plugins/modules']
  ansible python module location = /usr/local/Cellar/ansible/11.8.0/libexec/lib/python3.13/site-packages/ansible
  ansible collection location = /sat_example_playbook/collections:/usr/share/ansible/collections:/.ansible/collections
  executable location = /usr/local/bin/ansible
  python version = 3.13.5 (main, Jun 11 2025, 15:36:57) [Clang 17.0.0 (clang-1700.0.13.3)] (/usr/local/Cellar/ansible/11.8.0/libexec/bin/python)
  jinja version = 3.1.6
  libyaml = True
```

##### COLLECTION VERSION

```
Collection                               Version
---------------------------------------- -------
redhat.satellite                         5.5.0
```

##### KATELLO/FOREMAN VERSION

```
rubygem-katello-4.16.0.6-1.el9sat.noarch
foreman-3.14.0.4-1.el9sat.noarch
```

##### STEPS TO REPRODUCE

```shell
ansible-playbook -i localhost, playbook.yaml
```
```yaml
   ---
  - name: 'Create (Composite-) Content Views in a Red Hat Satellite'
    hosts: 'all'
    connection: local
    gather_facts: false
    vars:
      satellite_server_url: 'https://192.168.100.1'
      satellite_username: 'admin'
      satellite_password: 'password'
      satellite_organization: 'Default'
      satellite_validate_certs: false

    satellite_content_views:
      - name: 'cv-os-rhel9-latest'
        description: 'Last version of RHEL9.4 using latest RHEL9 repositories and filters'
        lifecycle_environments:
          - Library
        repositories:
          - name: 'Red Hat Enterprise Linux 9 for x86_64 - BaseOS RPMs 9'
            product: 'Red Hat Enterprise Linux for x86_64'
          - name: 'Red Hat Enterprise Linux 9 for x86_64 - AppStream RPMs 9'
            product: 'Red Hat Enterprise Linux for x86_64'
        filters:
          - name: 'All packages without errata'
            filter_type: 'rpm'
            inclusion: true
            original_packages: true
          - name: 'All modules without errata'
            filter_type: 'modulemd'
            inclusion: true
            original_module_streams: true
          - name: 'Include errata until date'
            filter_type: 'erratum'
            inclusion: true
            date_type: updated
            end_date: '2024-11-10'

  roles:
    - name: 'redhat.satellite.content_views'
```

##### EXPECTED RESULTS

Create content view without error

##### ACTUAL RESULTS
content view is created with required filters but ansible playbook failed be cause it's trying to add rules to filter with original_packages and original_module_streams set to true

tinsjourney:
When using original_packages or original_module_streams filters we don't need to add rules.
Updated to match Release 5.10.0

Fix theforeman#1896

### DIFF
--- a/roles/content_views/tasks/_create_content_view.yml
+++ b/roles/content_views/tasks/_create_content_view.yml
@@ -60,4 +60,6 @@
     rule_state: "{{ item.rule_state | default(omit) }}"
   when:
     - content_view.state is not defined or content_view.state != 'absent'
+    - not (item.original_packages | default(omit)) | bool
+    - not (item.original_module_streams | default(omit)) | bool
   loop: "{{ content_view.filters | default([]) }}"


evgeni:
I think this change is not entirely correct, but my package filtering knowledge is severely lacking, so maybe @ianballou can jump in here.

The example in #1896 is:
```
        filters:
          - name: 'All packages without errata'
            filter_type: 'rpm'
            inclusion: true
            original_packages: true
```
and I agree, for that example the current role will produce an error (as it tries to add a rule without any specification), while your change fixes it.

However, the way I read the description of `original_packages` and `original_module_streams` is that these are *additional* properties of the filter that are applied *after* all rules were executed. So for example the following would be totally valid (even tho might not be too sensible):
```
        filters:
          - name: 'All packages without errata or arch x86_64'
            filter_type: 'rpm'
            inclusion: true
            original_packages: true
            architecture: x86_64
```
As this will first find all packages with the matching architecture and *then* add all without errata.

If setting `original_*` and adding additional rules is something "people don't usually do", I think your change is fine. Otherwise we'd need to make the when condition more explicit.

ianballou:
Katello does support both `original_*` filters and further tailoring the filter rules. Maybe I want to include all RPMs without errata plus some other one, for example. If Katello supports it, FAM should support it too.

PR #1966, opened 2026-03-26
Add support for setup_container_registry_certs in registration_command

pablomh:
## Summary

- Add `setup_container_registry_certs` parameter to the `registration_command` module
- When set to `true`, container registry certificates will be installed on the host during registration
- Marked as requiring the Katello plugin, as it depends on subscription-manager entitlement certificates

## Test plan

- [ ] Updated `tasks/registration_command.yml` to accept `setup_container_registry_certs`
- [ ] Added a second test invocation in `registration_command.yml` with `setup_container_registry_certs: true`
- [ ] VCR fixture `registration_command-1.yml` needs to be recorded against a real Foreman+Katello instance

🤖 Generated with [Claude Code](https://claude.com/claude-code)

### DIFF
--- a/plugins/modules/registration_command.py
+++ b/plugins/modules/registration_command.py
@@ -132,6 +132,12 @@ options:
     - If this is set to true, pull provider client will be deployed on the host.
     required: false
     type: bool
+  setup_container_registry_certs:
+    description:
+    - If this is set to C(true), container registry certificates will be installed on the host.
+    required: false
+    type: bool
+    version_added: 5.11.0
   smart_proxy:
     description:
     - Name of Smart Proxy.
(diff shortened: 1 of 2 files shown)


tests/test_playbooks/registration_command.yml
```diff
@@ -27,6 +27,11 @@
       vars:
         rc_activation_keys:
           - "Test Activation Key"
+    - include_tasks: tasks/registration_command.yml
```
evgeni: if you add new tests (which I think is not necessary in this case), you need to re-record the fixtures :)

plugins/modules/registration_command.py
```diff
@@ -132,6 +132,11 @@
     - If this is set to true, pull provider client will be deployed on the host.
     required: false
     type: bool
+  setup_container_registry_certs:
+    description:
+    - If this is set to C(true), container registry certificates will be installed on the host.
+    required: false
+    type: bool
```
evgeni, line 139:
```suggestion
    type: bool
    version_added: 5.11.0
```

evgeni:
a changelog snippet would be cool for that

evgeni (approved):

PR #1967, opened 2026-03-27
Assert default of login_text, not value in module_defaults tests

evgeni:
That way the tests also pass on setups that have actually changed the value


arvind4501 (approved):

arvind4501:
Yes, thats better than value(which can be changed easily)

evgeni:
Technically, we don't need to assert anything - the test at hand is supposed to verify module defaults were able to provide authentication parameters properly, and if they didn't the call of the module would've already failed.

But it still feels safer :woman_shrugging:

PR #1973, opened 2026-04-10
Add Ansible 2.21 to test matrix

evgeni:


.github/workflows/main.yml
```diff
@@ -23,6 +23,7 @@ jobs:
           - stable-2.18
           - stable-2.19
           - stable-2.20
+          - stable-2.21
```
Gauravtalreja1, line 26: As per Ansible roadmap, Beta/pre-release version of Ansible-core v2.21 is out recently, but its release is planned for `2026-05-18`, aren't we too early for this? 😄 
https://docs.ansible.com/projects/ansible-core/devel/roadmap/ROADMAP_2_21.html

evgeni: Nah. If we add it to the matrix now, we'll be able to make sure it's always tested, regardless of beta or GA.

ehelms (approved):

.github/workflows/main.yml
Gauravtalreja1: fair enough, thanks!

Issue #1463, opened 2022-08-24
theforeman.foreman.smart_proxy lifecycle_environments not working with duplicate LCEs names


lpther:
##### SUMMARY

Using the theforeman.foreman.smart_proxy lifecycle_environments parameter, having multiple lifecycle environments with the same name doesn't work.

##### ISSUE TYPE
 - Bug Report

##### ANSIBLE VERSION

```
ansible [core 2.13.3]
  config file = /data/home/lpt069/workspace/ansible.cfg
  configured module search path = ['/data/home/lpt069/.ansible/plugins/modules', '/usr/share/ansible/plugins/modules']
  ansible python module location = /data/home/lpt069/.local/lib/python3.8/site-packages/ansible
  ansible collection location = /data/home/lpt069/.ansible/collections:/data/home/lpt069/workspace/ansible/collections
  executable location = /data/home/lpt069/.local/bin/ansible
  python version = 3.8.10 (default, Mar 15 2022, 12:22:08) [GCC 9.4.0]
  jinja version = 3.1.2
  libyaml = True

```

##### COLLECTION VERSION

```
# /data/home/lpt069/.ansible/collections/ansible_collections
Collection            Version
--------------------- -------
ansible.posix         1.3.0  
community.general     5.0.0  
community.vmware      2.6.0  
theforeman.foreman    3.4.0  
theforeman.operations 1.2.0  

```

##### KATELLO/FOREMAN VERSION

```
[root@foreman-dev1 ~]# rpm -q tfm-rubygem-katello foreman
package tfm-rubygem-katello is not installed
foreman-3.2.1-1.el8.noarch

```

##### STEPS TO REPRODUCE

```yaml

Define two different organization
Add a smart proxy

- name: Add a lifecycle environment that is present multiple times
  theforeman.foreman.smart_proxy:
    username: "{{ foreman_helper__foreman_username }}"
    password: "{{ foreman_helper__foreman_password }}"
    server_url: "{{ foreman_helper__foreman_url }}"
    validate_certs: "{{ foreman_helper__foreman_validate_certs }}"
    name: proxy.example.com
    url: https://proxy.example.com:9090
    lifecycle_environments:
      - Library

```

##### EXPECTED RESULTS

Add all organization libraries to the smart proxy. Or even better, having a way to set the organization alongside the name like when doing the same via hammer:

hammer capsule content add-lifecycle-environment --name proxy.example.com --lifecycle-environment Library --organization Organization1

[...]

PR #1980, opened 2026-06-29
Add org scoping to lifecycle_environments variable

tyderian1978:
This fixes the `Found too many (3) results while searching for lifecycle_environments with name="Library"` error message seen when specifing lifecycle_environments for both the `content_view` and `smart_proxy modules`.

This related to [1463](https://github.com/theforeman/foreman-ansible-modules/issues/1463) _theforeman.foreman.smart_proxy lifecycle_environments not working with duplicate LCEs names_.

### DIFF
--- a/plugins/modules/content_view.py
+++ b/plugins/modules/content_view.py
@@ -200,7 +200,7 @@ def main():
             solve_dependencies=dict(type='bool'),
             import_only=dict(type='bool'),
             components=dict(type='nested_list', foreman_spec=cvc_foreman_spec, resolve=False),
-            lifecycle_environments=dict(type='entity_list', flat_name='environment_ids'),
+            lifecycle_environments=dict(type='entity_list', flat_name='environment_ids', scope=['organization']),
             repositories=dict(type='entity_list', elements='dict', resolve=False, options=dict(
                 name=dict(required=True),
                 product=dict(required=True),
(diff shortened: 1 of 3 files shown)


plugins/modules/smart_proxy.py
```diff
@@ -115,7 +115,7 @@ def main():
         foreman_spec=dict(
             name=dict(required=True),
             url=dict(required=True),
-            lifecycle_environments=dict(required=False, type='entity_list'),
+            lifecycle_environments=dict(required=False, type='entity_list', scope=['organization']),
```
evgeni: Does this work given the smart proxy module gets a list of organizations, not a single one?

tyderian1978: Hi @evgeni the lifecycle_environments variable is a list in all of its invocations (content_view, content_view_version and smart_proxy modules).  I have tested against the content_view module with both a single string "Library" and also a list with multiple items and can confirm it works ok.

evgeni: So whu are tests failing with 
[ERROR]: Task failed: Module failed: Failed to lookup scope ['organization'] while searching for lifecycle_environments.

tyderian1978: Hi evgeni, I've fixed the unit tests for content_view.
The fix for smart_proxy is more involved, so I thought it best to separate the fixes.  It looks like the remaining failing tests are not related to the changes I have made.

tyderian1978:
@evgeni any updates on reviewing this one?
Once the content_view fix is accepted (it's a blocker for my playbooks in AAP), I'll raise a pull request for the smart_proxy fix.

evgeni (approved):

PR #1985, opened 2026-07-17
Implement LDAP CA certs support

adamlazik1:
Made with Cursor.

Requires https://github.com/theforeman/foreman/pull/11107

### DIFF
--- a/plugins/modules/auth_source_ldap.py
+++ b/plugins/modules/auth_source_ldap.py
@@ -99,12 +99,12 @@ options:
     type: bool
   cacert:
     description:
-      - Path to a PEM file with CA certificate(s) used to verify the LDAP server when LDAPS is enabled.
+      - PEM-encoded CA certificate(s) used to verify the LDAP server when LDAPS is enabled.
       - When set, only these certificates are trusted and the system trust store is not used.
       - Omit to use the system trust store. Pass an empty string to clear a previously set CA certificate.
       - In containerized deployments, the container's trust store is used, so the CA certificate(s) may need to be provided here explicitly.
     required: false
-    type: path
+    type: str
   groups_base:
     description: Base DN where groups reside.
     required: false
@@ -196,7 +196,7 @@ EXAMPLES = '''
     host: "ldap.example.org"
     port: 636
     tls: true
-    cacert: /path/to/ldap-ca.pem
+    cacert: "{{ lookup('file', '/path/to/ldap-ca.pem') }}"
     server_type: free_ipa
     server_url: "https://foreman.example.com"
     username: "admin"
@@ -216,8 +216,6 @@ entity:
       elements: dict
 '''
 
-
-from ansible.module_utils.common.text.converters import to_native
 
 from ansible_collections.theforeman.foreman.plugins.module_utils.foreman_helper import ForemanTaxonomicEntityAnsibleModule
 
@@ -243,7 +241,7 @@ def main():
             onthefly_register=dict(type='bool'),
             usergroup_sync=dict(type='bool'),
             tls=dict(type='bool'),
-            cacert=dict(type='path'),
+            cacert=dict(),
             groups_base=dict(),
             server_type=dict(choices=["free_ipa", "active_directory", "posix"]),
             ldap_filter=dict(),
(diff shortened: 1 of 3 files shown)


plugins/modules/auth_source_ldap.py
```diff
@@ -97,6 +97,12 @@
     description: Whether or not to use TLS when contacting the LDAP server.
     required: false
     type: bool
+  cacert:
+    description:
+      - PEM-encoded CA certificate(s) used to verify the LDAP server when LDAPS is enabled.
+      - Leave empty to use the system trust store.
```
adamruzicka: This sort of reads like it is either-or, while the foreman-side code seems to be additive.

adamlazik1: You mean, system trust store is used even when adidtional cacerts are provided via param?

adamruzicka: Yes and that's fine, but we should be clear about it

adamlazik1: Fixed.

plugins/modules/auth_source_ldap.py
```diff
@@ -181,6 +187,19 @@
...
     password: "changeme"
     state: present
+
+- name: LDAPS with a custom CA certificate
+  theforeman.foreman.auth_source_ldap:
+    name: "Example LDAPS"
+    host: "ldap.example.org"
+    port: 636
+    tls: true
+    cacert: "{{ lookup('file', '/path/to/ldap-ca.pem') }}"
```
lhellebr, line 199: Shouldn't it take a file path rather than the cert itself, so the user won't need to do this? Purely from UX POV, it looks better.

adamlazik1: Applied.

lhellebr:
ACK as per https://github.com/theforeman/foreman/pull/11107#issuecomment-5142155899

In case of FAM, I've used the version I got from `gh pr checkout` of this PR.

plugins/modules/auth_source_ldap.py
adamruzicka: I wouldn't be so sure about that, especially when it comes to various integrations. Passing a string around is much easier than ensuring that a file with specific contents will be available on the ansible control node, whichever machine that may end up being.

@evgeni opinions?

evgeni: Of course Evgeni has opinions! ;)

1. Yes, passing a string that came from $somewhere is much easier if $somewhere is not a file system (read: Vault, Bitwarden, SOPS, etc).
2. When you have a `path` parameter, the file is loaded on the target node, while a `lookup` happens on the controller node. I do preach "just use `hosts: localhost` when using the API", but not everyone listens and might end up with `hosts: foreman`, which would mean you first need to copy the file to the Foreman machine and *then* load it with a `path` parameter¹. Or you could have done it with `lookup` directly.

¹: There are modules that have a `remote_src` param which allows you to trigger between a remote source (= on the target node) and a local one (= on the controller), but that's awful and you don't want to see the module code needed to achieve that.

Given you're the people for LDAP auth, I'll leave the decision up to you. But if I would implement it, `cacert` would be a string, fed by `lookup` in the examples.

plugins/modules/auth_source_ldap.py
```diff
@@ -239,6 +263,13 @@ def main():
     if 'ldap_group_membership' in module.foreman_params and module.foreman_params['ldap_group_membership'] == 'rfc4519' and server_type != 'posix':
         module.fail_json(msg=f'ldap_group_membership=rfc4519 cannot be used when server_type={server_type}')
 
+    if module.foreman_params.get('cacert'):
```
evgeni:
```suggestion
    if module.foreman_params.get('cacert') and not module.desired_absent:
```

that allows the module to delete an ldap source even when the referenced file is absent (not needed if you switch back to strings)

plugins/modules/auth_source_ldap.py
adamlazik1: Reverted.

evgeni (approved):

adamruzicka (approved):

PR #1996, opened 2026-09-01
Add role to add and manage host collections

nbeernink:
Saw this role was missing, figured I added it. :)
Did I forget to add anything?


nbeernink:
Saw that `organization` is actually a required field for the host_collection module so made it not default to `omit` and updated README.md.

nbeernink:
Added a test play and made `organization` default to `foreman_organization` similar to how the activation_keys role does it.

PR #2003, opened 2026-09-21
Add role to manage webhooks

nbeernink:
Another role I was missing and would like to contribute  back.


nbeernink:
Tried adding a test play and test against a forklift centos9-katello-devel-stable box but then it fails with:
```
TASK [webhooks : Create webhooks] ******************************************************************
[ERROR]: Task failed: Module failed: Failed to ensure entity state: ForemanApiException: Error while performing create on webhooks: 
422 Client Error: Unprocessable Content for url: https://192.168.122.20/api/webhooks - {'id': None, 'errors': {'events': ["can't be blank"]}, 'full_messages': ["Events can't be blank"]}
```

Even though the event is filled in in the test-playbook like so:
```yaml
# tests/test_playbook/webhooks_role.yml
---
- hosts: localhost
  collections:
    - theforeman.foreman
  gather_facts: false
  vars_files:
    - vars/server.yml
  tasks:
    - name: create webhook
      include_role:
        name: webhooks
      vars:
        foreman_webhooks:
          - name: Test webhook
            event: 'actions.katello.content_view.promote_succeeded.event.foreman'
            target_url: "http://example.org:3128"
            http_method: POST
            webhook_template: Katello Publish
            enabled: true
``` 

Get the same thing when trying to run the already included `tests/test_playbook/webhook.yml`

